PR #2, opened 2016-10-16
Set tags on AWS resources as expected by K8s

mikkeloscar:
This patch adds the unique tag `KubernetesCluster` on the AWS resources in a
cluster as expected by K8s [1,2].

Fixes the issue where a PVC creates an EBS volume in the wrong AZ.

[1] https://github.com/kubernetes/kubernetes/blob/master/docs/design/aws_under_the_hood.md#tagging
[2] https://github.com/kubernetes/kubernetes/blob/7dcae5edd84f06e7d71a041931391a256b34dec9/pkg/cloudprovider/providers/aws/aws.go#L58-L60

### DIFF
--- a/cluster/senza-definition.yaml
+++ b/cluster/senza-definition.yaml
@@ -50,6 +50,9 @@ SenzaComponents:
           InstanceProtocol: SSL
           InstancePort: 443
           LoadBalancerPort: 443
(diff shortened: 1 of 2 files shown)


Raffo:
👍

PR #3, opened 2016-10-17
Handle systemd deps more reliable

mikkeloscar:
Sometimes the download from quay.io is so slow that `flanneld.service`
will be restarted by systemd. If this happens at the same time
`install-kube-system.service` is started then systemd will see a failing
dependency chain (docker.service -> flanneld.service) and will not try
to restart `install-kube-system.service`, effectively not installing the
expected k8s pods (heapster, dashboard, etc.).

This patch fixes the issue by handling the dependency check in
`ExecStartPre` allowing for a restart if it fails the first times.

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -93,14 +93,12 @@ coreos:
       command: start
       runtime: true
       content: |
-        [Unit]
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

PR #5, opened 2016-10-17
Fix etcdctl command

mikkeloscar:
#4 introduced a stray dash in the cloud config causing the master node to not come up.

This fixes it.


Raffo:
👍

PR #8, opened 2016-10-18
[postponed] split masters and workers in two cf stacks

linki:


linki:
postponed for now. may not be done at all.

PR #12, opened 2016-10-20
[wip] add mate to kube stack

linki:
**work in progress**

this adds mate by default to the stack 

discussion needed:
* opensource mate?
* hosted zone needs to be configured correctly
* `--no-sync` is the default but explicitly defined for clarity
* `route:53*` policy stuff must obviously be tweaked

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -748,7 +748,7 @@ write_files:
                 name: credentials
                 readOnly: true
             - name: kubectl
-              image: quay.io/coreos/hyperkube:v1.3.6_coreos.0
+              image: quay.io/coreos/hyperkube:v1.3.9_coreos.0
               command:
               - /hyperkube
               args:
@@ -784,7 +784,7 @@ write_files:
               - --aws-hosted-zone=HOSTED_ZONE.
               - --no-sync
             - name: kubectl
-              image: quay.io/coreos/hyperkube:v1.3.6_coreos.0
+              image: quay.io/coreos/hyperkube:v1.3.9_coreos.0
               command:
               - /hyperkube
               args:
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -776,7 +776,7 @@ write_files:
           spec:
             containers:
             - name: mate
-              image: pierone.stups.zalan.do/teapot/mate:v0.0.4-32-g0af5e78
+              image: registry.opensource.zalan.do/teapot/mate:v0.0.4-34-gf2f1dd2
               args:
               - --producer=kubernetes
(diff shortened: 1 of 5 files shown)


cluster/userdata-master.yaml
```diff
@@ -756,6 +760,36 @@ write_files:
...
+          spec:
+            containers:
+            - name: mate
+              image: pierone.stups.zalan.do/teapot/mate:v0.0.4-31-gebb7747
+              args:
+              - --producer=kubernetes
+              - --kubernetes-domain=mate.teapot.zalan.do.
+              - --consumer=aws
+              - --aws-hosted-zone=mate.teapot.zalan.do.
+              - --no-sync
```
linki: mate knows about producers (source of records, e.g. kube services) and consumers (dns providers, e.g. route53)

it seems the zone is duplicated but that's not the full story. `--kubernetes-domain` is responsible for what dns records to create. `--aws-hosted-zone` just picks a zone on the aws side. (legacy from google)

szuecs:
:+1:

cluster/userdata-master.yaml
```diff
@@ -756,6 +760,36 @@ write_files:
...
+            - name: mate
+              image: pierone.stups.zalan.do/teapot/mate:v0.0.4-32-g0af5e78
+              args:
+              - --producer=kubernetes
+              - --kubernetes-domain=HOSTED_ZONE.
+              - --consumer=aws
+              - --aws-hosted-zone=HOSTED_ZONE.
+              - --no-sync
+            - name: kubectl
+              image: quay.io/coreos/hyperkube:v1.3.6_coreos.0
```
hjacobs: we should upgrade hyperkube version here

cluster/userdata-master.yaml
```diff
@@ -756,6 +760,36 @@ write_files:
...
+      spec:
+        replicas: 1
+        template:
+          metadata:
+            labels:
+              app: mate
+          spec:
+            containers:
+            - name: mate
+              image: pierone.stups.zalan.do/teapot/mate:v0.0.4-32-g0af5e78
```
hjacobs: I think mate should be in the OSS registry (and we should open source) --- having Pier One here depends on correct credentials

cluster/userdata-master.yaml
Raffo: 1.4.3 is the latest on quay, 1.4.4 not yet there.

cluster/userdata-master.yaml
szuecs: Then we should also publish mate

Raffo: We're on the same page: https://github.bus.zalan.do/teapot/mate/issues/23

hjacobs:
:+1:

PR #16, opened 2016-10-24
Cluster create update

hjacobs:
Preparation for cluster updates (updating launch config + respawn)

NOTE: cluster name is now dynamic, i.e. you can create "mycluster-1.myteam.example.org"


Raffo:
👍

PR #19, opened 2016-10-25
Deploy etcd cluster, update nodes

hjacobs:
* fix multi-AZ ELB (hardcode `KubernetesCluster` tag and tag all subnets)
* deploy etcd cluster if none exists
* allow updating master and worker nodes (updates CF and respawns nodes --- this causes a downtime right now!)
* wait for CF stack creation and API server


cluster/senza-definition.yaml
```diff
@@ -42,7 +42,7 @@ SenzaComponents:
           LoadBalancerPort: 443
       Tags:
         - Key: "KubernetesCluster"
-          Value: {Ref: 'AWS::StackName'}
+          Value: kubernetes
```
szuecs: Which advantage do we get if we change this?

hjacobs: we can run multiple versions in one account/subnet (e.g. for testing or migrations)

PR #20, opened 2016-10-25
wait for master node to be "in service" in ELB

hjacobs:
This should make master node updates painless (without downtime).


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #27, opened 2016-10-26
Allow node exporter access from internal network

szuecs:
We have to allow zmon to access all node_exporter services on all nodes.


hjacobs:
:+1:

PR #28, opened 2016-10-26
#26 overwrite manifest with PUT on startup

### LINKED ISSUE #26: Updated manifests are not applied when updating master nodes
Changing the manifests in the master user data will not actually update the pods.

We probably need to change the `curl -X POST` call.

hjacobs:
Fixes #26

Just did this for secretary to try out


szuecs:
:+1:

PR #29, opened 2016-10-26
Implement wait and drain when updating worker nodes

### LINKED ISSUE #24: Updating a cluster causes downtime for worker nodes
Updating the cluster with `./cluster.py update <stack-name> <version>` will currently lead to a downtime for worker nodes as we are not waiting for node readiness (we just wait for ASG `InService` which does not indicate whether the node is completely up and registered).

NOTE: there is no downtime for master nodes while updating as we already correctly wait/check the ELB status for the master API server.

mikkeloscar:
* Waits for nodes to be ready by querying the kubernetes API and checks
  Ready status.
* Drains node before terminating it with `kubectl drain`. (uses
  worker_shared_secret to authenticate to the API server).

Fix #24

### DIFF
--- a/cluster/cluster.py
+++ b/cluster/cluster.py
@@ -276,12 +276,51 @@ def update(stack_name, version, force):
                            'UserDataWorker={}'.format(user_data_worker), 'KmsKey=*'])
     # wait for CF update to complete..
     subprocess.check_call(['senza', 'wait', '--timeout=600', stack_name, version])
-    perform_node_updates(stack_name, version, 'Master', user_data_master)
+    perform_node_updates(stack_name, version, 'Master', user_data_master, variables)
     wait_for_api_server(variables['api_server'])
-    perform_node_updates(stack_name, version, 'Worker', user_data_worker)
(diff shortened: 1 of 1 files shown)


hjacobs:
:+1:

cluster/cluster.py
```diff
@@ -276,12 +276,51 @@ def update(stack_name, version, force):
...
+
+def get_k8s_node_name(instance_id: str, config: dict):
+    nodes = get_k8s_nodes(config["api_server"], config["worker_shared_secret"])
+    for node in nodes["items"]:
+        if node["spec"]["externalID"] == instance_id:
+            return node["metadata"]["name"]
+    return ""
+
+
+def drain_node(node_name: str, config: dict, grace_period=10):
```
linki: in kubernetes each pod has a termination grace period after which the kubelet will try harder to kill it. this period is `30s` by default. i think it makes sense to match that default here.

mikkeloscar: We could also find the pod with the longest grace period on the node and respect that (within reason).

PR #30, opened 2016-10-26
Put manifest

hjacobs:
Also use PUT for the other manifests.

Remove TLS decrypt (we are not using it, client cert auth is disabled).

** still testing **


linki:
turns out `kubectl` does check existence on the client side. `PUT` on a missing deployment will just fail. i'll rewrite this part to use `kubectl apply -f` directly.

hjacobs:
:+1:

PR #33, opened 2016-10-26
Fetch hyperkube from pierone instead of quay.io

mikkeloscar:
Note this also switches to coreos beta channel which has a `rkt` version
compatible with pierone.


szuecs:
:+1:

PR #32, opened 2016-10-26
some minor improvements

linki:
* tagged version of mate (support `--no-sync`)
* run two copies of dns server as done by gke

PR #35, opened 2016-10-27
Change rc to deployment

### LINKED ISSUE #31: Cleanup: Replace ReplicationController with Deployment
We have some "legacy" manifests (e.g. `kube-dns`) using `ReplicationController`, we should replace it with `Deployment`.

This is low prio.

mikkeloscar:
Changes all replicacontrollers to deployments (kube-dns and dashboard).

Fix #31

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -165,7 +165,7 @@ write_files:
 
       KUBECTL="docker run --net host -v /srv/kubernetes/manifests:/manifests:ro quay.io/coreos/hyperkube:v1.4.4_coreos.0 /hyperkube kubectl"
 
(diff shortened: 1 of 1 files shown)


linki:
👍

cluster/userdata-master.yaml
```diff
@@ -420,10 +416,11 @@ write_files:
             version: v19
             kubernetes.io/cluster-service: "true"
         spec:
-          replicas: 1
+          replicas: 2
```
hjacobs: any reason to have two replicas now?

linki: i did that in https://github.com/zalando-incubator/kubernetes-on-aws/pull/32

linki: the dns setup is to be known a bit flaky. somewhere it was recommend to run two copies and gke also does this. should not harm.

PR #36, opened 2016-10-27
Respect grace-period of pods when draining a node

mikkeloscar:
When draining a node we want to respect termination grace period of
pods.

This patch solves this by finding the longest termination grace period of
all pods on a node and uses this when draining the node. A max grace
period is still enforced (currently set to 60s).


hjacobs:
:+1:

PR #37, opened 2016-10-27
Keep ASG size when updating CF

hjacobs:
Quickfix to apply current `DesiredCapacity` (i.e. keep ASG size when updating CF stack).


mikkeloscar:
:+1:

PR #38, opened 2016-10-27
allow instance type to be configured

linki:
worker instance type can be configured with this

on create:
* provide optional flag `--instance-type` to specify instance type of worker nodes (defaults to `t2.micro`)

on update:
* provide optional flag `--instance-type` to change instance type of worker nodes (defaults to `current` which results in whatever type the workers currently have)
* if cloud-config didn't change one has to use the `--force` flag to trigger the update


hjacobs:
:+1:

PR #44, opened 2016-10-28
Use non burstable cpu instance type

szuecs:
Use the cheapest non burstable cpu instance type that http://www.ec2instances.info/ suggests


hjacobs:
That instance type does not exist in Frankfurt (m1 is old): http://www.ec2instances.info/?region=eu-central-1&cost_duration=monthly

PR #50, opened 2016-10-28
allow to specify node count on create/update for master/worker

linki:


cluster/cluster.py
```diff
@@ -289,13 +297,18 @@ def update(stack_name, version, force, instance_type):
         info('Neither worker nor master user data did change, not updating anything.')
         return
 
-    worker_nodes = get_current_worker_nodes(stack_name, version)
+    if master_nodes == -1:
+        master_nodes_nodes = get_current_master_nodes(stack_name, version)
```
hjacobs: wrong variable here

hjacobs:
:+1:

cluster/cluster.py
linki: didn't see this one. thx for spotting!

PR #55, opened 2016-10-31
integrate mate and secretary with kube2iam

linki:
this let's `mate` and `secretary` use their correct iam roles


mikkeloscar:
:+1:

PR #71, opened 2016-11-02
Start install-ssh-keys early

mikkeloscar:
Coreos runs `coreos-cloudinit` to start each systemd service
sequentially and waits for service to be started before starting the
next one. This means that the `install-kube-system.service` blocks
`install-ssh-keys.service` until the API server is up.

Solution is to start the `install-ssh-keys.service` before
`install-kube-system.service`.


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #72, opened 2016-11-02
add --postpone flag: do not update nodes

linki:
this makes the node restarts on `update` optional.

allows for much faster iterations when testing new cloud configs, e.g.

```
$ ./cluster.py update foo v1 --force --postpone
$ senza respawn-instances foo v1 --force --inplace
```


mikkeloscar:
:+1:

hjacobs:
:+1:

PR #75, opened 2016-11-02
Align vagrant setup with the real AWS

mikkeloscar:
This aligns the vagrant and AWS setup a bit, it's still slightly
different in that it doesn't use the webhook and obviously doesn't
deploy AWS dependent services like secretary, mate and kube2iam.


cluster/userdata-worker.yaml
```diff
@@ -60,11 +60,11 @@ coreos:
         --pod-manifest-path=/etc/kubernetes/manifests \
         --cluster_dns=10.3.0.10 \
         --cluster_domain=cluster.local \
-        --cloud-provider=aws \
         --kubeconfig=/etc/kubernetes/worker-kubeconfig.yaml \
         --require-kubeconfig \
         --tls-cert-file=/etc/kubernetes/ssl/worker.pem \
         --tls-private-key-file=/etc/kubernetes/ssl/worker-key.pem \
+        --cloud-provider=aws \
```
linki: shouldn't we drop this for the vagrant setup?

linki:
👍

cluster/userdata-worker.yaml
mikkeloscar: Yes, this is the userdata for AWS clusters, I just moved the line so it would be less different from the vagrant version. :)

mikkeloscar:
:+1:

PR #77, opened 2016-11-02
add calico network policy support

linki:
adds calico's network policy functionality based on how CoreOS set's up Kubernetes: https://coreos.com/kubernetes/docs/latest/deploy-master.html

it still uses `flannel` for the overlay network but adds firewalling capabilities between containers via calico/felix.

as it sits here, the functionality is opt-in. using this PR will not affect the user unless they add network policies to their clusters via `kubectl`

## components
* each node runs the `calico-node` agent that watches etcd and configures iptables on each node. (similar to how `kube-proxy` works but against etcd directly)
* there's one controller running as a pod that watches the kubernetes API for declaritive network policy rules (annotations as well `networkpolicy` api objects). it actually runs a copy of the controller on each master with leader election. (similar to `controller-manager` etc.)

## changes
* `kube2iam` needs to be configured to create rules for the calico interfaces
* CNI config needs to be changed for calico

## demo
you can try it out yourself on a calico-enabled cluster by following this walkthrough: http://kubernetes.io/docs/getting-started-guides/network-policy/walkthrough/

yes, but only tested with a simple example

> What are the benefits?

every container that listens on some port is accessible from every other container in the cluster. on bigger clusters with many people working on it this may not be desirable.

> Is it more complex?

yes

> Do we understand it on an equal level as flanneld (which looks simple enough)?

flannel stays as it is. calico adds some unknows

> Is it more brittle?

potentially

> What are the config / tuning knobs?

don't know

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -37,6 +37,42 @@ coreos:
             ExecStartPre=/usr/bin/etcdctl \
             --endpoint=http://localhost:2379 set /coreos.com/network/config \
             '{ "Network": "10.2.0.0/16", "Backend": {"Type": "vxlan"}}'
+        - name: 40-ExecStartPre-symlink.conf
+          content: |
+            [Service]
+            ExecStartPre=/usr/bin/ln -sf /etc/flannel/options.env /run/flannel/options.env
+
+    - name: calico-node.service
+      command: start
+      content: |
+        [Unit]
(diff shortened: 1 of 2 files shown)


hjacobs:
* Does it work?
* What are the benefits?
* Is it more complex?
* Do we understand it on an equal level as flanneld (which looks simple enough)?
* Is it more brittle?
* What are the config / tuning knobs?

PR #81, opened 2016-11-03
allow write access to S3

szuecs:
allow write access to arn:aws:s3:::teapot-etcd-backup to enable backup to s3 from etcd hosts


szuecs:
:+1:

PR #86, opened 2016-11-03
Autoscaler

Raffo:
I have here an arbitrary number for the max size of the autoscaler. I think I should just make it configurable, what do you think?

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -854,7 +854,7 @@ write_files:
                 privileged: true
 
 
-  - path: /srv/kubernetes/manifests/daemonsets/cluster-autoscaler.yaml
+  - path: /srv/kubernetes/manifests/deployments/cluster-autoscaler.yaml
     content: |
       apiVersion: extensions/v1beta1
       kind: Deployment
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -866,9 +866,6 @@ write_files:
           version: v1
       spec:
         replicas: 1
-        selector:
-          matchLabels:
-            app: cluster-autoscaler
         template:
           metadata:
             labels:
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -877,7 +877,7 @@ write_files:
           spec:
             hostNetwork: true
             containers:
-              - image: registry.opensource.zalan.do/teapot/aws-k8s-cluster-autoscaler:0.3
+              - image: registry.opensource.zalan.do/teapot/aws-k8s-cluster-autoscaler:0.3.2
                 name: cluster-autoscaler
                 resources:
                   limits:
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -853,6 +853,59 @@ write_files:
               securityContext:
                 privileged: true
 
+
+  - path: /srv/kubernetes/manifests/daemonsets/cluster-autoscaler.yaml
+    content: |
+      apiVersion: extensions/v1beta1
+      kind: Deployment
+      metadata:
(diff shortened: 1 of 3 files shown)


cluster/userdata-master.yaml
```diff
@@ -853,6 +853,59 @@ write_files:
...
+                    memory: 300Mi
+                command:
+                  - ./cluster-autoscaler
+                  - --v=9
+                  - --cloud-provider=aws
+                  - --skip-nodes-with-local-storage=false
+                  - --leader-elect=false
+                env:
+                  - name: AWS_REGION
+                    value: eu-central-1
```
hjacobs: hardcoded region :disappointed: --- the cluster-autoscaler should detect this automatically... (I really see no need why we should define it here)

hjacobs: We are also running at least one cluster in `eu-west-1`, so it's necessary for now :smirk:

cluster/userdata-master.yaml
```diff
@@ -853,6 +853,59 @@ write_files:
...
+                volumeMounts:
+                  - name: kubeconfig
+                    mountPath: /etc/kubernetes/worker-kubeconfig.yaml
+                    readOnly: true
```
mikkeloscar: This mount point is not used :)

cluster/userdata-master.yaml
```diff
@@ -853,6 +853,59 @@ write_files:
...
+          matchLabels:
+            app: cluster-autoscaler
+        template:
+          metadata:
+            labels:
+              app: cluster-autoscaler
+            annotations:
+              scheduler.alpha.kubernetes.io/critical-pod: ''
+          spec:
+            hostNetwork: true
```
mikkeloscar: This is only needed to get node IAM roles right? In that case I think it's better to leverage kube2iam and use a custom role to avoid "bloating" the node role with "random" roles.

cluster/userdata-master.yaml
Raffo: I'll drop it.

cluster/userdata-master.yaml
Raffo: Agreed, role created.

Raffo: it looks like there are problem with this approach. The SDK has an hard time determining the role, even though everything seems good.

````
E1104 14:41:04.846531       1 aws_cloud_provider.go:264] Error describing EC2 instances: NoCredentialProviders: no valid providers in chain
F1104 14:41:04.846555       1 cluster_autoscaler.go:167] Failed to create AWS cloud provider: NoCredentialProviders: no valid providers in chain
````

File: cluster/userdata-master.yaml [hunk 4]
```diff
@@ -853,6 +853,51 @@ write_files:
...
+          version: v1
+      spec:
+        replicas: 1
+        selector:
+          matchLabels:
+            app: cluster-autoscaler
+        template:
+          metadata:
+            labels:
+              app: cluster-autoscaler
```
mikkeloscar: The labels sections is not needed here, it will be inherited by the deployment metadata.

cluster/userdata-master.yaml
```diff
@@ -853,6 +853,51 @@ write_files:
...
+        name: cluster-autoscaler
+        namespace: kube-system
+        labels:
+          app: cluster-autoscaler
+          version: v1
+      spec:
+        replicas: 1
+        selector:
+          matchLabels:
+            app: cluster-autoscaler
```
mikkeloscar: You don't need to define the `matchLabels` because it's automatically derived from the `template.labels` below.

mikkeloscar:
:+1:

Raffo:
👍

PR #87, opened 2016-11-03
etcd fix

szuecs:
etcd needs more memory for its memory disk to store the data
etcd should only have write access to subfolders


szuecs:
:+1:

hjacobs:
:+1:

Issue #89, opened 2016-11-03
Hardcoded S3 bucket for etcd backup


hjacobs:
https://github.com/zalando-incubator/kubernetes-on-aws/blob/master/cluster/etcd-cluster.yaml#L110

PR #90, opened 2016-11-04
fix hardcoded S3 Bucket to backup ETCD to

szuecs:
Fix for #89 
Default is set in cluster.py and there is a new parameter for senza.

### DIFF
--- a/cluster/cluster.py
+++ b/cluster/cluster.py
@@ -81,9 +81,8 @@ def get_mint_bucket_name():
     s3 = boto3.resource('s3')
     parts = account_alias.split('-')
     prefix = parts[0]
-    my_session = boto3.session.Session()
-    my_region = my_session.region_name
-    bucket_name = '{}-stups-mint-{}-{}'.format(prefix, account_id, my_region)
+    region = get_region()
+    bucket_name = '{}-stups-mint-{}-{}'.format(prefix, account_id, region)
     bucket = s3.Bucket(bucket_name)
     try:
         bucket.load()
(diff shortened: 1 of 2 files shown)


cluster/cluster.py
```diff
@@ -231,6 +232,7 @@ def cli():
 @click.option('--instance-type', type=str, default='t2.micro', help='Type of instance')
 @click.option('--master-nodes', default=1, type=int, help='Number of master nodes')
 @click.option('--worker-nodes', default=1, type=int, help='Number of worker nodes')
+@click.option('--etcd_s3_bucket', default='teapot-etcd-backup', help='S3 Bucket to backup ETCD to')
```
mikkeloscar: Is `teapot-etcd-backup` the best default name? Maybe something like `k8s-etcd-backup` or `cluster-etcd-backup` is better?

hjacobs: please use `--etcd-s3-bucket` as option name to be consistent

szuecs: The problem is that a bucket has a global name, so a custom prefix makes sense to have.

hjacobs: S3 bucket names are globally unique, so a default value actually makes no sense

szuecs: done

hjacobs:
Let's make the backup work first, I would afterwards change the script to create a bucket for the account if needed so we need no option.

hjacobs:
I'll quickly do the change...

hjacobs:
:+1:

szuecs:
:+1:

PR #91, opened 2016-11-04
Use static worker node name

mikkeloscar:
Uses 'static' name for worker IAM role so it's easy to reference when
creating roles that should be delegated to nodes through kube2iam.


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #101, opened 2016-11-10
document tls termination and dns setup

linki:
fixes some open docs from https://github.bus.zalan.do/teapot/k8s-aws/issues/89

### DIFF
--- a/docs/user-guide/tls-termination.rst
+++ b/docs/user-guide/tls-termination.rst
@@ -8,44 +8,13 @@ Expose your app
 ===============
 
 Let's deploy a simple web server to test that our TLS termination works.
-Submit the following ``yaml`` files to your cluster.
 
-Create the deployment
----------------------
+Submit the following ``yaml`` files to your cluster. Note that this guide uses a
+top-down approach and starts with deploying the service first. You can, however,
+submit the files in any order you like.
 
-.. code-block:: yaml
-
-    apiVersion: extensions/v1beta1
-    kind: Deployment
-    metadata:
-      name: nginx
-    spec:
-      replicas: 2
-      template:
-        metadata:
-          labels:
-            app: nginx
-        spec:
-          containers:
-          - name: nginx
-            image: nginx
-            ports:
-            - containerPort: 80
-
-This creates a deployment called ``nginx`` that will ensure to run two copies
-of the nginx image from dockerhub listening on port 80.
-
-Make sure your pods are running.
-
-.. code-block:: bash
-
-    $ kubectl get pods
-    NAME                     READY     STATUS    RESTARTS   AGE
-    nginx-1447934386-iblb3   1/1       Running   0          7m
-    nginx-1447934386-jj559   1/1       Running   0          7m
-
-Create the service
-------------------
+Create a service
+----------------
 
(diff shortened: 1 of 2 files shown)


docs/user-guide/tls-termination.rst
```diff
@@ -0,0 +1,223 @@
...
+Make sure your pods are running.
+
+.. code-block:: bash
+
+    $ kubectl get pods
+    NAME                     READY     STATUS    RESTARTS   AGE
+    nginx-1447934386-iblb3   1/1       Running   0          7m
+    nginx-1447934386-jj559   1/1       Running   0          7m
+
+Create the service
```
mikkeloscar: As I understand, creating a service first will help kubernetes distribute the pods over several nodes http://kubernetes.io/docs/user-guide/config-best-practices/#services.

For this reason it might make sense to talk about the service first, to teach people about this approach.

docs/user-guide/tls-termination.rst
```diff
@@ -0,0 +1,223 @@
...
+      selector:
+        app: nginx
+
+This creates a logical service called ``nginx`` that forwards all traffic to any pods
+that match the label selector ``app=nginx``, which exactly maps to the two nginx replicas
+we created in the last step. The service (logically) listens on port 443 and forwards to
+port 80 on each of the upstream pods, which is where the nginx processes listen on.
+
+We also define the protocol that our upstreams use. Often your upstreams will just speak
+plain HTTP so the second annotation's value is actually its default value and can be left out.
```
mikkeloscar: > HTTP so the second annotation's value is actually **its** default value and can be left out

I would change to **the**

mikkeloscar:
Added a couple of comments, but very nice explanation of all the parts!

docs/user-guide/tls-termination.rst
linki: interesting, didn't know that. i think it makes sense to start with the service anyway to teach people a top down approach. thanks!

docs/user-guide/tls-termination.rst
linki: yeah, will do

docs/user-guide/tls-termination.rst
linki: done

docs/user-guide/tls-termination.rst
linki: changed

docs/user-guide/tls-termination.rst
```diff
@@ -8,44 +8,13 @@ Expose your app
...
-            app: nginx
-        spec:
-          containers:
-          - name: nginx
-            image: nginx
-            ports:
-            - containerPort: 80
+Submit the following ``yaml`` files to your cluster. Note that this guide uses a
+top-down approach and starts with deploying the service first. You can, however,
+submit the files in any order you like.
```
linki: done

hjacobs:
Maybe we should use a different example domain instead of `.zalan.do`?

szuecs:
@hjacobs Why? This will help most of our users and is not too complicated for all the others.

hjacobs:
:+1:

szuecs:
:+1:

PR #104, opened 2016-11-14
Feature/multi region etcd

szuecs:
Based on https://raw.githubusercontent.com/zalando/stups-etcd-cluster/multiregion-cluster/etcd-cluster-multiregion.yaml @CyberDem0n and me created a new etcd-cluster.yaml.
Known issues are that Kubernetes apiserver can not access the 2nd region, which we consider as ok.
If we really have to do a region failover we could also boostrap Kubernetes masters and worker nodes.


hjacobs:
:+1:

szuecs:
:+1:

PR #105, opened 2016-11-14
use microscope as default if you use CoreOS toolbox

szuecs:
this pulls microscope as oneshot, the goal of microsocope is to provide all tools that SREs may need to debug in most cases


linki:
👍

szuecs:
:+1:

PR #106, opened 2016-11-14
Example policy for S3 access

dryewo:


mikkeloscar:
Thanks!

What do you think about moving this part to before the Kubernetes manifest? I think it would be better to have it right after the reference to the IAM roles page.

mikkeloscar:
:+1:

dryewo:
👍

PR #110, opened 2016-11-16
Mark system pods as critical

mikkeloscar:
We want to make sure that our system pods (things running in kube-system
namespace) always gets scheduled.

This adds the annotations needed by the scheduler to determine what pods
are critical: http://kubernetes.io/docs/admin/rescheduler/#rescheduler-guaranteed-scheduling-of-critical-add-ons#marking-add-on-as-critical

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -691,6 +691,7 @@ write_files:
           ports:
           - port: 80
             targetPort: 9090
+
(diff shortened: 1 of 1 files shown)


Raffo:
👍

mikkeloscar:
:+1:

mikkeloscar:
https://github.com/kubernetes/kubernetes/blob/master/docs/proposals/rescheduling-for-critical-pods.md#interaction-with-scheduler

PR #112, opened 2016-11-16
Use same indentation for the whole file

mikkeloscar:
It annoys me that the indentation is not the same all over the place :)

This just uses the same indentation all over the place in `userdate-master.yaml`.


Raffo:
let the flame begin...

Raffo:
👍

mikkeloscar:
:+1:

PR #113, opened 2016-11-16
[wip] add appdynamics agent to cluster

linki:
adds a daemonset running the logwatcher and appdynamics machine agent. you have to provide the `appdynamics` key on the command line for create and update.

this is more of a draft to get something running. todo:
* make appd account configurable
* put secrets in a secret (although it won't be more secret because of it)

i do not like providing the appd key on each operation but in the past we decided not to introduce `if/else` templating in our cloud config. i don't see another way currently.

### DIFF
--- a/cluster/cluster.py
+++ b/cluster/cluster.py
@@ -108,7 +108,7 @@ def get_etcd_bucket_name():
     return bucket_name
 
 
-def get_cluster_variables(stack_name: str, version: str, worker_shared_secret=None):
+def get_cluster_variables(stack_name: str, version: str, appdynamics_access_key: str, worker_shared_secret=None):
     route53 = boto3.client('route53')
     all_hosted_zones = route53.list_hosted_zones()['HostedZones']
     hosted_zone = all_hosted_zones[0]['Name'].rstrip('.')
@@ -135,7 +135,8 @@ def get_cluster_variables(stack_name: str, version: str, worker_shared_secret=No
         'mint_bucket': mint_bucket,
         'etcd_bucket': etcd_bucket,
         'account_id': get_account_id(),
-        'region': get_region()
+        'region': get_region(),
+        'appdynamics_access_key': appdynamics_access_key
     }
     return variables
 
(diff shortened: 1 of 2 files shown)


cluster/userdata-master.yaml
```diff
@@ -906,6 +906,78 @@ write_files:
...
+            labels:
+              app: appdynamics
+              component: logging
+            annotations:
+              scheduler.alpha.kubernetes.io/critical-pod: ''
+              scheduler.alpha.kubernetes.io/tolerations: '[{"key":"CriticalAddonsOnly", "operator":"Exists"}]'
+          spec:
+            containers:
+            - name: log-watcher
+              image: registry.opensource.zalan.do/eagleeye/k8s-log-watcher:0.5-b2
```
mohabusama: 0.7 is the latest

cluster/userdata-master.yaml
```diff
@@ -906,6 +906,78 @@ write_files:
...
+              - name: CLUSTER_NODE_NAME
+                valueFrom:
+                  fieldRef:
+                    fieldPath: spec.nodeName
+              - name: WATCHER_KUBE_URL
+                value: http://127.0.0.1:8001
+              - name: WATCHER_DEBUG
+                value: "1"
+            - name: appdynamics-agent
+              image: pierone.stups.zalan.do/monitoring-appdynamics/appdynamics-machine-agent:0.2-b2
```
mohabusama: 0.3 is the latest

cluster/cluster.py
```diff
@@ -307,20 +309,24 @@ def same_user_data(enc1, enc2):
...
+def update(stack_name, version,  dry_run, force, instance_type, master_nodes, worker_nodes, postpone, max_worker_nodes, appdynamics_access_key):
     '''
     Update Kubernetes cluster
     '''
     existing_user_data_master = get_launch_configuration_user_data(stack_name, version, 'Master')
     existing_user_data_worker = get_launch_configuration_user_data(stack_name, version, 'Worker')
     worker_shared_secret = get_worker_shared_secret(existing_user_data_worker)
-    variables = get_cluster_variables(stack_name, version, worker_shared_secret)
+    variables = get_cluster_variables(stack_name, version, appdynamics_access_key, worker_shared_secret)
+    if dry_run:
```
mohabusama: May be a dry-run would print user_data_master and worker as well?

cluster/userdata-master.yaml
mohabusama: Actually ``0.8`` with multiple namspaces support

hjacobs:
What about worker nodes? We want to streamm app logs too, right?

hjacobs:
:+1:

mikkeloscar:
:+1:

cluster/cluster.py
mikkeloscar: Opened a separate issue for this: #123

linki: good point, thanks!

PR #118, opened 2016-11-17
only run flake8 on Travis (fixed failing builds)

hjacobs:
The [`tox-travis` plugin](https://pypi.python.org/pypi/tox-travis) automatically sets the env for tox according to Travis Python version, but we just want to run flake8 for now.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,9 @@
 [tox]
 envlist=flake8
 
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

hjacobs:
:+1:

PR #129, opened 2016-11-18
Use kubernetes API as flannel backend

mikkeloscar:
**We don't have to merge, but we can consider it if it makes sense**

Run flannel as a daemonset which uses the kubernetes API as subnet
manager instead of etcd.

Based on this: https://github.com/coreos/flannel/blob/master/Documentation/kube-flannel.yml

*note*:
Node pod CIDR allocation is left to the controller manager instead of flannel, I don't know if this could be a problem.

**Pros**
* By not depending on etcd we can secure the etcd from the worker nodes
* Less configuration on the nodes.

**Cons**
* This is still not in a stable release of flannel.
* When the node is fresh (on cluster create) the pod startup will be a bit slower because they now depend on the flannel pod to be ready to get a pod IP (In the current setup flannel is ready before docker).


hjacobs:
If it works: great way to get rid of worker -> etcd dependency.

mikkeloscar:
This is rebased on master and updated to use flannel [v0.7.0](https://github.com/coreos/flannel/releases/tag/v0.7.0) which has Kubernetes backend as experimental feature.

szuecs:
Can we try this out in our alpha channel for longer time?

PR #132, opened 2016-11-18
appdynamics needs pull secrets

linki:
forces appdynamics to use the correct secrets

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -968,6 +968,8 @@ write_files:
                 path: /var/lib/docker/containers
             - name: jobfiles
               emptyDir: {}
+            imagePullSecrets:
+            - name: pierone.stups.zalan.do
 
   - path: /etc/kubernetes/nginx/nginx.conf
     content: |


mikkeloscar:
:+1:

hjacobs:
@linki how does it work? Isn't this still prone to a race-condition? (Just trying to understand why this fix works)

mikkeloscar:
@hjacobs normally we just reference the `imagePullSecret` in the default service account. Here the race occurs because the default service account is attached to the pod before it has the `imagePullSecret`, thus the pod doesn't "inherit" any `imagePullSecret`s from the service account.

So by defining the `imagePullSecret` directly on the pod it doesn't need to inherit it from the service account, and the pod will not be scheduled before the actual `imagePullSecret` has been created.

linki:
sorry for not explaining it right way. @mikkeloscar is right, that was the original idea of the fix. turns out that it behaves differently: the pod is scheduled without `imagePullSecrets` and fails as before. but since it now has a direct reference to the `secret` kubernetes will eventually apply it to the pod. (maybe when `secretary` overwrites the `secret` every two minutes this triggers some kind of resync.)

hjacobs:
:+1:

PR #139, opened 2016-11-22
add agreed upon labels to core components

linki:
add `application` and `version` labels to core components. does `release` make sense?

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -812,7 +812,7 @@ write_files:
       metadata:
         name: prometheus-node-exporter
         labels:
-          application: prometheus
+          application: prometheus-node-exporter
           version: 0.12.0
           component: node-exporter
         namespace: kube-system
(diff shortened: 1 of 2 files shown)


cluster/userdata-master.yaml
```diff
@@ -791,15 +812,17 @@ write_files:
       metadata:
         name: prometheus-node-exporter
         labels:
-          app: prometheus
+          application: prometheus
```
mikkeloscar: maybe this should be changed to `prometheus-node-exporter`

hjacobs:
Does `k8s-app` have any special meaning which might break something (just checking..)?

hjacobs:
:+1:

linki:
@hjacobs I don't think the label "k8s-app" has any special meaning to kubernetes. double checked with kubernetes source code and there's nothing but test code using the label.

linki:
👍

mikkeloscar:
:+1:

PR #142, opened 2016-11-23
configure docker logging opts via env vars

linki:
the current command overrides the invocation of docker


szuecs:
:+1:

Raffo:
Sorry for the previous change, was actually pretty wrong. That said, the race condition could still be there, so we should investigate further.

Raffo:
👍

PR #144, opened 2016-11-23
yaml parse will convert numbers with dots to float

szuecs:
yaml parse will convert numbers with dots to float, we use prefix v as kubernetes does


szuecs:
:+1:

Raffo:
👍

PR #145, opened 2016-11-23
Feature add kube-state-metrics to check cluster state from zmon 

szuecs:
zmon needs kube-state-metrics on a predictable url

* kube-state-metrics is exposed as service type LoadBalancer for zmon
* deploy kube-state-metrics 
  * kube-state-metrics uses kube-apiserver-readonly-svc with a predictable clusterIP
  * kube-apiserver-readonly exposes 8082 to be a target of kube-apiserver-readonly-svc

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -818,10 +818,8 @@ write_files:
       apiVersion: v1
       kind: Service
       metadata:
-        name: apiserver-svc
+        name: apiserver-svc-ro
         namespace: kube-system
-        labels:
-          application: kube-apiserver-ro
       spec:
         type: ClusterIP
         ports:
@@ -854,7 +852,7 @@ write_files:
               command:
               - /kube-state-metrics
               - --port=8080
-              - --apiserver=http://apiserver-svc.kube-system.svc.cluster.local
+              - --apiserver=http://apiserver-svc-ro.kube-system.svc.cluster.local
 
   - path: /srv/kubernetes/manifests/services/kube-state-metrics.yaml
     content: |
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -342,6 +342,9 @@ write_files:
               value: stups_deployment-controller-ui=Administrator
         - name: nginx
           image: registry.opensource.zalan.do/teapot/nginx:1.11.5
+          ports:
+          - containerPort: 8082
+            hostPort: 8082
           volumeMounts:
           - name: config-volume
             mountPath: /etc/nginx
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -842,7 +842,8 @@ write_files:
         template:
           metadata:
             labels:
-              app: kube-state-metrics
+              application: kube-state-metrics
+              version: v0.3.3
           spec:
             containers:
             - image: registry.opensource.zalan.do/teapot/kube-state-metrics:v0.3.3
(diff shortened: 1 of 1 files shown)


hjacobs:
LGTM, did you test with a new cluster?

szuecs:
@hjacobs Yes cluster create works. I checked that the endpoint of the internal LB is also accessible from zmon using curl http://<internalLB>/metrics.

szuecs:
:+1:

cluster/userdata-master.yaml
```diff
@@ -810,6 +813,67 @@ write_files:
               - kubectl
               - proxy
 
+  - path: /srv/kubernetes/manifests/services/kube-apiserver-readonly-svc.yaml
+    content: |
+      apiVersion: v1
+      kind: Service
+      metadata:
+        name: apiserver-svc
```
linki: imo, this should also have `read-only` (`-ro`) in the name

cluster/userdata-master.yaml
```diff
@@ -810,6 +813,67 @@ write_files:
...
+        namespace: kube-system
+        labels:
+          application: kube-apiserver-ro
```
linki: you don't need this right? do we need to have a kio app matching this?

cluster/userdata-master.yaml
```diff
@@ -810,6 +813,67 @@ write_files:
...
+          spec:
+            containers:
+            - image: registry.opensource.zalan.do/teapot/kube-state-metrics:v0.3.3
+              name: kube-state-metrics
+              ports:
+              - containerPort: 8080
+              command:
+              - /kube-state-metrics
+              - --port=8080
+              - --apiserver=http://apiserver-svc.kube-system.svc.cluster.local
```
linki: the `kubernetes-ro` service was dropped intentionally. people should use the "real" tls enabled and protected endpoint (`kubernetes.default.svc.cluster.local`) with a service account that allows read-only. couldnt we just use this one and then drop all this custom service setup?

szuecs: A readonly endpoint is required for https://github.com/kubernetes/kube-state-metrics if we can not do this, we have to rethink the setup

cluster/userdata-master.yaml
```diff
@@ -810,6 +813,67 @@ write_files:
...
+          service.beta.kubernetes.io/aws-load-balancer-internal: 0.0.0.0/0
+      spec:
+        type: LoadBalancer
+        ports:
+        - port: 80
+          protocol: TCP
+          targetPort: 8080
+        selector:
+          application: kube-state-metrics
+          version: v0.3.3
```
linki: the other services in this file usually don't put the `version` in the selector of the service

cluster/userdata-master.yaml
```diff
@@ -810,6 +813,67 @@ write_files:
               - kubectl
               - proxy
 
+  - path: /srv/kubernetes/manifests/services/kube-apiserver-readonly-svc.yaml
+    content: |
+      apiVersion: v1
+      kind: Service
+      metadata:
+        name: apiserver-svc-ro
```
linki: why not `kubernetes-ro` to keep it similar to `kubernetes.default.svc...`?

cluster/userdata-master.yaml
```diff
@@ -810,6 +813,67 @@ write_files:
...
+        namespace: kube-system
+      spec:
+        type: ClusterIP
+        ports:
+          - port: 80
+            protocol: TCP
+            targetPort: 8082
+        selector:
+          application: kube-apiserver
+          version: v1.4.5_coreos.0
```
linki: the other services in this file usually don't put the `version` in the selector of the service

hjacobs: I agree, services should not select on `version` by default.

szuecs: fixed

szuecs:
@linki I dropped the kube-apiserver-readonly service endpoint and tested it

linki:
much simpler now :)

linki:
👍

hjacobs:
:+1:

PR #148, opened 2016-11-24
Add standard storageclass

mikkeloscar:
This adds a standard storageclass which can be used by users.

We still need to figure out what classes we really need, but for now I think this is enough.

http://kubernetes.io/docs/user-guide/persistent-volumes/#storageclasses


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #154, opened 2016-11-28
add PlatformCredentialsSet as ThirdPartyResource

hjacobs:
The PlatformCredentialsSet resource allows application owners to declare needed OAuth credentials.

```yaml
apiVersion: "zalando.org/v1"
kind: PlatformCredentialsSet
metadata:
   name: my-app-credentials
spec:
   application: my-app
   tokens:
     full-access:
       privileges:
         - foobar.write
         - acme.full
     read-only:
       privileges:
         - foobar.read
   clients:
     employee:
       # the allowed grant type, see https://tools.ietf.org/html/rfc6749
       # options: authorization-code, implicit, resource-owner-password-credentials, client-credentials
       # (values directly reference RFC section titles)
       grant: authorization-code
       realm: employees
```
The declared credentials will automatically be provided as a secret with the same name.

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -1063,6 +1063,17 @@ write_files:
       parameters:
         type: gp2
 
+  - path: /srv/kubernetes/manifests/thirdpartyresources/platform-i-a-m-credentials.yaml
+    content: |
+      apiVersion: extensions/v1beta1
+      kind: ThirdPartyResource
+      metadata:
+        # NOTE: dashes are needed to get proper PlatformIAMCredentials name
+        name: platform-i-a-m-credentials.zalando.org
+      description: "A specification to declare needed OAuth credentials (tokens, clients)"
+      versions:
+      - version: v1
+
   - path: /etc/kubernetes/nginx/nginx.conf
     content: |
       user nginx;
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -1063,14 +1063,14 @@ write_files:
       parameters:
         type: gp2
 
-  - path: /srv/kubernetes/manifests/thirdpartyresources/platform-i-a-m-credentials.yaml
+  - path: /srv/kubernetes/manifests/thirdpartyresources/platform-credentials-set.yaml
     content: |
       apiVersion: extensions/v1beta1
       kind: ThirdPartyResource
       metadata:
-        # NOTE: dashes are needed to get proper PlatformIAMCredentials name
-        name: platform-i-a-m-credentials.zalando.org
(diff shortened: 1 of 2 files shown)


cluster/userdata-master.yaml
```diff
@@ -1063,6 +1063,17 @@ write_files:
       parameters:
         type: gp2
 
+  - path: /srv/kubernetes/manifests/thirdpartyresources/platform-i-a-m-credentials.yaml
+    content: |
+      apiVersion: extensions/v1beta1
+      kind: ThirdPartyResource
+      metadata:
+        # NOTE: dashes are needed to get proper PlatformIAMCredentials name
+        name: platform-i-a-m-credentials.zalando.org
```
linki: that's a weird name. how about `iam.zalando.org` as a namespace?

linki: how about this:

```
apiVersion: iam.zalando.org/v1
kind: Credentials
```

hjacobs: My gut feeling is that we should stick with only a simple namespace like `zalando.org` as everything else opens discussions on when to open a new namespace and what granularity a namespace has to have. The name `PlatformIAMCredentials` was suggested by @sarnowski and I'm OK with it but I would be also open for alternatives --- nobody else raised any issues with the name `PlatformIAMCredentials` yet, so right now I would stay with it.

hjacobs: BTW: we might want to change it to a singular form like e.g. `PlatformIAMCredentialSet` as Kubernetes pluralizes the resource name in the HTTP path (leading to constructs like `platformiamcredentialses`).

mohabusama: +1 for `PlatformIAMCredentialSet` or `IAMCredentialSet`

hjacobs:
:+1:

Raffo:
👍

yskopets:
Hi guys!

I'm wondering how `PlatformCredentialsSet` is implemented under the hood.

Is it a Kubernetes Controller ?

Is it an open source component ?

hjacobs:
@yskopets there is an internal component "CredentialsProvider" which is not open source. The CredentialsProvider reads the CRD and injects OAuth clients/tokens into Kubernetes secrets. See https://www.slideshare.net/try_except_/large-scale-kubernetes-on-aws-at-europes-leading-online-fashion-platform-container-days-hamburg/21 for some high level view on how it interacts with our proprietary Platform IAM (OAuth provider) and the Kubernetes API.

Also a snippet from our docs for informational purposes: http://kubernetes-on-aws.readthedocs.io/en/latest/user-guide/zalando-iam.html

yskopets:
👍

PR #156, opened 2016-11-28
add documentation regarding volumes and claims

linki:
This PR describes how to use a `persistentVolumeClaim` together with a `Deployment`.

Not covered
* using `volumeClaimTemplates` with `StatefulSets`

### DIFF
--- a/docs/user-guide/using-volumes.rst
+++ b/docs/user-guide/using-volumes.rst
@@ -1,0 +1,272 @@
+==================
+Persistent Storage
+==================
+
+Some of your pods need to persist data across pod restarts (e.g. databases). In order to facilitate this we can mount
+folders into our pods that are backed by EBS volumes on AWS.
+
+Deploying PostreSQL
+===================
+
+In this example we're going to deploy a non high-available but persistent PostgreSQL container.
+
+We start out by deploying a non-persistent version first and then extend it to keep our data across pod and node
+restarts. Submit the following two manifests to your cluster to create a deployment and a service for your postgres
+instance. Note that we expose our deployment via a service to the internet by using ``type: Loadbalancer``.
+
+.. Note::
+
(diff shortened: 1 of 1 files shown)


hjacobs:
@linki maybe using PostgreSQL as an example might be confusing as we want to promote/encourage Spilo for PostgreSQL.

linki:
@hjacobs yeah, could be. I'll switch it with something else. any recommendations?

hjacobs:
@linki maybe Redis?

linki:
changed the example from postgres to redis

hjacobs:
Creating a public ELB is probably a bad idea for Redis as it does not provide real authentication.

Can we simply change it to use `ClusterIP` service and `kubectl run ...` to connect to it (also has the advantage that users don't need to install `redis-cli` locally)?

linki:
changed to using clusterIP instead of exposing the service

hjacobs:
:+1:

docs/user-guide/using-volumes.rst
```diff
@@ -0,0 +1,247 @@
...
+
+Find out where your pod currently runs:
+
+.. code-block:: bash
+
+    $ kubectl get pods -o wide
+      NAME                        READY     STATUS    RESTARTS   AGE       IP          NODE
+      redis-3548935762-qevsk      1/1       Running   0          2m        10.2.1.66   ip-172-31-15-65.eu-central-1.compute.internal
+
+The node it landed on is ``ip-172-31-15-65.eu-central-1.compute.internal``. Connect to your exposed redis endpoint and create a table:
```
mikkeloscar: > Connect to your exposed redis endpoint...

I guess this should be removed.

linki: changed it accordingly

hjacobs:
:+1:

mikkeloscar:
:+1:

PR #157, opened 2016-11-29
Multiple changes to versions

Raffo:
This does: 

- Change CoreOS back to stable
- Changes the Hyperkube version to the one that has the volume fix (self compiled)
- Add the `--debug` flag to docker


hjacobs:
:+1:

Raffo:
👍

PR #164, opened 2016-12-02
Use Docker 1.13 RC

hjacobs:
To fix "Docker hanging" issue (https://github.com/docker/docker/issues/28889).

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -19,7 +19,42 @@ coreos:
       command: start
       runtime: true
 
+    - name: install-docker.service
+      command: start
+      runtime: true
+      content: |
+        [Service]
+        Type=oneshot
+        ExecStart=/opt/bin/install-docker
+
     - name: docker.service
+      content: |
+        [Unit]
+        Description=Docker Application Container Engine
+        Documentation=http://docs.docker.com
+        After=install-docker docker.socket early-docker.target network.target
+        Requires=install-docker docker.socket early-docker.target 
+
+        [Service]
+        EnvironmentFile=-/run/flannel_docker_opts.env
+        EnvironmentFile=/etc/profile.env
+        MountFlags=slave
+        ExecStart=/opt/docker/dockerd --host=fd:// $DOCKER_OPTS $DOCKER_CGROUPS $DOCKER_OPT_BIP $DOCKER_OPT_MTU $DOCKER_OPT_IPMASQ
+        LimitNOFILE=1048576
+        # Having non-zero Limit*s causes performance problems due to accounting overhead
+        # in the kernel. We recommend using cgroups to do container-local accounting.
+        LimitNPROC=infinity
+        LimitCORE=infinity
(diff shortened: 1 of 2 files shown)


cluster/userdata-worker.yaml
```diff
@@ -102,7 +138,28 @@ coreos:
         Type=oneshot
         ExecStart=/usr/bin/docker pull registry.opensource.zalan.do/teapot/microscope:v0.0.2
 
+
+ 
 write_files:
+
+  - path: "/etc/profile.d/docker-path.sh"
+    content: |
+      PATH="/opt/docker:$PATH"
```
mikkeloscar: This file doesn't seem to be used?

cluster/userdata-worker.yaml
```diff
@@ -102,7 +138,28 @@ coreos:
...
+
+  - path: "/etc/profile.env"
+    content: |
+      PATH="/opt/docker:/usr/sbin:$PATH"
```
mikkeloscar: is `/usr/sbin` not in `$PATH`, if not, why do we need it?

cluster/userdata-master.yaml
```diff
@@ -113,6 +148,26 @@ coreos:
         ExecStart=/usr/bin/docker pull registry.opensource.zalan.do/teapot/microscope:v0.0.2
 
 write_files:
+
+  - path: "/etc/profile.d/docker-path.sh"
+    content: |
+      PATH="/opt/docker:$PATH"
```
mikkeloscar: This file doesn't seem to be used? (same as for the worker userdata).

cluster/userdata-worker.yaml
Raffo: we need it for iptables.

cluster/userdata-master.yaml
szuecs: @mikkeloscar there is a for loop on /etc/profile.d/*.sh to source all of them

cluster/userdata-worker.yaml
mikkeloscar: @szuecs Ok, but the file below that seem to do the same thing is sourced in the service with `EnvironmentFile=/etc/profile.env`, so I assume this one isn't needed.

szuecs:
:+1:

hjacobs:
:+1:

cluster/userdata-worker.yaml
```diff
@@ -19,7 +19,43 @@ coreos:
...
+        LimitCORE=infinity
+        # Uncomment TasksMax if your systemd version supports .
+        # Only systemd 226 and above support this version.
+        TasksMax=infinity
+        TimeoutStartSec=0
+        # set delegate yes so that systemd does not reset the cgroups of docker containers
+        Delegate=yes
+
+        [Install]
+        WantedBy=multi-user.targe
```
mlafeldt: `multi-user.target` 😄

hjacobs: Uups, thanks!

PR #171, opened 2016-12-05
add requests and limits based on current hackweek cluster

szuecs:
do not change the default from vendors if they set requests or limits, but make sure we have always requests and limits

added yerken to authorized_keys

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -614,8 +614,8 @@ write_files:
                   cpu: 200m
                   memory: 200Mi
                 requests:
-                  cpu: 100m
-                  memory: 100Mi
+                  cpu: 10m
+                  memory: 10Mi
               args:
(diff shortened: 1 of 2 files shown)


mikkeloscar:
@szuecs can you add a short note about how you came up with these limit values? I think that could be useful in the feature if we ever want to understand why those were chosen.

szuecs:
@mikkeloscar Sure!

I added requests as in hackweek was used and added a limit which I think could be reasonable for all pods. For example for node_exporter I did:

    kubectl get pods -n kube-system | grep exporter
    kubectl top pods -n kube-system --containers prometheus-node-exporter-5jzio --containers

I know this is not "measuring", but measuring is hard and have no production yet, so we have to first "understand" and "break" until we get realistic requests and limits.

Raffo:
Before merging, should we create a cluster and try to make "a lot of load" on it? I don't see other way to prove that they are reasonable (which doesn't still mean that they will be perfect or completely realistic).

szuecs:
Cluster create works now

szuecs:
:+1:

szuecs:
#125 make sure system relevant PODs have enough resources with requests and limits

Raffo:
👍

PR #173, opened 2016-12-05
Create ELB to health check nodes on kubelet/healthz

szuecs:
workaround dns naming issue you get for loadbalancers

### DIFF
--- a/cluster/senza-definition.yaml
+++ b/cluster/senza-definition.yaml
@@ -82,6 +82,8 @@ SenzaComponents:
          - {Ref: WorkerSecurityGroup}
       IamRoles:
          - {Ref: WorkerIAMRole}
+      ElasticLoadBalancer: WorkerLoadBalancer
+      HealthCheckGracePeriod: 480 # give worker node up to 8 min to start up
       AssociatePublicIpAddress: true
       UserData: "{{ Arguments.UserDataWorker }}"
       AutoScaling:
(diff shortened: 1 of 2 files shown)


cluster/senza-definition.yaml
```diff
@@ -92,6 +94,25 @@ SenzaComponents:
       Tags:
         - Key: "KubernetesCluster"
           Value: kubernetes
+  - WorkerLoadBalancer:
+      Type: Senza::WeightedDnsElasticLoadBalancer
+      NameSuffix: worker
+      MainDomain: ignore-main-{{SenzaInfo.StackVersion}}.teapot.zalan.do
```
hjacobs: You can remove the "WeightedDns" prefix from the `Type`.

cluster/senza-definition.yaml
```diff
@@ -92,6 +94,25 @@ SenzaComponents:
...
+      HealthCheckPort: 4194
+      Listeners:
+        - PolicyNames: []
+          Protocol: HTTP
+          InstanceProtocol: HTTP
+          InstancePort: 4194
+          LoadBalancerPort: 4194
+      Tags:
+        - Key: "Teapot"
+          Value: instance health check
```
mikkeloscar: I think we should also have the `KubernetesCluster` tag on this like we do with all the other kubernetes related resources.

szuecs: KubernetesCluster tag means "this is managed by Kubernetes".
We use it, not Kubernetes so I would not do that.

mikkeloscar: Could also mean "this is owned by the cluster". Ideally the `kubernetesCluster` tag value should be different for each cluster in an account so it's easy to identify which resource is owned by which cluster.

I know we currently don't do this because of issues with multi AZ iirc, but hopefully in the future :)

We don't have to discuss it now. If you don't want to add it for the reasons you put then I'm fine with it :)

szuecs: @mikkeloscar no this is hardcoded in Kubernetes, that is why we use it in the AWS Tags.

Raffo:
👍

szuecs:
:+1:

cluster/senza-definition.yaml
hjacobs: @mikkeloscar we are using the hardcoded value as Kubernetes assumes to own all AWS resources, but we are creating multiple clusters per subnet (at least for testing).

PR #174, opened 2016-12-06
changes from upstream stups-etcd-cluster

hjacobs:
update from upstream (https://github.com/zalando-incubator/stups-etcd-cluster/blob/master/etcd-cluster-multiregion.yaml):

* use latest Docker image
* use 4 instances (for Frankfurt)
* use same order for firewall rules


Raffo:
👍

Raffo:
👍

hjacobs:
:+1:

PR #176, opened 2016-12-06
Update webhook to use cluser_id for looking up access roles

mikkeloscar:
Now uses cluster id (`aws:{account_id}:{region}:{local_id}`) to lookup access roles in the webhook.


hjacobs:
:+1:

Raffo:
👍

PR #186, opened 2016-12-09
Added targetgroup and clusterID for ingress impl

Raffo:


Raffo:
Target group resource must be removed as this is useless.

cluster/senza-definition.yaml
```diff
@@ -250,6 +250,38 @@ Resources:
...
+                  -
+                    Ref: WorkerIAMRole
+        Version: '2012-10-17'
+      Path: /
+      Policies:
+      - PolicyDocument:
+          Statement:
+          - {Action: 'autoscaling:*', Effect: Allow, Resource: '*'}
+          - {Action: 'elasticloadbalancing:*', Effect: Allow, Resource: '*'}
+          - {Action: 'ec2:*', Effect: Allow, Resource: '*'}
```
hjacobs: do we need to write to ec2? I would limit to needed actions

cluster/senza-definition.yaml
```diff
@@ -250,6 +250,38 @@ Resources:
...
+                -
+                  - arn:aws:iam::{{ AccountInfo.AccountID }}:role/
+                  -
+                    Ref: WorkerIAMRole
+        Version: '2012-10-17'
+      Path: /
+      Policies:
+      - PolicyDocument:
+          Statement:
+          - {Action: 'autoscaling:*', Effect: Allow, Resource: '*'}
```
Raffo: No, I don't think we need to write there.

cluster/senza-definition.yaml
Raffo: No.

aryszka:
:+1:

hjacobs:
:+1:

aryszka:
:+1:

PR #193, opened 2016-12-13
Use custom clusterID for webhook

mikkeloscar:
Destinguish between clusterID which should be really unique and the ID
passed to the webhook which is used for authentication and isn't unique
in all cases.


Raffo:
👍

mikkeloscar:
:+1:

hjacobs:
:+1:

PR #194, opened 2016-12-13
create skipper ingress daemon set

aryszka:
adds the ingress skipper router to the cluster

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -1078,8 +1078,6 @@ write_files:
               - proxy
               ports:
               - containerPort: 8001
-            hostNetwork: true
-            hostPID: true
 
   - path: /srv/kubernetes/manifests/daemonsets/kube2iam.yaml
     content: |
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -1070,7 +1070,7 @@ write_files:
                   cpu: 25m
                   memory: 25Mi
             - name: kubectl
-              image: registry.opensource.zalan.do/teapot/hyperkube:v1.5.0-beta.2_custom.0
+              image: registry.opensource.zalan.do/teapot/hyperkube:v1.4.7_zalando.0
               command:
               - /hyperkube
               args:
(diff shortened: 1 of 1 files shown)


cluster/userdata-master.yaml
```diff
@@ -1028,6 +1028,59 @@ write_files:
...
+              command: ["skipper", "-application-log-level", "DEBUG", "-kubernetes-url", "http://localhost:8001", "-address", ":9999"]
+              resources:
+                limits:
+                  cpu: 200m
+                  memory: 200Mi
+                requests:
+                  cpu: 25m
+                  memory: 25Mi
+            - name: kubectl
+              image: registry.opensource.zalan.do/teapot/hyperkube:v1.5.0-beta.2_custom.0
```
hjacobs: we should use the same hyperkube version as everywhere else

mikkeloscar: It would make sense to use the same hyperkube image here as we do in all other places.

This is currently: `registry.opensource.zalan.do/teapot/hyperkube:v1.4.7_zalando.0`

aryszka: true

cluster/userdata-master.yaml
```diff
@@ -1028,6 +1028,59 @@ write_files:
...
+              image: registry.opensource.zalan.do/teapot/hyperkube:v1.4.7_zalando.0
+              command:
+              - /hyperkube
+              args:
+              - kubectl
+              - proxy
+              ports:
+              - containerPort: 8001
+            hostNetwork: true
+            hostPID: true
```
hjacobs: why do we use `hostPID`?

cluster/userdata-master.yaml
```diff
@@ -1028,6 +1028,59 @@ write_files:
...
+            - name: kubectl
+              image: registry.opensource.zalan.do/teapot/hyperkube:v1.4.7_zalando.0
+              command:
+              - /hyperkube
+              args:
+              - kubectl
+              - proxy
+              ports:
+              - containerPort: 8001
+            hostNetwork: true
```
hjacobs: `hostNetwork` means that the kubectl container port 8001 is opened on the host's "localhost"

cluster/userdata-master.yaml
aryszka: removed, thx

hjacobs:
:+1:

aryszka:
:+1:

PR #197, opened 2016-12-14
Add internal ELB for readonly API Server

mikkeloscar:
Creates internal ELB for kube-apiserver on the readonly port (8082).

Makes it possible to create ZMON checks that only run once for each
cluster instead of multiple times in case of multiple master nodes.

This also makes the internal ELB for `kube-state-metrics` somewhat obsolete since
this can now be proxied through the API server.

Proxying would work like this:

```
GET http://internal-elb/api/v1/proxy/namespaces/kube-system/services/kube-state-metrics/
```

Once ZMON is running inside the cluster and can access the in-cluster service URLs then this wouldn't be needed anymore.

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -428,6 +428,8 @@ write_files:
           volumeMounts:
           - name: config-volume
             mountPath: /etc/nginx
+          ports:
+          - containerPort: 8082
         volumes:
         - hostPath:
             path: /etc/kubernetes/ssl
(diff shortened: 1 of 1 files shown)


hjacobs:
Yes, proxying everything does not sound nice.

mikkeloscar:
I'll put an elb in front of `heapster` and leave one in front of `kube-state-metrics`.

szuecs:
:+1:

mikkeloscar:
@mohabusama is working on in cluster ZMON, which would make this PR obsolete. I'll let it sit for a bit.

PR #204, opened 2016-12-16
expose dashboard to the outside

linki:
first working version of an exposed dashboard. couple of open questions though:

* makes dashboard availabe at a specific url per cluster: e.g. check out https://kube-aws-test-linki53-dashboard.teapot.zalan.do/
  * since the goal is one cluster per account, should this rather be sth. like https://dashboard.hackweek.zalan.do?
* requires the certificate id for the exposed https endpoint
  * maybe we should infer this somehow?
* generates a random cookie secret for employeelogind
  * even on update if not provided, so everybody gets logged out. is that ok for now?
* requires a specific kio app for the dashboard
  * with platform IAM we may be able to avoid this?
* allows read and write for anybody whose authenticated
  * could be fixed by putting the dashboard under a read-only service account

this is a first full fledged working version. lot's of dependencies and a manual step (create kio app). good for hackweek i guess but really not so nice for after that.

please provide some feedback.

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -773,37 +773,8 @@ write_files:
             annotations:
               scheduler.alpha.kubernetes.io/critical-pod: ''
               scheduler.alpha.kubernetes.io/tolerations: '[{"key":"CriticalAddonsOnly", "operator":"Exists"}]'
-              iam.amazonaws.com/role: {{STACK_NAME}}-{{STACK_VERSION}}-app-kube-dashboard
           spec:
             containers:
-            - name: employeelogind
-              image: pierone.stups.zalan.do/teapot/employeelogind:0457e10
-              args:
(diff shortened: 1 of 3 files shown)


cluster/userdata-master.yaml
```diff
@@ -803,12 +838,27 @@ write_files:
...
+          zalando.org/dnsname: {{STACK_NAME}}-{{STACK_VERSION}}-dashboard.{{HOSTED_ZONE}}
       spec:
+        type: LoadBalancer
         selector:
           application: kubernetes-dashboard
         ports:
-        - port: 80
-          targetPort: 9090
+        - port: 443
+          targetPort: 3030
```
linki: that does make sense.

linki: i'll do that in a followup PR

linki:
good to go for hackweek-test, imo

PR #210, opened 2016-12-20
webhook: Allow readonly access to non-resource paths

mikkeloscar:
Allow readOnly access to non-resource paths like `/apis`.


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #211, opened 2016-12-22
preserve incoming hostname in ingress

aryszka:
set the flag to preserve the incoming host during ingress

### DIFF
--- a/cluster/senza-definition.yaml
+++ b/cluster/senza-definition.yaml
@@ -263,7 +263,7 @@ Resources:
 
   KubeAWSIngressControllerIAMRole:
     Properties:
-      RoleName: "{{SenzaInfo.StackName}}-{{SenzaInfo.StackVersion}}-app-kube-aws-ingress-controller"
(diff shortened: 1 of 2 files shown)


hjacobs:
@aryszka did you create a test cluster with this setting?

Raffo:
I will rebase master here to make sure because the changes in this PR work. The current status of the PR is not working in some setups with ingress.

hjacobs:
@Raffo what is not working atm?

Raffo:
I'm testing it after updating from master. Before, the ingress controller was not working at all. The current setup with Ingress has problems when running in setups with multiple clusters per account which affect the ease of tests, but I would address this in a separate issue and try only to focus on the feature of this PR.

Raffo:
This feature works apart from the bug found in the ingress controller. I will report and fix that separately.

Raffo:
👍

Raffo:
👍

hjacobs:
:+1:

PR #213, opened 2016-12-27
Integrate ZMON in kubernetes cluster

mohabusama:
- worker
- scheduler
- agent
- redis

Notes:
- Tested with ``t2.medium``

TBD:
- Using ``zmon-worker`` OS image, due to failure while pulling private registry image on ``kube-system``
- That lead to transferring all private env variables to the manifests, which pollutes ``userdata-master`` and complicates CLI args

For ref, the following private env variables were needed:
- {{ZMON_MAIL_HOST}}
- {{ZMON_MAIL_SENDER}}
- {{ZMON_MAIL_USER}}
- {{ZMON_MAIL_PASSWORD}}
- {{ZMON_HIPCHAT_URL}}
- {{ZMON_NOTIFICATIONS_SERVICE_KEY}}
- {{ZMON_PUSH_KEY}}
- {{ZMON_PAGERDUTY_KEY}}
- {{ZMON_OPSGENIE_KEY}}

Possible solutions:
- Workaround for private registry image pulling failure?
- Pass args to ``cluster create`` & ``cluster update`` (imho complex and error prone)
- Pass ZMON args as ``yaml`` config file.
- Separate ZMON as standalone deployment.
- ... ?

### DIFF
--- a/cluster/cluster.py
+++ b/cluster/cluster.py
@@ -131,6 +131,7 @@ def get_cluster_variables(stack_name: str, version: str, node_labels: str, worke
         'mint_bucket': mint_bucket,
         'etcd_bucket': etcd_bucket,
         'account_id': get_account_id(),
+        'account_alias': get_account_alias(),
         'region': get_region(),
         'node_labels': node_labels,
     }
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -226,7 +226,7 @@ write_files:
     owner: root:root
     content: |
       #!/bin/bash -e
-      for uid in hjacobs mkerk sszuecs rdifazio mlinkhorst mlarsen tsarnowski lmineiro aryszka apfeiffer ytussupbekov; do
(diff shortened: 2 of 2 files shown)


mikkeloscar:
To pull the private image you can add an image-pull-secret like this: https://github.com/zalando-incubator/kubernetes-on-aws/commit/448ffe5af80c4739682ae6dd654cc195040b8fea

mohabusama:
@mikkeloscar fixing with ``imagePullSecrets``

mikkeloscar:
Somewhat related to #197 we want to allow ZMON to read from the api-server of the cluster.

In #197 I proposed to add a read-only service which wouldn't require OAuth credentials. But I now think a better solution would be to use a custom zmon user (instead of secretary) and then allow this user ReadOnly access in the webhook.

The alternative would be to change #197 to just create a kubernetes service which exposes the api-server on the read-only port (Then we don't need an internal ELB).

@mohabusama @hjacobs what do you think?

hjacobs:
I added the "do-not-merge" label as we probably need to discuss some things first.

szuecs:
Within the cluster you can use the nginx read-only endpoint which proxies read-only calls to the API server

PR #217, opened 2017-01-09
This enables batch API and CronJobs

Raffo:
This is needed for CronJob, that we can use for our (internal) Od updater.


hjacobs:
Docs for reference: http://kubernetes.io/docs/user-guide/cron-jobs/

hjacobs:
:+1:

mikkeloscar:
👍

PR #222, opened 2017-01-16
Update Skipper to v0.1.1

hjacobs:
Fix Skipper returning "Internal Server Error" instead of "Service Unavailable", see https://github.com/zalando/skipper/issues/241


Raffo:
"This is fine" (cit), but just to make it clear, I discussed with @aryszka a strategy for tagging/releasing the OSS Skipper so that we don't have to hardcode versions.

Raffo:
👍

szuecs:
:+1:

PR #224, opened 2017-01-17
allow multiple flannels to use the same etcd

linki:
clusters in our AWS account share the same etcd cluster to store flannel's network config.

This leads to many unrelated nodes to fight over a subnet as well as terminated clusters to leave lots of garbage and claimed subnets. Now we hit the limit of 256 and flannel won't come up anymore.

This puts each cluster's flannel config under a separate prefix in etcd.

Note: becomes obsolete with a config map backend for flannel.

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -7,6 +7,7 @@ coreos:
   flannel:
     interface: $private_ipv4
     etcd_endpoints: http://$private_ipv4:2379
+    etcd_prefix: /{{STACK_VERSION}}/coreos.com/network
   etcd2:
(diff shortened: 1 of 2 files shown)


hjacobs:
:+1:

hjacobs:
What happens if we update a production cluster? We need to somehow migrate the data first to prevent downtimes, right?

hjacobs:
Better to be safe than sorry, so I added the "do-not-merge" flag until we know how to proceed for production clusters.

linki:
I agree that it's unnecessarily risky at the moment. I lost some connections to my nginx while updating my cluster.

Ideally there would be a way to specify that some things should only apply to new cluster's but not to already existing ones. This would be a candidate.

PR #226, opened 2017-01-17
Increased delay to consider node as unneeded

Raffo:
Straight from the CLI flags: 

````
--scale-down-unneeded-time=10m0s: How long the node should be unneeded before it is eligible for scale down
````
In GCP, we pay the first 10 minutes and then per minute. In AWS we pay per hour, so it makes no sense to drop an unused node too early (the default is 10 minutes as per CLI flag) as we might need it shortly after, thus wasting money.

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -1162,6 +1162,7 @@ write_files:
                   - --cloud-provider=aws
                   - --skip-nodes-with-local-storage=false
                   - --leader-elect=false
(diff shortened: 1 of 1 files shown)


Raffo:
I'm open for suggestions regarding the time. I put 45 mins as 60 would be not enough as we will be paying for the 2nd hour and we don't know how long it takes to consider nodes for removal. This is "kind of safe".

hjacobs:
:+1:

mikkeloscar:
:+1:

PR #228, opened 2017-01-19
Pass hostedZone from cluster.py

mikkeloscar:
This passes the hostedZone to senza from cluster.py instead of letting
senza find it. This way senza won't prompt the user for a hosted zone
selection in case there is more than one.


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #229, opened 2017-01-19
Fix e2e tests for v1.5

mikkeloscar:
Makes the e2e tests work with kubernetes 1.5 (petset -> statefulset) and increases the default timeout a bit.


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #231, opened 2017-01-23
determine nodes to update by labels

linki:
This allows passing node labels to cluster.py which are added to all nodes.

---

*This issue used to modify the rolling update behaviour to consider node labels. This is not beneficial and postponed for now. For reference, see the text below and find the commits in the history.*

**Determine nodes to update by labels**

EARLY VERSION: Pretty much done but needs some testing. I would love to get a review @mikkeloscar @hjacobs @ideahitme for any obvious mistakes.

This allows passing node labels to each node of the cluster. Node labels can then be used to determine which nodes need to be updated during a cluster update.

The rolling update process looks for nodes that either have old userdata *or* different values for the labels specified on `cluster.py update`. Initial labels can also be set on `cluster.py create`. When providing no labels it should behave as before.

It makes `cluster.py update` re-entrant so it can be called again if a cluster update was interrupted. On each run it can make some progress until the entire cluster is updated instead of going into an endless loop like now.

Note: it's still not perfectly equal to an uninterrupted run (the ASGs' desired capacity is increased with each run and might not be reduced later, but that should be fixed by the autoscaler eventually).

### DIFF
--- a/cluster/cluster.py
+++ b/cluster/cluster.py
@@ -98,7 +98,7 @@ def get_etcd_bucket_name():
     return bucket_name
 
 
-def get_cluster_variables(stack_name: str, version: str, scalyr_access_key: str, worker_shared_secret=None):
+def get_cluster_variables(stack_name: str, version: str, node_labels: str, scalyr_access_key: str, worker_shared_secret=None):
     route53 = boto3.client('route53')
     all_hosted_zones = route53.list_hosted_zones()['HostedZones']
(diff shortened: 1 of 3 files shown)


mikkeloscar:
lgtm. The only thing is if you create a cluster with two labels and then update it with one, you loose one label from the new nodes. But I wouldn't care so much about this here.

linki:
PR changed to just add the node label feature.

hjacobs:
:+1:

mikkeloscar:
:+1:

PR #234, opened 2017-01-25
#225 scale up only if at least one node needs to be updated

### LINKED ISSUE #225: cluster update --force is misleading and wrong
When updating a cluster `cluster.py` will refuse to update if the cloud-config didn't change. One can override this by providing the `--force` flag.

However, the rolling update behaviour further down the call stack will detect whether it finished updating all the nodes by comparing the new and current user-data/cloud-config. This will be true and the rolling update stops after the first node (master and worker respectively).

hjacobs:
Fixes #225 

Removes the `--force` option (it's still there for compatibility, but it does not do anything!) and fixed the check when performing the rolling update.

### DIFF
--- a/cluster/cluster.py
+++ b/cluster/cluster.py
@@ -495,6 +495,8 @@ def perform_node_updates(stack_name, version, name_filter, desired_user_data, co
             time.sleep(5)
             act.progress()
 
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

hjacobs:
:+1:

PR #236, opened 2017-01-26
Remove etcd-member from worker nodes

mikkeloscar:
**We don't have to merge, but we can consider it if it makes sense**

Run flannel as a daemonset which uses the kubernetes API as subnet
manager instead of etcd.

Based on this: https://github.com/coreos/flannel/blob/master/Documentation/kube-flannel.yml

*note*:
Node pod CIDR allocation is left to the controller manager instead of flannel, I don't know if this could be a problem.

**Pros**
* By not depending on etcd we can secure the etcd from the worker nodes
* Less configuration on the nodes.

**Cons**
* This is still an experimental feature in flannel.
* When the node is fresh (on cluster create) the pod startup will be a bit slower because they now depend on the flannel pod to be ready to get a pod IP (In the current setup flannel is ready before docker).


mikkeloscar:
I rolled this out in the playground cluster and collected the following issues:

* Since we are applying the manifests last we end up in a situation where we roll all the nodes without flannel, and then apply the flannel daemonset in the end.
* After the flannel daemonset was running most things came back except one flannel pod, which caused kube2iam and the logging-agent on the same node to also not come up. But restarting the flannel pod fixed this problem.
Secretary never came up by itself (gerry was trying to get AWS credentials forever). I killed the secretary pod and it was scheduled to the same node, and that time it worked.

szuecs:
Sounds that you have to restart the containers to get a renew of the network adapter/configuration, to get it up.

Do you think we should treat flannel daemonset special, or do you want to try again different steps to roll it out?

mikkeloscar:
Updating the playground cluster I saw another issue where flannel on one of the nodes failed to start:

```
E0207 17:22:12.107965       1 main.go:127] Failed to create SubnetManager: error retrieving pod spec for 'kube-system/kube-flannel-k24vb': Get https://10.3.0.1:443/api/v1/namespaces/kube-system/pods/kube-flannel-k24vb: dial tcp 10.3.0.1:443: i/o timeout
```

This cause kube2iam to also fail:

```
time="2017-02-07T17:24:04Z" level=fatal msg="route ip+net: no such network interface"
```

and anything depending on kube2iam will then also fail.


I suspect this is an issue only on master nodes where the flannel pod tries to read from the API server before it has setup the pod network..

linki:
@mikkeloscar rebased it on the latest `dev`.

linki:
@mikkeloscar rebased it on the latest dev again. This now only contains the final piece of the puzzle: removing etcd-proxy from the workers.

linki:
TODO: also change security groups

linki:
I guess our etcd backup job requires access to etcd from the workers as well 😅

mikkeloscar:
> I guess our etcd backup job requires access to etcd from the workers as well 

Quick fix would be to run it on the masters..


Another thing is etcd metrics which we should be able to get from the workers (from ZMON). It's possible to expose metrics on a different port in etcd, but we need to configure this.

linki:
I think this is good to go now? @mikkeloscar @aermakov-zalando

aermakov-zalando:
LGTM, but let's run the tests just in case.

aermakov-zalando:
👍

linki:
👍

PR #237, opened 2017-01-27
ELB-unhealthy nodes should not be terminated

hjacobs:
Do not use ELB health check for ASGs: http://docs.stups.io/en/latest/components/senza.html#senza-taupageautoscalinggroup

This prevents nodes from getting terminated on ELB health check failure.


mikkeloscar:
:+1:

hjacobs:
:+1:

PR #240, opened 2017-01-30
make kubectl apply work with daemonsets

linki:
This deploys Mirantis' `daemonupgradecontroller` controller and annotates daemonsets to opt-in to being managed by it.

It looks for outdated pods of a daemon set (via `podTemplateHash` annotations) and kills them one by one.

Note: This may not be enough to have a smooth upgrade for skipper.

### DIFF
--- a/cluster/manifests/daemonupgrader/deployment.yaml
+++ b/cluster/manifests/daemonupgrader/deployment.yaml
@@ -1,0 +1,29 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: daemonupgrader
+  namespace: kube-system
+  labels:
+    application: daemonupgrader
+    version: 8264ca0
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      application: daemonupgrader
+  template:
+    metadata:
+      labels:
+        application: daemonupgrader
+        version: 8264ca0
+    spec:
+      containers:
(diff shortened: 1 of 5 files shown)


hjacobs:
@linki it looks like we have no internal CI/CD pipeline for the daemonupgrader Docker image (?), we should set this up first.

linki:
@hjacobs pipeline setup, all green.

hjacobs:
:+1:

cluster/manifests/daemonupgrader/deployment.yaml
```diff
@@ -0,0 +1,29 @@
...
+      application: daemonupgrader
+  template:
+    metadata:
+      labels:
+        application: daemonupgrader
+        version: 8264ca0
+    spec:
+      containers:
+      - name: daemonupgrader
+        image: pierone.stups.zalan.do/teapot/k8s-daemonupgradecontroller:8264ca0
```
mikkeloscar: The image should be pulled from the opensource registry, otherwise it will fail to launch when setting up a new cluster.

hjacobs: Yes, I agree

linki: done

hjacobs:
:+1:

Raffo:
👍

PR #242, opened 2017-01-31
remove hardcoded s3:// prefix

szuecs:
gerry defaults to s3://, so we do not have to change anything else and we easily support other cloud providers.


szuecs:
:+1:

mikkeloscar:
:+1:

linki:
@mikkeloscar @szuecs this should work but I currently facing problems where it complaints about missing region when I remove the prefix. Let's keep it in mind and check secretary in the next cluster update.

PR #243, opened 2017-01-31
Pre-generate SSH authorized_keys

hjacobs:
To avoid a runtime dependency to Even. Removed some non-authorized users and added 24x7 support users.


tuxlife:
:+1:

hjacobs:
Still testing, but seems to work :smile:

hjacobs:
:+1:

PR #244, opened 2017-01-31
Remove hosted zone from senza definition and add NodePool tag to ASGs

mikkeloscar:
These are the changes required for getting rid of cluster.py in the CLM.

This removes the need for constructing the apiserver url from the hosted zone and instead passes the apiserver url directly to the senza definition.

It also introduces `NodePool` tags on the autoscaling groups (master and worker) so it's possible to lookup the ASG by the tagged node pool name. Currently the node pool names are just hardcoded in cluster.py so it's still compatible. I can also pass these values in flags if you prefer.

### DIFF
--- a/cluster/senza-definition.yaml
+++ b/cluster/senza-definition.yaml
@@ -22,8 +22,8 @@ SenzaInfo:
           Description: "Type of instance"
       - ClusterID:
           Description: "ID of the cluster"
-      - APIServerHostName:
-          Description: "API Server host name"
+      - HostedZone:
+          Description: "Hosted DNS zone"
       - MasterNodePoolName:
           Description: "Name of the master node pool (ASG)"
       - WorkerNodePoolName:
@@ -51,8 +51,8 @@ SenzaComponents:
           LoadBalancerPort: 443
       ConnectionSettings:
         IdleTimeout: 300
-      VersionDomain: "{{ Arguments.APIServerHostName }}"
-      MainDomain: "{{ Arguments.APIServerHostName }}"
+      VersionDomain: "{{SenzaInfo.StackName}}-{{SenzaInfo.StackVersion}}.{{Arguments.HostedZone}}"
+      MainDomain: "{{SenzaInfo.StackName}}.{{Arguments.HostedZone}}"
       Tags:
         - Key: "KubernetesCluster"
           Value: kubernetes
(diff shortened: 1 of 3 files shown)


linki:
dependency of https://github.bus.zalan.do/teapot/cluster-lifecycle-manager/pull/69

cluster/senza-definition.yaml
```diff
@@ -47,8 +51,7 @@ SenzaComponents:
           LoadBalancerPort: 443
       ConnectionSettings:
         IdleTimeout: 300
-      VersionDomain: "{{SenzaInfo.StackName}}-{{SenzaInfo.StackVersion}}.{{Arguments.HostedZone}}"
-      MainDomain: "{{SenzaInfo.StackName}}.{{Arguments.HostedZone}}"
+      VersionDomain: "{{ Arguments.APIServerHostName }}"
```
hjacobs: I think we have to define `MainDomain` here too, otherwise Senza will ask the user what domain to choose in case of multiple hosted zones!

hjacobs:
@mikkeloscar can you test whether `VersionDomain` == `MainDomain` actually works?

mikkeloscar:
@hjacobs yes, I just did and it doesn't

```
RRSet with DNS name kube-aws-test-1.teapot.zalan.do., type A, SetIdentifier kube-aws-test-1 cannot be created as a non-weighted set exists with the same name and type.
```

I'm not sure how to do it then. It seems wrong to construct a URL without the senza stack version since it isn't useful in our setup.

We could add it like `main-{{ Arguments.APIServerHostName }}` to make senza happy.

hjacobs:
@mikkeloscar we can simply configure the `Domains` property manually and use `ElasticLoadBalancer` instead of `WeightedDnsLoadBalancer`: https://github.com/zalando-stups/senza/blob/master/senza/components/weighted_dns_elastic_load_balancer.py

hjacobs:
Or alternatively configure the DNS record as a custom Cloud Formation resource below `Resources` (similar to what you did with the RDS database).

mikkeloscar:
Good idea, but then we need to specify the hosted zone again, which is otherwise automatically identified by senza :)

szuecs:
:+1:

mikkeloscar:
I reverted the change to use APIServerURL instead of HostedZone since the hosted zone is still needed. We should think about how to do it right in a separate issue.

mikkeloscar:
:+1:

PR #245, opened 2017-01-31
revert s3 prefix

linki:
no idea why it ends up here: https://github.bus.zalan.do/teapot/gerry/blob/master/downloaders/downloaders.go#L47-L49

but it does when removing the prefix.


mikkeloscar:
:+1:

hjacobs:
:+1:

linki:
and here's the fix: https://github.bus.zalan.do/teapot/gerry/pull/30

it's basically sugar, specifying the `--region` like you normally have to would have fixed it as well.

PR #247, opened 2017-02-01
have netstat installed

szuecs:
found that I missed netstat which should be clearly installed, so bumped version of microscope


hjacobs:
:+1:

szuecs:
:+1:

Issue #239, opened 2017-01-28
Configure default container limits/requests


hjacobs:
This is helpful to enforce reasonable limits/requests as not every user defines resources (we don't need to set min/max).

https://github.com/kubernetes/community/blob/master/contributors/design-proposals/admission_control_limit_range.md

We should add a `limits.yaml` into every cluster which defines default resource requests like e.g. 200Mi memory.

Raffo:
This is a proposal for an admission controller, is there also an implementation in kubernetes already?

hjacobs:
Yes, it's already implemented!

PR #248, opened 2017-02-02
#239 define default requests for "default" namespace

hjacobs:
Define some reasonable defaults. This will not limit users, but it makes sure that all pods have at least 100m CPU and 100Mi memory requested (important for scheduler) and also some sane memory limit (1Gi). All values can be overwritten by users in the pod spec.


Raffo:
👍

hjacobs:
:+1:

hjacobs:
:+1:

hjacobs:
:+1:

hjacobs:
:+1:

PR #250, opened 2017-02-03
senza shouldn't fail if resources already exist

linki:
cluster.py crashes when creating the stack if it already exists. also done for delete.

https://github.bus.zalan.do/teapot/cluster-lifecycle-manager/issues/74
https://github.com/zalando-stups/senza/pull/434

### DIFF
--- a/cluster/cluster.py
+++ b/cluster/cluster.py
@@ -217,7 +217,9 @@ def deploy_etcd_cluster(hosted_zone, etcd_bucket, region):
             bucket.create(CreateBucketConfiguration={'LocationConstraint': region})
             bucket.wait_until_exists()
 
-    subprocess.check_call(['senza', 'create', 'etcd-cluster.yaml', 'etcd', 'HostedZone={}'.format(hosted_zone), 'EtcdS3Backup={}'.format(etcd_bucket)])
+    subprocess.check_call(['senza', 'create', '--update-if-exists', 'etcd-cluster.yaml', 'etcd',
+                           'HostedZone={}'.format(hosted_zone),
+                           'EtcdS3Backup={}'.format(etcd_bucket)])
     # wait up to 15m for stack to be created
     subprocess.check_call(['senza', 'wait', '--timeout=900', 'etcd-cluster', 'etcd'])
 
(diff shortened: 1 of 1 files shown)


cluster/cluster.py
```diff
@@ -217,7 +217,7 @@ def deploy_etcd_cluster(hosted_zone, etcd_bucket, region):
             bucket.create(CreateBucketConfiguration={'LocationConstraint': region})
             bucket.wait_until_exists()
 
-    subprocess.check_call(['senza', 'create', 'etcd-cluster.yaml', 'etcd', 'HostedZone={}'.format(hosted_zone), 'EtcdS3Backup={}'.format(etcd_bucket)])
+    subprocess.check_call(['senza', 'create', '--update-if-exists', 'etcd-cluster.yaml', 'etcd', 'HostedZone={}'.format(hosted_zone), 'EtcdS3Backup={}'.format(etcd_bucket)])
```
linki: is it save to update the etcd cluster?

linki:
will do the same for CLM go if ackknowledged

cluster/cluster.py
```diff
@@ -217,7 +217,9 @@ def deploy_etcd_cluster(hosted_zone, etcd_bucket, region):
             bucket.create(CreateBucketConfiguration={'LocationConstraint': region})
             bucket.wait_until_exists()
 
-    subprocess.check_call(['senza', 'create', 'etcd-cluster.yaml', 'etcd', 'HostedZone={}'.format(hosted_zone), 'EtcdS3Backup={}'.format(etcd_bucket)])
+    subprocess.check_call(['senza', 'create', '--update-if-exists', 'etcd-cluster.yaml', 'etcd',
```
linki: is it safe to update the etcd cluster?

hjacobs: yes, I think so

hjacobs:
:+1:

cluster/cluster.py
linki: I think it doesn't matter anyway as `has_etcd_cluster()` has to be `false` to trigger it.

linki:
👍

PR #254, opened 2017-02-06
kube-aws-autoscaler

hjacobs:
Replace the contrib/cluster-autoscaler with [kube-aws-autoscaler](https://github.com/hjacobs/kube-aws-autoscaler). This is not necessarily the final solution, but it improves the current situation:

* we are already running `kube-aws-autoscaler` (deployed manually to `default` namespace) in all clusters, but it currently "competes" with `contrib/cluster-autoscaler`, i.e. it does most of the scaling, but at some point `contrib/cluster-autoscaler` might decide to scale down (as it does not incorporate any buffer) and  `kube-aws-autoscaler` will scale up one minute later again
* it applies a 10% resource buffer, i.e. it ensures that always at least 10% of the requested resources are free --- this prevents deployments from waiting on cluster scaling (the situation we observed before with  `contrib/cluster-autoscaler`)
* it quickly scales down (but at most one node at a time) which helps "recovering" from our CLM node update loop (scaling up by one instance every time)
* it's completely "deterministic" as it simply calculates the required number of nodes from the given resource requests, see https://github.com/hjacobs/kube-aws-autoscaler#how-it-works for details on the algorithm

### DIFF
--- a/cluster/manifests/cluster-autoscaler/deployment.yaml
+++ b/cluster/manifests/cluster-autoscaler/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: cluster-autoscaler
-    version: v0.3.2
+    version: v0.2
 spec:
   replicas: 1
   selector:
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

hjacobs:
:+1:

ideahitme:
does it respect nodes with local storage attached/with pods marked critical/makes sure that the pods from scaled down node can be redistributed in other nodes ?

hjacobs:
@ideahitme it does not care about individual nodes/instances as we use AWS Auto Scaling Groups, so we have no direct control over which EC2 instance is actually terminated on scale down.

ideahitme:
@hjacobs I see, it seems k8s-autoscaler uses https://docs.aws.amazon.com/AutoScaling/latest/APIReference/API_TerminateInstanceInAutoScalingGroup.html to terminate specific instance and adjust asg size. Might be something we should consider

hjacobs:
@ideahitme I created an issue for this: https://github.com/hjacobs/kube-aws-autoscaler/issues/4

In theory all nodes are equal, so while we can make a "smarter" decision for the node selection this is not a hard requirement for me.

Issue #167, opened 2016-12-05
Remove custom Docker 1.13 as soon as it's included in CoreOS


hjacobs:
Follow-up to #164.

We installed Docker 1.13 RC to fix our "docker hangs" (https://github.com/docker/docker/issues/28889) problem. We should get rid of our custom Docker install as soon as 1.13 is released and CoreOS includes it.

hjacobs:
We switched back to Docker 1.12.6 in CoreOS stable which also includes the fix for "docker ps" hang.

PR #257, opened 2017-02-11
Kube node drainer

### LINKED ISSUE #133: Check out kube-node-drainer
https://github.com/coreos/kube-aws/blob/5232aa225911641f9dd48bab571850c32718b0c5/config/templates/cloud-config-worker#L149-L176

is a systemd unit that runs on each worker and "drain[s] this k8s node to make running pods time to gracefully shut down before stopping kubelet".

looks like it moves draining from the client to the server side. therefore reducing client complexity and might also work nicely with node autoscaling and spot instances.

hjacobs:
Fixes #133 and #167.

* call `kubectl drain` before stopping Docker daemon in shutdown sequence
* use Docker 1.13 included in CoreOS alpha

Uses snippet from `kube-aws` for node draining: https://github.com/coreos/kube-aws/blob/9c77168c060e6584f871973f8a1a287f90137264/config/templates/cloud-config-worker#L164

I haven't found issues yet regarding Docker 1.13 and `iptables` as it was mentioned here: https://twitter.com/jbeda/status/826969113801093121

### DIFF
--- a/cluster/userdata-worker.yaml
+++ b/cluster/userdata-worker.yaml
@@ -36,10 +36,10 @@ coreos:
             After=flanneld.service
             [Service]
             EnvironmentFile=/etc/kubernetes/cni/docker_opts_cni.env
-        - name: 50-logging.conf
+        - name: 60-dockeropts.conf
           content: |
             [Service]
-            Environment="DOCKER_OPTS=--log-opt=max-file=2 --log-opt=max-size=50m"
+            Environment="DOCKER_OPTS=--iptables=false --log-opt=max-file=2 --log-opt=max-size=50m"
 
     - name: kubelet.service
       command: start
--- a/cluster/userdata-worker.yaml
+++ b/cluster/userdata-worker.yaml
@@ -81,7 +81,9 @@ coreos:
         --tls-private-key-file=/etc/kubernetes/ssl/worker-key.pem \
         --cloud-provider=aws \
         --low-diskspace-threshold-mb=1000 \
-        --feature-gates=AllAlpha=true
+        --feature-gates=AllAlpha=true \
+        --system-reserved=cpu=100m,memory=164Mi \
+        --kube-reserved=cpu=100m,memory=282Mi
         Restart=always
         RestartSec=10
         [Install]
--- a/cluster/userdata-worker.yaml
+++ b/cluster/userdata-worker.yaml
@@ -94,13 +94,14 @@ coreos:
       content: |
         [Unit]
         Description=drain this k8s node to make running pods time to gracefully shut down before stopping kubelet
-        After=multi-user.target
+        Requires=docker.service
+        After=docker.service
 
         [Service]
         Type=oneshot
         RemainAfterExit=true
         ExecStart=/bin/true
-        TimeoutStopSec=30s
+        TimeoutStopSec=40s
         ExecStop=/bin/sh -c '/usr/bin/rkt run \
         --volume=kube,kind=host,source=/etc/kubernetes,readOnly=true \
         --mount=volume=kube,target=/etc/kubernetes \
--- a/cluster/userdata-worker.yaml
+++ b/cluster/userdata-worker.yaml
@@ -27,43 +27,7 @@ coreos:
       command: start
       runtime: true
 
-
-    - name: install-docker.service
-      command: start
-      runtime: true
-      content: |
-        [Service]
-        Type=oneshot
-        ExecStart=/opt/bin/install-docker
-
     - name: docker.service
-      content: |
-        [Unit]
-        Description=Docker Application Container Engine
-        Documentation=http://docs.docker.com
-        After=install-docker docker.socket early-docker.target network.target
-        Requires=install-docker docker.socket early-docker.target
-
-        [Service]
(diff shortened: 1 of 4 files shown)


hjacobs:
I documented the "flapping" problem in https://github.com/hjacobs/kube-aws-autoscaler/issues/7

hjacobs:
Scaling is now much better and there are less service disruptions. Here the graphs for number of service endpoints and number of worker nodes (ready/cordoned) for the scenario "scale up nginx replicas from 10 to 100 and back down to 10 again":

screenshot_2017-02-11_22-19-20

The bottom blue line shows the cordoning of nodes (by the `kube-node-drainer.service`).

hjacobs:
This PR now also reserves CPU/memory for OS and Kubernetes (fixes #258).

hjacobs:
If you wonder why this PR contains so many seemingly unrelated changes:

* I could not get the right `systemd` order (during shutdown) for `kube-node-drainer` to work with our manually installed Docker 1.13 so I switched to official Container Linux alpha with 1.13
* I updated the `kube-aws-autoscaler` to fix the problem with nodes marked unschedulable (which `kubectl drain` does)
* I added the kubelet resource reservations as I updated the autoscaler anyway (and the change to `allocatable` is in there)

cluster/userdata-worker.yaml
```diff
@@ -117,12 +81,44 @@ coreos:
...
+        [Unit]
+        Description=drain this k8s node to make running pods time to gracefully shut down before stopping kubelet
+        Requires=docker.service
+        After=docker.service
+
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        ExecStart=/bin/true
+        TimeoutStopSec=40s
```
mikkeloscar: I guess this will limit the `terminationGracePeriodSeconds` to <40. We should document this limit and evaluate if this is a reasonable limit for our users.

hjacobs: Actually it will limit the grace period to "far" less as running `kubectl` via `rkt` takes some seconds to start and it first cordons the node.

hjacobs: @mikkeloscar increased to 120s --- default is actually 90s from what I can see:

```
/etc/systemd/system.conf:34:#DefaultTimeoutStartSec=90s

hjacobs:
Contents of this PR are now running in three different non-production clusters. I think we should get it merged @mikkeloscar @linki @Raffo

Raffo:
👍

szuecs:
@hjacobs can we have docker options --iptables=false as stated by Thokin in your twitter link?

linki:
that looks very nice. But I don't get who is actually calling the systemd unit?

Raffo:
@linki see here: https://github.com/zalando-incubator/kubernetes-on-aws/pull/257/files#diff-6b98c9fd8e92a9c18c299fad2a1ea0d8R107 

Maybe we can add a comment above so that we can remember about it easily in the future.

linki:
Thanks @Raffo. Didn't see it.

linki:
👍

szuecs:
:-1:

szuecs:
I need first an answer for

@hjacobs can we have docker options --iptables=false as stated by Thokin in your twitter link?

linki:
Does anyone know which iptables docker creates with that flag?

for reference: gke also disables this, kube-aws seems not to.

hjacobs:
@szuecs @linki that's why I added my comment above about not seeing any problems with Docker 1.13, I also do not see any `DROP` rule in `iptables-save`. So I would not add the `--iptables=false` option just because we saw it on Twitter. But checking it once (enabling it and checking `iptables-save` output) makes sense.

szuecs:
Docker docs

> Docker will never make changes to your system iptables rules if you set --iptables=false when the 
 daemon starts. Otherwise the Docker server will append forwarding rules to the DOCKER filter chain. Docker will not delete or modify any pre-existing rules from the DOCKER filter chain. This allows the user to create in advance any rules required to further restrict access to the containers.

While docker is using docker0 bridge and it will maintain rules for it, we do not need them. It's just additional lines in iptables and an additional source of complexity if you look at the output of iptables-save and another process which is changing iptables rules.

Kubernetes is using cni0 bridge and kube-proxy is maintaining all the rules to implement POD connectivity or service iptables based LB.

In my opinion it makes just sense to drop Docker iptables and set --iptables=false.

mikkeloscar:
> In my opinion it makes just sense to drop Docker iptables and set --iptables=false.

Agree

Raffo:
I agree with the above as long as we test it enough.

szuecs:
I changed the userdata to set dockeropts in one place, because you can not merge ENV variables.
I tested the change on 1 master and 1 worker:

1. changed the config file
1. systemctl daemon-reload
1. systemctl restart docker
1. journalctl -xfe -u docker
1. docker ps

For me it looks good, but someone should start a new cluster and test if all PODs are starting, besides node_exporter, which I try to fix next.

szuecs:
:+1:

mikkeloscar:
Added node drainer to master userdata. Will test that it actually works...

szuecs:
:+1:

Raffo:
👍

PR #261, opened 2017-02-13
Make kube-state-metrics service type clusterIP

mikkeloscar:
Since we have ZMON in the clusters there is no longer a need for exposing `kube-state-metrics` through an internal elb.

This changes the service to type `ClusterIP`.

This will cause problems on update because you can't change a `LoadBalancer` service to `ClusterIP` :X I do think however that it would be feasable to go to every cluster and delete the service manually for this update unless someone has a better proposal?

### DIFF
--- a/cluster/manifests/kube-state-metrics/service.yaml
+++ b/cluster/manifests/kube-state-metrics/service.yaml
@@ -3,10 +3,8 @@ kind: Service
 metadata:
   name: kube-state-metrics
   namespace: kube-system
-  annotations:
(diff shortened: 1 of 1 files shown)


Raffo:
👍

hjacobs:
@mikkeloscar the update will fail, but it will at least work still with `type=LoadBalancer` (as it also makes it available with cluster IP). So it's more an annoyance than a breaking change.

hjacobs:
:+1:

hjacobs:
Just checked that our internal ZMON check already uses the service. Looks good!

PR #262, opened 2017-02-14
Configure SQL user and pass in ZMON worker

mikkeloscar:
Configures sql user and pass for ZMON worker.

Things to note:

* Since the values are stored in a secret it will not update the pods when the secret is updated
* Being stored in a secret also means we have to base64 encode the password before encrypting it and storing it in the cluster-registry.

### DIFF
--- a/cluster/manifests/zmon-worker/deployment.yaml
+++ b/cluster/manifests/zmon-worker/deployment.yaml
@@ -44,6 +44,16 @@ spec:
               value: zmon:queue:default/16
             - name: WORKER_REDIS_SERVERS
               value: zmon-redis:6379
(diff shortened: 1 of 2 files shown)


mohabusama:
So, these creds will/should be used by all DB instances (RDS, Spilo, etc...) to provide access to cluster zmon-workers, correct?

szuecs:
:+1:

hjacobs:
:+1:

hjacobs:
@mikkeloscar please add the values to all clusters in our Cluster Registry before we merge.

mikkeloscar:
Added the keys to all clusters.

PR #263, opened 2017-02-14
securityContext missing for node_exporter

szuecs:
node exporter needs to use the privileged psp, otherwise it has no access to hostNetwork


Raffo:
👍

szuecs:
:+1:

PR #264, opened 2017-02-15
updated webhook to v.0.1.2

Raffo:


szuecs:
:+1:

Raffo:
👍

hjacobs:
Would be nice to always state in the PR what the version update actually contains to answer the question "why do we need it?".

PR #266, opened 2017-02-15
Add base64 encode function

mikkeloscar:
Base64 encode secret values in templates.

This depend on this feature being merged into the CLM.

### DIFF
--- a/cluster/manifests/zmon-worker/secret.yaml
+++ b/cluster/manifests/zmon-worker/secret.yaml
@@ -5,5 +5,5 @@ metadata:
   namespace: kube-system
 type: Opaque
 data:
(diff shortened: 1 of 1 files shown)


Raffo:
I'm against this, I don't like having logic in templates (as I was against the other function with use for the infrastructure ID) and as long as it is documented it should not be necessary.

hjacobs:
@Raffo I see your concerns about logic in templates, but encoding is not really logic, just necessary transformation. We cannot do this in a clean way in CLM as we don't know where `config_items` are really used in templates (secrets need base64, but using it in env vars not).

I'm for adding the base64 "filter" (that's what this would be called in Jinja2).

hjacobs:
:+1:

szuecs:
:+1:

PR #268, opened 2017-02-16
drop second runtime-config line which seem to be merged in again

szuecs:
Somehow there got an old, second runtime-config into apiserver flag


mikkeloscar:
:+1:

szuecs:
:+1:

PR #275, opened 2017-02-19
Stability fixes: memory limit, job cleanup

hjacobs:
* autoscaler had not enough memory for clusters with large number of pods (such as pods not being cleaned up :smirk: )
* job cleanup now deletes pods explicitly and also deletes failed pods: https://github.com/hjacobs/kube-job-cleaner/releases/tag/0.2


szuecs:
:+1:

mikkeloscar:
:+1:

PR #277, opened 2017-02-20
update kube-dns to the latest version

linki:
updating kube-dns based on https://github.com/coreos/kube-aws/blob/b1b4c38ed9c1099abe46396920c61d4f95cf3409/core/controlplane/config/templates/cloud-config-controller#L1048-L1249

comes with a newer version, prometheus metrics, and cluster-proportional autoscaling.


cluster/manifests/kube-dns/deployment.yaml
```diff
@@ -72,27 +96,86 @@ spec:
...
         ports:
         - containerPort: 8080
           protocol: TCP
       dnsPolicy: Default
+
+---
+
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
```
mikkeloscar: I think this should be in a separate file to follow our manifest directory layout.

linki: done

cluster/manifests/kube-dns/depl-kube-dns.yaml
```diff
@@ -0,0 +1,143 @@
...
+    kubernetes.io/cluster-service: "true"
+spec:
+  # replicas: not specified here:
+  # 1. In order to make Addon Manager do not reconcile this replicas parameter.
+  # 2. Default is 1.
+  # 3. Will be tuned in real time if DNS horizontal auto-scaling is turned on.
+  strategy:
+    rollingUpdate:
+      maxSurge: 10%
+      maxUnavailable: 0
```
mikkeloscar: I guess this will be a problem if there is only one instance of `kube-dns` when we update the nodes of a cluster.

linki: why, because of drain?

szuecs: As far as I understand the kube-dns-autoscaler is set to min=2, so there should not be only 1 POD.

mikkeloscar: > why, because of drain?

Yes, I mean if we only have one running at the time, then we will always be unavailable for a little while, when updating the nodes.

mikkeloscar: > As far as I understand the kube-dns-autoscaler is set to min=2, so there should not be only 1 POD.

I missed that, then it's no problem.

cluster/manifests/kube-dns/depl-kube-dns.yaml
```diff
@@ -0,0 +1,143 @@
...
+            memory: 170Mi
+          requests:
+            cpu: 100m
+            memory: 70Mi
+        livenessProbe:
+          httpGet:
+            path: /healthz-kubedns
+            port: 8080
+            scheme: HTTP
+          initialDelaySeconds: 60
```
szuecs: 60s is this not a bit too long?

cluster/manifests/kube-dns/depl-kube-dns.yaml
```diff
@@ -0,0 +1,143 @@
...
+          name: metrics
+          protocol: TCP
+      - name: dnsmasq
+        image: gcr.io/google_containers/kube-dnsmasq-amd64:1.4
+        livenessProbe:
+          httpGet:
+            path: /healthz-dnsmasq
+            port: 8080
+            scheme: HTTP
+          initialDelaySeconds: 60
```
szuecs: same here: too long?

cluster/manifests/kube-dns/depl-kube-dns.yaml
```diff
@@ -0,0 +1,143 @@
...
+            cpu: 150m
+            memory: 10Mi
+      - name: dnsmasq-metrics
+        image: gcr.io/google_containers/dnsmasq-metrics-amd64:1.0
+        livenessProbe:
+          httpGet:
+            path: /metrics
+            port: 10054
+            scheme: HTTP
+          initialDelaySeconds: 60
```
szuecs: same here: too long?

Raffo:
@linki can you answer the comments so that we get this merged?

cluster/manifests/kube-dns/depl-kube-dns.yaml
linki: so fine then

cluster/manifests/kube-dns/depl-kube-dns.yaml
linki: It comes straight from upstream. In general I would trust it. The readiness probe is much faster.

hjacobs:
@linki can you summarize the changes? What does the new `-metrics` container provide? What other changes might be relevant for us?

hjacobs:
@linki sorry, did not see your last sentence which is probably already a summary.

linki:
@hjacobs yes it is. I think it will be a valuable change. Says it now also supports zero-downtime kubedns updates but I don't know if we had ever problems with this. Can't be worse, though.

cluster/manifests/kube-dns/depl-kube-dns.yaml
linki: It has been like this before: https://github.com/zalando-incubator/kubernetes-on-aws/commit/0a3740c2645b7b616b9711dc66b536c56590287a#diff-1e798ff8608657e19108b3c2e0d27db0R40

linki:
@Raffo done

hjacobs:
:+1:

szuecs:
👍

PR #279, opened 2017-02-20
#278 set reasonable max CPU/memory

### LINKED ISSUE #278: Define max in default LimitRange
To prevent something like this (mixed up "m" and "Mi" in pod spec):
screenshot_2017-02-20_20-10-25

hjacobs:
Set maximum resource requests/limits to "reasonable" values. This prevents users from accidentally requesting more than they need e.g. by mixing up units ("m" vs "Mi").

Fixes #278.


hjacobs:
:+1:

szuecs:
:+1:

PR #281, opened 2017-02-21
Rollback docker networking

Raffo:
This is a fix to the docker issues we are experiencing. In the future we can think of reintroducing the flag as long as the related issues are fixed.


hjacobs:
So actually following recommendations from jbeda on Twitter broke stuff? :smirk:

Raffo:
It was not Joe, but yeah :D

hjacobs:
:+1:

hjacobs:
@tuxlife please comment your findings about the relevant `iptables` rules which were missing...

tuxlife:
I think `iptables=false` disables more than the unused iptables rules. It breaks the SNAT for answers
:+1:

tuxlife:
:+1:

PR #292, opened 2017-02-24
Always use strings for envars

mikkeloscar:
Envars always need to be strings :)

Otherwise kubectl will complain:

```
error: unable to decode "STDIN": [pos 527]: json: expect char '"' but got char '6' 
FATA[0025] exit status 1: error: unable to decode "STDIN": [pos 527]: json: expect char '"' but got char
```


Raffo:
👍

mikkeloscar:
:+1:

PR #293, opened 2017-02-27
Install kube2iam iptables rules on nodes

### LINKED ISSUE #291: kube2iam: install iptables rule manually on each node
Right now we configure kube2iam to install the required iptables rule for routing traffic for the meta data url: https://github.com/jtblin/kube2iam#iptables

This means that there is no rule, and thus no protection if kube2iam isn't running on the node.

This problem is often seen when setting up a new cluster where the secretary pod comes up before kube2iam and fails because it gets the worker node IAM role instead of the intended secretary role.

mikkeloscar:
Creates iptables rule for kube2iam on node startup to prevent other pods
from getting access to the node IAM role if it is started before
kube2iam.

Fix #291


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #302, opened 2017-03-07
add "anonymized" cheat sheet

hjacobs:
Add "anonymized" cheat sheet (converted with internal `kubernetes-cheat-sheet-to-example.sh` script) to serve as an example/template for others to create their own Kubernetes cheat sheets.


mikkeloscar:
Would it make sense to also remove **z** from `zkubectl`?

hjacobs:
@mikkeloscar the `list` and `login` commands won't work and I don't want to redesign the whole sheet for public use (also the architecture image with Skipper+Ingress is obviously specific to our setup) --- I consider this more a template for other people (to adapt for their company) and not a cheat sheet to use "out of the box".

mikkeloscar:
:+1:

hjacobs:
:+1:

PR #309, opened 2017-03-07
fix download link to pdf in cheet-sheet.rst

johannesreichard:


hjacobs:
@johannesreichard the original version/link is correct when rendered with Sphinx, see here: http://kubernetes-on-aws.readthedocs.io/en/latest/user-guide/cheat-sheet.html

johannesreichard:
ah, sry. never mind

PR #310, opened 2017-03-08
Update alpha with latest dev changes

mikkeloscar:
This will include:

* Update to webhook/controller manager identity
* Skipper without kubectl proxy
* Update of ZMON agent (ACID is waiting for this in the db cluster).
* Open skipper metrics endpoint port in security group.


Raffo:
👍

mikkeloscar:
Was closed because we shouldn't do a merge directly from dev to alpha (as new changed to dev will automatically get added to the PR).

PR #315, opened 2017-03-13
Scale down agent. Change filter to kubernetes entities only.

Jan-M:
- Scale down AWS agent
 - Change filter for kube entities only


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #317, opened 2017-03-13
enable host header based metrics

szuecs:
To enable teams to check their own applications, which are backends for skipper we have to enable http host header based metrics in skipper.

### DIFF
--- a/cluster/manifests/skipper/daemonset.yaml
+++ b/cluster/manifests/skipper/daemonset.yaml
@@ -40,6 +40,7 @@ spec:
           - "-kubernetes-in-cluster"
           - "-address=:9999"
           - "-proxy-preserve-host"
+          - "-serve-host-metrics"
         resources:
           limits:
             cpu: 200m


hjacobs:
@szuecs can you add some examples here on how this changes the `/metrics` response?

szuecs:
For example, I have a route:

    first: Host(/^first.local$/) -> "https://www.ccc.de/"

and I set first.local to localhost in my /etc/hosts and do curl http://first.local:9999/  , then I get the following result from metrics endpoint:

    "timers": {
      "skipper.servehost.first_local__9999.GET.404": {
        "15m.rate": 0.8,
        "1m.rate": 0.8,
        "5m.rate": 0.8,
        "75%": 27486.5,
        "95%": 28333,
        "99%": 28333,
        "99.9%": 28333,
        "count": 4,
        "max": 28333,
        "mean": 24530.25,
        "mean.rate": 2.6399468703300495,
        "median": 23898,
        "min": 21992,
        "stddev": 2444.580881766852
      }
    },
    ...

servehost is the key everyone should use to query their application metrics

hjacobs:
:+1:

szuecs:
:+1:

PR #329, opened 2017-03-21
Merge to alpha

szuecs:
This will:
- deploy Kubernetes to v1.5.4 
- update ingress controller such that http to http redirects are automatically deployed
- fix mate canonical zone id
- fix the webhook such that the default account in non kube-system namespaces have read access to the apiserver
- instance limits are raised from 1 Core to 3 Cores

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,7 +4,7 @@ Kubernetes on AWS
 
 **WORK IN PROGRESS**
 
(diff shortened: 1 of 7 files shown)


szuecs:
:+1:

Raffo:
👍

PR #332, opened 2017-03-27
Update to kubernetes v1.5.5

mikkeloscar:
Updates to kubernetes v1.5.5

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md#v155

We need to verify that our PSP configuration still works with this before merge.


mikkeloscar:
:+1:

hjacobs:
:+1:

PR #333, opened 2017-03-28
Align db-alpha branch with alpha

mikkeloscar:
Aligns `db-alpha` branch with `alpha`.

Among other fixes, this brings in kubernetes v1.5.4 which should fix the EBS volume device naming/mounting issues experienced in the DB cluster.

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md#v154

/cc @Jan-M

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,7 +4,7 @@ Kubernetes on AWS
 
 **WORK IN PROGRESS**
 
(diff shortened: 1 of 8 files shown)


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #334, opened 2017-03-28
Feature/autopilot acm certs

szuecs:
ingress-controller automatically selects the right SSL ACM cert, if not specified


mikkeloscar:
:+1:

szuecs:
:+1:

PR #348, opened 2017-04-06
ingress controller with health period of 10s

hjacobs:
Set ALB health period to 10s to become unhealthy within default pod grace period (30s).

See https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/48

### DIFF
--- a/cluster/manifests/ingress-controller/deployment.yaml
+++ b/cluster/manifests/ingress-controller/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: app-ingress-controller
-    version: v0.2.0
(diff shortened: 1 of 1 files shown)


hjacobs:
e2e test failure: "Should create valid https and http ALB endpoint [Ingress] [Zalando]"

@szuecs can you check whether this is caused by the PR or a flake / 1.6 issue?

mikkeloscar:
> @szuecs can you check whether this is caused by the PR or a flake / 1.6 issue?

Timeout was too low, I bumped it from 2 min. to 5 min.

hjacobs:
:+1:

mikkeloscar:
:+1:

PR #349, opened 2017-04-07
Dev to alpha

hjacobs:
* handle `TERM` signal in Skipper
* change health check period of ALB Target Group to 10s (instead of default 30s)

=> node shutdown should now be graceful (tested manually by deleting `skipper-ingress` pod and watching AWS TargetGroup health state)


mikkeloscar:
:+1:

szuecs:
:+1:

PR #350, opened 2017-04-07
Disable anonymous-auth

mikkeloscar:
Set `--anonymous-auth=false` on the API server to block anonymous access to `/api` `/healthz` etc.


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #353, opened 2017-04-10
Secretary with IID secret source

mikkeloscar:
Add new version of secretary which uses IID as secret.

Bonus, we free up some bytes in the CF stack :)


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #355, opened 2017-04-10
Use real initContainers field

mikkeloscar:
`initContainers` are GA in 1.6, so we can get rid of the annotation and use the real spec field instead.


mohabusama:
👍

mohabusama (approved):

hjacobs:
:+1:

PR #356, opened 2017-04-10
Don't schedule skipper on master nodes

mikkeloscar:
This uses node anti-affinity to ensure the skipper daemonset pods are not scheduled on master nodes. Since no traffic is going to the pods it doesn't make sense to have them running there. (https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#node-affinity-beta-feature)

I think eventually taints/tolerations is a better approach for these kind of things, but as of 1.6 there is no way, afaik, to prevent other non-system pods from being scheduled on master nodes if we use taints.

### DIFF
--- a/cluster/manifests/skipper/daemonset.yaml
+++ b/cluster/manifests/skipper/daemonset.yaml
@@ -23,6 +23,13 @@ spec:
       annotations:
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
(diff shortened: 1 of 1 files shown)


mikkeloscar:
Wanted to confirm that the daemonset behaves correctly:

```
NAME                       DESIRED   CURRENT   READY     UP-TO-DATE   AVAILABLE   NODE-SELECTOR   AGE
kube2iam                   8         8         8         8            8           <none>          10m
prometheus-node-exporter   8         8         8         8            8           <none>          10m
skipper-ingress            6         6         6         6            6           <none>          10m
```

It does. :)

mikkeloscar:
:+1:

hjacobs:
:+1:

PR #359, opened 2017-04-12
add kube-dns-metrics

szuecs:
add kube-dns-metrics deployment which does dns queries every 10s and expose metrics calculated over the last 60s


szuecs:
:+1:

mikkeloscar:
:+1:

PR #369, opened 2017-04-21
add compatibility mode and upsert only policy

ideahitme:
add compatibility mode and upsert only policy

`--compatibility true` enables external-dns consider mate annotations `zalando.org/dnsname` to figure out the needed dnsname. It is only needed for services[type=LoadBalancer]. 
`--policy=upsert-only`will only create/update records as needed. If ingress/service is deleted, corresponding dns record will still exist. Adding this flag to make sure that this `alpha` release of external-dns functions as expected in a more sophisticated environment with multiple hosted-zones. Will be removed once we are sure.

### DIFF
--- a/cluster/manifests/external-dns/deployment.yaml
+++ b/cluster/manifests/external-dns/deployment.yaml
@@ -37,8 +37,8 @@ spec:
         - --dry-run=false 
         - --domain=
         - --zone=""
-        - --compatibility
-        - --policy=upsert-only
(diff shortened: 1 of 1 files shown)


hjacobs:
Why do we restrict to `upsert`?

hjacobs:
@ideahitme can you add a comment above the `--upsert-only` and `--compatibility` options so we know that they are temporary and should be removed later on (as soon as we are confident enough)? Thanks!

hjacobs:
:+1:

ideahitme:
👍

PR #371, opened 2017-04-23
Add imagePullSecret for private external-dns sidecar

mikkeloscar:
Makes sure the pod can become ready even if it's scheduled before secretary has attached the `imagePullSecret` to the serviceaccount.


mikkeloscar:
:+1:

hjacobs:
I would actually prefer to push `external-dns-migration` to our OSS registry as it (should be) completely Open Source anyway (migration for [mate](https://github.com/zalando-incubator/mate) users)..

hjacobs:
:+1:

PR #388, opened 2017-05-02
Alpha to db alpha

hjacobs:
Notable changes:

* Kubernetes v1.6.2
* External DNS replacing Mate
* Webhook version bump
* kube2iam version bump


hjacobs:
According to @ideahitme we might need to change some External DNS setting (FQDN template) for the "db" cluster...

ideahitme:
afaik `db` cluster is using mate's "dns name generation from template" feature. We will probably have to update external-dns manifest to simulate this behaviour

linki:
I count 34 services with an ELB. All are using the `mate` annotation and don't use the template.

double-checking with the source code. The operator does create the services with the annotation: https://github.bus.zalan.do/acid/postgres-operator/blob/master/pkg/cluster/k8sres.go#L264

linki:
This rollout will be supported by some manual work. So please don't merge.

hjacobs:
:+1:

linki:
👍

PR #390, opened 2017-05-02
fix SG for ELBs

szuecs:
use static SG to allow all local VPC IPs including ELBs to access worker ASG nodePorts

for reference there is a CloudConfig struct which can be configure with --cloud-config params on the controller-manager:
https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/aws.go\#L386-L426


mikkeloscar:
:+1:

szuecs:
:+1:

PR #400, opened 2017-05-08
remove obsolete imagePullSecret

hjacobs:
External DNS is in OSS registry, so we need no credentials to pull from Docker registry :smile:


hjacobs:
:+1:

ideahitme:
yup it was added because of external-dns-migration script :)

ideahitme:
👍

PR #401, opened 2017-05-09
Update kubelet-wrapper environment variables

mikkeloscar:
From kubelet logs:
```
KUBELET_VERSION environment variable is deprecated, please use KUBELET_IMAGE_TAG instead
KUBELET_ACI environment variable is deprecated, please use the KUBELET_IMAGE_URL instead
RKT_OPTS environment variable is deprecated, please use the RKT_RUN_ARGS instead
```

Not super important right now, but no need to run with deprecated configuration.

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -75,9 +75,9 @@ coreos:
       runtime: true
       content: |
         [Service]
-        Environment=KUBELET_VERSION=v1.6.2_coreos.0
-        Environment=KUBELET_ACI=docker://registry.opensource.zalan.do/teapot/hyperkube
-        Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf \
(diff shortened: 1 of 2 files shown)


hjacobs:
`kube-aws` cloud config for reference: https://github.com/kubernetes-incubator/kube-aws/blob/7a5a4bf36f531288de48311a174ebdef0b4453c2/core/controlplane/config/templates/cloud-config-worker#L137

hjacobs:
:+1:

hjacobs:
@mikkeloscar maybe we can revise the whole cloud config once before merging (to not trigger node updates for only such a non relevant change).

mikkeloscar:
> (to not trigger node updates for only such a non relevant change).

Yes, no need to trigger node update for a trivial change like this, just wanted to have it here, since I saw these messages in the logs today.

hjacobs:
We can do this together with the update to Kubernetes 1.6.3 #403

hjacobs:
Closing, merged into #404

PR #417, opened 2017-05-18
use CF to reduce API calls to AWS

szuecs:
To reduce AWS API calls we use now kube-aws-ingress-controller in version v0.3.0. 
We have to manually cleaning up AWS ALBs after traffic switching and watch out for DNS entries and kube-external-dns changes

### DIFF
--- a/cluster/manifests/ingress-controller/deployment.yaml
+++ b/cluster/manifests/ingress-controller/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: app-ingress-controller
-    version: v0.2.2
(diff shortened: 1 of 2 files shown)


hjacobs:
Removing the ELB permissions from the IAM role will not work. You need to have all IAM permission (CF stack create/update AND all used AWS services).

szuecs:
Conformance test was ok, but not the delete of the stack, because I created manually an ingress which created a CF stack, which I should delete before the delete of the cluster

mikkeloscar:
:+1:

szuecs:
:+1:

PR #422, opened 2017-05-22
Replace etcd2 proxy with etcd3 gateway

mikkeloscar:
Replaces the default etcd2 proxy with the etcd3 gateway
(etcd-member.service). This makes it possible to later switch to etcd3
while still being etcd2 compatible.

Using etcd gateway should also fix the issues we have seen where the etcd2 proxy doesn't re-resolve the hostnames of the etcd cluster after the members has been updated.

etcd version is set to v3.1.6 because the default version included in
coreos stable (v3.0.10) doesn't correctly signal systemd, making the unit
hang (https://github.com/coreos/etcd/commit/01dd60c0f774886193310180b6d62f47f11a0c24).

This also changes the apiserver flags etcd-prefix, storage-backend and
storage-media-type to be configured through template variables. This
enables different storage configuration for different clusters (pending
change in CLM).

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -24,7 +24,7 @@ coreos:
           content: |
             [Service]
             Environment="ETCD_IMAGE_TAG=v3.1.6"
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

linki:
👍

PR #418, opened 2017-05-18
Changed rkt mount and cni 

Raffo:
This change should fix some of the issue with mount that we saw with rkt. This comes from: 

https://github.com/kubernetes-incubator/kube-aws/pull/555

https://github.com/coreos/tectonic-installer/blob/1.6.2-tectonic.1/platforms/metal/cl/bootkube-worker.yaml.tmpl#L44-L83

For the CNI part, the changes are explained here: 

https://github.com/kubernetes-incubator/kube-aws/issues/60

PR #423, opened 2017-05-22
Fix quote in wrong place

mikkeloscar:
#418 introduced a small bug because a `"` was in the wrong place.

It didn't cause any problems because the `--mount volume=var-lib-cni` was just ignored by systemd.


Raffo:
Nice catch!

Raffo:
👍

mikkeloscar:
:+1:

Raffo:
👍

PR #426, opened 2017-05-22
Update to kubernetes v1.6.4

mikkeloscar:
Update to Kubernetes v1.6.4

Fixes a few bugs (which I don't think we hit yet): https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md#changelog-since-v163


mikkeloscar:
:+1:

linki:
👍

PR #433, opened 2017-05-29
Make sure secret values are quoted

mikkeloscar:
This ensures that empty values for the config items result in an empty string value rather than a `null` value.

Fixes an issue were the data values are not returned when they are both `null` when using etcd3.


mikkeloscar:
:+1:

Raffo:
👍

PR #451, opened 2017-06-12
Sample manifests with appdynamics integration

vibhory2j:
Providing sample daemonset file for machineagent
Providing sample deployment files for sample java application with appdynamics java agent configuration
Not adding the secrets manifests files here as this is a public repository
zdeploy templates with appdynamics is done separately


mikkeloscar:
@vibhory2j only the daemonset (machine-agent) should be here, the example app should be documented in our docs (internally).

cluster/manifests/appdynamics/daemonset.yml
```diff
@@ -0,0 +1,53 @@
...
+  template:
+    metadata:
+      name: appdynamics-machinenagent
+      labels:
+        application: appdynamics-machineagent
+    spec:
+      hostNetwork: true
+      containers:
+      - name: appdynamics-machineagent-prod
+        image: pierone.stups.zalan.do/monitoring-appdynamics/appdynamics-machine-agent-4.2:latest
```
mikkeloscar: Please use a specific version instead of `latest`.

cluster/manifests/appdynamics/daemonset.yml
```diff
@@ -0,0 +1,53 @@
...
+        ports:
+        - name: analytics-agent
+          containerPort: 9090
+          hostPort: 9090
+        env:
+          - name: APPDYNAMICS_AGENT_UNIQUE_HOST_ID
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+      - name: appdynamics-machineagent-test
```
mikkeloscar: What's the difference between prod and test? I guess there should only be one instance of the machine agent running.

vibhory2j: It's for production and testing tenants.

cluster/manifests/appdynamics/daemonset.yml
mikkeloscar: Something like `4.2-1` where the first part is the app-dynamics version and the second part (after the dash) is a counter could make sense as a version.

cluster/manifests/appdynamics/daemonset.yml
```diff
@@ -0,0 +1,53 @@
...
+    application: appdynamics-machineagent
+    version: v1
+  name: appdynamics-machineagent
+  namespace: kube-system
+spec:
+  template:
+    metadata:
+      name: appdynamics-machinenagent
+      labels:
+        application: appdynamics-machineagent
```
mikkeloscar: here the `version` label should also be defined.

cluster/manifests/appdynamics/daemonset.yml
```diff
@@ -0,0 +1,53 @@
...
+metadata:
+  labels:
+    name: monitor
+    application: appdynamics-machineagent
+    version: v1
+  name: appdynamics-machineagent
+  namespace: kube-system
+spec:
+  template:
+    metadata:
```
mikkeloscar: Here the update strategy and match selector should also be defined. See for an example: https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/logging-agent/daemonset.yaml#L12-L15

cluster/manifests/appdynamics/daemonset.yml
vibhory2j: removed test container

cluster/manifests/appdynamics/daemonset.yml
vibhory2j: updated manifest to include strategy and match selector

cluster/manifests/appdynamics/daemonset.yml
vibhory2j: Updated manifest to include version label

cluster/manifests/appdynamics/daemonset.yml
```diff
@@ -0,0 +1,45 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  labels:
+    name: monitor
+    application: appdynamics-machineagent
+    version: 0.4
```
mikkeloscar: this must be a string e.g `v0.4` or `"0.4"`

cluster/manifests/appdynamics/daemonset.yml
```diff
@@ -0,0 +1,45 @@
...
+  name: appdynamics-machineagent
+  version: 0.4
```
mikkeloscar: This is in the wrong place, should be below under labels, see my comment below.

cluster/manifests/appdynamics/daemonset.yml
```diff
@@ -0,0 +1,45 @@
...
+    type: RollingUpdate
+  template:
+    metadata:
+      name: appdynamics-machinenagent
+      labels:
+        application: appdynamics-machineagent
+    spec:
+      hostNetwork: true
+      containers:
+      - name: appdynamics-machineagent-prod
```
mikkeloscar: nit: remove `-prod` from the name.

cluster/manifests/appdynamics/secrets/appdynamics-prod-secrets.yaml
```diff
@@ -0,0 +1,14 @@
...
+type: Opaque
+data:
+  access_key: "{{ .ConfigItems.appd_prod_acesskey | base64 }}" 
+  account_name: "{{ .ConfigItems.appd_prod_account | base64 }}"
+  event_service_key: "{{ .ConfigItems.appd_prod_es_key | base64 }}"
+  global_account_name: "{{ .ConfigItems.appd_prod_global_account | base64 }}"
+  uri: "{{ .ConfigItems.appd_prod_uri | base64 }}"
+  es_port: "{{ .ConfigItems.appd_prod_es_port | base64 }}"
+  es_admin_port: "{{ .ConfigItems.appd_prod_es_admin_port | base64 }}"
+  es_uri: "{{ .ConfigItems.appd_prod_es_uri | base64 }}"
```
mikkeloscar: How many of these values must be secret? Stuff like port and uri probably doesn't have to be, so I would just hardcode it here, so we don't have to maintain all of this configuration separately.

cluster/manifests/appdynamics/secrets/appdynamics-sandbox-secrets.yaml
```diff
@@ -0,0 +1,14 @@
...
+type: Opaque
+data:
+  access_key: "{{ .ConfigItems.appd_prod_acesskey | base64 }}" 
+  account_name: "{{ .ConfigItems.appd_prod_account | base64 }}"
+  event_service_key: "{{ .ConfigItems.appd_prod_es_key | base64 }}"
+  global_account_name: "{{ .ConfigItems.appd_prod_global_account | base64 }}"
+  uri: "{{ .ConfigItems.appd_prod_uri | base64 }}"
+  es_port: "{{ .ConfigItems.appd_prod_es_port | base64 }}"
+  es_admin_port: "{{ .ConfigItems.appd_prod_es_admin_port | base64 }}"
+  es_uri: "{{ .ConfigItems.appd_prod_es_uri | base64 }}"
```
mikkeloscar: This looks identical to the secret above (`appdynamics-prod-secrets.yaml`), let's reduce it to only one, the content can still be different depending on environment.

cluster/manifests/appdynamics/secrets/appdynamics-test-secrets.yaml
```diff
@@ -0,0 +1,14 @@
...
+type: Opaque
+data:
+  access_key: "{{ .ConfigItems.appd_prod_acesskey | base64 }}" 
+  account_name: "{{ .ConfigItems.appd_test_account | base64 }}"
+  event_service_key: "{{ .ConfigItems.appd_prod_es_key | base64 }}"
+  global_account_name: "{{ .ConfigItems.appd_test_global_account | base64 }}"
+  uri: "{{ .ConfigItems.appd_prod_uri | base64 }}"
+  es_port: "{{ .ConfigItems.appd_test_es_port | base64 }}"
+  es_admin_port: "{{ .ConfigItems.appd_test_es_admin_port | base64 }}"
+  es_uri: "{{ .ConfigItems.appd_prod_es_uri | base64 }}"
```
mikkeloscar: Same here, let's reduce to one secret.

vibhory2j: Updated the above comments in the manifest files.

vibhory2j:
Closing this pull request as duplicated and new pull request created to address this. 
https://github.com/zalando-incubator/kubernetes-on-aws/pull/521

PR #457, opened 2017-06-15
Working on Postgres Operator dependencies.

Jan-M:
Adding postgres operator with dependencies:

- [ ] new etcd cluster for postgres / patroni
- [ ] s3 backup for postgres
- [ ] iam role for pods
- [ ] iam role for operator

### DIFF
--- a/cluster/manifests/postgres-operator/configmap.yaml
+++ b/cluster/manifests/postgres-operator/configmap.yaml
@@ -1,0 +1,37 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: postgres-operator
+data:
+  service_account_name: operator
+  cluster_labels: application:spilo
+  cluster_name_label: version
+  pod_role_label: spilo-role
+  db_hosted_zone: {{ .ConfigItems.hosted_zone }}
+  debug_logging: "true"
+  dns_name_format: '%s.%s.%s'
+  docker_image: registry.opensource.zalan.do/acid/spiloprivate-9.6:1.2-p7
+  etcd_host: etcd-acid-cluster.etcd.{{ .ConfigItems.hosted_zone }}:2379
+  infrastructure_roles_secret_name: postgresql-infrastructure-roles
+  oauth_token_secret_name: postgresql-operator
+  pam_configuration: |
+    https://info.services.auth.zalando.com/oauth2/tokeninfo?access_token= uid realm=/employees
+  pam_role_name: zalandos
+  pod_deletion_wait_timeout: 10m
+  pod_label_wait_timeout: 10m
+  ready_wait_interval: 3s
+  ready_wait_timeout: 30s
+  replication_username: replication
+  resource_check_interval: 3s
+  resource_check_timeout: 10m
+  resync_period: 5m
+  resync_period_pod: 5m
+  super_username: postgres
+  teams_api_url: https://teams.auth.zalando.com/api/
+  wal_s3_bucket: {{ .ConfigItems.postgres_backup_s3_bucket }}
+  kube_iam_role: app-spilo
+  default_cpu_request: 100m
+  default_memory_request: 100Mi
+  default_cpu_limit: 3000m
+  default_memory_limit: 1Gi
+  workers: "16"
--- a/cluster/manifests/postgres-operator/credentials.yaml
+++ b/cluster/manifests/postgres-operator/credentials.yaml
@@ -1,0 +1,13 @@
+apiVersion: "zalando.org/v1"
+kind: PlatformCredentialsSet
+metadata:
+  name: postgresql-operator
+spec:
+  application: postgresql-operator
+  tokens:
+    read-only:
+      privileges:
+    cluster-registry-rw:
+      privileges:
+    cluster-rw:
+      privileges:
--- a/cluster/manifests/postgres-operator/operator.yaml
+++ b/cluster/manifests/postgres-operator/operator.yaml
@@ -1,0 +1,25 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: postgres-operator
+  namespace: acid
+spec:
+  replicas: 1
+  template:
+    metadata:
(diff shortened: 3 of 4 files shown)


cluster/manifests/postgres-operator/configmap.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: postgres-operator
+data:
+  service_account_name: operator
```
linki, line 6: We don't pre-define the `operator` account anywhere. So it has to be created here as well. Alternatively we can also use `kube-system` namespace and the `system` user.

Jan-M: Lets create it then and have an example with it. should not be the same as system imho

cluster/manifests/postgres-operator/configmap.yaml
```diff
@@ -0,0 +1,37 @@
...
+  cluster_labels: application:spilo
+  cluster_name_label: version
+  pod_role_label: spilo-role
+  db_hosted_zone: {{ .ConfigItems.hosted_zone }}
```
linki, line 10: TODO: add this to CLM

cluster/postgres-operator-dependencies.yaml
```diff
@@ -0,0 +1,175 @@
...
+      Policies:
+      - PolicyName: postgres-pod-policy
+        PolicyDocument:
+          Version: "2012-10-17"
+          Statement:
+            - Effect: "Allow"
+              Action:
+              - "s3:*"
+              Resource:
+              - "arn:aws:s3:::{{ Arguments.postgres_backup_bucket }}"
```
Jan-M, line 161: This also needs to come from CLM or we need to build a bucket name here

cluster/manifests/postgres-operator/credentials.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: "zalando.org/v1"
+kind: PlatformCredentialsSet
+metadata:
+  name: postgresql-operator
+spec:
+  application: postgresql-operator
+  tokens:
+    read-only:
+      privileges:
+    cluster-registry-rw:
```
linki, line 10: I'm wondering why the operator, which is per cluster, needs access to the cluster-registry?

cluster/manifests/postgres-operator/operator.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: postgres-operator
+  namespace: acid
```
linki, line 5: This is the only Kubernetes resource that has a namespace. Is that correct?

cluster/manifests/postgres-operator/operator.yaml
```diff
@@ -0,0 +1,25 @@
...
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        name: postgres-operator
+        application: postgres-operator
+        version: {{ version }}
```
linki, line 13: That must be a specific version

cluster/manifests/postgres-operator/operator.yaml
```diff
@@ -0,0 +1,25 @@
...
+    spec:
+      containers:
+      - env:
+        - name: MY_POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: CONFIG_MAP_NAME
+          value: "postgres-operator"
+        image: pierone.stups.zalan.do/acid/postgres-operator:ba6529bba6529b
```
linki, line 23: `ba6529bba6529b` should probably be `ba6529b`

cluster/manifests/postgres-operator/configmap.yaml
```diff
@@ -0,0 +1,37 @@
...
+  ready_wait_timeout: 30s
+  replication_username: replication
+  resource_check_interval: 3s
+  resource_check_timeout: 10m
+  resync_period: 5m
+  resync_period_pod: 5m
+  super_username: postgres
+  teams_api_url: https://teams.auth.zalando.com/api/
+  wal_s3_bucket: {{ .ConfigItems.postgres_backup_s3_bucket }}
+  kube_iam_role: app-spilo
```
alexeyklyukin, line 32: the name better reflect that it belongs to the operator (we use postgres-operator in the db cluster).

cluster/manifests/postgres-operator/operator.yaml
linki: also `registry.opensource.zalan.do`

cluster/postgres-operator-dependencies.yaml
```diff
@@ -0,0 +1,175 @@
...
+      - PolicyName: AllowEC2Modify
+        PolicyDocument:
+          Version: "2012-10-17",
+          Statement:
+            - Effect: Allow
+              Action:
+                - ec2:Describe*
+                - ec2:Modify*
+
+  PostgresPodRole:
```
linki, line 140: This probably needs to be configured via an annotation on the deployment as well.

cluster/manifests/postgres-operator/credentials.yaml
alexeyklyukin: We neither read nor write it. I have no idea why it is there in the first place :-)

cluster/postgres-operator-dependencies.yaml
```diff
@@ -0,0 +1,175 @@
+SenzaComponents:
```
mikkeloscar, line 1: Why do we need this stack?

It seems like it's quite out of scope for our base Kubernetes infrastructure to also provision extra *non-cluster* related stacks.

We already have the problem of manually updating the existing etcd stack when taupage is outdated etc. Adding another etcd is not making it simpler.

mikkeloscar:
Closing as the postgres operator is already being deployed via CDP.

PR #461, opened 2017-06-20
Dev to db alpha

mikkeloscar:
Replaces: #460

This will align `db-alpha` with `dev`. It will cause an update to the CF stack which will trigger AZRebalance and potential downtime for services running in the cluster.

It will also trigger full update of master and worker nodes.

/cc @Jan-M @avaczi


avaczi:
:+1:

mikkeloscar:
:+1:

PR #462, opened 2017-06-20
[WiP] change image of ingress controller

linki:
We need to change the image of the ingress controller. It changed to make it more consistent.

This is a reminder. Currently CDP somehow doesn't build it.

### DIFF
--- a/cluster/manifests/ingress-controller/deployment.yaml
+++ b/cluster/manifests/ingress-controller/deployment.yaml
@@ -26,7 +26,7 @@ spec:
         operator: Exists
       containers:
       - name: controller
-        image: registry.opensource.zalan.do/teapot/kube-aws-ingress-controller:v0.3.5
+        image: registry.opensource.zalan.do/teapot/kube-ingress-aws-controller:v0.3.5
         env:
         - name: AWS_REGION
           value: {{ .Region }}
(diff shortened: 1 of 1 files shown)


cluster/manifests/ingress-controller/deployment.yaml
```diff
@@ -26,7 +26,7 @@ spec:
         operator: Exists
       containers:
       - name: controller
-        image: registry.opensource.zalan.do/teapot/kube-aws-ingress-controller:v0.3.5
+        image: registry.opensource.zalan.do/teapot/kube-ingress-aws-controller:v0.3.5
```
szuecs: the former looks more correct, not?

mikkeloscar: I also think the former sounds better, but the repository is called `kube-ingress-aws-controller` so it makes sense to keep it consistent like @linki did :+1:

linki: Exactly @mikkeloscar. Other option is to rename the repo but then I would go straight to `alb-ingress-controller`.

szuecs:
:-1:

szuecs:
Let's close this in favour of https://github.com/zalando-incubator/kubernetes-on-aws/pull/463

PR #464, opened 2017-06-22
Set custom log parsers

mikkeloscar:
Define custom scalyr log parsers for skipper and webhook.

This will allow us to create custom parsers to make it much easier to query requests in skipper, and make it easier to inspect the webhook logs e.g. to get information about number of unique users pr day/week/month.


Raffo:
👍

mikkeloscar:
:+1:

PR #467, opened 2017-06-25
[WIP] Use RBAC for authorization

mikkeloscar:
Initial effort to replace authz part of our webhook with RBAC while being
compatible with the current expectations e.g. `operator` serviceaccount, 
PSPs and `kube-system` `system` serviceaccount.

Roles and bindings are defined as follows:

* Components in `kube-system` requiring priviliged PSP must use the `system`
serviceaccount (or define a custom role/serviceaccount). I kept the `system`
serviceaccount to not change everything at once. It will be easy to migrate
individual components later on.
* `system:serviceaccount:kube-system:system` gets `cluster-admin` role.
* `system:serviceaccount:kube-system:default` gets `view` role.
* All serviceaccounts can use the `restricted` PSP.
* Users with `PowerUser` group gets the `edit` role.
* Users with `ReadOnly` group gets the `view` role.
* `Administrator` group maps to `system:masters` and gets `cluster-admin` role.
* `system:serviceaccount:default:operator` gets `edit` role + priviliged PSP.
* `kubelet` user gets `system:node` role.
* `kubelet` user gets `system:node-proxier` role. This is needed because
    kube-proxy is currently using kubelet credentials in our setup.

This requires a minor change to the authentication part in the webhook:

* `Administrator` group must be converted to `system:masters`.
* business-partner support needs to be modified to convert priviliges to groups, such that you can create unique rolebindings based on group name.

And also a change in the CLM to rename user `kube-controller-manager`
to `system:kube-controller-manager`.

## TODO

* [ ] Validate that `operator` role has all needed permissions.
* [ ] Validate that `view` and `edit` roles map to what we expect for `ReadOnly` and `PowerUser` roles.
* [ ] Prevent creation of `daemonsets`.
* [ ] Optional: Add separate serviceaccounts for `kube-system` components
  * [x] heapster
  * [x] kube-dns, kube-dns-autoscaler
  * [ ] dashboard
  * [x] zmon
  * [x] kube-state-metrics
  * [ ] external-dns
  * [ ] ingress-controller
  * [ ] skipper
  * [ ] others...
* [x] Validate permissions needed for ZMON.
* [ ] Review billing role binding (has view permissions)
* [ ] Consider how RBAC can be used by users to replace `operator` going
      foward. For instance, if the deployment system is allowed to create roles
      and bindings, then you can give yourself admin rights via a deployment,
      which may or may not be an issue.
* [x] Allow kubelet to create kube-proxy mirror-pod
    ```
    pods "kube-proxy-ip-172-31-12-63.eu-central-1.compute.internal" is forbidden: no providers available to validate pod request
    ```

## Future improvements

* Move kube-proxy to a daemonset ~so it can use a custom serviceaccount.~ This is not possible as it obviously need to start before cluster service IPs become available. So it needs to bootstrap using a non-cluster IP. (Should still be a daemonset, but thats unrelated).
* Limit nodes to access only the relevant pods/secrets. I.e. per node identity (https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/).
* Deprecate and later remove operator role.

RBAC reference: https://kubernetes.io/docs/admin/authorization/rbac/

### DIFF
--- a/cluster/manifests/heapster/deployment.yaml
+++ b/cluster/manifests/heapster/deployment.yaml
@@ -19,7 +19,7 @@ spec:
         version: v1.5.4
     spec:
       priorityClassName: system-cluster-critical
-      serviceAccountName: system
+      serviceAccountName: heapster
       containers:
         - image: registry.opensource.zalan.do/teapot/heapster:v1.5.4
           name: heapster
--- a/cluster/manifests/heapster/rbac.yaml
+++ b/cluster/manifests/heapster/rbac.yaml
@@ -1,0 +1,18 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: heapster
+  namespace: kube-system
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: heapster
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:heapster
+subjects:
+- kind: ServiceAccount
+  name: heapster
+  namespace: kube-system
--- a/cluster/manifests/kube-dns/rbac.yaml
+++ b/cluster/manifests/kube-dns/rbac.yaml
@@ -1,0 +1,46 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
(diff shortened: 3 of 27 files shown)


mikkeloscar:
Done in #1550

PR #470, opened 2017-06-27
Kube node ready

Raffo:
Changes needed for kube-node-ready and additionally removed some lines to keep up with the lack of space (which we will address soon!). 

EDIT: wait for review as I have to change the flag as well.

### DIFF
--- a/cluster/manifests/kube-node-ready/daemonset.yaml
+++ b/cluster/manifests/kube-node-ready/daemonset.yaml
@@ -24,7 +24,7 @@ spec:
         operator: Exists
       containers:
       - name: kube-node-ready
-        image: registry.opensource.zalan.do/teapot/kube-node-ready:8d13c11-dirty-4
+        image: registry.opensource.zalan.do/teapot/kube-node-ready:6ee38a6
         args:
         - --master-lifecycle-hook={{ .Outputs.MasterAutoscalingLifecycleHook }}
         - --worker-lifecycle-hook={{ .Outputs.WorkerAutoscalingLifecycleHook }}
--- a/cluster/manifests/kube-node-ready/daemonset.yaml
+++ b/cluster/manifests/kube-node-ready/daemonset.yaml
@@ -24,7 +24,7 @@ spec:
         operator: Exists
       containers:
       - name: kube-node-ready
-        image: registry.opensource.zalan.do/teapot/kube-node-ready:6ee38a6
+        image: registry.opensource.zalan.do/teapot/kube-node-ready:9799c3d
         args:
         - --master-lifecycle-hook={{ .Outputs.MasterAutoscalingLifecycleHook }}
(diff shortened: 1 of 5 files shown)


cluster/userdata-master.yaml
```diff
@@ -600,10 +588,4 @@ write_files:
...
     content: |
       [global]
-      DisableSecurityGroupIngress = true
-
-  - path: /home/core/.toolboxrc
-    owner: core
-    content: |
-        TOOLBOX_DOCKER_IMAGE=registry.opensource.zalan.do/teapot/microscope
-        TOOLBOX_DOCKER_TAG=v0.0.3
+      DisableSecurityGroupIngress = true
```
mikkeloscar, line 591: even saving the trailing newline :D

cluster/manifests/kube-node-ready/service.yaml
```diff
@@ -0,0 +1,19 @@
...
+    application: kube-node-ready
+    kubernetes.io/cluster-service: "true"
+    kubernetes.io/name: "kube-node-ready"
+  annotations:
+    service.beta.kubernetes.io/external-traffic: OnlyLocal
+spec:
+  type: NodePort
+  ports:
+    - port: 80
+      nodePort: 30797
```
linki: we picked this because of the `7979` convention 😄  so this should rather be `30979` or `30079`.

linki: but since the target port is `8080` it should maybe be `30080`. It should be a number that makes sense because it will stick.

cluster/manifests/kube-node-ready/daemonset.yaml
```diff
@@ -0,0 +1,37 @@
...
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
+        iam.amazonaws.com/role: "{{ .LocalID }}-kube-node-ready"
+    spec:
+      tolerations:
+      - key: CriticalAddonsOnly
+        operator: Exists
+      containers:
+      - name: kube-node-ready
+        image: registry.opensource.zalan.do/teapot/kube-node-ready:8d13c11-dirty-4
```
linki: nit: I find `kube-node-ready` not very descriptive and would prefer something along the lines of `node-readiness-reporter` or `node-readiness-informer` etc.

mikkeloscar: I just suggested `kube-node-ready` but I don't mind if it's called something else.

Raffo: I also liked `kube-node-ready` frankly, but maybe I read the name too much in the last two days to properly judge.

Raffo:
Test passed!

Raffo:
👍

hjacobs:
:+1:

PR #473, opened 2017-06-29
Make master node instance type configurable

mikkeloscar:
Until now we hardcoded master nodes to use `m3.medium`, which is a problem now that we have the new 3rd which doesn't support `m3.medium`.

It was never the idea to hardcode it, since we have it defined in the cluster registry as well, so let's just make it configurable.


mikkeloscar:
:+1:

Raffo:
No need to wait for the tests for this, let's get this merged.

Raffo:
👍

mikkeloscar:
:+1:

Raffo:
👍

PR #480, opened 2017-07-04
Use latest etcd appliance version

Raffo:
This is needed only for compliance reason as the source repo was changed with the recent url and the scm source of the image has to be updated. From what @CyberDem0n says, there should be no differences, it's clearly still worth testing.

### DIFF
--- a/cluster/etcd-cluster.yaml
+++ b/cluster/etcd-cluster.yaml
@@ -14,7 +14,7 @@ SenzaComponents:
         2379: 2379
         2380: 2380
       runtime: Docker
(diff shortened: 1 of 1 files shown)


mikkeloscar:
Let's upade all dev clusters to this version to test it for a little bit.

mikkeloscar:
@Raffo I have updated the dev clusters to this version and also changed the instance type to t2.medium to roll out to 3rd AZ.

Maybe you could change the instance type as well in this PR (or I can do it) and then it lgtm.

Raffo:
👍

mikkeloscar:
:+1:

PR #484, opened 2017-07-05
Allow deployment role to manage other roles

mikkeloscar:
This matches the resource with what we have for the `PowerUser` role. Without this the deployment role is not allowed to manage other roles.


tuxlife:
:+1:

tuxlife (approved):

mikkeloscar:
:+1:

PR #485, opened 2017-07-06
Proposal: replace secretary with a 'static' docker config.

mikkeloscar:
This is a proposal (+implementation) to replace **secretary** with a 'static'
docker config generated a boot.

So far we have relied on Kubernetes [ImagePullSecrets](https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod) to provide authentication for pulling images from our private repository.
This method works well when using a single namespace and the `default` serviceaccount, because we can attach the secret to the serviceaccount once and then it will *just work* for all pods deployed in the namespace.

This however becomes a bit of a problem with multiple namespaces or service accounts. Each time a new namespace is created Kubernetes will automatically create a `default` serviceaccount, and secretary will attach the `imagePullSecret`, but if a pod is spawned before the secret is attached, then it will never get it and will thus fail to pull an image from a private repository.

Similarly if users specify their own serviceaccount, they always have to manually specify the `imagePullSecret`.

Secretary was initially designed to rotate credentials form a mint-bucket, but since we switched to using Instance Identity Documents for authenticating pulls, there is no need anymore to rotate it during the lifetime of a node.

By using a static docker config on the node instead we can just initialize it on boot, and it will *just work* no matter the namespace or serviceaccount that is used and users will no longer have to care about this detail.

*BONUS*: Cluster creation can be a bit faster because it doesn't depend on secretary (and thereby kube2iam), to be ready, before it can start pulling private images.


hjacobs:
This sounds great :smile: 

@Raffo let's get this tested (e2e already works) and merged? I don't see any blockers for it...

Raffo:
I've seen it and looks good to me.

Raffo:
👍

linki:
👍

PR #490, opened 2017-07-10
Dev to db alpha

ideahitme:
Updating DB cluster with the following changes: 

1. Enable Deployment via CDP
2. Updating manifests including skipper, webhook, external-dns etc
3. Use latest etcd-appliance version p17
4. Integrate local PlatformIAM  
5. @gargravarr has now SSH access 

/cc @alexeyklyukin  @CyberDem0n @Jan-M 
/cc @Raffo @linki please review, I will update the status and notify DB team once reviewing is done


linki:
👍

Jan-M:
:+1:

PR #493, opened 2017-07-17
Alpha to beta

szuecs:
To keep track on changes from alpha we should merge them to beta with the following changes: 

1. Changed instance type of etcd cluster to use t2.medium (more CPUs, less money, but having 4h CPU burst)
1. Enable Deployment via CDP
1. Use latest etcd-appliance version p17
1. Integrate local PlatformIAM to lower rtt for token verification (fix kube-apiserver latency SLOs)


hjacobs:
@szuecs would be generally good to summarize the changes in the PR description (easier than looking through many many commits).

ideahitme:
@szuecs it is essentially same as https://github.com/zalando-incubator/kubernetes-on-aws/pull/490 you can just copy (minus the `2.`, `.5`)

szuecs:
:+1:

ideahitme:
👍

PR #498, opened 2017-07-19
Tell kubernetes how many master nodes it can expect

linki:
was wondering for a while about the usage of this and now found https://github.com/mantl/mantl/issues/1811.

If you look at our teapot cluster, which has two nodes:
```console
kubectl --namespace=default describe svc kubernetes | grep Endpoints
Endpoints:		172.31.20.166:443
```
However, the internal Kubernetes service only exports one of the nodes.

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -287,6 +287,7 @@ write_files:
           command:
           - /hyperkube
           - apiserver
+          - --apiserver-count={{ APISERVER_COUNT }}
           - --bind-address=0.0.0.0
           - --insecure-bind-address=0.0.0.0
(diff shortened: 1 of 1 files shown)


hjacobs:
:+1:

Raffo:
Any reason this has a "do-not-merge" label? It looks good to me and tests are passing!

linki:
@Raffo The number of masters is configurable and I would like to use that number instead.

linki:
@Raffo Done but it requires a change in CLM before: https://github.bus.zalan.do/teapot/cluster-lifecycle-manager/pull/183

linki:
👍

Raffo:
👍

PR #503, opened 2017-07-21
completely get rid of secretary :-( bye bye

linki:
This removes secretary's presence from our setup. Ideally zmon should use platform credentials set or another bucket.

We removed secretary but still relied on its mint bucket which may lead to confusion.


linki:
👍

ideahitme:
👍

PR #506, opened 2017-07-25
Adding ports needed for cross master communication

Raffo:
This fixes: 

- the issue with a communication from master to master dealing with some master failures spotted with kube-node-ready
- master to master communication is now fixed to allow streaming logs 
- kube-node-ready port (30080) is not accessible from zmon so that we can use it for monitoring


Raffo:
👍

mikkeloscar:
:+1:

PR #507, opened 2017-07-26
Add etcd backup bucket to the etcd stack

mikkeloscar:
This makes the etcd backup bucket part of the etcd stack such that it's owned by the stack instead of being owned by CLM.

It uses `DeletionPolicy: Retain` to make sure the bucket sticks around even if the etcd stack is accidentally deleted.


Raffo:
👍

mikkeloscar:
:+1:

PR #515, opened 2017-07-28
Alpha to beta

szuecs:
* This will use the new ingress controller version that will fix the issue for all ingress CF stacks, that have  targetGroups in eu-central-1-c, but are not in use, because the ALB has no Network interface in that AZ.
* EtcdBackupBucket added to support etcd cluster backups

### DIFF
--- a/cluster/etcd-cluster.yaml
+++ b/cluster/etcd-cluster.yaml
@@ -70,6 +70,13 @@ Resources:
         Fn::GetAtt:
           - EtcdSecurityGroup
           - GroupId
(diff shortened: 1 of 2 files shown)


Raffo:
conformance-tests failed.

mikkeloscar:
Test are failing because they expect kubernetes v1.7 now. I would say it doesn't make sense to test alpha-beta PRs since this exact config has already been tested when it was merged into dev.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #521, opened 2017-08-03
added manifests and appdynamics machine agent and secret

vibhory2j:
Added manifest files for Appdynamics machine agent integration with Kubernetes cluster. 
Added the secret which contains configurations like access keys, end points for Appdynamics controller. The secret is communicated with Mikkel so that it can be encrypted and added to CLM.

This machine agent will be used by Appdynamics application agents for communicating analytics transactions. 

For previous comments please refer old pull request -https://github.com/zalando-incubator/kubernetes-on-aws/pull/451. (Will be closing this pull request it was created on a forked repository)


linki:
👍

cluster/manifests/appdynamics/daemonset.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  labels:
+    name: monitor
+    application: appdynamics-machineagent
+    version: "4.3-0.74"
+  name: appdynamics-machineagent
+  namespace: kube-system
```
szuecs, line 9: Why it has to be in kube-system?

szuecs:
Why do we want to deploy appdynamics if we want to try other POCs soon, in which we want to remove appdynamics again?

szuecs:
:-1:

cluster/manifests/appdynamics/daemonset.yaml
vibhory2j: As per our understanding daemonsets are supposed to run in kube-system. Is that correct?

vibhory2j:
@szuecs Some of the teams are starting to deploy on k8s and using appdynamics application agents. The machine agent will be required for certain features to work properly.

szuecs:
@vibhory2j What about the point of the Instana POC?
The problem is if they have the jvm agent running, then we can create live incidents with installing Instana sensors.

vibhory2j:
@szuecs This one is a machine agent and doesn't instantiates a JVM agent to instrument applications. I think the Instana POC can be treated as a separate topic. Nevertheless, this machine agent should be turned off in case Instana agent are installed.

hjacobs:
This PR is pretty stale, should we close it?

szuecs:
From my side, yes, because it has no Limits defined and we know already that teams had problems with the appdynamics agent, causing too much unreaonable load on a node.
Maybe @vibhory2j has a different opinion, and I am ok to have clear Limits defined and test appdynamics, if our customers want to have it.

hjacobs:
@vibhory2j @szuecs closing this PR for now (no activity for a long time) --- feel free to reopen as needed in the future.

PR #527, opened 2017-08-17
db-alpha: resync internal changes with public repo

linki:
Things that piled up in our internal repo over the last weeks
* manage s3 bucket via CF stack
* stop restarting cronjobs
* add pod count quota to default and kube-system namespace
* update skipper ingress daemon set
* add region information to zmon
* controller-manager: connect to local apiserver and use randomly generated secret
* add back the custom toolbox image
* remove secretary


mikkeloscar:
:+1:

linki:
👍

PR #535, opened 2017-08-22
scalyr: hide scalyr secrets from api users

linki:
This uses `secretKeyRef` to populate scalyr access credentials into environment variables in order to prevent people from seeing it when using `kubectl describe pod`.


mikkeloscar:
:+1:

linki:
👍

PR #537, opened 2017-08-22
fix: remove duplicate content field

linki:
There's a duplicate `content` field. Seems like it doesn't have any effect but let's remove it anyways.


linki:
👍

mikkeloscar:
:+1:

PR #548, opened 2017-08-28
merge latest dev into dev-to-db-alpha

linki:
merge latest dev into db-alpha so we only require one userdata-master update.


linki:
👍

szuecs:
:+1:

PR #550, opened 2017-08-29
Add support for Emergency role in webhook

mikkeloscar:
Adds support for the `Emergency` role temporarily granted by the emergency-access-service. This role has the same permissions as `PowerUser`.


Raffo:
👍

mikkeloscar:
:+1:

PR #553, opened 2017-08-29
Add CriticalAddonsOnly toleration to static pods

mikkeloscar:
Adds `CriticalAddonsOnly` toleration to static pods to ensure they won't get evicted in case the node runs out of resources.

We had this on all the components in the manifests folder already, but forgot to add it to the pods defined in the userdata.


linki:
👍

hjacobs:
:+1:

PR #554, opened 2017-08-29
do not bind clients to a particular apiserver instance

linki:
This overwrites the default definition of the kubernetes service to employ no session affinity, as opposed to the default on "ClientIP".

This allows clients to pick another apiserver instance in case one is down. We had trouble with this in the past when clients wouldn't pick the other apiserver in case one is down. I believe this is due its default definition of `sessionAffinity: ClientIP`.

### DIFF
--- a/cluster/manifests/kubernetes/svc-kubernetes.yaml
+++ b/cluster/manifests/kubernetes/svc-kubernetes.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: kubernetes
+  namespace: default
+spec:
+  sessionAffinity: None


Raffo:
Did you test that this actually solves the problem that we hit in the past? I remember an issue in kubernetes that I can't find right now where the thing seemed a little bit more complex than that.

linki:
It basically times out as before but on the next try has a chance to get the other endpoint. Which is different from before I believe.

szuecs:
@linki 
Do I understand correctly that you believe this will have the chance or do you know or even tested that it has the behaviour of a chance in the next connect?
If you don't know please check if this is really changing behaviour you expect. From the name I would also expect the changed behaviour.

linki:
@szuecs I only checked using `curl`. I'll try to reproduce the original issue.

linki:
The underlying issue has been fixed and will be included in 1.8. It allows using `--apiserver=2` but will deregister non-existing ones even below that number.

PR #558, opened 2017-08-31
Update image policy hook with simulate and metrics

linki:
Updates imagePolicyWebhook with the `--simulate` feature and a `/metrics` endpoint.

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -429,10 +429,11 @@ write_files:
               value: https://identity.zalando.com/.well-known/openid-configuration
             - name: ENABLE_INTROSPECTION
               value: "true"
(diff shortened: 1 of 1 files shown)


linki:
The test that fails is the test that tests the imagePolicyWebhook. It fails because it runs in `simulation` mode and thus doesn't reject the image. @mikkeloscar what do you think on how to proceed? An additional config item seems overkill.

linki:
Temporarily disable image-policy check: https://github.bus.zalan.do/teapot/kubernetes-e2e-tests/pull/17

Revert this PR once simulation mode is disabled.

Raffo:
👍

szuecs:
:+1:

PR #559, opened 2017-09-01
Disable AllAlpha=true flag to not run with experimental alpha features

mikkeloscar:
This disables the `--AllAlpha=true` flag on all components so we don't run with experimental alpha features. One example is the certificatesigningrequest feature which we intended not to have enabled at this time.

https://kubernetes.slack.com/archives/C09NXKJKA/p1504200253000371

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -134,6 +134,7 @@ coreos:
         --tls-private-key-file=/etc/kubernetes/ssl/worker-key.pem \
         --cloud-provider=aws \
         --low-diskspace-threshold-mb=1000 \
+        --feature-gates=ExperimentalCriticalPodAnnotation=true,TaintBasedEvictions=true \
         --system-reserved=cpu=100m,memory=164Mi \
         --kube-reserved=cpu=100m,memory=282Mi
         ExecStop=-/usr/bin/rkt stop --uuid-file=/var/run/kubelet-pod.uuid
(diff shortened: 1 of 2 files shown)


linki:
I guess this is the stuff that gets disabled with this (apiserver):

```
Accelerators=true|false (ALPHA - default=false)
AdvancedAuditing=true|false (ALPHA - default=false)
AffinityInAnnotations=true|false (ALPHA - default=false)
DynamicKubeletConfig=true|false (ALPHA - default=false)
ExperimentalCriticalPodAnnotation=true|false (ALPHA - default=false)
ExperimentalHostUserNamespaceDefaulting=true|false (BETA - default=false)
LocalStorageCapacityIsolation=true|false (ALPHA - default=false)
PersistentLocalVolumes=true|false (ALPHA - default=false)
RotateKubeletClientCertificate=true|false (ALPHA - default=false)
RotateKubeletServerCertificate=true|false (ALPHA - default=false)
TaintBasedEvictions=true|false (ALPHA - default=false)
```

szuecs:
@linki we only use AffinityInAnnotations, ExperimentalCriticalPodAnnotation and TaintBasedEvictions right now, or do I miss something?
According to https://kubernetes.io/docs/admin/kube-apiserver/ we can enable it by key=value, so:

    --feature-gates AffinityInAnnotations=true,ExperimentalCriticalPodAnnotation=true,TaintBasedEvictions=true

mikkeloscar:
I configured `--feature-gates=ExperimentalCriticalPodAnnotation=true,TaintBasedEvictions=true`. Did not configure `AffinityInAnnotations=true` as we are not using it anywhere afaik, we use the actual spec fields instead of the annotations.

szuecs:
:+1:

Raffo:
👍

PR #565, opened 2017-09-05
Upgrade zmon-worker

mohabusama:
Upgrade zmon-worker with updated opsgenie notification that excludes alert event type in alert title.


linki:
👍

szuecs:
:+1:

PR #568, opened 2017-09-06
Update image policy webhook and configure failure policy

linki:
Notable features:
* Fix a bug that causes mirror pods not to be registered.
* Use a restricitive failure policy that rejects images in case of failed dependencies (like before)


linki:
👍

mikkeloscar:
:+1:

Issue #138, opened 2016-11-22
Correct cluster teardown (clean up AWS resources: ELBs, DNS, ..)


linki:
on cluster delete we should delete resources that were created by kubernetes and friends. this can be done as part of `cluster.py delete`

what do delete
* ELBs created via services of type LoadBalander
* DNS records created via services of type LoadBalander (mate)
* EBS volumes created by Kubernetes PVs
* ... ?

or just everything?

mikkeloscar:
related to #94

Raffo:
@mikkeloscar suggests that we can tag every resource so that we can get rid of things.

szuecs:
@mikkeloscar can we close this issue or do we have something left?

mikkeloscar:
@szuecs we still need to implement proper teardown in the CLM, but we don't necessarily need to track it here.

szuecs:
Close this, because our cluster-lifecycle-manager is not yet public

PR #569, opened 2017-09-07
Tag resources with cluster ID

mikkeloscar:
Tags resources with the `kubernetes.io/cluster/<cluster-id>` tag introduced in https://github.com/kubernetes/kubernetes/pull/41695 and removes the deprecated `KubernetesCluster: kubernetes` tag.

This is part of the fix for https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/77 and will also allow us to fix the problem of ELBs created for a service of type LoadBalancer not attaching to worker nodes in all subnets. The subnet issue also requires a change in the CLM to attach this tag to existing subnets. There is a PR for this in the CLM repo.

The tag `ClusterID` is still required by the ingress-controller, but can be removed once https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/77 is fixed.

Additionally we will eventually be able to solve #138 by having consisting tagging on all resources.


szuecs:
:+1:

linki:
👍

PR #572, opened 2017-09-08
Bump version for cluster autoscaler healthcheck

Raffo:
**DO NOT MERGE, STILL WIP** 

The goal of this PR is to update the cluster-autoscaler to add an healthcheck. We want to use this to identify when scaling attempts fail and take action on that. This is a very simple approach that could probably be improved to export proper metrics that are more actionable than just an health endpoint. 

Related PR: https://github.com/hjacobs/kube-aws-autoscaler/pull/35


Raffo:
This is stale and outdated.

PR #578, opened 2017-09-12
use ntp to set time

szuecs:
use ntpdate + ntpd to setup the time, such that opentracing can connect the request times better

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -52,14 +52,6 @@ coreos:
             --endpoint=http://127.0.0.1:2379 set /coreos.com/network/config \
             '{ "Network": "10.2.0.0/16", "Backend": {"Type": "vxlan"}}'
 
-    - name: ntpdate.service
-      command: start
-      enable: true
-
-    - name: ntpd.service
-      command: start
-      enable: true
-
     - name: kube2iam-iptables.service
       command: start
       runtime: true
@@ -88,6 +80,17 @@ coreos:
         Type=oneshot
         ExecStartPre=/usr/bin/mkdir -p /root/.docker
         ExecStart=/opt/bin/dockercfg.sh
+
+        [Install]
+        WantedBy=multi-user.target
+
+    - name: timesynced-enable-network-time.service
+      command: start
+      runtime: true
+      content: |
+        [Service]
+        Type=oneshot
+        ExecStart=/usr/bin/timedatectl set-ntp true
 
         [Install]
         WantedBy=multi-user.target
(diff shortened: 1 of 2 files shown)


linki:
The [CoreOS docs](https://coreos.com/os/docs/latest/configuring-date-and-timezone.html) about this use a slightly different approach.

Their approach is via `systemd-timesyncd`. And for the `ntpd` approach it seems putting config file is enough and one should mask `systemd-timesyncd`.

szuecs:
@linki I don't get it from their configuration example. Do you know what systemd-timesyncd is doing? From the systems I checked we had more than 0.1s difference in time, which is really not what we want for opentracing. Can you suggest something that adjust times, which is why I used ntpdate for?

mikkeloscar:
@szuecs afaict it should be enough to configure the amazon pools in `/etc/systemd/timesynced.conf`.

e.g. by writing the file:

```yaml
- path: /etc/systemd/timesynced.conf
  owner: root
  content: |
    [Time]
    NTP=0.amazon.pool.ntp.org 1.amazon.pool.ntp.org 2.amazon.pool.ntp.org 3.amazon.pool.ntp.org
```

szuecs:
@mikkeloscar the renaming of the ntp pools is not the issue I want to solve.
The issue is that a started ntp will not adjust the clock to the time of the other nodes, so the ntpdate is required to once get the time somehow "equal" (rolling updates will not do that correctly anyways).
I am not familiar with timesynced, but if you are or @linki is, then we should make sure it has the right configuration to not have the clock out of sync (0.1s diff is too much for opentracing-api to be really useful).

linki:
`systemd-timesyncd.service` is already running in our clusters and each node syncs time with some `*.coreos.pool.ntp.org`. Is 0.1s difference between nodes expected in this setup?

szuecs:
@linki I don't know expectations of timesyncd. We should somehow run ntpdate or timesyncd should do a similar thing, but right now it does not work as expected, that's why I created the PR. If you don't agree than please show a solution that use timesyncd to set the initial clock to a known state, similar to ntpdate.

linki:
`systemd-timesyncd.service` syncs the time from the configured endpoints similar to what this PR is about. It should already be enabled on our clusters so the question is: "Is a skew of 0.1s acceptable and if not, will this PR make it better?"

Maybe just changing the timeservers to amazon's is enough?

linki:
We have to do `timedatectl set-ntp true` to get `Network time on: yes` in `timedatectl status`.

https://wiki.archlinux.org/index.php/systemd-timesyncd
https://forums.gentoo.org/viewtopic-p-8115408.html#8115408

mikkeloscar:
:+1:

linki:
👍

PR #580, opened 2017-09-14
Add CoreOS's update operator

linki:
This adds CoreOS's update operator which is the original auto-restart service adapted to take advantage of Kubernetes. It coordinates updates across the cluster via node labels. It also uses taints/tolerations to prevent daemonsets from being scheduled on master nodes by default.

### Update Operator

Currently, we update the CoreOS AMI whenever we update the userdata which happens somewhat around once a week. However, the goal should be to make the CoreOS layer as thin as possible and run the same images across all clusters. This can be achieved by moving more parts into the environment (e.g. pulling arguments from etcd) and into Kubernetes itself (e.g. via self-hosted components).

In this scenario, the in-place updates will become more relevant as nodes get replaced less often by userdata changes. Another advantage of in-place updates is that they are faster in general as docker images don't need to be re-pulled.

When a new CoreOS image is realesed for the current channel, the agent marks each outdated node with "reboot-needed". The operator watches theses labels and allows one node at a time to reboot via setting another label on the node. A human can prevent nodes from being restarted with yet another label (see https://github.com/coreos/container-linux-update-operator/blob/master/pkg/constants/constants.go for a full list of available labels). The update operator drains and cordons the node before restarting it.

### Taints for master nodes

The PR also adds node taints for the master's which I think is much better to work with. If you want to run a pod on a master node you have to explicitely opt-in instead of opt-out. This makes much more sense as can be seen in the changes to the `skipper` daemonset.

### Caveats

`SchedulingDisabled` is removed from the master nodes and replaced by the taint. With the changes in this PR this works as expected and similar as before.

However, when `cordon`ing a node it gets correctly set to `SchedulingDisabled` but if for some reason the `update-agent` on this node restarts it makes it scheduleable again like after a successful reboot.

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -11,8 +11,6 @@ ssh_authorized_keys:
   - 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCqAikaHsZWMuJvKZphZPZG0fnKMvVCRfBAbIS6e0Y+YqM0PfsWgB5e4f5TrbisQHdKopbfZVwYIaV/NegEuinrYPKC7t2ese/HjxgjHR95zHOcDP19Cbo+xeyH8zbRd9K3iRSyCUSMNRw5NL6zN8JOSl12m8QWQA4hTjFTmt870fIT4RLxu9qGlbQipUm57E/SotsNC41MQ/PsLQzOAviKrkS1rei2vzRHzAcjz1Z7GT5oH+dFVUC66kKa0XWDvq+VtkRVoLvS2chrIPCgESeeZAyOKyiOoyJxFFFiMVK48MWDBBIYTIsHE0qs/RwBi9+8lQGiHK5Rpk2djcloO0c7 sandor.szuecs@zalando.de'
   - 'ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAINkh52Py+FvH9CRLDQg0gzvjEIrzwA45yMTXTsl2BVxV alexey.ermakov@zalando.de'
 coreos:
-  update:
-    reboot-strategy: "off"
   units:
     - name: etcd-member.service
       command: start
@@ -178,6 +176,9 @@ coreos:
 
         [Install]
         WantedBy=multi-user.target
+
+    - name: locksmithd.service
+      mask: true
 
 write_files:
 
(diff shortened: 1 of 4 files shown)


cluster/userdata-master.yaml
```diff
@@ -173,6 +172,9 @@ coreos:
         [Install]
         WantedBy=multi-user.target
 
+    - name: locksmithd.service
+      mask: true
```
linki, line 181: The `update-operator` takes over the role of `locksmithd`. `update-engine` which polls `coreos.com` for updates is running on each node.

cluster/userdata-master.yaml
```diff
@@ -123,8 +121,9 @@ coreos:
         --rkt-path=/usr/bin/rkt \
         --register-schedulable=false \
         --allow-privileged \
-        --node-labels=master=true \
+        --node-labels=node-role.kubernetes.io/master,node.alpha.kubernetes.io/role=master \
```
linki: These labels are taken from `kube-aws`.

With this you can better see the type in the node list:

```console
$ kubectl get nodes -L node.alpha.kubernetes.io/role
NAME                                             STATUS    AGE       VERSION           ROLE
ip-172-31-0-127.eu-central-1.compute.internal    Ready     9m        v1.7.5+coreos.0   worker
ip-172-31-1-130.eu-central-1.compute.internal    Ready     1h        v1.7.5+coreos.0   master
```

But still be able to filter without using a key/value pair

```console
$ kubectl get nodes -l node-role.kubernetes.io/master
NAME                                             STATUS    AGE       VERSION
ip-172-31-1-130.eu-central-1.compute.internal    Ready     1h        v1.7.5+coreos.0
ip-172-31-16-219.eu-central-1.compute.internal   Ready     1h        v1.7.5+coreos.0
```

We could think about dropping one of the labels.

mikkeloscar:
> The PR also adds node taints for the master's which I think is much better to work with. If you want to run a pod on a master node you have to explicitely opt-in instead of opt-out. This makes much more sense as can be seen in the changes to the skipper daemonset.

This also allows anyone to schedule applications on the master nodes. I don't think it's a huge issue as only malicious actors would use it, but worth mentioning.

linki:
@mikkeloscar That's right. I still think it's correct and that instead we would not allow unpriviliged users to use that particular toleration in their applications.

linki:
This feature hasn't found the necessary support to be merged. Closing.

PR #582, opened 2017-09-18
Add emergency-access-service

mikkeloscar:
This deploys the emergency-access-service to every cluster.

There is currently one problem with this approach related to the PlatformCredentialsSet.
Since we just apply all manifests in random order we might apply the PlatformCredentialsSet resource before the actual definition as defined in: https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/platformcredentialsset/thirdpartyresource.yaml

I tried forcing the thirdpartyresource to be created first, but even in this case the other manifest gets applied too fast, before Kubernetes really understands the issue.

This is not a problem for existing clusters, but will be an issue for every new cluster e.g. e2e tests..

I'll think about a way to solve this.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #583, opened 2017-09-19
Add critical-pod annotation

mikkeloscar:
Adds critical-pod annotation to mirror pods such that they won't be evicted by the kubelet in case of too few resources available on the node.

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/preemption/preemption.go#L233

This was forgotten in #553


linki:
👍

mikkeloscar:
:+1:

PR #584, opened 2017-09-19
custom opentracing instana POC skipper

szuecs:
Replace normal ingress skipper with the new skipper plugin opentracing flavour instana.
This test will probably show logs that show failing connections to instana-agent from skipper-ingress


szuecs:
:+1:

szuecs:
Already tested in stups-test and the former tests already were ok, now I rebased to dev and it would trigger again test run which does not makes so much sense.

mikkeloscar:
:+1:

PR #586, opened 2017-09-20
Update zmon-agent-core to 0.1-a30

aermakov-zalando:
This should allow us to see the resource usage of individual containers.


mikkeloscar:
:+1:

mikkeloscar:
Since we are getting ratelimited (which makes the tests fail) right now in the test account, I think it's fine to merge this simple change without running e2e tests first.

aermakov-zalando:
👍

PR #587, opened 2017-09-20
Update emergency-service

mikkeloscar:
Adds structured logging for requests and configurable token-url.

### DIFF
--- a/cluster/manifests/emergency-access-service/deployment.yaml
+++ b/cluster/manifests/emergency-access-service/deployment.yaml
@@ -16,6 +16,9 @@ spec:
       labels:
         application: emergency-access-service
         version: master-30
+      annotations:
+        kubernetes-log-watcher/scalyr-parser: |
+          [{"container": "emergency-access-service", "parser": "json-structured-log"}]
     spec:
       serviceAccountName: system
       containers:
--- a/cluster/manifests/emergency-access-service/deployment.yaml
+++ b/cluster/manifests/emergency-access-service/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: emergency-access-service
-    version: master-28
+    version: master-30
 spec:
   replicas: 1
   selector:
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

mikkeloscar:
Added custom scalyr log parser.

mikkeloscar:
:+1:

hjacobs:
:+1:

cluster/manifests/emergency-access-service/deployment.yaml
```diff
@@ -15,18 +15,22 @@ spec:
...
       - name: emergency-access-service
-        image: "pierone.stups.zalan.do/teapot/emergency-access-service:master-28"
+        image: "pierone.stups.zalan.do/teapot/emergency-access-service:master-30"
         args:
         - --insecure-http
         - --community={{ .Owner }}
         - --cluster-id={{ .ID }}
         - --teams-api-url=https://teams.auth.zalando.com
         - --debug
+        - --token-url=https://info.services.auth.zalando.com/oauth2/tokeninfo
```
hjacobs: this should better be called "tokeninfo-url" (we have the convention of `TOKENINFO_URL`) to avoid confusion (Token Endpoint is something else in OAuth, it's the URL where access tokens are created)

mikkeloscar: This was inspired by what it's called in swagger:

```yaml
securityDefinitions:
  OAuth2:
    type: oauth2
    flow: password
    tokenUrl: https://info.services.auth.zalando.com/oauth2/tokeninfo
    scopes:
      uid: Unique identifier of the user accessing the service.
```

But I now realise that this is just wrong. We have been doing this in all our projects....

I'll fix it. But not in this PR.

hjacobs: Yes, exactly, the Swagger `tokenUrl` is not the Token Info URL (there is no such thing as "Token Info" in the OAuth RFC).

hjacobs: @mikkeloscar maybe you can add a comment in the K8s YAML (in this PR) to make it clear that we already know of the "confusing" option name and that it should be "tokeninfo"?

mikkeloscar: Added a comment.

mikkeloscar:
:+1:

ideahitme:
👍

PR #590, opened 2017-09-21
Update zmon agent dev to alpha

aermakov-zalando:
This is working on dev, let's roll out to alpha.


aermakov-zalando:
Let's just wait and deploy this with the emergency service update.

PR #596, opened 2017-09-25
set authentication ttl to 10 sec

ideahitme:
This sets authentication webhook cache TTL to 10 seconds. Authentication caching happens based on tokens, as used [here in kubernetes code](https://github.com/kubernetes/apiserver/blob/19667a1afc13cc13930c40a20f2c12bbdcaaa246/plugin/pkg/authenticator/token/webhook/webhook.go#L75)

This should make emergency access service more user-friendly by not forcing the user to wait for 2minutes (current default ttl) before new role is visible by API server


ideahitme:
👍

aermakov-zalando:
👍

hjacobs:
Can somebody give me an overview of the current authentication and authorization method and where caching is still performed?

PR #598, opened 2017-09-25
use new opentracing aware skipper without catchall route

szuecs:
I reverted the last change from us, which made it possible to use a different default http status code.
This introduced a new created catchall route for every ingress having not a catchall route, which might introduced a bug that happens if you have 2 ingress with the same hostname and different paths/service endpoints without catchall route.

Related issue in skipper: https://github.com/zalando/skipper/pull/437

Open problem: tracing instana module does not load correctly

    [APP]time="2017-09-26T15:48:28Z" level=fatal msg="module tracing_instana's InitTracer function has wrong signature"


szuecs:
:+1:

mikkeloscar:
:+1:

PR #601, opened 2017-09-26
Update zmon-worker deployment

mohabusama:
Changes:

- Increase number of processes to 32
- Increase number of replicas to 4
- Increase cpu requests/limit
- Upgrade zmon-worker to zv225 with EBS wrapper and Hipchat notification link

### DIFF
--- a/cluster/manifests/zmon-worker/deployment.yaml
+++ b/cluster/manifests/zmon-worker/deployment.yaml
@@ -25,11 +25,11 @@ spec:
           image: "pierone.stups.zalan.do/stups/zmon-worker:zv225"
           resources:
             limits:
-              cpu: 800m
+              cpu: 500m
               memory: 2Gi
             requests:
               cpu: 500m
-              memory: 1Gi
+              memory: 2Gi
 
           ports:
             - containerPort: 8080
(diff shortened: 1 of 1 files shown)


szuecs:
@mohabusama is zmon-worker CPU bound?
If so did you considered an hpa?

https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/

mohabusama:
@szuecs I wouldn't consider it CPU bound, but I only increased the requests/limits as I increased the number of processes. I would need to investigate this further.

We wanted to use HPA indeed, but haven't given it time yet. One option is to trigger based on redis queue length.

mikkeloscar:
:+1:

mikkeloscar:
> One option is to trigger based on redis queue length.

This would be good, but I think increasing the number of workers manually is good enough for now.

aermakov-zalando:
@mohabusama: please check the memory request & limits as well. It's not a good idea to keep them different unless you have bursty workloads, and zmon-worker has been running at 200% of request (and close to limit) recently.

aermakov-zalando:
According to the statistics, 32 processes doesn't work since the pods are getting OOM killed in a day or less, so more investigation is needed.

aermakov-zalando:
Also, HPA should definitely be investigated since the amount of resources varies a lot between the clusters. We don't want to have zmon-worker consume 32GB of memory in every single cluster just because one of them needs it.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #612, opened 2017-09-29
new version of emergency-access-service

ideahitme:
this makes sure debug role request cannot be approved by the request issuer and debug role can only be requested only for oneself


ideahitme:
👍

mikkeloscar:
Can you update the version labels as well? I think we forgot it a couple of times...

mikkeloscar:
:+1:

ideahitme:
👍

PR #615, opened 2017-10-01
Add readinessprobe to skipper deamonset

mikkeloscar:
Adds readinessprobe to the skipper deamonset to detect unhealthy skippers during a rolling update and automatically stop the update.

### DIFF
--- a/cluster/manifests/skipper/daemonset.yaml
+++ b/cluster/manifests/skipper/daemonset.yaml
@@ -57,3 +57,9 @@ spec:
           requests:
             cpu: 25m
             memory: 25Mi
+        readinessProbe:
+          httpGet:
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

Raffo:
Did you test this with the case where the skipper command is "wrong"? I guess the e2e tests don't cover this by default.

mikkeloscar:
> Did you test this with the case where the skipper command is "wrong"? I guess the e2e tests don't cover this by default.

I manually tested the case we saw were skipper was running with the `skipper skipper` command and thus serves no health endpoint. In this case the pod never gets into ready state and the rolling update stops.

Raffo:
👍

PR #619, opened 2017-10-02
use default dns policy for cluster autoscaler

ideahitme:
This is required in scenarios when kube-dns fails to be scheduled on a node, when there is no space available, in which case cluster-autoscaler fails to communicate with AWS ASG API, which causes a deadlock - kube-dns cannot be scheduled on a node, because not enough nodes. Cluster-autoscaler fails to add nodes, because kube-dns is not accessible


mikkeloscar:
Both tests got rate-limited in the delete step after the tests ran successfully.

mikkeloscar:
:+1:

ideahitme:
👍

PR #621, opened 2017-10-04
Replace KubeDNS with CoreDNS

linki:
Initial try to replace KubeDNS with CoreDNS. Works fine in my test setup.

It's based on the migration script here: https://github.com/coredns/deployment/tree/master/kubernetes

I took the same requests/limits as well as added readiness probe and the cluster proportial autoscaling.

TODO:
- [ ] Move docker image to Pierone
- [ ] Check the impact of migrating to CoreDNS for applications
- [ ] Get rid of all the kube-dns related stuff
- [ ] Re-add any existing metrics

### DIFF
--- a/cluster/manifests/coredns/cm-coredns.yaml
+++ b/cluster/manifests/coredns/cm-coredns.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: coredns
+  namespace: kube-system
+  labels:
+    application: coredns
+data:
+  Corefile: |
+    .:53 {
+        errors
+        log stdout
+        health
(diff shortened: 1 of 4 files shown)


szuecs:
@linki Can you show us benefits from a migration to CoreDNS?

linki:
@szuecs Functionality-wise I'm not sure if there are any. It could even be less reliable but someone has to find out I guess. Non-functional benefit is the much less configuration needed (no dnsmasq sidecars).

linki:
This feature hasn't found the necessary support to be merged. Closing.

I think there's still a desire to replace KubeDNS with CoreDNS as the default cluster-internal DNS provider amongst the Kubernetes community. If that becomes more evident and useful to prefer CoreDNS we can follow-up on this PR.

PR #616, opened 2017-10-02
Flannel v0.9.0

Raffo:
Initial experimental branch to update the flannel version with several bugfixes. Default CoreOS stable uses v0.7.1. 

Still need to figure out why flannel is started first with the wrong version: 

```
UUID		APP		IMAGE NAME							STATE	CREATED		STARTED		NETWORKS
12ef5fd2	etcd		quay.io/coreos/etcd:v3.1.6					running	12 minutes ago12 minutes ago
7a92bbbf	hyperkube	registry.opensource.zalan.do/teapot/hyperkube:v1.7.6_coreos.0	running	9 minutes ago9 minutes ago
bc1d0a7f	flannel		quay.io/coreos/flannel:v0.9.0					running	29 seconds ago29 seconds ago
ea727f7e	flannel		quay.io/coreos/flannel:v0.7.1					exited	29 seconds ago29 seconds ago
```

PR #623, opened 2017-10-04
Use etcd image version provided by channel configuration

linki:
This is more to raise awareness, keep track and ask how we want to deal with that in the future.

We overwrote the `ETCD_IMAGE_TAG` at some point in the past (I think it was due to [a bug in `v3.0.x`](https://github.com/zalando-incubator/kubernetes-on-aws/commit/ddd67156d1c23575d017c4906aad4fbb76ffebd4)).

However, by now the version provided in the stable channel is newer than what we define (https://coreos.com/releases/#1465.8.0):

```
# /usr/lib/systemd/system/etcd-member.service
Environment="ETCD_IMAGE_TAG=v3.1.8"
...
# /etc/systemd/system/etcd-member.service.d/40-etcd-gateway.conf
[Service]
Environment="ETCD_IMAGE_TAG=v3.1.6"
...
```

We'll face the same issue with [`flannel`](https://github.com/zalando-incubator/kubernetes-on-aws/pull/616) in the future.

I think it's not desireable to keep track of this manually and rather let CoreDNS's channels decide. However, this can lead to surprising updates of the version which could lead to issues (even in stable channel).

What's your opinion on that?

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -27,7 +27,6 @@ coreos:
         - name: 40-etcd-gateway.conf
           content: |
             [Service]
-            Environment="ETCD_IMAGE_TAG=v3.1.6"
             Environment="ETCD_OPTS=gateway start --listen-addr=127.0.0.1:2379 --endpoints={{ ETCD_ENDPOINTS }}"
 
     - name: docker.service
--- a/cluster/userdata-worker.yaml
+++ b/cluster/userdata-worker.yaml
@@ -27,7 +27,6 @@ coreos:
         - name: 40-etcd-gateway.conf
           content: |
             [Service]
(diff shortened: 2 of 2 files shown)


mikkeloscar:
I think it makes sense to NOT hardcode the version unless we see a problem with the version provided by coreos. You are right that the etcd version was only hardcoded because we needed >=3.1.6 for the etcd-member service to work correctly with systemd.

So for me this is :+1:

Raffo:
+1 on my side as well. Let's queue this for testing. Labelling `do not merge` for now.

mikkeloscar:
:+1:

szuecs:
@linki this was successfully tested, yesterday, right?

linki:
👍

PR #626, opened 2017-10-05
Alpha to beta

szuecs:
* Add readiness probes to skipper DaemonSet
* Emergency Service update
* remove systemd lifecycle dependency from kube-node-ready to docker
* remove cfn-signal, because we use instead kube-node-ready


szuecs:
:+1:

mikkeloscar:
:+1:

PR #627, opened 2017-10-05
Add kube-proxy daemonset

mikkeloscar:
This is an effort to move the static kube-proxy pod out of the userdata and make it into a daemonset. The reason for this is to have more stuff managed by Kubernetes and have less node specific configuration in the config.

I kept the mirror pod in the userdata for now in order to have a smooth upgrade path. It was renamed to `kube-proxy-userdata` in order to ensure that the daemonset doesn't try to manage the mirror pods.

Still need to verify a few things before this is ready:

* [x] ~~Running both mirror and daemonset pods should not have any impact (e2e should be enough to verify this)~~ we ensure that only one of them is running
* [x] Try running without the mirror pods (tested with e2e tests on [this branch](https://github.com/zalando-incubator/kubernetes-on-aws/tree/kube-proxy-rm-mirror))

### DIFF
--- a/cluster/manifests/kube-proxy/daemonset.yaml
+++ b/cluster/manifests/kube-proxy/daemonset.yaml
@@ -6,8 +6,6 @@ metadata:
   labels:
     application: kube-proxy
     version: v1.7.9_coreos.0
-  annotations:
-    rkt.alpha.kubernetes.io/stage1-name-override: coreos.com/rkt/stage1-fly
 spec:
   selector:
     matchLabels:
--- a/cluster/manifests/kube-proxy/daemonset.yaml
+++ b/cluster/manifests/kube-proxy/daemonset.yaml
@@ -6,6 +6,8 @@ metadata:
   labels:
     application: kube-proxy
     version: v1.7.9_coreos.0
+  annotations:
+    rkt.alpha.kubernetes.io/stage1-name-override: coreos.com/rkt/stage1-fly
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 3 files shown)


linki:
Let's apply the node labels + nodeSelector approach described in https://github.bus.zalan.do/teapot/issues/issues/683 to safely rollout only one kube-proxy process per node at any point in time.

mikkeloscar:
> Let's apply the node labels + nodeSelector approach

@linki this is what I added in the latest commit or do you mean it should be done somewhat differently?

linki:
Sorry @mikkeloscar, didn't see that. All good.

I included a couple of changes I saw in https://github.com/kubernetes-incubator/kube-aws/pull/966/files as well.

cluster/manifests/kube-proxy/daemonset.yaml
```diff
@@ -6,8 +6,6 @@ metadata:
   labels:
     application: kube-proxy
     version: v1.7.9_coreos.0
-  annotations:
-    rkt.alpha.kubernetes.io/stage1-name-override: coreos.com/rkt/stage1-fly
```
linki: This is only used when Kubernetes runs via `rkt` as container runtime?

mikkeloscar: Yup: https://kubernetes.io/docs/getting-started-guides/rkt/

linki:
👍

mikkeloscar:
:+1:

PR #632, opened 2017-10-06
Add the PodCIDR controller

linki:
Populate Node's `PodCIDR` field with data from etcd (in order to switch from etcd-flannel to kube-flannel).

## Why

When migrating from an `etcd`-backed `flannel` to a Kubernetes API-backed `flannel` daemon it's necessary to keep both backends synchronized so that old nodes and already updated nodes don't enter a split brain situation resulting in a malfunctioning Pod network.

---

Check out the [README.md](https://github.bus.zalan.do/teapot/pod-cidr-controller/blob/fc87d046db367c8668a85f4383f6852eed44fcaa/README.md) if you want to know more.

### DIFF
--- a/cluster/manifests/pod-cidr-controller/deployment.yaml
+++ b/cluster/manifests/pod-cidr-controller/deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1beta1
 kind: Deployment
 metadata:
   name: pod-cidr-controller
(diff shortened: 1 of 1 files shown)


cluster/manifests/pod-cidr-controller/deployment.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: extensions/v1beta1
```
szuecs: extensions? apps!

hjacobs:
How do we proceed here? Is this safe to merge?

linki:
There should be no effect but I would wait until we start working on `kube-flannel` again, i.e. after BF.

linki:
I would like to get this in now so that the controller can run for a while. This way I can check whether it has correctly assigned the matching `PodCIDR`s from etcd to the nodes.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #645, opened 2017-10-16
fix the SLO metrics - use exp-decay

szuecs:
to fix the SLO metrics we have to use the updated version with exp-decay


aermakov-zalando:
👍

szuecs:
:+1:

PR #646, opened 2017-10-16
Metrics: use exponential decay

aermakov-zalando:
This should help with the SLOs and give us better data.


szuecs:
we already have https://github.com/zalando-incubator/kubernetes-on-aws/pull/645

PR #647, opened 2017-10-17
Dev to alpha

szuecs:
* externa-dns update from v0.4.4 to v0.4.6 which enables setting TTL for reference see: https://github.com/kubernetes-incubator/external-dns/releases/tag/v0.4.5 and https://github.com/kubernetes-incubator/external-dns/releases/tag/v0.4.6
* skipper-ingress with upgrade header support, which supports protocols like websockets or SPDY
* skipper-ingress update uses -metrics-exp-decay-sample flag which will fix calculation of SLOs for not high frequently used services

### DIFF
--- a/cluster/manifests/external-dns/deployment.yaml
+++ b/cluster/manifests/external-dns/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: external-dns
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
👍

szuecs:
:+1:

PR #648, opened 2017-10-18
Add pdb-controller deployment

mikkeloscar:
Adds https://github.com/mikkeloscar/pdb-controller for generating default
Pod Disruption Budgets which is needed as part of the new update procedure.


linki:
👍

Raffo:
👍

PR #649, opened 2017-10-19
pdb-controller needs more privileges

linki:
The controller needs to run with the system service account in order to create PDBs.


Raffo:
👍

Raffo:
No real need to run e2e here as they will not spot anything as they didn't do in the previous PR #648

linki:
👍

PR #657, opened 2017-10-27
update skipper to latest version without instana

szuecs:
drop opentracing support from skipper, because POC will be stopped soon

### DIFF
--- a/cluster/manifests/skipper/daemonset.yaml
+++ b/cluster/manifests/skipper/daemonset.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: skipper-ingress
-    version: v0.10.1-alpha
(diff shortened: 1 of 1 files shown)


hjacobs:
Shouldn't we roll forward instead of backward? Would this remove the rate limit feature?

mikkeloscar:
@hjacobs this is the latest skipper release, I assume the versioning was just different/wrong for the other image with tracing support.

szuecs:
@hjacobs yes I did this 0.10.x-alpha, because it was a skipper bundled with opentracingAPI support, which this PR will not use anymore. So it makes sense to go back to 0.9.y., in my opinion.

szuecs:
:+1:

ideahitme:
:+1:

PR #659, opened 2017-10-30
robuster cronjobs

szuecs:
add more robust options to kube-job-cleaner cronjob
add cron namespace and add limits and quotas to omit problems with cronjobs.

### DIFF
--- a/cluster/manifests/kube-job-cleaner/cronjob.yaml
+++ b/cluster/manifests/kube-job-cleaner/cronjob.yaml
@@ -11,8 +11,6 @@ spec:
   concurrencyPolicy: Forbid
   failedJobsHistoryLimit: 1
   jobTemplate:
-    metadata:
-      creationTimestamp: null
     spec:
       activeDeadlineSeconds: 120
       template:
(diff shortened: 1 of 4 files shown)


szuecs:
:+1:

cluster/manifests/kube-job-cleaner/cronjob.yaml
```diff
@@ -8,8 +8,13 @@ metadata:
     version: "0.3"
 spec:
   schedule: "17 * * * *"
+  concurrencyPolicy: Forbid
+  failedJobsHistoryLimit: 1
   jobTemplate:
+    metadata:
+      creationTimestamp: null
```
mikkeloscar: nit: Does this `creationTimestamp: null` do anything? Otherwise I would just remove it to keep it simple as this manifest could be used as example for others.

szuecs: good catch seems to be a leftover from dumping cronjob :)

szuecs:
:+1:

mikkeloscar:
:+1:

PR #660, opened 2017-11-01
Add etcd-backup cronjob

mikkeloscar:
This adds a Cronjob for making snapshots of etcd3 and pushing it to an S3 bucket once an hour.

Also allows ZMON `s3:ListBucket` access so we can make a zmon check that verifies that the backups are created every hour.

### DIFF
--- a/cluster/manifests/etcd-backup/cronjob.yaml
+++ b/cluster/manifests/etcd-backup/cronjob.yaml
@@ -1,0 +1,38 @@
+apiVersion: batch/v2alpha1
+kind: CronJob
+metadata:
+  name: etcd-backup
+  namespace: kube-system
+  labels:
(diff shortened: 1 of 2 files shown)


hjacobs:
This would stop working with kube-flannel, right? Maybe we can run it on master nodes?

Raffo:
It would be okay for now and we can move it on master nodes later on when we move to kube-flannel.

mikkeloscar:
I agree with @Raffo let's not solve it now but good to keep in mind.

mikkeloscar:
:+1:

linki:
👍

Raffo:
Apart from the random number for requests and limits, this looks good to me.

Raffo:
👍

PR #665, opened 2017-11-02
Update to Kubernetes v1.7.9

mikkeloscar:
* https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.7.md#v179
* https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.7.md#v178

Notable change:

> kubelet to master communication when doing node status updates now has a timeout to prevent indefinite hangs (#52176, @liggitt)

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -121,7 +121,7 @@ coreos:
       runtime: true
       content: |
         [Service]
-        Environment=KUBELET_IMAGE_TAG=v1.7.7_coreos.0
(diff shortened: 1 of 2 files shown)


Raffo:
Original upstream issue which makes us want to upgrade: https://github.com/kubernetes/kubernetes/issues/48638 . We had the problem described several times in production and `1.7.9` should contain a fix for it.

hjacobs:
:+1:

szuecs:
:+1:

PR #670, opened 2017-11-08
set logging level of controller manager to 7

ideahitme:
this increases the logging level of the controller manager to monitor issues with volume mounts. Didn't set it to any higher number because there it starts to dump all request and with leader election it is a huge logging load.


ideahitme:
👍

szuecs:
:+1:

PR #675, opened 2017-11-09
update zmon-worker to zv227 ￼…

vetinari:
same as pushed to appliances


szuecs:
test fails, because we got an external PR, maybe we should just merge it.

szuecs:
@mikkeloscar @Raffo any opinions on that?

szuecs:
@vetinari why we should merge it, maybe you can explain what will be changed and if this is required for black friday preparation?

szuecs:
no response

PR #681, opened 2017-11-10
fix skipper-ingress edge case in predicate/filter handling 

szuecs:
fix >1 ingress with the same host header has custom predicates and filters it would only recognize them in the first route


szuecs:
:+1:

mikkeloscar:
:+1:

PR #666, opened 2017-11-02
update ingress-controller

szuecs:
we have to follow the upgrade guide or we break https://github.com/zalando-incubator/kube-ingress-aws-controller

PR #684, opened 2017-11-13
Remove ClusterID tag

mikkeloscar:
Since #666 we no longer depend on the custom `ClusterID` tag and can
instead use the official `kubernetes.io/cluster/<cluster-id>` tag.

This is not of high priority and can wait till after BF.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #687, opened 2017-11-16
Dev to db alpha

szuecs:
* [feature] "cron" namespace to run cronjobs in to limit problems causing from jobs
* [feature] cluster-autoscaler support >50 nodes
* [feature] emergency-access-service handle jira ticket, metrics endpoint, add uid in error message
* [feature] etcd-backup cronjob
* [feature] skipper-ingress arbitrary Filters and Predicates can be used from ingress
* [feature] scalyr preparation for API switch from us to eu
* [feature] pdb-controller to support stateful rolling node updates
* remove instana POC
* update kubernetes from 1.7.7 to 1.7.9
  * Ignore pods marked for deletion that exceed their grace period in ResourceQuota (#46542, @derekwaynecarr)
  * kubelet to master communication when doing node status updates now has a timeout to prevent indefinite hangs (#52176, @liggitt)
  * When performing a GET then PUT, the kube-apiserver must write the canonical representation of the object to etcd if the current value does not match. That allows external agents to migrate content in etcd from one API version to another, across different storage types, or across varying encryption levels. This fixes a bug introduced in 1.5 where we unintentionally stopped writing the newest data. (#48394, @smarterclayton)
  * Use separate client for leader election in scheduler to avoid starving leader election by regular scheduler operations. (#53793, @wojtek-t)
  * Address a bug which allowed the horizontal pod autoscaler to allocate desiredReplicas > maxReplicas in certain instances. (#53690, @mattjmcnaughton)
  * BugFix: Exited containers are not Garbage Collected by the kubelet while the pod is running (#53167, @dashpole)


szuecs:
:+1:

szuecs:
:+1:

mikkeloscar:
:+1:

PR #693, opened 2017-11-20
Bump --node-monitor-grace-period as well

aermakov-zalando:
`pod-eviction-timeout` is ignored when using taint-based evictions apparently.

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -526,6 +526,7 @@ write_files:
           - --cloud-config=/etc/kubernetes/cloud-config.ini
           - --feature-gates=ExperimentalCriticalPodAnnotation=true,TaintBasedEvictions=true
           - --use-service-account-credentials=true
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
👍

aermakov-zalando:
This is the channel we use when upgrading the etcd to v3. The problem is that once etcd is up, there's a race condition: if kubelet manages to send a heartbeat first, the node is unaffected by the upgrade. If, however, controller-manager manages to notice the node first, it'll consider the node unhealthy (because from its POV the node was down for 10 minutes already) and reschedule all pods on other nodes, which is something we'd like to avoid. We initially tried to avoid this by setting `pod-eviction-timeout` to 1 hour, but apparently this setting is completely ignored when using taint-based evictions.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #704, opened 2017-11-23
Update zmon-worker

mohabusama:
Changes:

**zmon-worker**

- Send extra kubernetes related fields in check results: ``application``, ``version``, ``cluster_alias``, ``alias``.
- Fields come from entities discovered by ZMON kubernetes [agent](https://github.com/zalando-zmon/zmon-agent-core).

### DIFF
--- a/cluster/manifests/zmon-worker/deployment.yaml
+++ b/cluster/manifests/zmon-worker/deployment.yaml
@@ -12,7 +12,7 @@ spec:
     metadata:
       labels:
         application: zmon-worker
-        version: "zv227"
+        version: "zv228"
       annotations:
         iam.amazonaws.com/role: "{{ .LocalID }}-app-zmon"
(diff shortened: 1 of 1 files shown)


hjacobs:
@mohabusama can you explain the different fields? How are they filled? What is the benefit of "sending" them?

mikkeloscar:
> @mohabusama can you explain the different fields? How are they filled? What is the benefit of "sending" them?

These fields will be added as tags in kariosdb so we can use them for queries in grafana.

hjacobs:
How are they filled? What is the meaning of "alias" (the other fields I can guess)?

mikkeloscar:
`alias` is the cluster alias. It's there to be compatible with all the kubernetes entities which have alias defined.

szuecs:
:+1:

hjacobs:
:+1:

PR #705, opened 2017-11-27
Sync alpha with blackfriday changes

linki:
So we don't get an old version of the etcd backup when switching channels back.


mikkeloscar:
:+1:

linki:
👍

PR #716, opened 2017-11-28
Run flannel with etcd backend as daemonset

linki:
This changes `flannel` to run as a Daemonset but still keeping the etcd backend in order to prepare for the migration, see [Migration strategy](https://github.bus.zalan.do/teapot/issues/issues/683#issue-251780).

This is the first stage of a two-stage rollout which deploys but doesn't activate the flannel daemonset, yet.

- [x] This is PR is intended to see if e2e passes. The actual rollout will require to split this PR in two stages.

### DIFF
--- a/cluster/manifests/flannel/daemonset.yaml
+++ b/cluster/manifests/flannel/daemonset.yaml
@@ -1,0 +1,81 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: kube-flannel
+  namespace: kube-system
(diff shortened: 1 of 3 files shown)


linki:
e2e for daemonset flannel passed.

Changed the PR to make the first stage of the multi-stage rollout:
* deploy flannel daemonset with non-matching node-selector <=
* remove node-flannel while changing the node-labels

gargravarr:
:+1:

linki:
👍

PR #717, opened 2017-11-29
Activate Daemonset kube-proxy

linki:
Stage two of rolling out kube-proxy as a daemonset ([Stage one](https://github.com/zalando-incubator/kubernetes-on-aws/pull/627))

This rolls all nodes. The nodes don't have kube-proxy in the userdata and they can be identified so by checking the node labels. Each new node will instead run the daemonset kube-proxy setup in [stage one](https://github.com/zalando-incubator/kubernetes-on-aws/pull/627) due to the node-affinity now matching. The dormant daemonset kube-proxy will slowly be rolled-in. At last, the node-affinity is removed from the daemonset.


linki:
@mikkeloscar Any blockers?

mikkeloscar:
My only blocker was to make sure we are aware to not merge this and then merge everything from dev into alpha at some point. If you take care to setup the right PRs for dev -> alpha so it's in two steps, then it's fine by me :+1:

linki:
Tested and found one issue:
* a worker pod cannot connect to a master pod (with hostNetwork) via a Service IP, have to check whether this is a regression

linki:
Same behaviour on `stups-test`.

How to reproduce:
* remove `nodeAffinity` from `skipper-ingress` (will start skipper on masters as well)
* hit skipper-ingress cluster ip from a pod on a worker
* see some requests working and then hanging.

linki:
👍

mikkeloscar:
:+1:

PR #721, opened 2017-11-30
Change etcd instance type to m4.large

mikkeloscar:
Changes the instance type to m4.large which is what we currently use until we get a reply from AWS regarding the t2.mediums having issues.


mikkeloscar:
:+1:

PR #723, opened 2017-11-30
role and policy and deployement

szuecs:
This will enable users to use static egress for defined targets in configmap with label egress=static

### DIFF
--- a/cluster/manifests/kube-static-egress-controller/deployment.yaml
+++ b/cluster/manifests/kube-static-egress-controller/deployment.yaml
@@ -20,10 +20,6 @@ spec:
         scheduler.alpha.kubernetes.io/critical-pod: ''
         iam.amazonaws.com/role: "{{ .LocalID }}-static-egress-controller"
     spec:
-      serviceAccountName: system
-      tolerations:
-      - key: CriticalAddonsOnly
-        operator: Exists
       containers:
       - name: controller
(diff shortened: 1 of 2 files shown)


cluster/manifests/kube-static-egress-controller/deployment.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: extensions/v1beta1
```
mikkeloscar: apps/v1beta1 :)

szuecs: yeah, sure :)

cluster/manifests/kube-static-egress-controller/deployment.yaml
```diff
@@ -0,0 +1,48 @@
...
+  template:
+    metadata:
+      labels:
+        application: kube-static-egress-controller
+        version: v0.0.3
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
+        iam.amazonaws.com/role: "{{ .LocalID }}-static-egress-controller"
+    spec:
+      serviceAccountName: system
```
mikkeloscar: The controller only needs to read from the API right? In that case I would drop this and just use the default service account which has readOnly access.

szuecs:
:+1:

linki:
👍

PR #725, opened 2017-12-04
Activate Flannel Daemonset deployment

linki:
Stage two of rolling out `flannel` as a daemonset, but still with etcd backend ([Stage one](https://github.com/zalando-incubator/kubernetes-on-aws/pull/716))

This rolls all nodes. The nodes don't have flannel in the userdata anymore and they can be identified so by checking the node labels (flannel=local). Each new node will instead run an instance of flannel described by the daemonset setup in [stage one](https://github.com/zalando-incubator/kubernetes-on-aws/pull/716) due to the node-affinity now matching. The dormant flannel daemonset will slowly be rolled-in. At last, the node-affinity is removed from the daemonset.

### DIFF
--- a/cluster/manifests/flannel/daemonset.yaml
+++ b/cluster/manifests/flannel/daemonset.yaml
@@ -66,15 +66,6 @@ spec:
         effect: NoExecute
       - operator: Exists
         key: CriticalAddonsOnly
-      affinity:
-        nodeAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-            nodeSelectorTerms:
-            - matchExpressions:
-              - key: flannel
-                operator: NotIn
-                values:
-                - local
       volumes:
       - name: runflannel
         hostPath:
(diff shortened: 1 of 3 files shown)


linki:
Tested and found one issue:
* two master pods cannot reach each other, have to check whether this is a regression

What happens when kubelet dies and cannot reach api server afterwards
* kube-proxy will not run as kubelet doesn't know about it
* iptables rules stay the same which should be the exact behaviour as of now

linki:
> two master pods cannot reach each other, have to check whether this is a regression

same behaviour on `stups-test` so I guess not.

linki:
This has been running on `teapot` for two weeks without issues to my knowledge. Can we merge it to `dev`?

linki:
I see, there's one un-green test...

linki:
👍

mikkeloscar:
:+1:

linki:
@mikkeloscar Need a :+1: again, there was a conflict with dev.

szuecs:
retriggered tests

linki:
👍

gargravarr:
:+1:

PR #756, opened 2017-12-16
Configure kube-proxy via config file (configmap)

mikkeloscar:
Configures kube-proxy via a config file (provided as configmap) instead of using flags which are deprecated.

```
W1215 09:33:56.758561       1 server.go:190] WARNING: all flags other than --config, --write-config-to, and --cleanup-iptables are deprecated. Please begin using a config file ASAP.
```


mikkeloscar:
The config version used depends on Kubernetes v1.9.0, will wait for that before this is ready to test again.

mikkeloscar:
:+1:

linki:
👍

PR #762, opened 2017-12-18
Don't configure kubelet to read from /etc/kubernetes/manifests on wor…

mikkeloscar:
…ker nodes

Since we no longer have any manifests on the node of worker nodes the
kubelet doesn't need to read from it.

From the kubelet logs:
```
Unable to read manifest path "/etc/kubernetes/manifests": path does not exist, ignoring
```


mikkeloscar:
:+1:

hjacobs:
:+1:

PR #764, opened 2017-12-19
Update authnz webhook to v0.3.8

mikkeloscar:
Adds a realm group when authenticating users such that we can track the realm in audit events.


mikkeloscar:
:+1:

Raffo:
👍

PR #767, opened 2017-12-20
Add the audittrail adapter

linki:
Adds the audittrail adapter


Raffo:
I added some of the config as taken from the kubernetes repo which contains some interesting observation: https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/configure-helper.sh#L734

szuecs:
@Raffo why this should be a daemonset?

Raffo:
@szuecs Because we decided not to have the adapter in the apiserver pod and we need the apiserver container to know the adapter's URL. Originally in this PR it was a deployment with a service, but this doesn't work as our apiserver is using `hostnetwork: true`. In this way, I solved the issue by changing it to a daemonset (running only on master nodes) and have each apiserver talk only to the local adapter. Feel free to change it if you want to implement it in a different way.

cluster/manifests/audittrail-adapter/daemonset.yaml
```diff
@@ -25,8 +25,10 @@ spec:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
             - matchExpressions:
-              - key: master
```
Raffo: Just note that this is also what we use in the skipper daemonset, I did it for consistency, maybe you can change it there as well (https://github.com/zalando-incubator/kubernetes-on-aws/blob/7acbcb75a1e30cba5c51254954fd231774a40b45/cluster/manifests/skipper/daemonset.yaml#L33 )

mikkeloscar: Done #776

mikkeloscar:
:+1:

szuecs:
:+1:

PR #768, opened 2017-12-22
Added docker watchdog

Raffo:
This PR adds a silly docker restarter in a condition when docker is stuck. We have been experiencing problems with docker 17 and Kubernetes 1.8, similarly to what documented in this issue: https://github.com/moby/moby/issues/33820 . 

This will help by restarting docker when docker cannot anymore terminate containers reliably.

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -210,6 +210,28 @@ coreos:
         [Install]
         WantedBy=multi-user.target
 
+    - name: docker-watchdog.service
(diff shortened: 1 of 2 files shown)


Raffo:
We merged the downgrade of Docker in #769, keeping this open in the meantime if the problem will still be there after the upgrade.

mikkeloscar:
We will close this since the problem was solved by the docker 1.12 downgrade so far.

PR #769, opened 2017-12-27
Downgrade Docker to 1.12

aermakov-zalando:
We've had a lot of issues with 17.09, so let's downgrade.

Since we're still using cloud-config which runs very late in the boot process, we need to reboot once after creating the config file.


aermakov-zalando:
👍

Raffo:
👍

Raffo:
Reference issue for the reader: https://github.com/moby/moby/issues/33820

PR #771, opened 2017-12-28
Revert allowPrivilegeEscalation changes

aermakov-zalando:
This doesn't work with Docker 1.12:

 * https://github.com/google/metallb/issues/21
 * https://github.com/moby/moby/issues/27983


Raffo:
This will leave it to the default. Do we want to be explicit and set the value that work for us, maybe with a comment such that we don't "fix it" by mistake?

Raffo:
👍

aermakov-zalando:
👍

PR #773, opened 2017-12-28
Revert allowPrivilegeEscalation changes

Raffo:
This will make sure that we have this change to beta before rolling the latest version of our config in beta that will downgrade docker and cause problems.


aermakov-zalando:
👍

Raffo:
👍

PR #776, opened 2018-01-02
Use kubernetes.io/role label for nodeAffinity

mikkeloscar:
Uses the more "official" kubernetes.io/role label rather than our custom
master=true label.

To align with the node affinity used in #767


mikkeloscar:
:+1:

linki:
👍

PR #777, opened 2018-01-02
Prevent scheduling on master nodes with taints

aermakov-zalando:
Use a taint to prevent unwanted stuff from running on the master nodes. It's still possible to work around this with tolerations, but at least you'll have to be explicit now. See https://github.com/kubernetes/kubernetes/pull/41835 and related issues for the discussion of the taint & label names.

### DIFF
--- a/cluster/manifests/kube-proxy/daemonset.yaml
+++ b/cluster/manifests/kube-proxy/daemonset.yaml
@@ -28,6 +28,8 @@ spec:
         effect: NoExecute
       - operator: Exists
         key: CriticalAddonsOnly
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
       hostNetwork: true
       containers:
       - name: kube-proxy
--- a/cluster/manifests/kube2iam/daemonset.yaml
+++ b/cluster/manifests/kube2iam/daemonset.yaml
@@ -23,6 +23,8 @@ spec:
       tolerations:
       - key: CriticalAddonsOnly
         operator: Exists
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
       hostNetwork: true
       containers:
       - image: registry.opensource.zalan.do/teapot/kube2iam:0.7.0
--- a/cluster/manifests/logging-agent/daemonset.yaml
+++ b/cluster/manifests/logging-agent/daemonset.yaml
@@ -26,6 +26,8 @@ spec:
       tolerations:
       - key: CriticalAddonsOnly
         operator: Exists
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
       initContainers:
       - name: init-scalyr-config
         image: registry.opensource.zalan.do/stups/alpine:3.5-cd10
--- a/cluster/manifests/prometheus-node-exporter/daemonset.yaml
+++ b/cluster/manifests/prometheus-node-exporter/daemonset.yaml
@@ -26,6 +26,8 @@ spec:
       tolerations:
       - key: CriticalAddonsOnly
         operator: Exists
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
       containers:
       - image: registry.opensource.zalan.do/teapot/prometheus-node-exporter:v0.14.0
         name: prometheus-node-exporter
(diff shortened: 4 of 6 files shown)


cluster/manifests/kube-proxy/daemonset.yaml
```diff
@@ -28,6 +28,8 @@ spec:
         effect: NoExecute
       - operator: Exists
         key: CriticalAddonsOnly
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
```
szuecs, line 32: why?

cluster/manifests/kube2iam/daemonset.yaml
```diff
@@ -23,6 +23,8 @@ spec:
       tolerations:
       - key: CriticalAddonsOnly
         operator: Exists
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
```
szuecs, line 27: why?

cluster/manifests/logging-agent/daemonset.yaml
```diff
@@ -26,6 +26,8 @@ spec:
       tolerations:
       - key: CriticalAddonsOnly
         operator: Exists
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
```
szuecs, line 30: why?

cluster/manifests/prometheus-node-exporter/daemonset.yaml
```diff
@@ -26,6 +26,8 @@ spec:
       tolerations:
       - key: CriticalAddonsOnly
         operator: Exists
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
```
szuecs, line 30: why?

cluster/manifests/kube-proxy/daemonset.yaml
aermakov-zalando: So the corresponding pod would be scheduled, see example at [link](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#how-daemon-pods-are-scheduled).

cluster/manifests/kube2iam/daemonset.yaml
aermakov-zalando: So the corresponding pod would be scheduled, see example at [link](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#how-daemon-pods-are-scheduled).

cluster/manifests/logging-agent/daemonset.yaml
aermakov-zalando: So the corresponding pod would be scheduled, see example at [link](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#how-daemon-pods-are-scheduled).

mikkeloscar:
Seems like there is no consensus on the label names (the references PR), but it makes sense to use what's in kubeadm as that's as close to "official" as possible.

LGTM.

cluster/manifests/prometheus-node-exporter/daemonset.yaml
aermakov-zalando: So the corresponding pod would be scheduled, see example at [link](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#how-daemon-pods-are-scheduled).

aermakov-zalando:
@mikkeloscar I think `node-role.kubernetes.io/master` is what most of the tooling converged on, so it makes sense for us to use it as well. I've kept the old labels in case somebody still relies on them.

mikkeloscar:
:+1:

cluster/manifests/kube-proxy/daemonset.yaml
szuecs: ah ok, the config "sounded" exactly the opposite

szuecs:
:+1:

szuecs:
feel free to merge

aermakov-zalando:
👍

linki:
@aermakov-zalando can we have a follow-up PR to remove the `master=true` label?

PR #778, opened 2018-01-02
Add tolerations for daemonsets running on masters

aermakov-zalando:
#777 has to be rolled out in two parts, since otherwise kube-node-ready and kube2iam won't be able to run on the new master nodes.


aermakov-zalando:
👍

linki:
👍

PR #780, opened 2018-01-03
fix egress controller finding dmz route tables

szuecs:
static egress found also internal route tables and this mixed up static routes


mikkeloscar:
:+1:

szuecs:
:+1:

PR #790, opened 2018-01-05
Fetch gerry credentials once before starting zmon-worker

linki:
This is a quick fix for https://github.com/zalando-zmon/zmon-worker/issues/270 which haunted pitchfork and possibly others.

Using an `initContainer` here is similar to how taupage makes sure credentials are loaded before anything else: https://github.com/zalando-stups/taupage/blob/Taupage-AMI-20180102-114355/runtime/opt/taupage/init.d/02-start-berry.sh#L7


mikkeloscar:
:+1:

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #794, opened 2018-01-09
Remove limits for master components

aermakov-zalando:
The limits on Kubernetes system components, especially the API server, are very restrictive and don't really make sense. We probably don't want to limit it to 20% of a single CPU and then resize every time we change the instance type. Let's just remove the limits for now and add them back only if we run into issues.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #795, opened 2018-01-11
Dev to alpha

szuecs:
* remove limits for master components
* flannel as daemonset
* fix zmon crash on start with initContainers

### DIFF
--- a/cluster/manifests/flannel/daemonset.yaml
+++ b/cluster/manifests/flannel/daemonset.yaml
@@ -66,15 +66,6 @@ spec:
         effect: NoExecute
       - operator: Exists
         key: CriticalAddonsOnly
-      affinity:
-        nodeAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-            nodeSelectorTerms:
-            - matchExpressions:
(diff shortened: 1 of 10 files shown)


cluster/manifests/flannel/daemonset.yaml
```diff
@@ -66,15 +66,8 @@ spec:
...
-        nodeAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-            nodeSelectorTerms:
-            - matchExpressions:
-              - key: flannel
-                operator: NotIn
-                values:
-                - local
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
```
linki, line 70: I believe this toleration must be applied before rolling the nodes.

linki: No, scratch that. Normally yes, but the Pod also has a "tolerate everything" toleration above. This is therefore not doing anything and was just added to be similar to the other DaemonSets.

linki:
👍

aermakov-zalando:
👍

PR #797, opened 2018-01-11
Add secret with cluster-inf secrets to default ns

mikkeloscar:
This adds a secret `cluster-infrastructure-secrets` to the default
namespace with standard credentials like scalyr-access-key and zmon-sql
credentials such that they can be consumed by other infrastructure
services like postgres-operator.

ACID needs this for the postgres-operator such that they can automatically ship logs of the postgres clusters to scalyr (not just the stdout logs which the logging-agent ship) and they also need the zmon-sql credentials in order to create a user in the postgres clusters which zmon can use for running sql checks.

### DIFF
--- a/cluster/manifests/infrastructure-secrets/secret.yaml
+++ b/cluster/manifests/infrastructure-secrets/secret.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cluster-infrastructure-secrets
+  namespace: default
+type: Opaque
+data:
+  zmon-sql-user: "{{ .ConfigItems.zmon_worker_plugin_sql_user | base64 }}"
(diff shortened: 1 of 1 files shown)


Jan-M:
:+1:

mgomezch:
Will these secrets be correctly populated in all clusters, including existing clusters?  Some appear to have empty contents in the Cluster Registry API, e.g.

```
$ echo; if [[ '' == "$(zttp cluster-registry.stups.zalan.do/kubernetes-clusters?alias=corptech | jq -r '.items[].config_items.zmon_worker_plugin_sql_pass')" ]]
then
  echo 'empty string'
fi

empty string
```

Others seem to have values that look more like placeholders, e.g.

```
echo; zhttp cluster-registry.stups.zalan.do/kubernetes-clusters?alias=merchant-center-test | jq -r '.items[].config_items.zmon_worker_plugin_sql_pass'

robot_zmon_merchant_center_test_pass
```

mikkeloscar:
:+1:

PR #801, opened 2018-01-15
Enable Kubernetes Backend for Flannel

linki:
This modifies the flannel Daemonset to switch its backend to the `Node` resource of the Kubernetes API server and allows to prevent workers from accessing etcd at all.

The new flannels will pick up their subnet allocations from the API server and - with the help of [PodCIDR controller](https://github.bus.zalan.do/teapot/pod-cidr-controller) - will find the same allocations as are currently present in etcd. To make this work the update is required to happen without a worker node update so that previous node allocations stay valid. Because of this, we migrated it to an etcd backed daemonset first in order to make the update without a userdata change.

During update, the responsibility of allocating PodCIDRs is shifted from flannel to Kubernetes' controller manager. Once the new controller manager becomes the leader it will allocate PodCIDRs and the PodCIDR controller as well as etcd access can be removed.

For more details on the entire upgrade process read the [Migration strategy proposal](https://github.bus.zalan.do/teapot/issues/issues/683)

Part of https://github.com/zalando-incubator/kubernetes-on-aws/pull/236

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -4,5 +4,14 @@
 #   namespace: kube-system
 #   kind: deployment
 
-pre_apply: [] # everything defined under here will be deleted before applying the manifests
+pre_apply: # everything defined under here will be deleted before applying the manifests
+# let kubernetes handle podCIDR allocations.
+- name: pod-cidr-controller
+  namespace: kube-system
+  kind: deployment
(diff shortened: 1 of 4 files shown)


aermakov-zalando:
:

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #802, opened 2018-01-16
Fix node draining during termination

aermakov-zalando:
Resolves the issues with the graceful termination unit:
 * use correct paths for kubeconfig
 * fix DNS issues in the container

### DIFF
--- a/cluster/userdata-master.yaml
+++ b/cluster/userdata-master.yaml
@@ -173,26 +173,7 @@ coreos:
         RemainAfterExit=true
         ExecStart=/bin/true
         TimeoutStopSec=120s
-        ExecStop=/bin/sh -c '/usr/bin/rkt run --insecure-options=image \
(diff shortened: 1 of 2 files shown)


linki:
👍

aermakov-zalando:
👍

mikkeloscar:
:+1:

linki:
Since we immediately restart the node (for the docker downgrade) a working node-drainer marks the node as Unschedulable and there's nothing to make it scheduleable again after reboot, right?

aermakov-zalando:
@linki the downgrade script is a dependency for `docker.service`, so the node-drainer isn't even started until after the first reboot.

linki:
Ah, nice :)

aermakov-zalando:
👍

linki:
👍

PR #803, opened 2018-01-16
etcd-proxy: bump version, remove restart limit

aermakov-zalando:
This should avoid the transient DNS errors we get sometimes.


linki:
👍

aermakov-zalando:
👍

PR #804, opened 2018-01-16
Run multiple kube-dns deployments

aermakov-zalando:
We run `kube-dns` backed by a `ClusterIP` service which is a simple round-robin to pod IPs without any failover. This means that whenever a node running one of kube-dns instances dies, applications start experiencing DNS resolution failures until the rest of the cluster catches up and removes the offending cluster IP from the nodes, which leads to a lot of alerts popping up and is rather inconvenient for the users. We can mitigate this somewhat by running 2 `kube-dns` deployments and configuring kubelet to use both of them instead of just one, the normal DNS retry logic in the runtimes/libraries will handle the failures for us.

The `kube-dns` and `kube-dns-backup` pods use an anti-affinity to avoid scheduling them on the same nodes unless absolutely necessary. I didn't do this for the autoscaler because affinities are expensive and it doesn't really matter if both of them are down for a short time. The cluster IP subnet is also extended to /16 to avoid overlap issues.

### DIFF
--- a/cluster/manifests/kube-dns/svc-kube-dns-backup.yaml
+++ b/cluster/manifests/kube-dns/svc-kube-dns-backup.yaml
@@ -1,0 +1,20 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: kube-dns-backup
+  namespace: kube-system
+  labels:
+    application: kube-dns-backup
+    kubernetes.io/cluster-service: "true"
+    kubernetes.io/name: "KubeDNS"
+spec:
+  selector:
+    application: kube-dns-backup
+  clusterIP: 10.3.0.11
+  ports:
+  - name: dns
+    port: 53
+    protocol: UDP
+  - name: dns-tcp
+    port: 53
+    protocol: TCP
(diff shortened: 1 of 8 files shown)


linki:
Thanks @aermakov-zalando.

The `-backup` is not really a backup since both DNS servers are used all the time. I would prefer to call them `kube-dns-a` and `kube-dns-b` or similar in this setup.

cluster/manifests/kube-dns/svc-kube-dns-backup.yaml
```diff
@@ -0,0 +1,20 @@
...
+  name: kube-dns-backup
+  namespace: kube-system
+  labels:
+    application: kube-dns-backup
+    kubernetes.io/cluster-service: "true"
+    kubernetes.io/name: "KubeDNS"
+spec:
+  selector:
+    application: kube-dns-backup
+  clusterIP: 10.3.0.11
```
mikkeloscar: This could potentially conflict with other services in existing clusters. I don't see a good way to roll this out other than fix it one at a time in case there are conflicts.

aermakov-zalando: @mikkeloscar I've extended the cluster IPs to /16 in the previous commit. Since they seem to be allocated randomly, the chances of getting a conflict are rather low.

aermakov-zalando:
@linki Renaming `kube-dns` would be rather annoying since I'll have to split it into 2 PRs. I'll rename the second one to `kube-dns-b` as you've suggested, then once it's deployed we can figure out if we want to rename the first one as well.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #806, opened 2018-01-18
Add secret with cluster-inf secrets to default ns

mikkeloscar:
This adds a secret `cluster-infrastructure-secrets` to the default
namespace with standard credentials like scalyr-access-key and zmon-sql
credentials such that they can be consumed by other infrastructure
services like postgres-operator.

cherry-picked to let ACID use these credentials with the postgres-operator.


Jan-M:
:+1:

mikkeloscar:
:+1:

PR #813, opened 2018-01-23
Promote alpha to beta

linki:
This is an out-of-order 😉  update from alpha to beta. It includes all changes from alpha with the exception of switching the flannel backend from etcd to kubernetes.

Note, this is unsafe as this particular combination isn't running in alpha or dev at this point. It was requested to speed up the cluster update by combining all pending userdata changes in one update.

Notable changes:
* Removing ssh key for raffaele.di.fazio@zalando.de
* disable image-policy simulation mode
* Fix node draining during termination
* etcd-proxy: bump version, remove restart limit
* Increase the service IP range to /16
* Run 2 deployments for kube-dns so applications
* update skipper and use SourceFromLast() predicate
* use a newer and compliant alpine image for logging agent


mikkeloscar:
:+1:

linki:
👍

PR #820, opened 2018-01-25
Handle special zmon config for db cluster

mikkeloscar:
Add special config for zmon components when applied to the DB cluster.

This will allow us to get rid of the db-alpha branch and should be a temporary fix until ACID switches to the in-kubernetes zmon appliance.


mikkeloscar:
:+1:

linki:
👍

linki:
@mikkeloscar sweet 👍

PR #830, opened 2018-02-05
Disable audit events from credentials-provider

mikkeloscar:
Disables audit events from credentials-provider as agreed with Torch. It creates most of the current events which are less important since it's an infrastructure component.


mikkeloscar:
:+1:

linki:
👍

PR #832, opened 2018-02-05
Drop emergency access service from test clusters

mikkeloscar:
Drops the emergency access service from all test clusters where it can't be used anyway. This will save us an ALB in each test cluster where no other ingress resources are defined.

Also bumps the emergency-access-service to version: master-49 which includes support for the special Zalando Payments community format.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #835, opened 2018-02-06
Update audittrail-adapter to use new schema

mikkeloscar:
Updates the audittrail schema to 1.2 which is less strict on the type of the requestObject field. Before it was restricted to objects but that is too strict since the value can also be an array.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #838, opened 2018-02-07
Switch to CLC/Ignition based userdata

mikkeloscar:
Migrate userdata from cloud-config to [Container Linux Config/ignition](https://coreos.com/ignition/docs/latest/).

The point of switching from cloud-config to ignition is that it solves the following problems:

* We can have more than 16K userdata by putting it on s3 (allowing us to experiment with stuff like aggregated API servers where we need more configuration on the master nodes :D).
* Ignition runs earlier in the boot process of Coreos allowing a more predictable startup process. In cloud-config the systemd units are enabled AND started in the order they appear in config. In ignition the units are only enabled and systemd handles the dependencies in a reliable way.
* We don't need the hack with restarting the nodes to downgrade to docker 1.12. "It just works" (tm) **This fixes a current problem where some nodes doesn't come up correctly after being restarted**

Depends on changes in CLM, see internal PR.


hjacobs:
What's the benefit/impact of this and what are the implications? Why the change?

mikkeloscar:
@hjacobs added some reasons to the description.

mikkeloscar:
This is good to go from my side. It has been running in teapot and stups-test since last week. We found one issue with the hostname not being set correctly causing problems with kube-proxy, but this was fixed.

mikkeloscar:
:+1:

aermakov-zalando:
We can revert to the old metadata if we find any issues, so let's merge it.

aermakov-zalando:
👍

PR #841, opened 2018-02-12
Switch to the official Kubernetes autoscaler

aermakov-zalando:
The cluster autoscaler we're using has a lot of issues:
 * Scaling doesn't consider the actual nodes, only the total usage and node size. Deploying 10 pods each requesting 75% of a node's CPU would lead to just 8 nodes being created, not 10.
 * Only CPU and memory requests are taken into account. Pod affinity/anti-affinity, volume/zone restrictions, node selectors are completely ignored.
 * Multi-ASG setups are not supported (they might work, but I don't think anyone has ever tested it).
 * Nodes are terminated by decreasing the desired size. Since we don't have any special setup in place, this will ignore PDBs and graceful termination configuration.

We could of course implement all of this in our autoscaler, but I don't think it's worth the effort. This PR switches to the [official cluster autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) instead. 

Notable changes:
 * Old cluster autoscaler replaced with the official one.
 * Pod priorities enabled, these are needed for the buffer pods.
 * Since the official autoscaler doesn't have the concept of spare nodes, there are now buffer pods instead. These are simply pods with running the pause container and consuming a configured amount of CPU/memory requests. Their priority is set very low, so they'll be evicted immediately to free up capacity for the users' pods, and the autoscaler will then resize the cluster to match.
 * The autoscaler is running mostly with the default settings with a couple of tweaks (system pods can be evicted and logging is slightly more verbose). Let's discuss if we need to change something.

Issues:
 * The worker ASG has two additional tags now, one of them so we can easily disable the autoscaler if something goes wrong, and another one because the autoconfiguration if the `k8s.io/cluster` tag has any colons inside. Ugly, but it works.
 * ASG min/max size updates are ignored (bug in the autoscaler, see https://github.com/kubernetes/autoscaler/issues/209). We don't really change them often, so for now we'll have to manually restart it.


cluster/senza-definition.yaml
```diff
@@ -114,6 +114,10 @@ SenzaComponents:
           Value: owned
         - Key: "NodePool"
           Value: "{{ Arguments.WorkerNodePoolName }}"
+        - Key: "k8s.io/cluster-autoscaler/enabled"
+          Value: ""
+        - Key: "zalando.de/cluster-local-id/{{SenzaInfo.StackName}}-{{SenzaInfo.StackVersion}}"
```
mikkeloscar: If the parsing is fixed in the autoscaler we could get rid of this right?

aermakov-zalando: Yes, it's not used by anything else.

hjacobs:
How does the official cluster autoscaler behave now? Does it up and downscale fast enough? How many problems (as mentioned) did we have before, i.e. what is the user impact?

aermakov-zalando:
Upscaling is fast enough, especially with buffer pods. With the default settings nodes are downscaled when they're not used for 10+ minutes, but this can be changed. Regarding the user impact, I think it'll be significant, considering that the adoption is increasing. A lot of teams are now running applications with large requests, which can easily trigger the first bug, and we've ran into volume conflicts at least several times with ACID's databases. The termination behaviour of the current autoscaler is also something that negatively affects the databases and we had several requests from ACID to change it. Teams are also deploying stateful applications of their own.

aermakov-zalando:
👍

aermakov-zalando:
Autoscaler in action:

screen shot 2018-02-16 at 12 12 33

aermakov-zalando:
Autoscaler in action:

screen shot 2018-02-16 at 12 12 33

aermakov-zalando:
👍

szuecs:
:+1:

PR #844, opened 2018-02-14
Update audittrail-adapter

mikkeloscar:
Fixes a problem where it would not re-resolve DNS for the audittrail endpoint when the audittrail endpoint was updated.


mikkeloscar:
:+1:

linki:
👍

PR #849, opened 2018-02-15
Resurrect kube-job-cleaner

aermakov-zalando:
We don't need the job cleaner part of it, but it had a side effect of removing completed pods as well, which we still need. Let's just resurrect it for now and rename/cleanup later.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #855, opened 2018-02-20
Updates image-policy-webhook to fix caching invalid SCM source

njuettner:
Fixes that image-policy-webhook can cache invalid SCM Source


linki:
👍

njuettner:
:+1:

PR #857, opened 2018-02-21
Add ingress-template-controller deployment

mikkeloscar:
This adds the ingress-template-controller which will be used for traffic switching support along with the CRD describing the new `IngressTemplate` resource.


linki:
👍

mikkeloscar:
:+1:

PR #863, opened 2018-02-22
Update ExternalDNS to include multi-target changes

linki:
* Includes tentative multi-target support: code structure in place but still only one target created on AWS


mikkeloscar:
:+1:

linki:
👍

PR #870, opened 2018-02-26
Update ExternalDNS to v0.5.0-alpha.1

linki:
* Includes required code to add support for multiple targets. Multiple targets itself isn't supported on AWS yet.


mikkeloscar:
Is this not the same as #863 ?

linki:
@mikkeloscar yes 😆 closing

PR #866, opened 2018-02-23
Enable pod priorities

aermakov-zalando:
Enable the feature gate for [Pod Priority](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/). A prerequisite switching to the official autoscaler (#841).

PR #874, opened 2018-02-26
Alpha to beta: enable pod priorities on workers

aermakov-zalando:
#866 needs to be rolled out in two phases (which we learned in alpha unfortunately). This is part 1.


linki:
👍

aermakov-zalando:
👍

PR #878, opened 2018-02-28
Update flannel to v0.10.0

linki:
https://github.com/coreos/flannel/releases/tag/v0.10.0

Mostly bug and security fixes for us.

However:
* `portmap` plugin might allow using local ports without specify also `hostNetwork: true` which CNI didn't support until now, afaik https://github.com/kubernetes/kubernetes/issues/31307

It also updates it to latest suggested configuration over at: https://github.com/coreos/flannel/blob/master/Documentation/kube-flannel.yml
* ~~enables the `portmap` plugin~~ follow-up PR if we want it.
* allows to move cni config from userdata to configmap

### Notes to Merger ###

* ~~This must be split into two PRs~~ follow-up PR will be made

### DIFF
--- a/cluster/manifests/flannel/daemonset.yaml
+++ b/cluster/manifests/flannel/daemonset.yaml
@@ -62,8 +62,8 @@ spec:
         securityContext:
           privileged: true
         volumeMounts:
-        - name: flannelcfg
-          mountPath: /etc/kube-flannel
+        - name: flannel-cfg
+          mountPath: /etc/kube-flannel/
         - name: run
           mountPath: /run
       hostNetwork: true
@@ -77,7 +77,7 @@ spec:
       - key: node-role.kubernetes.io/master
         effect: NoSchedule
       volumes:
-      - name: flannelcfg
+      - name: flannel-cfg
         configMap:
           name: kube-flannel-cfg
       - name: cni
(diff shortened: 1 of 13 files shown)


linki:
This is green now. `kubelet` picks up the cni config even if it's added after it started.

mikkeloscar:
@linki this is not safe to merge as is right? It would make the new nodes unable to schedule until the flannel daemonset is updated in the end as far as I understand.

linki:
@mikkeloscar Yes, I wanted to first discuss if we want this change or not. I did remove the userdata changes so that it could be merged. Let's see what e2e says.

cluster/manifests/flannel/daemonset.yaml
```diff
@@ -16,21 +16,33 @@ spec:
...
       serviceAccountName: system
+      initContainers:
+      - name: install-cni
+        image: registry.opensource.zalan.do/teapot/flannel:v0.10.0
+        command:
+        - cp
+        args:
+        - -f
+        - /etc/kube-flannel/cni-conf.json
+        - /etc/cni/net.d/10-flannel.conflist
```
linki: todo: probably better to use the same name as the existing file during the upgrade so that we don't have two configs in the folder.

linki:
👍

mikkeloscar:
:+1:

PR #887, opened 2018-03-06
fix skipper endpoint issue

szuecs:
Skipper did not consider kubernetes services with more than one port.
see https://github.com/zalando/skipper/issues/563


szuecs:
Closed in favour of #861

PR #890, opened 2018-03-07
autoscaling: make utilization threshold configurable

aermakov-zalando:
- Utilization threshold (for downscaling) is now configurable with the
`autoscaling_utilization_threshold` config item.
 - Production clusters have buffer pods, so raise the default threshold to 0.75 instead of 0.5.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #891, opened 2018-03-07
Cherrypick dev to alpha

aermakov-zalando:
* etcd 3.3.0 -> 3.3.1
 * Increase default autoscaler utilization threshold and make it configurable

### DIFF
--- a/cluster/etcd-cluster.yaml
+++ b/cluster/etcd-cluster.yaml
@@ -14,7 +14,7 @@ SenzaComponents:
         2379: 2379
         2380: 2380
       runtime: Docker
-      source: registry.opensource.zalan.do/acid/etcd-cluster:3.3.0-p19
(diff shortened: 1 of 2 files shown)


hjacobs:
How did we come up with the 0.75 number for `--scale-down-utilization-threshold` and why do we change it in the first place? Is this a guesstimate or based an data/experiments?

aermakov-zalando:
@hjacobs it's a less conservative guesstimate than the default value of 0.5, which should make the downscaling a bit more aggressive. This change also makes it configurable by changing a config item, which would allow us to experiment on live clusters to better tune the autoscaler. Considering that the production clusters should have buffer pods enabled, we could probably bump it all the way up to 0.9, but I'd start with a safer default first.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #897, opened 2018-03-09
Protected etcd: part 1

aermakov-zalando:
This change exports the security group ID from the etcd stack and adds an explicit ingress definition as part of the Kubernetes stack to allow master->etcd communication. `etcd-backup` is also configured to run on master nodes.

This has to be applied separately from the follow-up change, and the etcd stacks have to be updated manually.


linki:
👍

szuecs:
:+1:

PR #918, opened 2018-03-14
Inject TOKENINFO_URL to anyone who cares based on environment

linki:
This uses a `PodPreset` which injects the correct value for `TOKENINFO_URL` depending on the cluster's environment in the `default` namespace.

It results in adding the following to a Pod's `PodSpec`:

In Production clusters:
```yaml
spec:
  env:
  - name: TOKENINFO_URL
    value: https://info.services.auth.zalando.com/oauth2/tokeninfo
```

In Test clusters:
```yaml
spec:
  env:
  - name: TOKENINFO_URL
    value: https://sandbox.identity.zalando.com/oauth2/tokeninfo
```

In order to trial it for some time all Pods that want this environment variable to be injected must be labbelled with the following label:
```yaml
metadata:
  labels:
    zalando.org/uses-tokeninfo: "true"
```

Later, we could omit the label selector to apply this environment variable to all Pods. Even then it's possible to opt-out from any Presets by setting an annotation.

### DIFF
--- a/cluster/manifests/pod-presets/pp-tokeninfo.yaml
+++ b/cluster/manifests/pod-presets/pp-tokeninfo.yaml
@@ -11,5 +11,5 @@ spec:
 {{ if eq .Environment "production" }}
     value: https://info.services.auth.zalando.com/oauth2/tokeninfo
 {{ else }}
-    value: https://sandbox.identity.zalando.com
+    value: https://sandbox.identity.zalando.com/oauth2/tokeninfo
 {{ end }}
--- a/cluster/manifests/pod-presets/pp-tokeninfo.yaml
+++ b/cluster/manifests/pod-presets/pp-tokeninfo.yaml
@@ -1,0 +1,15 @@
+apiVersion: settings.k8s.io/v1alpha1
+kind: PodPreset
+metadata:
+  name: tokeninfo
+spec:
+  selector:
+    matchLabels:
+      zalando.org/uses-tokeninfo: "true"
+  env:
+  - name: TOKENINFO_URL
+{{ if eq .Environment "production" }}
+    value: https://info.services.auth.zalando.com/oauth2/tokeninfo
+{{ else }}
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
Will this work for pods in all namespaces, or just the default one?

linki:
@aermakov-zalando Just for the `default` namespace.

szuecs:
Cool @linki makes so much sense to reduce in cluster latency of token lookup and usage. We could even create issues in used token libraries, if they read this env variable transparently.

hjacobs:
I added the `do-not-merge` label --- not because I don't like the approach, but because I think we should align the user/developer experience before adding this option. Let's discuss.

linki:
It's a bit of a chicken and egg situation. I would like to get some experience with that feature before advocating it to users.

My proposed way would be to bring this to `alpha` (behind the annotation) and then try it out on our own services, e.g. on ClusterRegistry to get rid of https://github.bus.zalan.do/teapot/cluster-registry-deploy/search?utf8=%E2%9C%93&q=TOKEN_INFO&type=.

hjacobs:
@linki OK, I'm fine with that if you change the label to make it clear that it's experimental (e.g. `experimental.zalando.org/...`. I want to avoid that users depend on this feature right now (maybe they just see, copy & use it).

aermakov-zalando:
We can use templating to not deploy it past alpha, which will allow us to play with it but won't expose it to the users. WDYT?

hjacobs:
@aermakov-zalando also fine, whatever prevents the feature to be exposed to users :smirk:

ruiaraujo:
Having this would allow a token info daemon set to rolled out very smoothly.

linki:
Closing this PoC. Find details about it in the first post.

If this feature is desired later on feel free to adopt this `PodPreset`-based approach. Alternatively, this can also be achieved by a MutatingWebhook, CDP or left to the user like now.

PR #920, opened 2018-03-14
add kubelet metrics to get disk metrics

szuecs:
This will grant access from all nodes in VPC access to port 10255, which is the default readonly port from kubelet.
The change is driven by getting ebs volume metrics attached as pv from a statefulset.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #921, opened 2018-03-15
Introduce kube-aws-iam-controller

mikkeloscar:
This is to introduce and use the [kube-aws-iam-controller](https://github.com/mikkeloscar/kube-aws-iam-controller) for all our components. It's intended as a replacement for kube2iam (the problem that it solves is described in the [README](https://github.com/mikkeloscar/kube-aws-iam-controller#ec2-metadata-service-solution-kube2iam)) but for now it can run side-by-side with kube2iam to not break apps depending on the kube2iam annotation.

I'm opening the PR now, mostly to run e2e tests. I'll follow up and update this with more motivations for introducing this component.

### DIFF
--- a/cluster/manifests/etcd-backup/cronjob.yaml
+++ b/cluster/manifests/etcd-backup/cronjob.yaml
@@ -17,8 +17,6 @@ spec:
           labels:
             application: etcd-backup
             version: "master-5"
-          annotations:
-            iam.amazonaws.com/role: "{{ .LocalID }}-etcd-backup"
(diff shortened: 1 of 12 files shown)


mikkeloscar:
After more tests I realized this approach doesn't actually work as expected since the AWS SDKs doesn't support refreshing temporary credentials from a file as was the intention.

I'll follow up with AWS/sig-aws about this as I think it would be a great feature and a solution to many problems.

Closing this for now.

PR #923, opened 2018-03-15
reduce latency of ingress 5-10 times

szuecs:
Reduce latency of ingress 5-10 times, because of 100ms time slices set by Kubernetes for having the limits.

limits were set to `200m` before, which sets:

	cfs_quota_us /* quota = 20ms */
	cpu.cfs_period_us /* period = 100ms */

This means that we get 20ms in every 100ms slice and if a request is processed not within one time period of 100ms it will use >100ms for processing the request.

### DIFF
--- a/cluster/manifests/skipper/daemonset.yaml
+++ b/cluster/manifests/skipper/daemonset.yaml
@@ -60,7 +60,6 @@ spec:
           - "-metrics-flavour=codahale,prometheus"
         resources:
           limits:
-            cpu: 200m
             memory: 200Mi
           requests:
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

szuecs:
:+1:

hjacobs:
Related: #1335

hjacobs:
For more context see my DevOps Gathering talk about Kubernetes resource requests and limits: https://www.youtube.com/watch?v=4QyecOoPsGU

PR #924, opened 2018-03-15
Instruct prometheus to scrape kube-state-metrics

linki:
Since we're using Prometheus now 🎉 let's also scrape `kube-state-metrics` to get [some more great metrics](https://github.com/kubernetes/kube-state-metrics/tree/master/Documentation#documentation).

/cc @szuecs

### DIFF
--- a/cluster/manifests/kube-state-metrics/service.yaml
+++ b/cluster/manifests/kube-state-metrics/service.yaml
@@ -3,6 +3,10 @@ kind: Service
 metadata:
   name: kube-state-metrics
   namespace: kube-system
+  annotations:
+    prometheus.io/path: /metrics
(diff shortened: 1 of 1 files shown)


szuecs:
@linki what's the use case?
Can you check if this is required for custom hpa or something that makes sense for us?
I do not want to overrun prometheus right now.

linki:
@szuecs I see. Our prometheus are already quite memory hungry if I recall correctly from last week. I do tend to get some ideas for new checks and alerts when seeing all the metrics being available, rather than specifically looking for what I need when I need to.

That was the only reason. I'm closing this, but let's just keep in mind that there's many more metrics to be easily scraped.

PR #928, opened 2018-03-15
Alpha to beta

szuecs:
- update kubernetes core components to 1.9.4
- use our patched kube2iam
- add kubelet metrics to monitor EBS volumes
- remove cpu limits of skipper-ingress to reduce the latency with factor 20-50x


mikkeloscar:
:+1:

szuecs:
:+1:

PR #930, opened 2018-03-15
Add EFS as possible storage provider

linki:
This is work in progress and allows to use EFS as backend for PVCs.

It allows two Pods on different nodes to share the same volume. Use it like so:

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: my-data
  annotations:
    volume.beta.kubernetes.io/storage-class: aws-efs
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 1Mi

---

apiVersion: v1
kind: Pod
metadata:
  name: test-pod
spec:
  containers:
  - name: test-pod
    image: gcr.io/google_containers/busybox:1.24
    command:
      - "/bin/sh"
    args:
      - "-c"
      - "touch /mnt/SUCCESS && exit 0 || exit 1"
    volumeMounts:
      - name: efs-pvc
        mountPath: "/mnt"
  restartPolicy: "Never"
  volumes:
    - name: efs-pvc
      persistentVolumeClaim:
        claimName: my-data
```

Todo:
* [ ] provision an EFS and pass its endpoint to the configmap instead of via a config item


cluster/manifests/efs-provisioner/cm-efs-provisioner.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: efs-provisioner
+  namespace: kube-system
+data:
+  file.system.id: "{{ .ConfigItems.efsid }}"
```
mikkeloscar: Maybe all these resources should only be deployed to clusters which has this item defined.

shansfolder:
👍

elgalu:
Hi, is this moving forward?

szuecs:
@elgalu not planned for this week and I am not sure how much work is left.

mikkeloscar:
Maybe we should also add a security group which only allows access from the worker nodes.

cluster/manifests/efs-provisioner/cm-efs-provisioner.yaml
mikkeloscar: I updated it to only deploy is `efsid` is set. This way we can enable for selected clusters.

mikkeloscar:
I think this is ready as an alpha feature. We could deploy it like it is and enable it for selected clusters. But this being an alpha feature I would not promise that we will not loose any data at some point :) @elgalu would this be good enough for you?

cluster/manifests/efs-provisioner/cm-efs-provisioner.yaml
```diff
@@ -0,0 +1,11 @@
+{{ if index .ConfigItems "efsid" }}
```
aermakov-zalando: Nit: can we rename this to `efs_id`?

aermakov-zalando:
We might want to think about conditional deletions. Currently, provisioning EFS in a cluster and then removing the config item will still all the resources in place.

cluster/manifests/efs-provisioner/cm-efs-provisioner.yaml
mikkeloscar: changed.

aermakov-zalando:
👍

aermakov-zalando:
Do we want to whitelist the config item for the users in the cluster registry?

elgalu:
:+1:

elgalu:
Yes Mikkel this is good enough for us. Our use case is to use EFS as a temporary elastic space to save costs avoiding 1 EBS per user. Our users are strongly encouraged to store code in GHE and company data in S3 or the Data Lake. The use case should never be to use EFS as their long term storage just because is "easier" to leave stuff there.

mikkeloscar:
> Do we want to whitelist the config item for the users in the cluster registry?

As a start I would just let us configure it in each account for the users who want it. Let's give it to data first and and see if it's good for GA.

Missing one thing, we need to re-host the docker image in our registry. I'll fix that.

mikkeloscar:
:+1:

linki:
👍

elgalu:
How/where do I set `"efs_id"` ?

elgalu:
Ah you already create the EFS resource and set it as part of the cluster config so is all transparent to me, awesome!

PR #932, opened 2018-03-16
fix handle endpoints more safely

szuecs:
- fix handle endpoints more safely
- fix ingress route id formatting
- fix support for legacy predicates
- do not use lb targets on svc targets -> we did not found endpoints for


szuecs:
:+1:

mikkeloscar:
:+1:

PR #938, opened 2018-03-20
HPA: use the old API for getting metrics

aermakov-zalando:
We broke the horizontal pod autoscaler when updating to 1.9. It now uses the new metrics server APIs by default, which we don't deploy. See [a similar issue](https://github.com/kubernetes/kubernetes/issues/57673) for details.


szuecs:
:+1:

aermakov-zalando:
👍

PR #947, opened 2018-03-22
Alpha to beta

aermakov-zalando:
* Downgrade ZMON because of a bug in the new version
* Remove the default etcd cluster access from worker nodes (this will not be applied by CLM, so it's safe to merge, and it's easier than cherry-picking)


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #949, opened 2018-03-22
Update authnz webhook to v0.4.0

mikkeloscar:
This adds support for RBAC (which we don't need yet) but also adds logging to the authn part of the webhook so it's easier to see what groups are assigned to a user.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #950, opened 2018-03-22
prometheus visibility and increase limits and requests

szuecs:
add monitoring for prometheus itself and increase cpu and memory limits and requests as required to run in bigger clusters


mikkeloscar:
I think we should go for higher defaults. Here are two of the bigger clusters:

mct

sp

I would go for 4Gi Memory and 500-1000m CPU request to be on the safe side.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #951, opened 2018-03-23
Dev to alpha

mikkeloscar:
* Add support for new clm token uid
* Update authnz webhook to v0.4.0
* Add monitoring for prometheus itself and increase cpu and memory limits and requests as required to run in bigger clusters


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #952, opened 2018-03-23
prometheus version update

szuecs:
prometheus version update that fixes database corruption and compaction issues. It also seems to scrap much better with the same amount of cpu and memory consumption.


szuecs:
:+1:

aermakov-zalando:
👍

PR #956, opened 2018-03-25
LB connection metrics

szuecs:
We should enable connection metrics to see active and new connections.
These are important loadbalancer metrics to understand better the behaviour of the system and the requests.


szuecs:
:+1:

gargravarr:
:+1:

PR #957, opened 2018-03-26
Update skipper version

ruiaraujo:
Adds cors filter.


szuecs:
we can not test external branches. I opened https://github.com/zalando-incubator/kubernetes-on-aws/pull/959 instead

PR #962, opened 2018-03-28
Update zmon-agent to a49-zv1

avaczi:
To roll out the same logic from a48-zv1, but in a way that does not break clusters without the PostgreSQL operator


szuecs:
:+1:

avaczi:
:+1:

PR #972, opened 2018-04-04
Update ingress-template-controller to version with custom per service…

mikkeloscar:
… ingress support

This will automatically generate ingresses for each of the services selected by an ingress template such that the services are exposed on a non traffic switched endpoint for QA/manual testing before switching traffic on the main endpoint.


mikkeloscar:
:+1:

hjacobs:
:+1:

PR #974, opened 2018-04-05
Update to Kubernetes v1.10.2

### LINKED ISSUE #975: Kubernetes 1.10
https://github.com/kubernetes/kubernetes/releases/tag/v1.10.0

mikkeloscar:
**WORK IN PROGRESS**

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.10.md#v1100

### Notable changes
* Removes rescheduler and uses Priority classes instead. (https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/#rescheduler-guaranteed-scheduling-of-critical-add-ons)
  * Adds the system defined PriorityClasses to kube-system pods:
    * `system-node-critical` for daemonset pods and static pods on masters (this is the highest possible priority in Kubernetes).
    * `system-cluster-critical` for deployment pods (second highest).
    * Removes the `scheduler.alpha.kubernetes.io/critical-pod` annotation as setting the priorityClass to a system-critical one has the same effect: https://github.com/kubernetes/kubernetes/blob/v1.10.0/pkg/kubelet/types/pod_update.go#L143-L175
  * Makes `kube-dashboard` non-critical.
  * Makes `prometheus` cluster-critical.
* Uses the recommmended Admission Controllers as defined here: https://kubernetes.io/docs/admin/admission-controllers/#is-there-a-recommended-set-of-admission-controllers-to-use

Fixes #975

### DIFF
--- a/cluster/manifests/kube-proxy/configmap.yaml
+++ b/cluster/manifests/kube-proxy/configmap.yaml
@@ -23,7 +23,10 @@ data:
       tcpCloseWaitTimeout: 1h0m0s
       tcpEstablishedTimeout: 24h0m0s
     enableProfiling: false
-    featureGates: "ExperimentalCriticalPodAnnotation=true,TaintBasedEvictions=true,PodPriority=true"
+    featureGates:
+      ExperimentalCriticalPodAnnotation: true
+      TaintBasedEvictions: true
+      PodPriority: true
     healthzBindAddress: 0.0.0.0:10256
     hostnameOverride: ""
     iptables:
--- a/cluster/manifests/kube-proxy/configmap.yaml
+++ b/cluster/manifests/kube-proxy/configmap.yaml
@@ -24,7 +24,6 @@ data:
       tcpEstablishedTimeout: 24h0m0s
     enableProfiling: false
     featureGates:
-      ExperimentalCriticalPodAnnotation: true
       TaintBasedEvictions: true
       PodPriority: true
     healthzBindAddress: 0.0.0.0:10256
--- a/cluster/manifests/10-vertical-pod-autoscaler/admission-controller-deployment.yaml
+++ b/cluster/manifests/10-vertical-pod-autoscaler/admission-controller-deployment.yaml
@@ -17,6 +17,7 @@ spec:
         application: vpa-admission-controller
         version: v0.1.0
     spec:
+      priorityClassName: system-cluster-critical
       serviceAccountName: system
       containers:
       - name: admission-controller
--- a/cluster/manifests/10-vertical-pod-autoscaler/recommender-deployment.yaml
+++ b/cluster/manifests/10-vertical-pod-autoscaler/recommender-deployment.yaml
@@ -17,6 +17,7 @@ spec:
         application: vpa-recommender
         version: v0.1.0
     spec:
+      priorityClassName: system-cluster-critical
       serviceAccountName: system
       containers:
       - name: recommender
--- a/cluster/manifests/10-vertical-pod-autoscaler/updater-deployment.yaml
+++ b/cluster/manifests/10-vertical-pod-autoscaler/updater-deployment.yaml
@@ -17,6 +17,7 @@ spec:
         application: vpa-updater
         version: v0.1.0
     spec:
+      priorityClassName: system-cluster-critical
       serviceAccountName: system
       containers:
       - name: updater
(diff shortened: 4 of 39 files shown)


cluster/manifests/kube-proxy/configmap.yaml
```diff
@@ -23,7 +23,10 @@ data:
       tcpCloseWaitTimeout: 1h0m0s
       tcpEstablishedTimeout: 24h0m0s
     enableProfiling: false
-    featureGates: "ExperimentalCriticalPodAnnotation=true,TaintBasedEvictions=true,PodPriority=true"
+    featureGates:
+      ExperimentalCriticalPodAnnotation: true
```
aermakov-zalando: This is no longer needed since we've switched to priorities. Or am I missing something?

mikkeloscar: True, not needed. Removed it!

hjacobs:
Any breaking changes in the 1.10 changelog (like making some volume read-only? :smirk:)?

szuecs:
All mentioned changes should be good to go.

## They also posted relevant known issues
* If you need to downgrade from 1.10 to 1.9.x, downgrade to v1.9.6 to ensure PV and PVC objects can be deleted properly.
* Use of subPath module with hostPath volumes can cause issues during reconstruction (#61446) and with containerized kubelets (#61456). The workaround for this issue is to specify the complete path in the hostPath volume. Use of subPathmounts nested within atomic writer volumes (configmap, secret, downwardAPI, projected) does not work (#61545), and socket files cannot be loaded from a subPath (#62377). Work on these issues is ongoing.
* Some users, especially those with very large clusters, may see higher memory usage by the kube-controller-manager in 1.10. (#61041)

Not sure about the subPath usage by our users

## Mentioned deprecations that we should change soon:

* The ability to use the insecure flags --insecure-bind-address, --insecure-port in the apiserver has been deprecated and will be removed in a future release. Use --secure-port and --bind-address instead. (#59018, hzxuzhonghu)
* The PodSecurityPolicy API has been moved to the policy/v1beta1 API group. The PodSecurityPolicy API in the extensions/v1beta1 API group is deprecated and will be removed in a future release. Authorizations for using pod security policy resources should change to reference the policy API group after upgrading to 1.11. (#54933, php-coder)
* Add --enable-admission-plugins --disable-admission-plugins flags and deprecate --admission-control. When using the separate flag, the order in which they're specified doesn't matter. (#58123, hzxuzhonghu)
* The kubelet's cadvisor port has been deprecated. The default will change to 0 (disabled) in 1.12, and the cadvisor port will be removed entirely in 1.13. (#59827, dashpole)

## AWS

*    Add AWS cloud provider option to use an assumed IAM role. For example, this allows running Controller Manager in a account separate from the worker nodes, but still allows all resources created to interact with the workers. ELBs created would be in the same account as the worker nodes for instance.(#59668, brycecarman)
*    AWS EBS volume plugin now includes block and volumeMode support. (#58625, screeley44)
*    On AWS kubelet returns an error when started under conditions that do not allow it to work (AWS has not yet tagged the instance), rather than failing silently. (#60125, vainu-arto)
*    AWS Security Groups created for ELBs will now be tagged with the same additional tags as the ELB; that is, the tags specified by the "service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags" annotation. This is useful for identifying orphaned resources. (#58767, 2rs2ts)
*    AWS Network Load Balancers will now be deleted properly, including security group rules. Fixes #57568 (#57569, micahhausler)
*    Time for attach/detach retry operations has been decreased from 10-12s to 2-6s (#56974, gnufied)

hjacobs:
@szuecs we are using `subPath` (I checked via git search), but I don't know whether we use it in combination with `hostPath`.

hjacobs:
To check:

- [x] subPath/hostPath usage
- [x] "The PodSecurityPolicy API has been moved to the policy/v1beta1 API group"
- [x] `--admission-control` deprecated
- [x] kubelet's cadvisor

linki:
I checked and no Kubernetes manifest currently on GHE uses `subPath` **and** `hostPath`.

mikkeloscar:
For cadvisor we will leave it as is for now. It will be disabled in 1.12 at which point we might consider deploying a daemonset to expose the metrics we need.

hjacobs:
Merge?

mikkeloscar:
:+1:

szuecs:
:+1:

PR #982, opened 2018-04-10
Worker ASG: add a lifecycle-status node template tag

aermakov-zalando:
This will make it possible to use a node selector targeting lifecycle-status: ready nodes without breaking the autoscaling.


linki:
👍

aermakov-zalando:
👍

PR #987, opened 2018-04-13
Disable automatic updates

aermakov-zalando:
Let's avoid sudden breakage if a node reboots unexpectedly.


aermakov-zalando:
👍

aermakov-zalando:
@mikkeloscar suggested to only do it on the ignition branch.

PR #988, opened 2018-04-16
fix uneven choice of backends and reduce log noise

szuecs:
fix uneven loadbalancing on low traffic backends, because of too often triggered route updates
fix reduce log events
fix do not leak decision header to backends


szuecs:
:+1:

mikkeloscar:
:+1:

PR #989, opened 2018-04-16
Dev to alpha

szuecs:
* Add support for finding multiple certificates per ingress
* Fix uneven load balancing of low traffic applications
* Fix empty names for NAT routing tables
* Have consistent naming of ALB stacks
* Worker ASG: add a lifecycle-status node template tag
* remove deprecated annotation and use field
* Update pdb-controller


mikkeloscar:
:+1:

szuecs:
:+1:

PR #990, opened 2018-04-16
Disable creating parameters on the etcd stack

mikkeloscar:
This just disables the creation of cloudformation stack parameters for the etcd stack. This change is needed to make the CLM update the etcd stack as part of cluster provisioning. Before we had to update the etcd stacks manually outside of CLM.

The parameters are not needed since the values passed to senza will be injected into the right place in the template anyway.

There's an internal PR for CLM that depends on this change.

### DIFF
--- a/cluster/etcd-cluster.yaml
+++ b/cluster/etcd-cluster.yaml
@@ -1,6 +1,7 @@
 SenzaComponents:
 - Configuration:
     Type: Senza::StupsAutoConfiguration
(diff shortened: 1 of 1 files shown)


mikkeloscar:
Will not run e2e since it won't touch this change.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #991, opened 2018-04-16
Use static names for lifecycle hooks

mikkeloscar:
This uses static names for lifecycle hooks which is now supported in
cloudformation: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-as-lifecyclehook.html#cfn-as-lifecyclehook-lifecyclehookname

This means that we don't have to pass the dynamic lifecycle hooks around
which was just a stupid workaround because of this missing feature in
cloudformation in the past.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #1000, opened 2018-04-19
Scrape metrics from kube-apiserver pods

mikkeloscar:
This enables scraping metrics of kube-apiserver pods (a lot of metrics) as well as the audittrail-adapter pods (very few metrics).

The purpose of this is to be able to answer important questions like rps, latency, error rate etc. for apiservers and additionally monitor the audit events in a better way.

Additionally I added a commit to enable scraping cadavisor metrics on master nodes because I saw this was failing in prometheus because the port was not open in the SG.


linki:
👍

aermakov-zalando:
👍

szuecs:
:+1:

PR #1001, opened 2018-04-19
Add support for spot worker instances

aermakov-zalando:
This adds an optional argument to the cluster definition that allows setting a spot price for worker nodes which is then used to automatically provision spot instances instead of on-demand ones.


hjacobs:
LGTM and could be a very simple thing for non-production clusters.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1005, opened 2018-04-23
Add node pool support

### LINKED ISSUE #115: Support node pools
Support having multiple node pools as defined in the cluster registry:

```yaml
NodePool:
    type: object
    properties:
      name:
        type: string
        example: pool-1
        description: Name of the node pool
      profile:
        type: string
        example: worker/default
        description: Profile used for the node pool. Possible values are "worker/default", "worker/database", "worker/gpu", "master". The "master" profile identifies the pool containing the cluster master
      instance_type:
        type: string
        example: m4.medium
        description: Type of the instance to use for the nodes in the pool. All the nodes in the pool share the same instance types
      discount_strategy:
        type: string
        example: none
        description: |
          A discount strategy indicates the type of discount to be associated with the node pool. This might affect the availability of the nodes in the pools in case of preemptible or spot instances.
          Possible values are "none", "aggressive", "moderate", "reasonable" #TODO naming should be "reasonable" :-D

### PR DESCRIPTION
**Pull request by mikkeloscar**
This adds node pool support as described in the (internal narrative/CLM PR, see that for context).

Basically we add the folder structure:

```
node-pools
├── master-default
│   ├── stack.yaml
│   └── userdata.clc.yaml
└── worker-default
    ├── stack.yaml
    └── userdata.clc.yaml
```

Which is then used for provisioning node pools based on the profile (`master-default` or `worker-default` in this case.)

Fix #115

### DIFF
--- a/cluster/node-pools/master-default/stack.yaml
+++ b/cluster/node-pools/master-default/stack.yaml
@@ -1,0 +1,70 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: Kubernetes default master node pool
+
+Mappings:
+  Images:
+    eu-central-1:
+      LatestCoreOSImage: ami-604e118b
+      StableCoreOSImage: ami-862140e9
+
+Resources:
+  AutoScalingGroup:
+    CreationPolicy:
+      ResourceSignal:
+        Count: '0'
+        Timeout: PT15M
+    Properties:
+      HealthCheckGracePeriod: 480
+      HealthCheckType: EC2
+      LaunchConfigurationName: !Ref AutoScalingConfig
+      LoadBalancerNames:
+      - !ImportValue '{{ .Cluster.ID }}:master-load-balancer'
+      MinSize: '{{ .NodePool.MinSize }}'
+      MaxSize: '{{ .NodePool.MaxSize }}'
+      VPCZoneIdentifier: !Split [",", "{{ .Cluster.ConfigItems.subnets }}"]
+    Type: 'AWS::AutoScaling::AutoScalingGroup'
+  AutoScalingConfig:
+    Properties:
+      AssociatePublicIpAddress: true
+      BlockDeviceMappings:
+      - DeviceName: /dev/xvda
+        Ebs:
+          VolumeSize: 20
+          VolumeType: standard
+      EbsOptimized: false
+      IamInstanceProfile: !Ref AutoScalingInstanceProfile
+      # ImageId: " .AMI }}" # TODO: find
+      ImageId: "ami-862140e9"
+      InstanceType: "{{ .NodePool.InstanceType }}"
+      SecurityGroups:
+      - !ImportValue '{{ .Cluster.ID }}:master-security-group'
+      UserData: "{{ .UserData }}"
+    Type: 'AWS::AutoScaling::LaunchConfiguration'
+  AutoScalingInstanceProfile:
+    Properties:
+      Path: /
+      Roles:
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by hjacobs** [comment]
What is `userdata.yaml.bak`?

**Comment by hjacobs** [comment]
The migration path is probably the most interesting question..

**Comment by mikkeloscar** [comment]
@hjacobs oops, that was a local change not committed, but included in the `tree` output. Updated the PR description.

> The migration path is probably the most interesting question..

This is described in the internal documents/CLM pr.

File: cluster/node-pools/master-default/stack.yaml [hunk 1]
```diff
@@ -0,0 +1,70 @@
...
+      AssociatePublicIpAddress: true
+      BlockDeviceMappings:
+      - DeviceName: /dev/xvda
+        Ebs:
+          VolumeSize: 20
+          VolumeType: standard
+      EbsOptimized: false
+      IamInstanceProfile: !Ref AutoScalingInstanceProfile
+      # ImageId: " .AMI }}" # TODO: find
+      ImageId: "ami-862140e9"
```
linki: Should this use one of the mappings above?

cluster/node-pools/worker-default/stack.yaml
```diff
@@ -0,0 +1,97 @@
...
+          VolumeSize: 50
+          VolumeType: standard
+      EbsOptimized: false
+      IamInstanceProfile: !Ref AutoScalingInstanceProfile
+      # ImageId: !FindInMap
+      #   - Images
+      #   - !Ref 'AWS::Region'
+      #   - StableCoreOSImage
+      # ImageId: " .AMI }}" # TODO: find
+      ImageId: "ami-862140e9" # TODO: find
```
linki: Can we fix this TODO?

linki:
@mikkeloscar Without the corresponding changes in CLM this should be safe to merge.

linki:
👍

mikkeloscar:
@linki Would be good to run e2e with the updated CLM IMO, but it can ofc. also live side by side with the legacy stuff.

linki:
@mikkeloscar Which order do you prefer?

cluster/node-pools/master-default/stack.yaml
mikkeloscar: This is open because we don't have a clear process for how to define the AMI since the incident some weeks ago caused by a AMI unable to boot. For now I would just hardcode it here (and remove the mappings on top) and then we need to figure out how to go forward with this. This would be in line with the current approach were we hardcode it in a mapping because of senza (https://github.com/zalando-incubator/kubernetes-on-aws/pull/965).

Thoughts are welcome! (maybe this is not the right place to discuss it).

mikkeloscar: Changed it to look this up from the mapping. Still we need to iterate on how to do this moving forward.

cluster/node-pools/worker-default/stack.yaml
mikkeloscar: Fixed.

mikkeloscar:
This is now ready for dev. Note it will not be enabled before we add the `node_pool_feature_enabled=true` config_items to the clusters, as described in: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/3

Merging to dev will make it possible to switch this on per cluster basis.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #1009, opened 2018-04-24
fix the uneven startup of roundrobin

szuecs:
As requested by a user, which did benchmarks and saw some weird numbers in the beginning of POD starts


mikkeloscar:
:+1:

linki:
👍

PR #1012, opened 2018-04-25
Improve prometheus setup

aermakov-zalando:
* Scrape `kube-state-metrics` explicitly
* Change labeling rules so that pod/node/container labels are consistent
* Bump requests to match limits (let's avoid overcommitting)
* Force an update after a configmap change

### DIFF
--- a/cluster/manifests/prometheus/configmap.yaml
+++ b/cluster/manifests/prometheus/configmap.yaml
@@ -17,17 +17,47 @@ data:
       scheme: http
       kubernetes_sd_configs:
       - role: pod
+        namespaces:
+          names:
+            - kube-system
       relabel_configs:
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
The env var to force a configmap update is awful. Unfortunately, Prometheus doesn't reload configuration automatically. You can enable a special reloading endpoint, but the same setting also enables an endpoint to stop it, and it also looks like there's memory leaks related to config reloading.

linki:
Helm Charts use an annotation instead of an env var to make the PodSpec different when the configmap changes which is a better fit IMO.

Furthermore, the annotations value is a `sha256` of its content calculated at apply-time so that no manual effort is required. I believe, we could add this to our setup as well as to CDP for our users, see:

https://github.com/kubernetes/charts/search?utf8=%E2%9C%93&q=%22checksum%2Fconfig%22&type=

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #1016, opened 2018-04-27
Fix kube-state-metrics OOM

njuettner:
This will fix the OOM issue. kube-state-metrics consumes more memory because it fetches more data


aermakov-zalando:
Can you bump the requests to match?

aermakov-zalando:
👍

njuettner:
👍

PR #1022, opened 2018-05-02
Increase CPU limit for kube-state-metrics

aermakov-zalando:
`100m` is not enough for clusters with a large number of pods so we get no metrics.


aermakov-zalando:
👍

njuettner:
👍

PR #1030, opened 2018-05-08
modify prometheus cpu and mem ressources by changing via config items

njuettner:
some clusters seem to need more ressources otherwise prometheus is crashing all the time. 
by changing the ressources via config items we are more flexible to increase cpu and mem based on the cluster.


mikkeloscar:
:+1:

njuettner:
👍

PR #1036, opened 2018-05-11
Enable metrics-server (heapster replacement)

mikkeloscar:
Introduce [metrics-server](https://github.com/kubernetes-incubator/metrics-server) as a replacement for heapster and to further enable custom and external metrics for HPA later on.

I didn't find any of our checks that really depend on heapster, but I'll let it stay there in case someone still depends on it. We can clean that up separately.

I tested manually that HPA works with this configuration (we previously broke it #938)

### TODO

* [x] Mirror image to our registry.
* [x] Add proxy-client certs/keys to e2e tests.
* [x] Add proxy-client certs/keys to account creator.
* [x] Add proxy-client certs/keys to all existing cluster in cluster-registry.

### DIFF
--- a/cluster/manifests/metrics-server/apiservice.yaml
+++ b/cluster/manifests/metrics-server/apiservice.yaml
@@ -1,0 +1,13 @@
+apiVersion: apiregistration.k8s.io/v1beta1
+kind: APIService
+metadata:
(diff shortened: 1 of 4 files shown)


mikkeloscar:
:+1:

njuettner:
👍

PR #1037, opened 2018-05-11
Add VPA control plane

mikkeloscar:
Adds the control plane for the VerticalPodAutoscaler.

See https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler to understand how it works.

~I have added a VPA just for prometheus for now, since we need to understand how this actually works in practice. Maybe we even need to try a less important component first.~

~Added a VPA only to `pdb-controller`. This component is not super critical so a good test case.~

Dropped the VPA from kube-system components for now, we will test this first with a single service (green-snail-trail).

#### TODO

* [x] mirror docker images to our registry.
* [x] Add new vpa-webhook certs/keys to all clusters.

### DIFF
--- a/cluster/manifests/10-vertical-pod-autoscaler/admission-controller-deployment.yaml
+++ b/cluster/manifests/10-vertical-pod-autoscaler/admission-controller-deployment.yaml
@@ -1,0 +1,40 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: vpa-admission-controller
+  namespace: kube-system
+  labels:
+    application: vpa-admission-controller
+    version: v0.1.0
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      application: vpa-admission-controller
+  template:
+    metadata:
+      labels:
+        application: vpa-admission-controller
+        version: v0.1.0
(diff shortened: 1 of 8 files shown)


szuecs:
Yes please use some other component first, if it does not work as expected people are blind.

cluster/manifests/prometheus/vpa.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: "poc.autoscaling.k8s.io/v1alpha1"
+kind: VerticalPodAutoscaler
+metadata:
+  name: prometheus
+  namespace: kube-system
+spec:
+  selector:
+    matchLabels:
+      application: prometheus
```
szuecs: not prometheus

szuecs:
:-1:

cluster/manifests/prometheus/vpa.yaml
mikkeloscar: Changed it to use the `pdb-controller` which is not super critical.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #1038, opened 2018-05-14
Enable AWS Glue for the deployment role (to enable deployment via CDP)

mikkeloscar:
This was enabled some time ago for PowerUsers in our AWS accounts, but it should also be enabled in the deployment role so people can deploy [Glue](https://aws.amazon.com/glue/) via cloudformation in CDP.


njuettner:
👍

mikkeloscar:
:+1:

PR #1039, opened 2018-05-14
Enable PersistentVolumeClaimResize admission plugin

mikkeloscar:
This allows users to expand psersisten volume claims assuming they use a
storage class where this is enabled.

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#expanding-persistent-volumes-claims

It doesn't enable expansion of the default (`standard`) storage class
since this is still an alpha feature. For now users will have to opt-in
by creating their own storage class.


mikkeloscar:
:+1:

njuettner:
👍

PR #1040, opened 2018-05-14
Move zmon related components to visibility namespace

linki:
So that EagleEye can fix it in case of emergency: https://github.bus.zalan.do/teapot/issues/issues/1030


cluster/manifests/visibility/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
mikkeloscar: I think you need to name this file `01-namespace.yaml` or something like that, to ensure CLM creates the namespace before creating resources in it. :X

aermakov-zalando: The directory needs to be renamed as well, otherwise logging-agent change above would fail.

mikkeloscar:
:+1:

njuettner:
👍

PR #1044, opened 2018-05-15
[EXPERIMENT] Disable CPU quota

### LINKED ISSUE #1026: Consider disabling CPU quota via Kubelet flag --cpu-cfs-quota=false
To mitigate the CPU throttling problem (https://github.com/kubernetes/kubernetes/issues/51135), we should consider disabling CPU quota/throttling all together. That's what @vishh from Google recommends.

hjacobs:
** EXPERIMENT, DO NOT MERGE **

Disable CPU quota via `kubelet` flag as proposed in https://github.com/kubernetes/kubernetes/issues/51135

Potentially fixes #1026.

### DIFF
--- a/cluster/node-pools/master-default/userdata.clc.yaml
+++ b/cluster/node-pools/master-default/userdata.clc.yaml
@@ -173,6 +173,7 @@ systemd:
       --cloud-provider=aws \
       --feature-gates=ExperimentalCriticalPodAnnotation=true,TaintBasedEvictions=true,PodPriority=true \
       --pod-infra-container-image=registry.opensource.zalan.do/teapot/pause-amd64:3.1 \
+      --cpu-cfs-quota=false \
       --system-reserved=cpu=100m,memory=164Mi \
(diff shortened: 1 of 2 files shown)


hjacobs:
@szuecs why not run the tests? I want to see them running before I manually launch a cluster :smirk:

hjacobs:
Test fail because of AWS rate limits :disappointed: "An error occurred (RequestLimitExceeded) when calling the DescribeSubnets operation (reached max retries: 4): Request limit exceeded."

mikkeloscar:
@hjacobs yes :(, we have seen this for the past days. I opened a support case earlier to ask why it's so bad now.

(could potentially be related to changes in the CLM for the node-pools, but I just don't see the pattern from the failures here. Really looks like we are more rate-limited than normally. I'll wait for an answer from AWS before I make the conclusion though :P)

aermakov-zalando:
Rolled out in #1335.

PR #1046, opened 2018-05-16
Enable non-ready-ttl for pdb-controller

mikkeloscar:
This updates the pdb-controller to version v0.0.7 which enables support
for removing managed PDBs when the `--non-ready-ttl` flag is set.

See: https://github.com/mikkeloscar/pdb-controller/pull/14

The default ttl is set to one hour, but can be tweaked per cluster via a
config item: `pdb_non_ready_ttl`.

Signed-off-by: Mikkel Oscar Lyderik Larsen <mikkel.larsen@zalando.de>


mikkeloscar:
:+1:

njuettner:
👍

PR #1052, opened 2018-05-17
add channel handling scripts

szuecs:
add channel handling scripts


njuettner:
👍

aermakov-zalando:
Can you add a hashbang and `set -e`? This will happily continue on errors.

alpha-to-beta.sh
```diff
@@ -0,0 +1,12 @@
+git co alpha
```
mikkeloscar: I think `co` is an alias? Please spell out the full command `checkout` :)

Same for `br` -> `branch` etc.

szuecs:
I added a bit more automation
* prepares PR message and title
* if "hub" is installed then it creates a PR with title and body and adds our testing label automatically

szuecs:
:+1:

aermakov-zalando:
Thanks!

Shellcheck prints some warnings, at least one of those looks that it matters:

    In alpha-to-beta.sh line 16:
    echo "alpha-to-beta\n"> $MSG
          ^-- SC2028: echo won't expand escape sequences. Consider printf.

The rest looks relatively benign, but I'd still fix the reported issues.

Additionally, the scripts only differ in the branch names. Would you consider having a `prepare-pr` script instead that would accept source and target branches as arguments?

szuecs:
@aermakov-zalando changed as commented

szuecs:
:+1:

mikkeloscar:
:+1:

PR #1053, opened 2018-05-17
Make it possible to not downgrade docker

mikkeloscar:
This makes it possible to not downgrade docker to v1.12 by setting the config item: `upstream_docker=yes` in the cluster registry.

This way we can easily try out the default docker version from CoreOS in selected clusters.

Note: Only works for the new node pools.

Reference for the docker downgrade file: https://coreos.com/blog/toward-docker-17-in-container-linux

### DIFF
--- a/cluster/node-pools/master-default/userdata.clc.yaml
+++ b/cluster/node-pools/master-default/userdata.clc.yaml
@@ -894,11 +894,13 @@ storage:
           --table nat \
           --to-destination ${PRIVATE_IPV4}:8181
 
+  {{if not (index .Cluster.ConfigItems "upstream_docker")}}
   - filesystem: root
     path: /etc/coreos/docker-1.12
     mode: 0644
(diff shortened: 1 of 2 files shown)


hjacobs:
:+1:

mikkeloscar:
:+1:

szuecs:
not sure if it did not make more sense to have the docker version string specified as config item

aermakov-zalando:
@szuecs you can't, I think. you only get a choice between 1.12 and the latest.

mikkeloscar:
> @szuecs you can't, I think. you only get a choice between 1.12 and the latest.

Correct, I also wanted to put the docker version, but this is currently not possible.

See https://coreos.com/blog/toward-docker-17-in-container-linux for the timeline when this would be possible. I'm not sure if this is still true though, since they became part of Redhat.

PR #1068, opened 2018-05-18
Pdb controller beta

mikkeloscar:
* Enable non-ready-ttl for pdb-controller
* Create default PDBs with maxUnavailable: 1
* move zmon related components to visibility namespace

We want to roll this to beta before #1050 to ensure CLM doesn't get stuck on broken test deployments and to provide more conservative PDBs.

Also had to add the ZMON move to visibility in order for the e2e to run (they assume logging-agent is in visibility).


mikkeloscar:
:+1:

szuecs:
:+1:

PR #1074, opened 2018-05-24
Only enable auditing in production clusters

mikkeloscar:
Only enable auditing in production clusters or if the `audittrail_url`
config item is explicitly set in a cluster. This makes it easy to
integrate with test audittrail to automatically test new versions of the
audittrail-adapter.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #1076, opened 2018-05-24
Fix templating in userdata

mikkeloscar:
Fixes an issue introduced in #1074. This was not caught by the e2e tests because they are not yet testing the new node-pool implementation.

Once all clusters has node pools we can fix the e2e.


mikkeloscar:
**Not running e2e tests on this because they would not test this change**

mikkeloscar:
:+1:

hjacobs:
:+1:

szuecs:
:+1:

PR #1078, opened 2018-05-24
Mount /data into kubelet container

mikkeloscar:
We need to mount the `/data` directory into the kubelet container for it to see it as a hostPath. To make this simpler I changed the naming convention from `/data{n}` to `/data/disk{n}` such that we can mount the whole `/data` directory and not have to specify the mounts individually.


otrosien:
👍

mikkeloscar:
:+1:

mikkeloscar:
This will not be covered by e2e tests, but I tested it in a custom cluster. It works! :)

hjacobs:
@mikkeloscar "I tested it manually and it works" :trollface:

njuettner:
👍

PR #1082, opened 2018-05-25
Automatically restart autoscaler on node pool changes

aermakov-zalando:
This is needed because the autoscaler doesn't handle ASG min/max size changes automatically.


aermakov-zalando:
👍

mikkeloscar:
:+1:

hjacobs:
@aermakov-zalando neat trick :smile:

PR #1084, opened 2018-05-28
[WIP] Introduce CoreDNS to replace kube-dns

### LINKED ISSUE #1019: Consider using CoreDNS as default
Docs: https://kubernetes.io/docs/tasks/administer-cluster/coredns/
KEP for GA: https://github.com/kubernetes/community/pull/1956

mikkeloscar:
The purpose of this is to introduce CoreDNS as a replacement for kube-dns. Fixes #1019 

Since we are currently investigating problems with kubedns, it makes sense to compare the two and validate if it makes sense to switch to CoreDNS.

If we want to switch the idea would be to roll it out as a replacement for one of the kube-dns instances and then replace the kube-dns-b instance later on. This way we should have no downtime for dns.

I have added an extra label `component=cluster-dns` which enables a pod anti-affinity rule to attempt to spread the dns pods as much as possible.

The configuration is from: https://github.com/coredns/deployment/blob/master/kubernetes/coredns.yaml.sed#L48-L61

The plugins used are:

* errors - [enable error logging](https://coredns.io/plugins/errors/)
* health - [enable health check endpoint](https://coredns.io/plugins/health/)
* kubernetes - [enable kube-dns replacement](https://coredns.io/plugins/kubernetes/)
  * `pods insecure` - Always return an A record with IP from request (without checking k8s). This option is is vulnerable to abuse if used maliciously in conjunction with wildcard SSL certs. This option is provided for backward compatibility with kube-dns.
  * `upstream` - [ADDRESS…] defines the upstream resolvers used for resolving services that point to external hosts (aka External Services aka CNAMEs). If no ADDRESS is given, CoreDNS will resolve External Services against itself. ADDRESS can be an IP, an IP:port, or a path to a file structured like resolv.conf.
  * `fallthrough` - [ZONES…] If a query for a record in the zones for which the plugin is authoritative results in NXDOMAIN, normally that is what the response will be. However, if you specify this option, the query will instead be passed on down the plugin chain, which can include another plugin to handle the query. If [ZONES…] is omitted, then fallthrough happens for all zones for which the plugin is authoritative. If specific zones are listed (for example in-addr.arpa and ip6.arpa), then only queries for those zones will be subject to fallthrough.
* prometheus - [enable prometheus metrics](https://coredns.io/plugins/health/)
* proxy - [proxy everything using host's resolv.conf](https://coredns.io/plugins/proxy/)
* cache - [enables a frontend cache](https://coredns.io/plugins/cache/)
* reload - [automatic reload of changed Corefile](https://coredns.io/plugins/reload/)

## TODO

* [ ] Compare to kube-dns
* [ ] Validate default configuration.
* [x] Mirror upstream docker image.


szuecs:
@mikkeloscar Can you please add information about why we change and why we use exactly this configfile?
All options in the configfile has to be understood and if these are only defaults we have to document it somehow.

cluster/manifests/deletions.yaml
```diff
@@ -4,5 +4,14 @@
 #   namespace: kube-system
 #   kind: deployment
 
-pre_apply: [] # everything defined under here will be deleted before applying the manifests
+pre_apply: # everything defined under here will be deleted before applying the manifests
```
hjacobs: wouldn't this leave us without any DNS during the cluster update procedure? (old kube-dns deleted, new CoreDNS not applied yet)

mikkeloscar:
@szuecs added a reference to the plugins defined in the config file.

cluster/manifests/deletions.yaml
mikkeloscar: > If we want to switch the idea would be to roll it out as a replacement for one of the kube-dns instances and then replace the kube-dns-b instance later on. This way we should have no downtime for dns.

@hjacobs maybe you are not aware, but we run two instances of kube-dns to have a fallback for each pod in the `relsolv.conf.` We have DNS servers on `10.3.0.10` and `10.3.0.11`. This will only delete the first one, so traffic should go to the second while coredns is being applied.

szuecs:
I tested CoreDNS and kube-dns with dnsblast. CoreDNS does not provide a better alternative considered to resolver problems, but it has a simpler setup and good metrics, which I consider to be worth the change.

mikkeloscar:
Sorry, I managed to rebase on dev in a way that I removed all the changes here. `git reflog` saved me..

szuecs:
I will close this in favour of https://github.com/zalando-incubator/kubernetes-on-aws/pull/1108

PR #1091, opened 2018-05-30
Remove obsolete static node pools (Legacy pools step 2/4)

linki:
Step 2/4 to remove legacy node pools:
* make CLM tolerate missing legacy node pool stack and userdata files (https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/49)
* update stack to remove legacy resources and set "empty" default values for to-be-removed stack parameters (https://github.com/zalando-incubator/kubernetes-on-aws/pull/1091) <==
* stop specifying stack parameters in CLM. Default stack values from step 2 will take over. (https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/46)
* finally remove unused stack parameters from stack config (https://github.com/zalando-incubator/kubernetes-on-aws/pull/1121)

Purpose of this PR:
* removes obsolete resources but keeps all stack parameters so that CLM will successfully invoke senza. It also sets defaults to each obsolete parameter so the final version of CLM can leave them out.


cluster/senza-definition.yaml
```diff
@@ -921,8 +796,3 @@ Outputs:
       Ref: WorkerIAMRole
     Export:
       Name: "{{ Arguments.ClusterID }}:worker-iam-role"
-  MasterLoadBalancerSecurityGroup:
-    Value:
-      Ref: MasterLoadBalancerSecurityGroup
-    Export:
-      Name: "{{ Arguments.ClusterID }}:master-load-balancer-security-group"
```
mikkeloscar, line 928: This we still need :)

linki: I removed it because I couldn't find a place where it's referenced 😕

mikkeloscar: @linki true, then it's not needed as an export :+1:

linki:
👍

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1097, opened 2018-05-31
update skipper to add timeouts to calls to apiserver

szuecs:
update skipper to fix the problem of hanging skipper in apiserver call could take long and result in inconsistent routing tables, which could lead to 504 GW timeouts from ALB or skipper depending on the timeout settings used


szuecs:
:+1:

mikkeloscar:
:+1:

PR #1103, opened 2018-05-31
Add missing priority classes on system components

aermakov-zalando:
We should avoid preemption of kube-system components, and now that some of them have priority all of them should have it.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #1108, opened 2018-06-01
redo kubernetes DNS setup to fix latency problems

szuecs:
* switch dns resolver to coredns, which has better metrics support
* enable node local dnsmasq cache to scale out dns
* node creation: lookup node ip to create a node local resolv.conf target to hit the daemonset dnsmasq cache before falling back to coredns service ip
* adjust node-local svc for dnsmasq monitoring
* coredns set resolve: pods disabled


szuecs:
We might need to delete the kube-dns deployment and service to make this rolling update work correctly.

hjacobs:
@szuecs any benchmark results for this change (yet)?

szuecs:
@hjacobs yes. No matter what I did dnsblast with 200req/s destroyed the dns setup:

vertical or horizontal scale or use more cache or use --no-negcache or a kube-dns replaced by coredns, I got 5s timeouts:

image

New setup looks much better with 100ms:
image
New setup is coredns resolver and dnsmasq local dns cache with more cache size and --no-ngecache. All pods lookup first from local dnsmasq with caching.

cluster/manifests/kube-dns/node-local-daemonset.yaml
```diff
@@ -0,0 +1,99 @@
...
+        volumeMounts:
+        - name: kube-dns-config
+          mountPath: /etc/k8s/dns/dnsmasq-nanny
+      - name: sidecar
+        image: registry.opensource.zalan.do/teapot/k8s-dns-sidecar-amd64:1.14.10
+        securityContext:
+          privileged: true
+        livenessProbe:
+          httpGet:
+            path: /metrics
```
mikkeloscar, line 78: @szuecs Should we change this to not call the metrics endpoint for liveness? You mentioned it's a lot of data for a liveness probe :)

cluster/manifests/kube-dns/node-local-daemonset.yaml
```diff
@@ -0,0 +1,99 @@
...
+        - containerPort: 53
+          name: dns
+          protocol: UDP
+        - containerPort: 53
+          name: dns-tcp
+          protocol: TCP
+        # see: https://github.com/kubernetes/kubernetes/issues/29055 for details
+        resources:
+          requests:
+            cpu: 150m
```
mikkeloscar: @szuecs do you have any metrics on how much CPU is needed? It seems like a lot to use `150m` on each node for part of the base infrastructure.

(we should see how much we use for all daemonsets combined).

cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -150,7 +150,7 @@ systemd:
       --node-labels=kubernetes.io/node-pool={{ .NodePool.Name }} \
       --node-labels={{ .Values.node_labels }} \
       --pod-manifest-path=/etc/kubernetes/manifests \
-      --cluster-dns=10.3.0.10,10.3.0.11 \
+      --cluster-dns=127.0.0.1:30053,10.3.0.10,10.3.0.11 \
```
mikkeloscar: I guess this is just left by mistake.

mikkeloscar:
We probably need to split this up and roll it out in two steps. First rolling out the daemonset and then rolling out the nodes where kubelet is configured to use the node ip in `resolv.conf`.

szuecs:
I get this log:

```time="2018-06-01T17:28:43Z" level=fatal msg="Fail to provision: run kubectl failed: exit status 1: The Service "coredns" is invalid: spec.clusterIP: Invalid value: "10.3.0.10": provided IP is already allocated```

@mikkeloscar do you know how to fix it and where?

cluster/manifests/kube-dns/node-local-daemonset.yaml
szuecs: I would go for the defaults first

cluster/manifests/kube-dns/node-local-daemonset.yaml
szuecs: I will test with less requests, I think it's less than 50m.

szuecs:
step 1: https://github.com/zalando-incubator/kubernetes-on-aws/pull/1112

PR #1109, opened 2018-06-04
Use dnsPolicy: Default for cluster autoscaler

aermakov-zalando:
Using kube-dns means that we can't scale-up from 0, since the autoscaler won't be able to resolve any addresses. Since it doesn't really use any addresses we can safely run it with `dnsPolicy: Default`.


szuecs:
:+1:

aermakov-zalando:
👍

PR #1110, opened 2018-06-04
Move the logging-agent secret as well

aermakov-zalando:
This was only done in alpha and beta, need to port to dev as well.


aermakov-zalando:
👍

linki:
👍

PR #1112, opened 2018-06-04
step1 redo DNS - rollout coredns

szuecs:
* replace kube-dns with coredns, fallback to kube-dns-b exists and will be deleted in step2
* cleanup kube-dns
* switch svc ip from kube-dns to coredns
* rollout inactive node-local dnsmasq daemonset, which will be activated in step2, which will change resolv.conf for pods

First step for #1019.


cluster/manifests/deletions.yaml
```diff
@@ -5,13 +5,14 @@
...
 
 # everything defined under here will be deleted before applying the manifests
-pre_apply:
-- name: logging-agent
-  namespace: visibility
-  kind: daemonset
-- name: logging-agent
-  namespace: visibility
-  kind: secret
+pre_apply: []
```
mikkeloscar, line 8: The `kube-dns` service needs to be deleted here in order to overwrite it with the new `coredns` service which uses the same ip.

szuecs: @mikkeloscar would create a downtime

PR #1115, opened 2018-06-04
redo DNS step1 - enable fallback to coredns

szuecs:
* replace kube-dns-b with coredns as fallback wil be changed in step2
* cleanup kube-dns-b
* switch svc ip from kube-dns-b to coredns
* rollout inactive node-local dnsmasq daemonset, which will be activated in step2, which will change resolv.conf for pods


szuecs:
:+1:

cluster/manifests/kube-dns/node-local-daemonset.yaml
```diff
@@ -0,0 +1,99 @@
...
+        volumeMounts:
+        - name: kube-dns-config
+          mountPath: /etc/k8s/dns/dnsmasq-nanny
+      - name: sidecar
+        image: registry.opensource.zalan.do/teapot/k8s-dns-sidecar-amd64:1.14.10
+        securityContext:
+          privileged: true
+        livenessProbe:
+          httpGet:
+            path: /metrics
```
mikkeloscar, line 80: @szuecs what about changing this to not do a liveness probe on the metrics endpoint which is quite a lot of data? (my comment in the other PR probably got lost).

szuecs: I replied before that this is the default and I am not sure if it makes sense to change it. 
Maybe they thought about it, maybe not....

mikkeloscar: Probably not :)

But we can optimize later.

cluster/manifests/kube-dns/node-local-svc.yaml
```diff
@@ -0,0 +1,23 @@
...
+    kubernetes.io/name: "KubeDNS"
+  annotations:
+    prometheus.io/path: /metrics
+    prometheus.io/port: "9054"
+    prometheus.io/scrape: "true"
+spec:
+  selector:
+    application: dnsmasq-node
+  type: NodePort
+  externalTrafficPolicy: Local
```
mikkeloscar: I guess we can just use `ClusterIP` and drop the `externalTrafficPolicy` as this service is only needed for the metrics collection now.

szuecs: no ClusterIP is not possible because of the port range

cluster/manifests/kube-dns/node-local-daemonset.yaml
```diff
@@ -0,0 +1,99 @@
...
+        - -logtostderr
+        - -configDir=/etc/k8s/dns/dnsmasq-nanny
+        - -restartDnsmasq=true
+        - --
+        - -k
+        - --cache-size=10000
+        - --no-negcache
+        - --server=/cluster.local/10.3.0.10#53
+        - --server=/in-addr.arpa/10.3.0.10#53
+        - --server=/ip6.arpa/10.3.0.10#53
```
mikkeloscar: Should we use `10.3.0.11` to go to coredns from the beginning? Then we can safely remove the kube-dns service `10.3.0.10` after dnsmasq has been enabled.

cluster/manifests/kube-dns/node-local-svc.yaml
mikkeloscar: But afaict we don't need the node ports anymore, since we use hostNetwork on the pods?

szuecs: Is this hurting you?
From my side I don't see a value in changing and as far as I remember there was a port assign problem.

mikkeloscar: I just think it makes sense to remove unnecessary configuration which is not used. It's confusing to have `externalTrafficPolicy: Local` explicitly defined if we are not even using it. The service is only used for gathering metrics, so it should be as simple as possible IMO.

szuecs: I will change it. I reread https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport and it was our hack with the 10.4.0.1 I think why we had this service nodeport.

mikkeloscar: >  it was our hack with the 10.4.0.1 I think why we had this service nodeport.

yes, that I was I meant :)

szuecs:
:+1:

mikkeloscar:
:+1:

PR #1116, opened 2018-06-04
Set default memory scale to 0.9

aermakov-zalando:
We actually have less memory available, which is a problem on large instances. Use 90% or 1500Mi, which should cover both cases.


hjacobs:
:+1:

aermakov-zalando:
👍

PR #1117, opened 2018-06-04
Reduce the memory scaling coefficient to 0.85

aermakov-zalando:
m5 instances have even less memory available, so 0.9 wouldn't work for m5.xlarge. Let's just reduce to 0.85 and see if we can get this information from somewhere.


aermakov-zalando:
👍

hjacobs:
:+1:

PR #1120, opened 2018-06-05
Dev to alpha

szuecs:
- redo DNS step1 - enable fallback to coredns

Update zmon-worker
- Add fields for monitoring replication in redis wrapper
- Add align param to Scalyr wrapper to fix count queries with float results
- Add redis connect/read timeouts for most redis connections
- Add alert exception for any alert returning None

instance types
- Reduce the memory scaling coefficient to 0.85
- m5 instances have even less memory available, so 0.9 wouldn't work for
- m5.xlarge. Let's just reduce to 0.85 and see if we can get this

Add muaaz's pubkey to node pools
Use dnsPolicy: Default for cluster autoscaler
Move the logging-agent secret as well
Autoscaling: resize buffer pods based on instance type
Update kube-job-cleaner
Add missing priority classes on system components
Fix critical logging pods
kube-job-cleaner update
 - Update to v0.7
 - Run every 5 minutes instead of every hour
update skipper to add timeouts to calls to apiserver
Move logging-agent back to kube-system
Revert "Remove unneeded ExperimentalCriticalPodAnnotation feature gate"

Unfortunately, the daemonset controller doesn't really implement
priorities (<https://github.com/kubernetes/kubernetes/blob/v1.10.2/pkg/controller/daemon/daemon_controller.go#L1219>),
and the pod criticality check is only enabled if the annotation feature
gate is enabled.

update skipper to fix the problem of hanging skipper in apiserver call could take long and result in inconsistent routing tables, which could lead to 504 GW timeouts from ALB or skipper depending on the timeout settings used

Skip unauthorized audit events
eas: Expose expiry time in GET /access-requests
Skip unauthorized audit events
Update Emergency Access Service to include list access requests endpoint


szuecs:
:+1:

mikkeloscar:
:+1:

linki:
👍

PR #1127, opened 2018-06-05
Alpha to beta without v1.10

szuecs:
* redo dns setup step1: add coredns resolver local dnsmasq cache rolled out not in use, enable in step2
* rollout new members key
* update zmon-worker


mikkeloscar:
:+1:

linki:
👍

PR #1128, opened 2018-06-06
prepare-pr.sh: correctly format commits with no body

aermakov-zalando:
Previously, these were formatted as `****`, which broke the lists completely.


mikkeloscar:
:+1:

linki:
👍

PR #1130, opened 2018-06-06
Search alpha align beta

mikkeloscar:
Align `search-alpha` with `beta`.

The change-set is so big because it was based on `alpha`, but going forward it will be based on `beta`. The actual changes are smaller:

* Autoscaling: resize buffer pods based on instance type
* Set default memory scale to 0.9
* Update Emergency Access Service to include list access requests endpoint
* Move logging-agent back to kube-system
* ZMON: use priorities instead of critical pod annotations
* Only enable auditing in production clusters
* replace kube-dns-b with coredns as fallback wil be changed in step2


mikkeloscar:
:+1:

linki:
👍

PR #1134, opened 2018-06-07
Upgrade logging-agent

mohabusama:
Changes:

- Fix logs with missing pod metadata labels and annotations
- log-watcher will now make an API call for every pod instead of using cached pod per namespace. This is to overcome some race condition were container config files are available before their corresponding pods. Log-watcher will ignore containers with no pods until the next watcher iteration.
- This issue is mainly visible in cronjob/job pods.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1139, opened 2018-06-08
use path-prefix match instead of kubernetes regexp path match

szuecs:
v0.10.20 
* will fix tokeninfo filters to work for longer time
* switch to -kubernetes-path-mode=path-prefix will change the behaviour of path matches to be non kubernetes api compliant, but kubernetes user doc compliant and more obvious


szuecs:
:+1:

mikkeloscar:
:+1:

njuettner:
👍

PR #1142, opened 2018-06-11
Fix azID in templates

aermakov-zalando:
`azId` was renamed to `azID`, but I forgot to update the template.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1145, opened 2018-06-12
Add support for node pool taints via config items

linki:
* add taints from a specific config item to master and worker node pools
* tolerate the `dedicated` taint for core components (where not already present)

todo:
- [x] node-labels (we use the node-pool label for node selection)
- [x] taints exposed as launch config tags

### DIFF
--- a/cluster/manifests/skipper/daemonset.yaml
+++ b/cluster/manifests/skipper/daemonset.yaml
@@ -25,6 +25,7 @@ spec:
           [{"container": "skipper-ingress", "parser": "skipper-access-log"}]
     spec:
       priorityClassName: system-node-critical
+      tolerations:
       - key: dedicated
         effect: NoSchedule
       affinity:
--- a/cluster/manifests/skipper/daemonset.yaml
+++ b/cluster/manifests/skipper/daemonset.yaml
@@ -26,17 +26,10 @@ spec:
     spec:
       priorityClassName: system-node-critical
       tolerations:
-      - key: dedicated
+      - operator: Exists
         effect: NoSchedule
-      affinity:
-        nodeAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-            nodeSelectorTerms:
-            - matchExpressions:
-              - key: kubernetes.io/role
-                operator: NotIn
-                values:
-                - master
+      nodeSelector:
+        kubernetes.io/role=worker
       hostNetwork: true
       containers:
       - name: skipper-ingress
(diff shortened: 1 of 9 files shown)


aermakov-zalando:
LGTM. Not sure about skipper, though. If the main use case for this is dedicated compute nodes or other special setups, do we really want to run skipper on those?

cluster/manifests/skipper/daemonset.yaml
```diff
@@ -25,6 +25,8 @@ spec:
           [{"container": "skipper-ingress", "parser": "skipper-access-log"}]
     spec:
       priorityClassName: system-node-critical
+      - key: dedicated
```
mikkeloscar: Shouldn't the key name come from the config item?

linki: I didn't know this was intended. The key is arbitrary and lives solely in the cluster registry.

The current approach:
* config items allow setting arbitrary taints on node pools.
* `skipper` tolerates any dedicated nodes in the cluster which means nodes dedicated to anything (gpus, local disks, particular teams, used for performance tests, ...) will still run `skipper` because it's a cluster component (the meaning of dedicated is derived from https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/ "Example Use Cases")

If we wanted to tolerate it dynamically, it would essentially mean: "tolerate any taint that is defined in the cluster registry via a config item on a worker node pool" (because we don't know the name we have to take all of them).

Is this what we want?

mikkeloscar: > Is this what we want?

That's what I imagined. Either skipper is special in the sense that it should not be scheduled on "special" nodes or it's like all the other daemonsets IMO.

WDYT?

aermakov-zalando: I'd suggest not running it on dedicated nodes (and switch a normal deployment in the future, which would solve the issue completely).

linki: @mikkeloscar DaemonSets support node selectors, so I wonder if we could achieve the same by running `skipper` DS with the `kubernetes.io/role=worker` node selector. (and tolerate all `NoSchedule` taints)

```yaml
      # skipper daemon set
      tolerations:
      - operator: Exists
        effect: NoSchedule
      nodeSelector:
        kubernetes.io/role=worker
```

njuettner:
👍

linki:
👍

njuettner:
👍

PR #1146, opened 2018-06-12
Run the custom cluster autoscaler build

aermakov-zalando:
Change the autoscaler to our own fork and enable the new behaviour where it creates template nodes using information from the launch configuration (see https://github.com/kubernetes/autoscaler/issues/789 for details). This fixes a number of issues:
 * autoscaling works correctly when we change the instance type
 * database pods will trigger scale-up so we don't need to baby-sit the clusters during upgrades
 * autoscaler will correctly handle multi-AZ ASGs for zone-restricted pods instead of just getting stuck.

Let's test this internally for now, I'll send a PR upstream later.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1147, opened 2018-06-13
Don't touch non-owned ingress resources

mikkeloscar:
This updates the ingress-template-controller to a version which does not manage ingress resources it doesn't own.


mikkeloscar:
e2e tests will not cover this...

mikkeloscar:
:+1:

hjacobs:
:+1:

PR #1149, opened 2018-06-13
Don't touch non-owned ingress resources (beta)

mikkeloscar:
Fast-track to beta:

* This updates the ingress-template-controller to a version which does not manage ingress resources it doesn't own. (#1147)


mikkeloscar:
:+1:

linki:
👍

PR #1150, opened 2018-06-13
Expose node labels as tags for the autoscaler

linki:
This exposes additonal node labels as AWS tags so that the autoscaler can create the correct node templates.

It also adds an additional role label with the - by now it seems - agreed upon name. The old one is used by our ingress controller and can be removed here once it's not used there anymore.


linki:
👍

aermakov-zalando:
👍

PR #1153, opened 2018-06-13
Run flannel via etcd-proxy

aermakov-zalando:
Kubernetes client-go apparently doesn't have any timeouts whatsoever, which causes issues when we roll the master nodes because flannel might get stuck for minutes until the kernel closes the apiserver connection. Let's run it through `etcd-proxy` as a mitigation.


aermakov-zalando:
**TODO**: rename it to something more generic!

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1155, opened 2018-06-14
Allow evicting any number of buffer pods

aermakov-zalando:
We didn't notice that pdb-controller created a PDB for the buffer pods, preventing users' stuff from getting scheduled ASAP. :(


aermakov-zalando:
👍

njuettner:
👍

PR #1160, opened 2018-06-14
Taint worker nodes not-ready on startup

mikkeloscar:
This is the follow up to #1151 for tainting worker nodes as not ready on startup. They will then be marked ready when the `kube-node-ready-controller` removes the taint after all expected daemonset pods has started.


aermakov-zalando:
We went with a different workaround, so closing this for now.

PR #1162, opened 2018-06-15
Enable volume mount for node pools with instance_storage

mikkeloscar:
This allows enabling local storage config for certain node pools. This is needed by search and we need to merge the search changes back in to the main channels before it gets out of hand to maintain a separate `search-alpha` branch.

This is a very simple way of allowing this feature. It's not very flexible, but I think we can iterate on this later.

It defines `/data` as the standard place for mounting storage, by default it will just be an empty directory on the root disk.

### DIFF
--- a/cluster/node-pools/worker-default/userdata.clc.yaml
+++ b/cluster/node-pools/worker-default/userdata.clc.yaml
@@ -22,6 +22,24 @@ systemd:
   - name: locksmithd.service
     mask: true
 
+  # this should only be enabled for instances with instance storage. e.g.
+  # i3.large instances
+  {{if index .NodePool.ConfigItems "instance_storage"}}
+  - name: data-disk0.mount
+    enable: true
+    contents: |
+      [Unit]
+      Before=local-fs.target
+
+      [Mount]
+      What=/dev/nvme0n1
+      Where=/data/disk0
+      Type=ext4
+
+      [Install]
+      WantedBy=local-fs.target
+  {{end}}
+
   - name: set-hostname.service
     enable: true
     contents: |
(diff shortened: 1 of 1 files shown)


cluster/node-pools/worker-default/userdata.clc.yaml
```diff
@@ -22,6 +22,24 @@ systemd:
   - name: locksmithd.service
     mask: true
 
+  # this should only be enabled for instances with instance storage. e.g.
```
aermakov-zalando, line 25: Please automate this, we already have the data in CLM. :)

cluster/node-pools/worker-default/userdata.clc.yaml
```diff
@@ -22,6 +22,24 @@ systemd:
...
+  # i3.large instances
+  {{if index .NodePool.ConfigItems "instance_storage"}}
+  - name: data-disk0.mount
+    enable: true
+    contents: |
+      [Unit]
+      Before=local-fs.target
+
+      [Mount]
+      What=/dev/nvme0n1
```
aermakov-zalando, line 35: The path should also be automatically determined based on whether the instance is NVMe-based  or not.

aermakov-zalando:
I still think that we should use instance storage by default as local storage for all containers if it's available. This will save us money and massively speed up everything which uses even a tiny bit of local storage without having to do complicated mounts. However, this can come in follow-up PRs.

mikkeloscar:
> I still think that we should use instance storage by default as local storage for all containers if it's available. This will save us money and massively speed up everything which uses even a tiny bit of local storage without having to do complicated mounts. However, this can come in follow-up PRs.

Sure, I wanted something which is already tested that we can roll out fairly soon. Then we can test your suggestion later on, which I think requires some more work.

szuecs:
@mikkeloscar @aermakov-zalando should we merge this and improve later?

aermakov-zalando:
Sure!

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #1163, opened 2018-06-15
fix traffic switching

szuecs:
fix blue green deployments creted 500s caused by a skipper route matching issue

### DIFF
--- a/cluster/manifests/skipper/daemonset.yaml
+++ b/cluster/manifests/skipper/daemonset.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: skipper-ingress
-    version: v0.10.20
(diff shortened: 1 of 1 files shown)


hjacobs:
Can you say some more words about the issue and how it was resolved?

szuecs:
@hjacobs basically we check now, that if a decision header is set, we do not match the lb_group predicate again, which could trigger an infinte loop.

Once we have a decision of the loadbalancer group made by one decision route, we made it impossible to match the decision route again, which triggered the infinite loop. Infinite loop was triggered in ~10% of the calls, dependent on the Traffic() predicate setting.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #1164, opened 2018-06-15
Use required affinity for autoscaling pods

aermakov-zalando:
This makes the scaling a bit more fair, since the autoscaler doesn't really handle preferred affinities. It won't allow us to use more than 3 pods, however, but if we ever want this we can just split into 3 different deployments, one for each AZ.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #1165, opened 2018-06-15
hotfix blue green deployments creted 500s caused by a skipper route matching issue

szuecs:
Once we have a decision of the loadbalancer group made by one decision route, we made it impossible to match the decision route again, which triggered the infinite loop. Infinite loop was triggered in ~10% of the calls, dependent on the Traffic() predicate setting.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #253, opened 2017-02-06
allow querying Skipper metrics from ZMON (in the cluster)

hjacobs:
Allow querying Skipper Ingress metrics on port 9911 from worker nodes.

PR #1168, opened 2018-06-15
external-dns v0.5.3

njuettner:
Includes:

* Print a message if no hosted zones match (aws provider) (#592) @svend
* Add support for NodePort services (#559) @grimmy
* Update azure.md to fix protocol value (#593) @JasonvanBrackel
* Add cache to limit calls to providers (#589) @jessfraz
* Add Azure MSI support (#578) @r7vme
* CoreDNS/SkyDNS provider (#253) @istalker2

Signed-off-by: Nick Jüttner <nick@zalando.de>

### DIFF
--- a/cluster/manifests/external-dns/deployment.yaml
+++ b/cluster/manifests/external-dns/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: external-dns
-    version: v0.5.2
+    version: v0.5.3
 spec:
   strategy:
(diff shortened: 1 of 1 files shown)


njuettner:
**Note:**
This includes now caching by default for 1h.

By changing the flag `--txt-cache-interval=` it can be overwritten.
It shouldn't effect the behaviour of external-dns.

mikkeloscar:
:+1:

njuettner:
👍

PR #1171, opened 2018-06-15
Autoscaling buffer: use nodeSelectors instead of anti-affinity

aermakov-zalando:
The current version autoscaler only pays attention to existing pods when handling anti-affinity predicates (see https://github.com/kubernetes/autoscaler/issues/257). This means that if multiple buffer pods are evicted at the same time, the autoscaler can provision two nodes in the same AZ and only rectify the situation by upscaling another AZ and downscaling the unnecessary node after one of those is ready and the buffer pod is scheduled. This issue will hopefully be fixed in the future, but for now we can avoid this by using node selectors instead of anti-affinities.

Note that this slightly changed the semantics of `autoscaling_buffer_pods` from the total count to per-az count, but we never actually overrode this setting to anything except 0.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1175, opened 2018-06-15
Drop kube-node-ready-controller, implement another workaround

aermakov-zalando:
Unfortunately, kube-node-ready-controller doesn't really work well with the cluster autoscaler. The autoscaler only looks at node conditions, so what actually happens is that a pod is marked Pending, the autoscaler provisions a new node, the node becomes Ready/Schedulable (even though the pods can't be scheduled there because of the taints) and then the autoscaler happily requests another node. This can then repeat for quite a long time, creating lots of useless nodes.

Let's try a simpler workaround. We remove the kube-node-ready-controller and instead replace flannel's init container with a normal container that sleeps first and then copies the CNI data onto the host. This results in the following sequence of events:

 * node is started, but is marked as NotReady because networking is not configured yet
 * flannel and the delayed init container start up at the same time
 * flannel gets the NodeCIDR and sets up pod networking on the node
 * flannel updates the node with the annotation that marks it as configured for flannels on other nodes, which update their configurations as well
 * sleep finishes, the delayed init container copies the networking config onto the host. Kubelet marks the node as fully ready and now the stuff that requires networking can actually start

This has the added benefit that applications which run with host networking can start immediately without waiting for anything.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1179, opened 2018-06-19
Better pool name for the multi-ASG pool

aermakov-zalando:
`multiaz` is exactly the opposite of what it actually does. I've added the new name as a symlink, so we can start using it and then clean up after we fix all existing clusters to not use the old name.

### DIFF
--- a/cluster/node-pools/worker-splitaz
+++ b/cluster/node-pools/worker-splitaz
@@ -1,0 +1,1 @@
+worker-multiaz


linki:
@aermakov-zalando Regarding the legacy node-pool cleanup my changes have all been merged to dev. Or am I missing something?

aermakov-zalando:
👍

linki:
👍

PR #1180, opened 2018-06-19
Port flannel changes from alpha

aermakov-zalando:
Let's just do it in one commit and fix with `merge -s ours` later. The critical annotation is kept because I don't remember if we need it in 1.9, so better safe than sorry.


linki:
👍

aermakov-zalando:
👍

PR #1181, opened 2018-06-19
Flannel to 1.9 (production 2/2)

aermakov-zalando:
Port flannel changes from alpha.

**WARNING** do not merge until #1183 is merged and the clusters are fully updated.

PR #1183, opened 2018-06-19
dnsmasq-node to 1.9 (production 1/2)

aermakov-zalando:
This needs to be merged before #1181 so that flannel doesn't evict dnsmasq.


linki:
👍

aermakov-zalando:
👍

PR #1187, opened 2018-06-19
Add flannel annotation back to avoid extra disruptions

aermakov-zalando:
Otherwise the 1.10 merge will cause network issues again.


aermakov-zalando:
👍

linki:
👍

PR #1188, opened 2018-06-19
flannel: use OnDelete

aermakov-zalando:
A temporary workaround until we have a mitigation for https://github.com/coreos/flannel/issues/1002 in place (which can be as simple as a wrapper that just kills flannel instead of gracefully shutting it down). Every trivial change to the flannel daemonset will cause disruption in production clusters otherwise.

### DIFF
--- a/cluster/manifests/flannel/daemonset.yaml
+++ b/cluster/manifests/flannel/daemonset.yaml
@@ -8,7 +8,7 @@ metadata:
     version: v0.10.0
 spec:
   updateStrategy:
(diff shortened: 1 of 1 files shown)


linki:
👍

linki:
There's a handy init system designed for containers which also allows signal rewriting: https://github.com/Yelp/dumb-init/tree/v1.2.1#signal-rewriting

aermakov-zalando:
👍

PR #1195, opened 2018-06-25
coredns - limit the number of roundtrips

njuettner:
Automatically detect queries being made with a known Kubernetes suffix

See https://github.com/coredns/coredns/issues/1752

There is also a comment which mentions enable sequential lookups for A and AAAA, using `single-request` or `single-request-reopen` in /etc.resolv.conf

...
>
              single-request (since glibc 2.10)
                     Sets RES_SNGLKUP in _res.options.  By default, glibc
                     performs IPv4 and IPv6 lookups in parallel since
                     version 2.9.  Some appliance DNS servers cannot handle
                     these queries properly and make the requests time out.
                     This option disables the behavior and makes glibc
                     perform the IPv6 and IPv4 requests sequentially (at the
                     cost of some slowdown of the resolving process).

              single-request-reopen (since glibc 2.9)
                     Sets RES_SNGLKUPREOP in _res.options.  The resolver
                     uses the same socket for the A and AAAA requests.  Some
                     hardware mistakenly sends back only one reply.  When
                     that happens the client system will sit and wait for
                     the second reply.  Turning this option on changes this
                     behavior so that if two requests from the same port are
                     not handled correctly it will close the socket and open
                     a new one before sending the second request.

### DIFF
--- a/cluster/manifests/coredns/configmap.yaml
+++ b/cluster/manifests/coredns/configmap.yaml
@@ -15,6 +15,7 @@ data:
             upstream
             fallthrough in-addr.arpa ip6.arpa
         }
+        autopath @kubernetes
         prometheus :9153
         proxy . /etc/resolv.conf
         cache 30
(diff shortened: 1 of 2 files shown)


szuecs:
I am not sure if this is a CoreDNS problem. It hsould be more a client resolver or dnsmasq issue which could be caused by a glibc/mussl issue, see:
https://github.com/kubernetes/kubernetes/issues/56903#issuecomment-359897058

PR #1197, opened 2018-06-25
Update hyperkube to 1.10.5

linki:
* and switch to gcr based hyperkube image which is updated more frequently

This update is important as Kubernetes 1.10.5 fixes the backoffLimit issue for CronJobs: https://github.com/kubernetes/kubernetes/pull/63650 (fix regression in `v1.JobSpec.backoffLimit` that caused failed Jobs to be restarted indefinitely.)


linki:
👍

gargravarr:
:+1:

PR #1198, opened 2018-06-26
Update to Kubernetes 1.11

linki:
Let's see what the e2es have to say about this.

Changes:
* Update Hyperkube to 1.11.x
* Update CoreOS to the latest stable (as of July 4)
* Remove PodPriority feature flag as it's enabled by default
* ~~Switch from KubeProxy's `iptables` mode to `ipvs` mode~~
* ~~Disable docker version change, therefore it uses 18.03.x with the AMI above which is **not** officially supported by Kubernetes 1.11 (let's see where we end up; 1.12 and 17.03.x work fine and are supported)~~
* Remove `PersistentVolumeClaimResize` feature flag ([enabled by default](https://kubernetes.io/blog/2018/07/12/resizing-persistent-volumes-using-kubernetes/?utm_source=cncf&utm_medium=email&utm_campaign=cloudnativemonthly&utm_content=july18))
* Remove `CustomResourceSubresources` feature gate ([enabled by default](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#subresources)).


cluster/manifests/10-vertical-pod-autoscaler/vpa-crd.yaml
```diff
@@ -27,25 +27,26 @@ spec:
...
-                      name:
+                      containerName:
                         type: string
                       mode:
                         type: string
                         enum:
-                        - "On"
-                        - "Off"
+                          - "Auto"
+                          - "Off"
```
linki: This file change is just for testing. Don't merge me.

hjacobs:
@linki what is the impact of switching to ipvs? We should have a good understanding of how it works and what the benefits/drawbacks are.

linki:
@hjacobs It claims to be more performant and seems to become the default option going forward. From [the release notes](https://kubernetes.io/blog/2018/06/27/kubernetes-1.11-release-announcement/): "This change delivers better network throughput, better programming latency, and higher scalability limits for the cluster-wide distributed load-balancer that comprises the Kubernetes Service model"

However, there's [a known-issue](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#known-issues) in Kubernetes 1.11 so I think we should wait for the next release.

I was just curious if e2e would pass: it does.

szuecs:
@linki maybe we want not to switch to ipvs https://github.com/kubernetes/kubernetes/issues/57841.
I was suggesting this to do but maybe we should rethink and review what bugs exists on ipvs and which in iptables.

linki:
@szuecs yes, it's mentioned as a known issue in the release notes as well. We should wait.

mikkeloscar:
Yes, this sounds like quite a serious problem :)

cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -868,14 +868,6 @@ storage:
           --table nat \
           --to-destination ${PRIVATE_EC2_IPV4}:8181
 
-  {{if not (index .Cluster.ConfigItems "upstream_docker")}}
-  - filesystem: root
-    path: /etc/coreos/docker-1.12
-    mode: 0644
-    contents:
-      inline: yes
-  {{end}}
```
linki: We should fix the docker version again. I would like to try out docker 17 as 18 isn't officially supported.

Furthermore the new (better) way of defining versions for those bundled binaries is using `torcx`: https://coreos.com/blog/toward-docker-17-in-container-linux

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1202, opened 2018-06-27
Bumping image-policy-webhook version

muaazsaleem:
A new image-policy-webhook version was tagged. So bumping the version here to be able to test

* Includes initial support for PierOne's `X-Trusted` header


linki:
👍

mikkeloscar:
:+1:

muaazsaleem:
@linki and I decided to not roll out these changes together with the new "trusted" policy so we can make the time spent rolling out more valuable

PR #1204, opened 2018-06-28
autoscaling_buffer_pools: use a more specific regex

aermakov-zalando:
This allows us to not tinker with config items every time we create a dedicated node pool.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1207, opened 2018-06-28
Update ExternalDNS to v0.5.4

linki:
Now that the cache has been fixed and is disabled by default, let's enable and use it 😃

### DIFF
--- a/cluster/manifests/external-dns/deployment.yaml
+++ b/cluster/manifests/external-dns/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: external-dns
-    version: v0.5.3
+    version: v0.5.4
 spec:
   strategy:
     type: Recreate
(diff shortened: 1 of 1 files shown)


njuettner:
👍

cluster/manifests/external-dns/deployment.yaml
```diff
@@ -16,20 +16,21 @@ spec:
...
       - name: external-dns
-        image: registry.opensource.zalan.do/teapot/external-dns:v0.5.3
+        image: registry.opensource.zalan.do/teapot/external-dns:v0.5.4
         args:
         - --source=service
         - --source=ingress
         - --provider=aws
         - --registry=txt
         - --txt-owner-id={{ .Region }}:{{ .LocalID }}
+        - --txt-cache-interval=1h
```
mikkeloscar: I suggest we fix the problem mentioned here: https://github.com/kubernetes-incubator/external-dns/pull/612#issuecomment-399253828 before we enable caching in our setup.

mikkeloscar:
:+1:

linki:
👍

PR #1208, opened 2018-06-28
Open dnsmasq metrics port within private network

mikkeloscar:
Opens the dnsmasq metrics port `9054` within the private network so we can collect metrics from any dnsmasq.

We were only collecting for the dnsmasq on the same node as prometheus giving unexpected data when querying.


linki:
👍

mikkeloscar:
:+1:

PR #1210, opened 2018-06-29
Remove the VPA control plane

aermakov-zalando:
We couldn't make VPA work on the simplest possible application and nobody really uses it anyway. The control plane is just needlessly wasting resources, so let's drop it for now.


aermakov-zalando:
👍

njuettner:
👍

PR #1212, opened 2018-06-29
autoscaler: enable --balance-similar-node-groups

aermakov-zalando:
We have 3 ASGs with different AZs but the same instance type and labels. The autoscaler should try to balance their sizes unless it's not possible for some reason, but this doesn't happen by default and we need to explicitly enable it by running the autoscaler with `--balance-similar-node-groups`.


hjacobs:
Can you describe the impact/purpose of this change?

aermakov-zalando:
@hjacobs done.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #1220, opened 2018-07-04
First draft of NVIDIA GPU support

muaazsaleem:
- Use deamonset to install gpu drivers
- Use k8s-device-plugin to expose GPUs to GPU enabled containers

### Test with a GPU deployment
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: gpu-tester
  labels:
    application: gpu-tester
spec:
  replicas: 1
  selector:
    matchLabels:
      application: gpu-tester
  template:
    metadata:
      labels:
        application: gpu-tester
    spec:
      containers:
      - name: gpu-tester
        image: gcr.io/google_containers/cuda-vector-add:v0.1
        resources:
          limits:
            nvidia.com/gpu: 1 # requesting 1 GPUs
```

`zkubectl logs -f <POD Name>`
```
[Vector addition of 50000 elements]
Copy input data from the host memory to the CUDA device
CUDA kernel launch with 196 blocks of 256 threads
Copy output data from the CUDA device to the host memory
Test PASSED
Done
```

TODO:
- [ ] documentation
- [x] mirror images to pierone
- [x] extensible resources admission controller

### DIFF
--- a/cluster/manifests/nvidia/nvidia-driver-installer.yaml
+++ b/cluster/manifests/nvidia/nvidia-driver-installer.yaml
@@ -1,3 +1,7 @@
+# based on:
+#
+#   https://github.com/shelmangroup/coreos-gpu-installer/blob/master/daemonset.yaml
+#
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
@@ -5,9 +9,10 @@ metadata:
   namespace: kube-system
   labels:
     application: nvidia-driver-installer
-  annotations:
-    scheduler.alpha.kubernetes.io/critical-pod: ''
+    version: c117d39
 spec:
+  updateStrategy:
+    type: RollingUpdate
   selector:
     matchLabels:
       application: nvidia-driver-installer
@@ -15,13 +20,12 @@ spec:
     metadata:
       labels:
         application: nvidia-driver-installer
-      annotations:
-        scheduler.alpha.kubernetes.io/critical-pod: ''
+        version: c117d39
     spec:
       tolerations:
-      - effect: NoSchedule
-        key: dedicated
-        value: gpu
+      - key: nvidia.com/gpu
+        effect: NoSchedule
+        operator: Exists
       nodeSelector:
         kubernetes.io/node-pool: worker-gpu
       priorityClassName: system-node-critical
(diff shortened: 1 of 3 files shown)


cluster/manifests/nvidia/nvidia-driver-installer.yaml
```diff
@@ -0,0 +1,66 @@
...
+spec:
+  selector:
+    matchLabels:
+      application: nvidia-driver-installer
+  template:
+    metadata:
+      labels:
+        application: nvidia-driver-installer
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
```
hjacobs: I think this annotation is deprecated

njuettner: @muaazsaleem https://github.com/zalando-incubator/kubernetes-on-aws/pull/974

linki:
Let's add documentation so that another team can try it out.

cluster/manifests/nvidia/nvidia-driver-installer.yaml
linki: We kept the manifests mostly like upstream. Will fix the missing things now.

linki:
We now have a better understanding of how this works:
* `nvidia-driver-installer` runs as a one-shot DaemonSet on GPU nodes and installs NVIDIA drivers on the host (compiled at runtime, host target path mounted via `hostPath`) [1]
* `nvidia-gpu-device-plugin` runs as a permanent DaemonSet on GPU nodes (i.e. GRPC server) and registers the GPUs attached to the node with the `kubelet`.
  * This registers the number of `nvidia.com/gpu` resources on that node as allocatable
* When a Pod is created that uses a resource limit of `nvidia.com/gpu` two things happen:
  * The scheduler schedules the Pod to nodes that have available `nvidia.com/gpu` (i.e. `nvidia-gpu-device-plugin` is running and no Pod is already using all the nodes GPUs. This will limit this Pod to GPU nodes)
  * `ExtendedResourceToleration` admission controller will automatically add tolerations for `nvidia.com/gpu` so that the Pod **can** be scheduled on GPU nodes.
* Furthermore, each container that declares a resource limit of `nvidia.com/gpu` gets additional content injected into their docker container by the kubelet:
  * A `hostPath` mount that mounts the NVIDIA drivers from the host into the container
  * Env vars such as `LD_LIBRARY_PATH` and other CUDA related variables that point to driver locations in the container [2]

Overall: The only thing users need to do is specify a `nvidia.com/gpu` resource limit.

[1] It currently selects the nodes via a nodeSelector using the node pool name. This should be changed to a selector that selects any node that has an NVIDIA GPU.
[2] `nvidia-gpu-device-plugin` sets this when registering the device with the `kubelet`. You only see this in `docker`-land, Kubernetes manifests don't reflect the mounts.

muaazsaleem:
👍

linki:
👍

PR #1221, opened 2018-07-09
Update webhook to allow aws-cloud-provider serviceaccount to patch nodes

mikkeloscar:
Allow aws-cloud-provider service account to patch nodes which is needed for setting a taint on nodes where volume attachment fails.

https://github.com/kubernetes/kubernetes/commit/514f219c229647ad2a3d6223fbcb596ea032363c


aermakov-zalando:
👍

hjacobs:
:+1:

PR #1223, opened 2018-07-10
Use latest compliant pause image everywhere

linki:
* Unifies and updates the `pause` container's version to `3.1` so we don't pull redundant versions.
* Instructs the `kubelet` to use this version for the magical hidden network pods as well.


szuecs:
:+1:

hjacobs:
:+1:

PR #1228, opened 2018-07-10
fix kernel race condition in DNAT rules for DNS

szuecs:
This adds a sidecar into flannel pod spec, that adds a 1-4ms delay for udp port 53 packets going through DNAT, such that the kernel race condition happens less likely.

Background see https://blog.quentin-machu.fr/2018/06/24/5-15s-dns-lookups-on-kubernetes/
https://github.com/kubernetes/kubernetes/issues/62628
https://github.com/coreos/flannel/pull/1001#issuecomment-403802099

### DIFF
--- a/cluster/manifests/flannel/daemonset.yaml
+++ b/cluster/manifests/flannel/daemonset.yaml
@@ -84,6 +84,25 @@ spec:
           mountPath: /etc/kube-flannel/
         - name: run
           mountPath: /run
+      - args:
+        - -c
+        - /tc-flannel.sh
+        command:
+        - /bin/bash
+        image: registry.opensource.zalan.do/teapot/flannel-tc:v0.0.2
+        name: flannel-tc
+        resources:
+          requests:
+            cpu: 25m
+            memory: 25Mi
+        securityContext:
+          privileged: true
+        stdin: true
+        volumeMounts:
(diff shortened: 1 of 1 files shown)


cluster/manifests/flannel/daemonset.yaml
```diff
@@ -100,3 +119,7 @@ spec:
       - name: run
         hostPath:
           path: /run
+      - hostPath:
+          path: /lib/tc
+          type: ""
+        name: lib-tc
```
mikkeloscar, line 125: What is this volume used for?

szuecs: For making "tc" possible to run

mikkeloscar: Maybe we could put this folder in the Dockerfile if it's just needed by `tc`. Then the configuration would be a bit less confusing :) But not a big deal.

szuecs: But if anyone else adds also something to tc, then it does not have the xtables.lock, which might do more harm. I think it's good to have it in the hostpath.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #1229, opened 2018-07-10
Support for adding node pool labels via config items

linki:
This allows configuring node labels per node pool, e.g.
```yaml
  - discount_strategy: none
    instance_type: p2.xlarge
    min_size: 1
    max_size: 1
    name: worker-gpu
    profile: worker-default
    config_items:
      labels: cloud.google.com/gke-accelerator=nvidia-tesla-k80,foo=bar
      taints: nvidia.com/gpu=present:NoSchedule,foo2=bar2
```

Like above, this can be used to mark nodes of a node pool with their special hardware such as GPUs so that users can schedule pods on a subset of all GPU nodes without using node pool names, e.g.

via `nodeSelector` for targetting a specific type of device as explained [here](https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/):
```yaml
  nodeSelector:
    cloud.google.com/gke-accelerator: nvidia-tesla-p100 # or nvidia-tesla-k80 etc.
```

or for targetting all types of devices like for [installing the drivers](https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/k8s-1.10/nvidia-driver-installer/cos/daemonset-preloaded.yaml#L42-L48) themselves for arbitrary node pool names:
```yaml
      affinity:
        nodeAffinity:
            nodeSelectorTerms:
            - matchExpressions:
              - key: cloud.google.com/gke-accelerator
                operator: Exists
```

### DIFF
--- a/cluster/node-pools/master-default/userdata.clc.yaml
+++ b/cluster/node-pools/master-default/userdata.clc.yaml
@@ -162,7 +162,7 @@ systemd:
       --register-with-taints=node-role.kubernetes.io/master=:NoSchedule{{if index .NodePool.ConfigItems "taints"}},{{.NodePool.ConfigItems.taints}}{{end}} \
       --allow-privileged \
       --node-labels=node-role.kubernetes.io/master,kubernetes.io/role=master,master=true \
-      --node-labels=kubernetes.io/node-pool={{ .NodePool.Name }} \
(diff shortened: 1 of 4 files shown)


mikkeloscar:
:+1:

linki:
👍

PR #1230, opened 2018-07-10
Don't use custom rootfs mount

linki:
Using a custom path for rootfs mount (different from the upstream config) isn't fully supported by the script. Falling back to the original path to make it work for now.


linki:
👍

mikkeloscar:
:+1:

PR #1232, opened 2018-07-10
Update kube-ingress-aws-controller to v0.7.2

mikkeloscar:
Fixes a bug when using custom target-port value

https://github.com/zalando-incubator/kube-ingress-aws-controller/releases/tag/v0.7.2

We don't depend on this particular feature in our setup, but it makes sense to run the latest version so we can eat our own dog food.


hjacobs:
:+1:

linki:
👍

PR #1241, opened 2018-07-16
coredns/dnsmasq: bump CPU requests

aermakov-zalando:
This might help with the DNS issues we're experiencing.

PR #1242, opened 2018-07-16
dev-to-alpha

aermakov-zalando:
* **coredns/dnsmasq: bump CPU requests**
   <sup>Merge pull request #1241 from zalando-incubator/bump-dns-requests</sup>


mikkeloscar:
I think we want an intermediate branch for this instead of dev directly into alpha?

aermakov-zalando:
@mikkeloscar thanks for noticing!

PR #1244, opened 2018-07-17
Increase the conntrack size to reduce chances of DNS packet loss

aermakov-zalando:
ZMON makes so many connections and DNS requests that the conntrack table on a node hosting its pods gets completely overflown. If one of coredns pods is running on the same node, the whole cluster will start experiencing issues.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1247, opened 2018-07-17
add node label aws.amazon.com/spot=true

hjacobs:
Mark EC2 Spot instances with `aws.amazon.com/spot=true`. I initially planned to use the same label as kops, they use "spot", see https://github.com/kubernetes/kops/blob/70118bda4d05a331dce2b6c4f5c82a4cf4b1a001/docs/instance_groups.md

Marking nodes with a "aws.amazon.com/spot" label allows to have adequate scheduling decisions and helps to build proper reporting, e.g. like https://github.com/hjacobs/kube-resource-report/issues/18

The `spot_price` value is set in https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/4efbc647b6364095d349230b10bc5f8a5d9300f4/provisioner/node_pools.go#L168

### DIFF
--- a/cluster/node-pools/worker-default/userdata.clc.yaml
+++ b/cluster/node-pools/worker-default/userdata.clc.yaml
@@ -160,6 +160,7 @@ systemd:
       --allow-privileged \
       --node-labels=kubernetes.io/role=worker \
       --node-labels=kubernetes.io/node-pool={{ .NodePool.Name }}{{if index .NodePool.ConfigItems "labels"}},{{.NodePool.ConfigItems.labels}}{{end}} \
+      --node-labels=spot={{if .Values.spot_price ne ""}}true{{else}}false{{end}} \
       --node-labels={{ .Values.node_labels }} \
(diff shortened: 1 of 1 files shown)


hjacobs:
:+1:

hjacobs:
I will change the label to `node-role.kubernetes.io/spot-worker=true` as this better fits with the other labels we have and is also used in https://github.com/pusher/k8s-spot-termination-handler

mikkeloscar:
@hjacobs I was also gonna suggest something more descriptive than just `spot=true`

I'm worried this one might conflict (in meaning) with our current label though: `kubernetes.io/role=worker`.

Since spot is Amazon specific it could make sense to namespace it like: `amazon.com/spot=true` or something like that.

hjacobs:
@mikkeloscar OK, your suggestion makes sense! I changed it to `aws.amazon.com/spot` :smile:

hjacobs:
:+1:

gargravarr:
:+1:

PR #1252, opened 2018-07-18
Set default image policy based on environment

linki:
Uses a default config to define a value for image policy if not present in cluster registry.


aermakov-zalando:
Shouldn't it be different for dev and production?

aermakov-zalando:
👍

szuecs:
:+1:

PR #1253, opened 2018-07-18
Use a custom flannel image so it's always killed

aermakov-zalando:
Mitigates https://github.com/coreos/flannel/issues/1002.

### DIFF
--- a/cluster/manifests/flannel/daemonset.yaml
+++ b/cluster/manifests/flannel/daemonset.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: flannel
-    version: v0.10.0
+    version: v0.10.0-8
 spec:
   updateStrategy:
     type: OnDelete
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

hjacobs:
@aermakov-zalando can you create an upstream PR for flannel with the fix?

aermakov-zalando:
@hjacobs I didn't fix the problem in flannel, just added a workaround.

gargravarr:
:+1:

notxcain:
@aermakov-zalando could you please describe the workaround you've used?

aermakov-zalando:
@notxcain we just kill `flannel` instead of letting it terminate gracefully.

linki:
@notxcain ...so that it doesn't have a chance to execute any shutdown action, which in our case somehow triggered some undesireable effects.

notxcain:
@aermakov-zalando I see, thanks. In our case flannel doesn't terminate, writes `shutdownHandler sent cancel signal...` to log and network stops working.

PR #1255, opened 2018-07-18
add k8s-spot-termination-handler

### LINKED ISSUE #1238: Handle EC2 Spot Termination Notice
We should automatically drain nodes which receive an EC2 Spot Termination Notice. Example for ECS (from [AWS blog post](https://aws.amazon.com/blogs/compute/powering-your-amazon-ecs-cluster-with-amazon-ec2-spot-instances/)):

```bash
#!/bin/bash

while sleep 5; do
  if [ -z $(curl -Isf http://169.254.169.254/latest/meta-data/spot/termination-time) ]; then
    /bin/false
  else
    # TODO: do kubectl drain instead
    ECS_CLUSTER=$(curl -s http://localhost:51678/v1/metadata | jq .Cluster | tr -d \")
    CONTAINER_INSTANCE=$(curl -s http://localhost:51678/v1/metadata \
      | jq .ContainerInstanceArn | tr -d \")
    aws ecs update-container-instances-state --cluster $ECS_CLUSTER \
      --container-instances $CONTAINER_INSTANCE --status DRAINING
  fi
done

hjacobs:
Fixes #1238. Uses https://github.com/pusher/k8s-spot-termination-handler

### DIFF
--- a/cluster/manifests/spot-termination-handler/daemonset.yaml
+++ b/cluster/manifests/spot-termination-handler/daemonset.yaml
@@ -29,5 +29,5 @@ spec:
             cpu: 100m
             memory: 100Mi
       nodeSelector:
-        "spot": "true"
+        "aws.amazon.com/spot": "true"
 
--- a/cluster/manifests/spot-termination-handler/daemonset.yaml
+++ b/cluster/manifests/spot-termination-handler/daemonset.yaml
@@ -1,0 +1,33 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: spot-termination-handler
+  namespace: kube-system
+  labels:
+    application: spot-termination-handler
+spec:
+  template:
+    metadata:
+      name: spot-termination-handler
+      labels:
+        application: spot-termination-handler
+    spec:
+      containers:
+      - name: spot-termination-handler
+        image: registry.opensource.zalan.do/teapot/k8s-spot-termination-handler:v0.1.0
+        imagePullPolicy: IfNotPresent
+        env:
+          - name: NODE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+        resources:
+          requests:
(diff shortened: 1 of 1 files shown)


mikkeloscar:
I don't think this will actually work. They run the drain command without `--delete-local-data` so it would complain if you try to delete pods that use host volumes (which some of our users do i.e. search).

Also seems a bit overkill to run a python process to spawn a kubectl drain command :)

Lastly I just noticed that `kubectl drain` doesn't handle statefulset pods very well. Something we explicitly handle in CLM code: https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/193e2c05cd6c0cc29097af66605dc9ad2937a952/pkg/updatestrategy/node_pool_manager.go#L502-L505

My point is that this needs some more work to be useful for us.

hjacobs:
@mikkeloscar I know that it's just a start, but I think stateful and local data is not so important as a first step (better to have "something" than nothing).

If you think Python is "overkill", we can also use the `kube-aws` project, but that's [a long Bash script](https://github.com/kube-aws/kube-spot-termination-notice-handler/blob/master/entrypoint.sh).. :smirk:  (I just want to avoid creating another project which does exactly the same just because we are not happy with details which could be fixed upstream).

mikkeloscar:
> @mikkeloscar I know that it's just a start, but I think stateful and local data is not so important as a first step (better to have "something" than nothing).

My point is the command will not work if just one pod on the node has local data. It will just exit with non-0.

> If you think Python is "overkill"

I admit this was a bit snarky, but I'm mostly thinking in terms of overhead for memory and the docker file size. I don't expect it to be so much, but it all adds up, especially if it runs, but doesn't work anyway.

aermakov-zalando:
Alternative: https://github.com/zalando-incubator/kubernetes-on-aws/pull/1296

hjacobs:
Closing as #1296 was merged.

PR #1256, opened 2018-07-19
fix: client closed request should not count as 5xx

szuecs:
fix: client closed request should not count as 5xx, instead of 504 we will now count a 499, similar to nginx
log: reduce logging in case of client closed request in the proxy


njuettner:
👍

linki:
👍

PR #1257, opened 2018-07-19
Reduce kube-controller-manager logging

mikkeloscar:
We enabled `--v=4` logging in controller manager once and never changed it again. It's now logging roughly 1GB per hour (reported by users) and rarely provides a lot of information.

In many clusters `kube-controller-manager` is logging more than any other application.

I propose we reduce the logging to default levels to reduce logging and save cost.


mikkeloscar:
:+1:

elgalu:
:+1:

PR #1264, opened 2018-07-20
fix: client closed request should not count as 5xx

szuecs:
fix: client closed request should not count as 5xx, instead of 504 we count it as 499


szuecs:
:+1:

szuecs:
conformance test was crashing while creating cluster in ou CI environment. Since this change does not affect any kubernetes core component, let's merge it

aermakov-zalando:
👍

PR #1266, opened 2018-07-20
Master nodes: enable T2 unlimited

aermakov-zalando:
* Switch master node pool to use LaunchTemplate instead of LaunchConfiguration
* Automatically enable T2 unlimited for master nodes

**Note**: depends on https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/66!

### DIFF
--- a/cluster/node-pools/master-default/stack.yaml
+++ b/cluster/node-pools/master-default/stack.yaml
@@ -15,7 +15,9 @@ Resources:
     Properties:
       HealthCheckGracePeriod: 480
       HealthCheckType: EC2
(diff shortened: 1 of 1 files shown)


mikkeloscar:
@aermakov-zalando Why not add it to all node pools while we are at it? Want to test it on masters first?

aermakov-zalando:
@mikkeloscar I wanted to experiment on masters first, yeah. We also don't use T2 for worker nodes, so the switch to launch templates there isn't that critical.

aermakov-zalando:
👍

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1273, opened 2018-07-23
Replace senza-definition with pure cloudformation stack definition

mikkeloscar:
This replaces the `senza-defintion.yaml` with a `cluster.yaml` defined in pure cloudformation (with go templating).

Depends on https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/67

One open question. Should we keep support for the "main" domain for the load balancer? That is, if the apiserver hostname is `kube-1.teapot.zalan.do` then the main domain is `kube.teapot.zalan.do`.

I want to get rid of it as it's just a leftover from how senza works and not something we ever used. I'm just worried that someone might depend on it?


hjacobs:
@mikkeloscar I think we can get rid of the main domain.

mikkeloscar:
~Also made the CIDR used in the etcd stack configurable. Changed the default value from `172.16.0.0/12` -> `172.31.0.0/16` since the default VPC network is not that big.~

Moved to separate PR #1274

mikkeloscar:
@hjacobs :+1: Removed the main domain.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #1274, opened 2018-07-24
Make Network CIDR used for etcd security groups configurable

mikkeloscar:
Make Network CIDR used for etcd security groups configurable

Part of #1273 

This needs to be rolled out to all clusters before we make the related changes in CLM.

Changed the default value from `172.16.0.0/12` -> `172.31.0.0/16` since the default VPC network is not that big.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #1283, opened 2018-07-26
[dev] Pass VpcID to etcd stack

mikkeloscar:
Passes the VpcID to the etcd stack which is needed when running in a non-default VPC.

This needs to be merged to all channels in the following steps:
* scale down CLM
* Merge to dev, alpha, beta (e2e test will not work for this because CLM can't work with both versions of the etcd stack). alpha PR: #1288 , beta PR: #1289  

* Merge change in CLM: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/68

Once the new CLM is deployed it will correctly update the ETCD stack (without rolling any nodes) now with the VPC ID passed to the stack.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #1284, opened 2018-07-27
Skipper: use more histogram buckets for latency metrics

aermakov-zalando:
The default buckets are capped at 10 seconds, which doesn't really work for real applications. Increase the number of the buckets so we have better numbers.


aermakov-zalando:
👍

aermakov-zalando:
JFYI: the default configuration is `.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10`.

szuecs:
:+1:

PR #726, opened 2017-12-04
Promote dev to alpha

linki:
Notable changes:
* Add support for our own etcd proxy
* Prepare kube-proxy daemonset rollout
* Prepare flannel daemonset rollout
* Add PodCIDR Controller for flannel migration
* Use latest version of ingress controller
* Remove ClusterID tag frome CF
* Expire old etcd backups
* Upgrade zmon-agent to include deep-link (clean tag)

PR #724, opened 2017-12-04
Promote alpha to beta

linki:
Notables changes:
* Remove third party resource for platform credentials set (Migrates PlatformCredentials to CRD)
* etcd-backup: reduce memory request/limits
* Update kube-aws-autoscaler to 0.11

PR #707, opened 2017-11-27
Convert PlatformCredentialsSets from TPR -> CRD

mikkeloscar:
This replaces the Third Party Resource defintion for the `PlatformCredentialsSets` to a Custom Resource Definition.

It's a prerequisite before merging that all clusters in the target branch has been converted manually following the guide here: https://kubernetes.io/docs/tasks/access-kubernetes-api/migrate-third-party-resource/

## Clusters

* [x] teapot
* [ ] stups-test
* [ ] playground

PR #1286, opened 2018-08-01
skipper update custom-routes change and OAuth tokeninfo request timeouts

szuecs:
This will use a released version:
  - Add -histogram-metric-buckets option (#722 @aermakov-zalando )
    If set, use a custom set of buckets for Prometheus histogram metrics
    instead of the default one.
  - Add flag for Oauth request timeout (#707 @addityasingh )
  - Add filter setQuery(string) #723 (#724) to set only the key without =
  - feature: Adds Gauges to metrics (#726 @topicus)
  - fix kube dataclient: (#720)
        custom-routes annotation should specify the path with pathmode if set in the ingress spec
        pathmode lookup is global for an ingress and should follow the other annotation handling, too. It will default to the global pathmode if it is not able to parse the mode. Find more information about pathmodes in our documentation https://opensource.zalando.com/skipper/kubernetes/ingress-usage/\#ingress-path-handling
        adds path to route id and add test cases for custom skipper routes via ingress annotations
        makes sure to test #720 (comment) object works as expected

### DIFF
--- a/cluster/manifests/skipper/daemonset.yaml
+++ b/cluster/manifests/skipper/daemonset.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: skipper-ingress
-    version: v0.10.45
+    version: v0.10.55
     component: ingress
 spec:
   selector:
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

mikkeloscar:
:+1:

PR #1292, opened 2018-08-05
Run kube-downscaler in test clusters

hjacobs:
Run [kube-downscaler](https://github.com/hjacobs/kube-downscaler) in all test clusters with some default configuration (to be overwritten via config items).

Needs internal documentation and announcement.

### DIFF
--- a/cluster/manifests/kube-downscaler/deployment.yaml
+++ b/cluster/manifests/kube-downscaler/deployment.yaml
@@ -1,0 +1,40 @@
+{{ if neq .Environment "production" }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kube-downscaler
+  namespace: kube-system
+  labels:
+    application: kube-downscaler
+    version: v0.2
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      application: kube-downscaler
+  template:
+    metadata:
+      labels:
+        application: kube-downscaler
+        version: v0.2
+    spec:
+      containers:
+      - name: downscaler
+        # see https://github.com/hjacobs/kube-downscaler/releases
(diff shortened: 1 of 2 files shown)


cluster/manifests/kube-downscaler/deployment.yaml
```diff
@@ -0,0 +1,41 @@
...
+        # see https://github.com/hjacobs/kube-downscaler/releases
+        image: hjacobs/kube-downscaler:0.3
+        args:
+          # run every 3 minutes
+          - --interval=180
+          - --exclude-namespaces=kube-system,visibility
+          - "--default-uptime={{ .ConfigItems.downscaler_default_uptime }}"
+        resources:
+          limits:
+            memory: 100Mi
```
mikkeloscar: Can we please avoid over-committing :)

hjacobs: @mikkeloscar done, fixed --- this was not intended and I copied the manifest from https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/external-dns/deployment.yaml . We should probably once clean up all manifests if we want to avoid overcommit.

muaazsaleem:
Since this is first of many changes to our test environment that won't affect prod clusters, should we also take this opportunity to create a "stable" channel for prod clusters? 
AFAIK we've already agreed to do it eventually but haven't decided on a time.

szuecs:
@muaazsaleem please have this a separate discussion from this PR. We had it already 1-2 times to discuss and we agreed on having more automations to support the channel flow, before creating more channels, but I think we can also discuss it again offline.

szuecs:
:+1:

hjacobs:
:+1:

PR #1296, opened 2018-08-07
Add a simple spot termination handler

aermakov-zalando:
The actual code is around 10 lines of shell, so I'm not sure if reusing code from other projects actually helps us. The existing handlers also don't log the information we'd be interested in (instance type/region), so it's easier to just implement what we want.


hjacobs:
I'm OK with that simple solution if the team accepts it :smile:

aermakov-zalando:
👍

szuecs:
:+1:

PR #1297, opened 2018-08-07
eas: use etcd-proxy to reduce timeouts

aermakov-zalando:
This should reduce the issues for the users until client-go actually fixes the timeout issues.


njuettner:
👍

aermakov-zalando:
👍

PR #1299, opened 2018-08-08
Kube 1.11 patch

njuettner:
Takes the latest changes from dev and bump Kubernetes version to 1.11.2.
Also changes the mode of kube-proxy back to iptables, because of the issue https://github.com/kubernetes/kubernetes/issues/57841 which will be fixed in the future


njuettner:
Cluster-Autoscaler needs an update to v.1.3.1, but we currently run a custom autoscaler. There's a PR from Alexey which is still open https://github.com/kubernetes/autoscaler/pull/1021.

njuettner:
👍

linki:
👍

PR #1302, opened 2018-08-09
Increase memory reservation for autoscaling buffer pods

aermakov-zalando:
Previous setting of 1.5Gi wasn't enough for c5 instances. Let's increase it to 2Gi for now.


aermakov-zalando:
👍

njuettner:
👍

PR #1305, opened 2018-08-09
switch from flannel to amazon-vpc-cni-k8s

tuxlife:
this would be the first step to get rid of the flannel layer.

the pr adds https://github.com/aws/amazon-vpc-cni-k8s (Networking plugin repository for pod networking in Kubernetes using Elastic Network Interfaces on AWS) as daemon set and removes the flannel settings.

_Open topics:_
- [ ] Build the amazon-vpc-cni-k8s container with CDP and provide the image with pierone
- [ ] Test with running pods
- [ ] Performance benchmark and compare the performance with flannel
- [ ] Find the right pod/node limit
- [ ] update the documentation
- [ ] update the documentation
- [ ] add dedicated service-accounts for the daemon set with webhook support (see the rbac config in the original repo)


mikkeloscar:
Closing for now, we haven't picked this up.

PR #1306, opened 2018-08-09
oauth filter in front of emergency-access-service 

szuecs:
add oauth filter in front of eas to make scanners not contributing to measured and alerted errors


szuecs:
:+1:

njuettner:
👍

PR #1307, opened 2018-08-10
increase memory to the conntrack exporter more reliable

szuecs:
increase memory to the conntrack exporter more reliable


szuecs:
:+1:

njuettner:
👍

PR #1310, opened 2018-08-10
ZMON: use a user-defined priority class

aermakov-zalando:
In 1.11, system priority classes [are restricted](https://github.com/kubernetes/kubernetes/pull/65593) to `kube-system` only. Fix by creating a separate priority class for ZMON components.


aermakov-zalando:
👍

njuettner:
👍

PR #1311, opened 2018-08-11
Remove Autobahn and Depoyment Jenkins permissions after shutting them down

dryewo:
As part of cleanup.


hjacobs:
@dryewo your PR did not work as the e2e tests don't work with PRs from forks. I created a new PR from a branch #1312 with the same contents.

PR #1313, opened 2018-08-13
Cluster Autoscaler: Deployment -> DaemonSet

aermakov-zalando:
We run 1 master node for test clusters, so it makes no sense to run 2 replicas of CA.

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -5,7 +5,10 @@
 #   kind: deployment
 
 # everything defined under here will be deleted before applying the manifests
-pre_apply: []
+pre_apply:
+- name: kube-cluster-autoscaler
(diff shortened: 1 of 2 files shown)


linki:
Another option is to keep it as a `Deployment` and set `replicas` to `${APISERVER_COUNT}`.

linki:
It needs a slight change in CLM as we currently hardcode that value to `1` even for multi-master clusters.

aermakov-zalando:
👍

linki:
👍

mikkeloscar:
Does it even make sense to run more than one replica of the autoscaler?

aermakov-zalando:
@mikkeloscar It uses leader election so it doesn't hurt, and if one of the nodes is broken somehow the autoscaling still works.

PR #1318, opened 2018-08-16
fix memory leak in skipper, that is triggered by using tokeninfo filters

szuecs:
fix memory leak in skipper, that is triggered by using tokeninfo filters


szuecs:
:+1:

njuettner:
👍

Issue #1217, opened 2018-07-02
Move logging-agent back to visibility namespace


linki:
Kubernetes 1.11 supports critical addons for all namespaces. This means we can move logging-agent to visibility.

mikkeloscar:
This was done

PR #1334, opened 2018-08-17
Critical pods don't have to run in kube-system

njuettner:
In Kubernetes 1.11 pod priority and preemption feature is enabled by default.
We can run critical pods in all namespaces. 

#1217 Moving the logging-agent into visibility

### DIFF
--- a/cluster/manifests/logging-agent/daemonset.yaml
+++ b/cluster/manifests/logging-agent/daemonset.yaml
@@ -3,7 +3,7 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: logging-agent
-  namespace: kube-system
+  namespace: visibility
   labels:
     application: logging-agent
     version: v0.19
@@ -22,7 +22,7 @@ spec:
         version: v0.19
         component: logging
     spec:
-      priorityClassName: system-node-critical
+      priorityClassName: visibility-zmon
       tolerations:
       - operator: Exists
         effect: NoSchedule


mikkeloscar:
@linki could you link the reference to this change? From what I see in the CHANGELOG it sounds like the opposite? https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#action-required-1 (third item from the top).

linki:
@mikkeloscar Looks like it changed again in `v1.11.1`. The `v1.11.0` has it here: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#urgent-upgrade-notes (second item)

I think we should be able to create additional priority classes that are on the same level as kube-system's critical ones to achieve the same effect (if we want to move it to the `visibility` namespace).

aermakov-zalando:
@linki that's actually forbidden, the highest user-definable priority is lower than the system ones. however, since nobody actually uses priorities in the clusters it should be fine. :)

linki:
Interesting. That means anything outside `kube-system` will always be of (minimally) lower priority.

Is that fine for the `visibility` components?

szuecs:
@linki yes it's ok, because cluster needs to be operable before, but we might want to drop some stuff into other namespaces.

linki:
Alright, then I guess we have a decision.

But `visibility-zmon` is a bad choice for a priority class name as it doesn't mean anything.

Since `system-node-critical` and `system-cluster-critical` (or any `system-*`) are reserved for being used exclusively in `kube-system` we'd have to introduce a new priority class for highly important "core components" that aren't living in `kube-system` (hopefully for a good reason).

How about `node-essential` or to be good citizen and avoid clashing `addon-node-essential`. `node` captures the scope and `essential` communicates its importance just below critical.

linki:
As for the value of the priority class. Here's an example of how IBM does it:

> ibm-app-cluster-critical (value: 900000000)
> 
> Select pods that are deployed into the ibm-system namespace when you create the cluster use this priority class to protect critical functionality for apps, such as the load balancer pods.

900000000 is a sane version of specifying `HighestUserDefinablePriority` - 1

https://console.bluemix.net/docs/containers/cs_pod_priority.html#pod_priority

PR #1335, opened 2018-08-17
Disable CFS quota

aermakov-zalando:
If `disable_cfs_quota` config item is set, disable CFS quota on all nodes. Supersedes #1044.


mikkeloscar:
:+1:

aermakov-zalando:
👍

hjacobs:
Another reason to disable CFS quota: https://github.com/kubernetes/kubernetes/issues/67577

And this gist (linked from the Kernel patch) is IMHO interesting: https://gist.github.com/bobrik/2030ff040fad360327a5fab7a09c4ff1

PR #1341, opened 2018-08-20
Configurable resources for the logging agent

arjunrn:
The logging agent turns out to be pretty expensive because it runs on every node in the cluster. With @linki  I have added some limits for the logging agent containers. If the are too low please leave a comment.


aermakov-zalando:
Since this is a daemonset, I'd keep `requests` the same as `limits`. Otherwise it's possible that kubelet will try to evict it when a node runs out of memory. I'm also not sure if using different settings for production and test clusters is useful — doesn't it just scale with the amount of data that's logged or the number of pods?

arjunrn:
@aermakov-zalando I wanted to make the values configurable through the cluster registry while keeping the current values as they are for existing clusters. Also I think it makes sense to have different values for both production and non-production environments because the quantity of logs produced in production environments is higher. Also if log shipping is slower in test clusters thats not an issue, so less resources would be ok. But I see your point about the over-committing. Would it be ok if I set the limits to the requests?

aermakov-zalando:
@arjunrn I've looked at a bunch of random clusters in our dashboard and it looks like this will just cause OOM kills in test clusters since a lot of these containers use 60 or more MiB in both test and production. This will also cause the agent to just crash if the users create many small pods, breaking log delivery and our alerting, just so we could save 50-100MiB of RAM per node (0.6-1% of the smallest instance type we use). Is it really worth it? :)

arjunrn:
@aermakov-zalando You are right. I've made the changes so they are the same for both production and test clusters.

aermakov-zalando:
`log-watcher` container currently uses more than 50MiB in a lot of clusters. Are you sure it's safe to reduce it?

arjunrn:
It's probably not. I will set it a higher value for now closer to the original limit.

arjunrn:
Closing this PR for now because it could disrupt normal logging operations. It can be revisited if there is a need to optimize these values on a per-cluster basis.

PR #1363, opened 2018-08-28
ZMON: use a user-defined priority class

aermakov-zalando:
If we don't do this, ZMON will be completely down for the whole 1.11 upgrade, which might take days.


aermakov-zalando:
👍

szuecs:
:+1:

PR #1365, opened 2018-08-31
Changed the skipper ingress to use image with lightstep integration.

arjunrn:
Changed the skipper ingress to use image with lightstep integration.


szuecs:
needs to have args and be templated with lightstep token

PR #1370, opened 2018-09-04
platformcredentialsset: add status subresource

aermakov-zalando:
This is needed so we can report the status from Credentials Provider.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1371, opened 2018-09-04
deploy lightstep skipper ingress

szuecs:
enable lightstep opentracing templated for test and production environment, replaces #1365

### DIFF
--- a/cluster/manifests/skipper/daemonset.yaml
+++ b/cluster/manifests/skipper/daemonset.yaml
@@ -38,6 +38,12 @@ spec:
         - name: ingress-port
           containerPort: 9999
           hostPort: 9999
+        env:
+        - name: LIGHTSTEP_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: skipper-ingress
+              key: lightstep-token
         args:
           - "-kubernetes"
           - "-kubernetes-in-cluster"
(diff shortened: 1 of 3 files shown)


cluster/manifests/skipper/daemonset.yaml
```diff
@@ -55,6 +54,7 @@ spec:
           - "-enable-connection-metrics"
           - "-oauth2-tokeninfo-url={{ .ConfigItems.tokeninfo_url }}"
           - "-histogram-metric-buckets=.01,.025,.05,.075,.1,.2,.3,.4,.5,.75,1,2,3,4,5,7,10,15,20,30,60,120,300,600"
+          - "-opentracing=lightstep component-name=skipper-ingress token={{ .ConfigItems.lightstep_token }} collector=tracing.zmon.zalan.do:8444 cmd-line=skipper-ingress max-buffered-spans=4096"
```
mikkeloscar: The token should be put in a secret. You could mount it as a environment variable and use the `$(TOKEN)` reference in the flag.

szuecs: no it can't be done like this, because the tracing plugins do not read it from ENV as @vetinari told me.

mikkeloscar: Nono, I mean you can use `$(TOKEN)` where you define the flag. Kubernetes should infer this when starting the pod.

mikkeloscar: @szuecs pushed a commit with the change I meant :) https://github.com/zalando-incubator/kubernetes-on-aws/pull/1371/commits/2cf5e0c10d76674ef97b0f4a3a99facbfb99df24

szuecs:
👍

mikkeloscar:
:+1:

PR #1386, opened 2018-09-10
Disable CFS quota by default

aermakov-zalando:
Let's add a config item to enable it back just in case. We can drop it a month or two later, once we're comfortable enough.


aermakov-zalando:
👍

szuecs:
:+1:

PR #1392, opened 2018-09-11
Kube 1.11 to dev

aermakov-zalando:
Syncing up now that 1.11 is the main version.


aermakov-zalando:
Oops, we already have #1198.

PR #1397, opened 2018-09-12
Go back to Kubelet's built-in cAdvisor

aermakov-zalando:
cAdvisor uses *WAY* too much memory, so let's go back to Kubelet's built-in metrics for now.


aermakov-zalando:
:+1:

arjunrn:
:+1:

PR #1400, opened 2018-09-13
Use v1.11 e2e image for conformance tests

mikkeloscar:
Since upgrading to v1.11 we should one the e2e image build from v1.11


aermakov-zalando:
👍

hjacobs:
:+1:

PR #1407, opened 2018-09-14
Use a fixed image

aermakov-zalando:
Since 1.11 tests don't work, let's use a fixed 1.10 image.


aermakov-zalando:
1.11 tests are fixed.

PR #1412, opened 2018-09-19
do not log the body of a request in auditLogs per default

szuecs:


cluster/manifests/skipper/daemonset.yaml
```diff
@@ -59,6 +59,7 @@ spec:
           - "-lb-healthcheck-interval=3s"
           - "-metrics-flavour=codahale,prometheus"
           - "-enable-connection-metrics"
+          - "--max-audit-body=0"
```
aryszka: should be a single dash ("-") instead of double ("--")

szuecs: thanks fixed

aryszka:
:+1:

szuecs:
:+1:

PR #1413, opened 2018-09-19
Run external-dns with limited privileges

linki:
https://github.com/kubernetes-incubator/external-dns/pull/714 replaces glog with a package that doesn't write to the filesystem.

It allows us to run `external-dns` with all the settings recommended by [https://kubesec.io/](https://kubesec.io/).

Note: this is a test and if desired can be added to the remaining components where applicable.

fixes https://github.com/kubernetes-incubator/external-dns/issues/595

requires: >v0.5.6


linki:
:+1:

mikkeloscar:
:+1:

PR #1419, opened 2018-09-20
kube-downscaler: increase memory limits

aermakov-zalando:
It's reading pods now, which requires a bit more memory.


aermakov-zalando:
👍

linki:
:+1:

PR #1420, opened 2018-09-20
More resilient Spot support

aermakov-zalando:
* If `autoscaler_experimental` config item is set, run CA built with https://github.com/aermakov-zalando/autoscaler/commit/d426c25feb2378c94a0f68279ac38552f6def504 and slightly reduce the max. node provisioning time.
 * Add a Spot label for CA so users can opt-out of Spot for some of their stuff
 * Downscale the ASG if we receive a pending termination so CA can fallback to another pool

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -229,6 +229,9 @@ Resources:
       Policies:
         - PolicyDocument:
             Statement:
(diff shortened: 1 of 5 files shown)


aermakov-zalando:
👍

gargravarr:
:+1:

PR #1428, opened 2018-09-21
Fix zmon-agent application label

sbrauner-z:
Fix zmon-agent application label
Signed-off-by: Shery Brauner <shery.brauner@zalando.de>

### DIFF
--- a/cluster/manifests/zmon-agent/deployment.yaml
+++ b/cluster/manifests/zmon-agent/deployment.yaml
@@ -4,17 +4,17 @@ metadata:
   name: "zmon-agent"
   namespace: "visibility"
   labels:
-    application: "zmon-agent"
+    application: "zmon-agent-core"
     version: "0.1-a56-zv2"
 spec:
   replicas: {{.ConfigItems.zmon_agent_replicas}}
   selector:
     matchLabels:
-      application: zmon-agent
+      application: zmon-agent-core
   template:
     metadata:
       labels:
-        application: "zmon-agent"
+        application: "zmon-agent-core"
         version: "0.1-a56-zv2"
       annotations:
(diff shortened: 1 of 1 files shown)


hjacobs:
:+1:

cluster/manifests/zmon-agent/deployment.yaml
```diff
@@ -4,17 +4,17 @@ metadata:
...
   labels:
-    application: "zmon-agent"
+    application: "zmon-agent-core"
     version: "0.1-a56-zv2"
 spec:
   replicas: {{.ConfigItems.zmon_agent_replicas}}
   selector:
     matchLabels:
-      application: zmon-agent
+      application: zmon-agent-core
```
mikkeloscar, line 13: This won't work. We need to delete the deployment first and re-apply.

nolar: BTW, that would be totally fine if you "move" the zmon-agent app from our account to yours — if this is possible — thus, no re-deployment will be needed. Anyway, we use the team-specific zmon-blah for our needs, as I have recently found in our repo.

mikkeloscar: @nolar thanks, then I propose we do that instead. Is that ok with you @sbrauner-z ?

sbrauner-z: @mikkeloscar @nolar sounds good to me if this is actually possible.

nolar: Oops. Just noticed it is a public GitHub. Removed the internal links & editing history.

jrake-revelant:
Closing this in favour of changing the ownership.

PR #1437, opened 2018-09-27
Configured AWS regions for kube-metrics-adapter.

arjunrn:
Forgot to add the region configuration for the adapter which would allow AWS SDK use for those regions.


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #1450, opened 2018-10-05
Default storage class: explicitly specify zones

aermakov-zalando:
We've switched to ASGs with min. size 0 recently, which means that some of the statefulsets, especially those created early on use just 1 AZ instead of all 3 (because the default is to use only the zones where there's at least one node).


aermakov-zalando:
👍

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1457, opened 2018-10-09
Run skipper as deployment with HPA

mikkeloscar:
## Problem statement

In our setup skipper-ingress runs as a daemonset. This means that we can only
scale our ingress "gateway" by scaling the number of nodes. If the resources
required by skipper scales linear with the cluster then this is fine, but if it
does not then we are unable to add more capacity to skipper.

There are multiple cases where this scale is not linear. One example was
when we tried to use skipper ingress to direct shadow traffic to something
running in Kubernetes but forwarding real requests to something still running
outside of Kubernetes. The solution here was to deploy a custom skipper as a
deployment scaling with an HPA (this included setting up all the metrics which
are normally provided "for free" in skipper ingress).

Another problem is when we try to reduce the overhead of system components by using
bigger instances but fewer of those, then we also have less skippers to serve
the traffic.

A third problem is that we over-provision skipper in most clusters by setting
resource memory requests lower than this limit. In some clusters this is fine,
but in others we are reaching the memory limit and skippers are dying. Because
these are daemonsets it's complicated to scale out, even vertically as we have
to account for what else is taking up resources on the nodes.

## Proposed solution

To work around the limitations of a daemonset based deployment the proposal is
to use a deployment instead and do Horizontal Pod Autoscaling based on CPU and
Memory. This way we can run a reduced number of pods in clusters without
traffic and auto scale to a high number in clusters with a lot of ingress
traffic.

### Load testing

To validate how a deployment based setup would compare to the current daemonset
based setup I ran the following "simple" load test:

1. Used locust to produce 12k req/s hitting a simple skipper deployment acting
   as the backend. Just a simple `GET https://example-app.example.org`. 

locust

2. Ran this against different configurations

### skipper-ingress daemonset

skipper-ingress-daemonset-150m

| Pods | Req/pod | req/millicore |
| ---- | ------- | ------------- |
| 15   | 800 | 1.3 |

### skipper-ingress deployment (500m/500Mi) with HPA

skipper-ingress-deployment-500m

| Pods | Req/pod | req/millicore |
| ---- | ------- | ------------- |
| 22   | 545 | 1.09 |

### skipper-ingress deployment (1000m/500Mi) with HPA

skipper-ingress-deployment-1000m

| Pods | Req/pod | req/millicore |
| ---- | ------- | ------------- |
| 9   | 1333 | 1.33 |

Note: Skipper-ingress was configured with `hostPort` in all cases meaning there
can only be one skipper per node even if the node has more resources available.
If we didn't use hostPort, then we could run multiple skippers per node and
possible get a better utilization of the node, but this would be at the cost of
more metrics because of more pods.

For the deployment based configuration an HPA configured to scale on `100%` CPU
and `80%` Memory was used. The argument for scaling on `100%` CPU is that a
node will often have a bit of spare CPU capacity and the pod thus won't get
throttled if it uses a little more than `100%`. If you scaled on say `80%` CPU
then you would always run with `20%` slack. Memory has to be below `100%`
because it helps combat OOM Kills if we scale before it reaches `100%`. Unless
skipper is being overrun memory will not be the upper bound for resources, so
the `20%` slack on Memory will not really be important relative to the CPU
resources consumed.

The numbers shown are not 100% accurate because of limited precision of the
collected data, but it gives a rough estimate showing that skipper, in this
scenario, scales linearly with req/s and CPU.

Furthermore it shows that running skipper as daemonset under these conditions
means that it is very much over-committing on CPU which is ok when no other
heavy workloads are running on a node, but becomes a problem when the nodes are
being utilized by something else.

Here is the problem clearly visible in a real cluster (past 7 days for cluster
with higest ingress load)

skipper-ingress-daemonset-prod

#### Example (based on "high-traffic" cluster)

Looking at the CPU peak over the last 7 days for this particular cluster we see
that skipper uses around `14200m` CPU spread over `23` `skipper-ingress` pods.
That is `~618m` per pod or an average over-commit of `312%`.

If we over the same period, for the same cluster, look at the min CPU usage for
skipper-ingress then we see it's around `1280m` over `17` `skipper-ingress`
pods, which is `~75m` per pod or a slack of `50%`.

If we had a deployment with `500m` per pod then we would only need to run `3`
pods during low traffic periods resulting in `1500m` total CPU (`14%` slack).
At peak this would require `29` pods with each `500m`. This would be `6` more
pods (and nodes) compared to the daemonset deployment, but it would also not be
over-committing on CPU as we are currently doing. Ideally we would have
Vertical scaling for these scenarios, but as it's not there yet, we could at
least manually scale to e.g. `1000m` per pod for clusters with this kind of
traffic pattern.

## Conclusion

Based on these observations I would suggest we run skipper as a deployment with
a default of `500m` and manually configure it to something like `1000m` for
high traffic clusters. This would solve two big problems we have right now 1)
we could scale skipper based on traffic (CPU) instead of only scaling it with
the cluster size. 2) we could reduce the number of skipper pods and thereby
reduce the number of metrics, thus taking load off prometheus.

Going with the deployment will also enable us to invest in Vertical Pod
Autoscaling of skipper-ingress later on (Has to be both VPA and HPA).


cluster/manifests/skipper/hpa.yaml
```diff
@@ -0,0 +1,19 @@
...
+  namespace: kube-system
+  labels:
+    application: skipper-ingress
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: skipper-ingress
+  minReplicas: 3
+  maxReplicas: 10
```
szuecs: max 10?
maybe {{{MAX_NUM_OF_NODES}}} to scale up to node max for a cluster?

cluster/manifests/skipper/hpa.yaml
```diff
@@ -0,0 +1,19 @@
...
+  metrics:
+  - type: Resource
+    resource:
+      name: cpu
+      targetAverageUtilization: 50
```
szuecs: We should target for 80%, similar to shop skipper. The good thing is we can scale faster than them.

aermakov-zalando:
Let's increase the CPU requests and memory limits a bit, now that we'll have a lot less pods? We can always tune it down later.

szuecs:
@aermakov-zalando we should have some calculation dependent on requests, instance type (--> how many cores and memory is available), number of instances, number of ingress and best also "filter usage", how much CPU/memory we set.
Any good idea for something like that?

aermakov-zalando:
I'd start with something more simple at first (bump the current values by 2x/3x) and then tune it based on actual usage. Maybe it's a good idea to move it to config items right now so we can at least tune it per-cluster.

mikkeloscar:
I bumped the resources and moved everything to config items. The default max is `100` which is pretty arbitrary.

I'll keep testing this in `stups-test`.

aermakov-zalando:
`100` sounds a bit too much, no?
Also, since it's using a `hostPort`, the pod should have an anti-affinity on the node level, because it won't be able to run anyway.

mikkeloscar:
> `100` sounds a bit too much, no?

True, we need to derive a better number either based on max nodes of the cluster or by estimating with some load tests.

> Also, since it's using a `hostPort`, the pod should have an anti-affinity on the node level, because it won't be able to run anyway.

The scheduler actually takes care of this as far as I understand:

> 0/18 nodes are available: 1 Insufficient memory, 1 node(s) didn't match node selector, **16 node(s) didn't have free ports for the requested pod ports**, 6 Insufficient cpu.

linki:
@mikkeloscar Could you also add to the issue description briefly what we want to achieve with this change?

arjunrn:
Ran a load test with around 10k requests/second with 3 skipper pods. The results are as follows:

### Host Networking

- `p99` 6 milliseconds
- `p95` 3 milliseconds
- `p90` 2 milliseconds
- `average` 2 milliseconds

### NodePort Service

- `p99` 11 milliseconds
- `p95` 6 milliseconds
- `p90` 4 milliseconds
- `average` 3 milliseconds

The size of the response was `1208` bytes.

Host Network

NodePort Service

mikkeloscar:
@arjunrn do you know how many nodes/pods were used in total and for the ingress skipper deployment during these tests? I think it relevant in terms of how likely a requests is to take the extra hub.

arjunrn:
21 m4.xlarge nodes 

### NodePort Service

- skipper pods: `17` _(HPA CPU usage 98%)_
- RequestCount	`573k`
- TargetResponseTime p99	`133ms`
- TargetResponseTime p95	`96ms`
- TargetResponseTime p90	`79ms`
- TargetResponseTime Average	`36ms`

### Host Networking
- skipper pods: `17` _(HPA CPU usage 112%)_
- RequestCount	`628k`
- TargetResponseTime p99	`3ms`
- TargetResponseTime p95	`2ms`
- TargetResponseTime p90	`2ms`
- TargetResponseTime Average	`1ms`

I also noticed that the usage across the pods when running as a NodePort service was not event. Some pods kept using much more CPU than others. I think this may be because of how load balancing is done at TCP level which means that existing connections are not immediately rebalanced.

NodePortLoadTest

szuecs:
product-analytics cluster runs skipper as daemonset with each ~400m CPU in use

szuecs:
@arjunrn how long the load test resulting in https://github.com/zalando-incubator/kubernetes-on-aws/pull/1457#issuecomment-438644667 was done?

arjunrn:
> @arjunrn how long the load test resulting in [#1457 (comment)](https://github.com/zalando-incubator/kubernetes-on-aws/pull/1457#issuecomment-438644667) was done?

@szuecs I ran the test for more than an hour. I've updated the comment with the graph from the load balancer.

arjunrn:
Reran the test with pod antiaffinities and got better results. I found that all pods were running on the same node and that would explain the high penalty for the `NodePort` service. Following are the results:

### Anti-Affinity Required

- Skipper pod: 21 _(Scaled by HPA)_
- RequestCount    `11,633 Request/Second`
- TargetResponseTime p99  `21ms`
- TargetResponseTime p95  `11ms`
- TargetResponseTime p90  `8ms`
- TargetResponseTime Average  `3ms`


### Anti-Affinity Preferred

- RequestCount    `11716 Requests/Second`
- TargetResponseTime p99  `21ms`
- TargetResponseTime p95  `11ms`
- TargetResponseTime p90  `7ms`
- TargetResponseTime Average  `3ms`

Both test were run for an hour.

cluster/manifests/skipper/deployment-a.yaml
```diff
@@ -30,6 +28,7 @@ spec:
         operator: Exists
       nodeSelector:
         kubernetes.io/role: worker
+        failure-domain.beta.kubernetes.io/zone: eu-central-1a
```
mikkeloscar: Let's use the `Region` variable here so we can also spin up clusters in other regions if needed in the future.

cluster/manifests/skipper/deployment-a.yaml
```diff
@@ -11,11 +11,9 @@ spec:
...
       application: skipper-ingress
-  updateStrategy:
-    type: RollingUpdate
   template:
     metadata:
-      name: skipper-ingress
+      name: skipper-ingress-a
       labels:
         application: skipper-ingress
         version: v0.10.121-1-g5f1806d
```
mikkeloscar: Maybe we should also have a `zone` label?

cluster/manifests/skipper/deployment-a.yaml
```diff
@@ -66,10 +65,10 @@ spec:
           - "-opentracing=lightstep component-name=skipper-ingress token=$(LIGHTSTEP_TOKEN) collector=tracing.zmon.zalan.do:8444 cmd-line=skipper-ingress max-buffered-spans=4096"
         resources:
           limits:
-            memory: {{ .ConfigItems.skipper_limits_mem }}
+            memory: {{ .ConfigItems.skipper_ingress_cpu }}
```
mikkeloscar: This should be `ConfigItems.skipper_ingress_memory` instead of `cpu`. (Same for `b` and `c`).

mikkeloscar (commented):
I wonder if we could simplify this by just making CLM output a slice of availability zones and loop through them in a single template file so we don't have to duplicate both the deployment and HPA. I could imagine that we might forget to change some of the flags if we have to always change it in three places.

cluster/manifests/skipper/deployment-a.yaml
szuecs: I don’t think it’s necessary, because we won’t run multi region clusters or do we have other limitations somewhere that needs it?

cluster/manifests/skipper/deployment-a.yaml
szuecs: To do what?

szuecs:
:+1:

cluster/config-defaults.yaml
```diff
@@ -96,6 +91,17 @@ horizontal_pod_autoscaler_sync_period: "30s"
 horizontal_pod_autoscaler_tolerance: "0.1"
 horizontal_pod_autoscaler_upscale_delay: "3m0s"
 
+# skipper ingress settings
+skipper_ingress_target_average_utilization_cpu: "100"
+skipper_ingress_target_average_utilization_memory: "80"
+skipper_ingress_max_replicas: "10"
```
mikkeloscar, line 100: This would be a max of `30`, maybe we want to use `5` as the max to limit it to `15` as we talked about, to prevent scaling up too much?

/cc @aermakov-zalando since you made a point about this in our meeting :)

szuecs: Our internal ticket, which we agreed on in the meeting with everyone interested was 10 per AZ.

mikkeloscar: Ok, I understood it wrong then, thanks!

szuecs:
:+1:

szuecs:
:+1:

mikkeloscar:
:+1:

szuecs:
stopped all ci pipelines to not burn money because of the deletion change.

PR #1462, opened 2018-10-11
Open a subset of ports on all workers for NLB support

linki:
In order to allow an NLB necessary for OpenTracing to work we need to open some ports on all workers (30443-30444).

see: https://github.bus.zalan.do/eagleeye/lightstep-collector-deploy/pull/21/files#diff-34b78ece998339dae031d09457f9bd2c

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -266,6 +266,11 @@ Resources:
           FromPort: 22
           IpProtocol: tcp
           ToPort: 22
+        # Allow NLBs to target these ports on all workers.
+        - CidrIp: 0.0.0.0/0
+          FromPort: 30443
+          IpProtocol: tcp
+          ToPort: 30444
         - CidrIp: "{{.Values.vpc_ipv4_cidr}}"
           FromPort: 9999
           IpProtocol: tcp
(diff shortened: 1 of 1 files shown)


cluster/cluster.yaml
```diff
@@ -266,6 +266,11 @@ Resources:
           FromPort: 22
           IpProtocol: tcp
           ToPort: 22
+        # Allow NLBs to target these ports on all workers.
```
mikkeloscar: Let's mention the service here as well, so it's clear.

mikkeloscar:
> (30443-30444).

Did you check that these ports are not used by any other services right now?

linki:
> Did you check that these ports are not used by any other services right now?

~~No~~ (Yes)

linki:
- [x] Checked that port isn't allocated in any cluster
  * checked Service `nodePort`s and Pod `hostPort`
  * currently only used in playground for our tests

jrake-revelant:
👍

mikkeloscar:
I think this depends on the white listing (I'm not sure about the status).

linki:
:+1:

jrake-revelant:
@mikkeloscar this was done by Hanno and followed up by me.

PR #1463, opened 2018-10-11
Add teapot admission controller config

aermakov-zalando:
Register the teapot admission controller for create ops on Pods and StorageClasses. We don't care about updates for now because the relevant parts are immutable and we don't care about metadata.

Only enable it by default in e2e for now.

### DIFF
--- a/cluster/manifests/admission-control/teapot.yaml
+++ b/cluster/manifests/admission-control/teapot.yaml
@@ -1,0 +1,29 @@
+{{ if eq .ConfigItems.teapot_webhook_enabled "true" }}
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: teapot-admission-controller
+  labels:
+    application: teapot-admission-controller
+webhooks:
+  - name: pod-admitter.teapot.zalan.do
+    clientConfig:
+      url: "https://localhost:8085/pod"
+      caBundle: "{{ .ConfigItems.ca_cert_decompressed }}"
+    failurePolicy: Fail
+    rules:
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
👍

mikkeloscar:
:+1:

cluster/manifests/admission-control/teapot.yaml
```diff
@@ -0,0 +1,29 @@
...
+metadata:
+  name: teapot-admission-controller
+  labels:
+    application: teapot-admission-controller
+webhooks:
+  - name: pod-admitter.teapot.zalan.do
+    clientConfig:
+      url: "https://localhost:8085/pod"
+      caBundle: "{{ .ConfigItems.ca_cert_decompressed }}"
+    failurePolicy: Fail
```
linki, line 13: Does this tell the API server what to do in case of this webhook being unavailable? Isn't it safer to allow Pod creation in those cases? Yes, they could circumentvent the (non security-relevant) restrictions but all pods get re-created on the next update anyways.

aermakov-zalando: I'd rather ensure that it's always available. If we start putting more helpful stuff into the webhook (e.g. TOKENINFO_URL injection), it's going to be weird for the users to suddenly get pods created without that. Plus the controller would retry pod creation again soon.

mikkeloscar: What do you mean by next update? The owning controller would retry in case the pod creation fails when the webhook is unavailable.

linki: Ok, I didn't think about the retries.

PR #1465, opened 2018-10-12
Change admission controller and enable it fully

aermakov-zalando:
If teapot_admission_controller_enabled is set to 'true', fully enable the admission controller:

 * don't create limits/default
 * enable pod and storageclass admitter
 * enable `--pod-process-resources`


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1472, opened 2018-10-19
Allow changing resources for metrics-service

mikkeloscar:
The metrics-service needs to scale in proportion to the cluster size especially for memory. Since we don't have VPA yet, make it possible to manually bump the resources if needed.


mikkeloscar:
:+1:

muaazsaleem:
👍

PR #1474, opened 2018-10-19
resolve conntrack container being killed.

dionb:
Conntrack uses more memory when it runs if there are more connections active on the machine. This was causing large machines to kill it due to it exceeding it's memory allowance. I have conditionally disabled it and increased the memory allocations.

### DIFF
--- a/cluster/manifests/prometheus-node-exporter/daemonset.yaml
+++ b/cluster/manifests/prometheus-node-exporter/daemonset.yaml
@@ -55,10 +55,10 @@ spec:
         resources:
           limits:
             cpu: 10m
-            memory: 40Mi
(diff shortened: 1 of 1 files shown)


dionb:
:+1:

dionb:
resolves https://github.bus.zalan.do/teapot/issues/issues/1416

mikkeloscar:
Since this provides very low level metrics that we haven't needed to look at since we introduced the actual DNAT fixes in flannel it makes sense as a quick fix to simply disable these metrics and enable them per cluster in case we need to investigate something in the future.

mikkeloscar:
:+1:

PR #1476, opened 2018-10-19
Only open lightstep ports if the option is enabled

mikkeloscar:
Put the open NLB ports for the lightstep collector behind a config item such that we don't have to enable it in every cluster where the collector might not be running.


mikkeloscar:
:+1:

linki:
👍

PR #1478, opened 2018-10-23
ExternalDNS: reduce batch size to 100 by default

linki:
To not run into the 32k character limit we could reduce the batch size to `100` by default without much impact (most changes should be handled in a single batch even with that limit). Using this feature also helps in making sure it works correctly.

### DIFF
--- a/cluster/manifests/external-dns/deployment.yaml
+++ b/cluster/manifests/external-dns/deployment.yaml
@@ -28,6 +28,7 @@ spec:
         - --source=service
         - --source=ingress
         - --provider=aws
+        - --aws-batch-change-size=100
         - --registry=txt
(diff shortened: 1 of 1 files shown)


linki:
Do you think we should do that?

mikkeloscar:
Is it correctly understood that the implications of this is that we for instance could end up adding only the `A` record but not the `TXT`? How would we know if something like this happens?

szuecs:
@linki can we merge this or do we need to work on this?

linki:
@mikkeloscar I think it's fine if we used an even number as the batch limit. Because the changes is an ordered list of A/TXT pairs. We should enforce this ExternalDNS.

However, I didn't have time to test this and since the batch size reduction isn't strictly needed here I'm closing this issue for now.

PR #1479, opened 2018-10-29
Enable zmon metrics HPA in production clusters

mikkeloscar:
This updates the kube-metrics-adapter and includes ZMON support as implemented in (https://github.com/zalando-incubator/kube-metrics-adapter/pull/2)

For now ZMON is only enabled in production clusters as it requires production credentials to talk to ZMON kariosdb. We need something like gerry, but I would like to avoid implementing token management in the kube-metrics-adapter.


arjunrn:
:+1:

mikkeloscar:
:+1:

PR #1483, opened 2018-10-30
system pods: don't overcommit on memory

aermakov-zalando:
Set explicit limits & don't overcommit. More to come.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1489, opened 2018-10-31
Fix heapster deployment

aermakov-zalando:
Removing `cpu.limit` from the deployment causes future updates to fail if the `nanny` container reduces the `limit` below the `request` in the manifest. Keep it for until we get rid of `nanny`.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1491, opened 2018-11-01
Automatically configure instance storage

aermakov-zalando:
If an instance has NVME instance storage, automatically format and mount the first disk. It's mounted to `/var/lib/docker` by default, but this can be overridden by changing `instance_storage_mount_path` node pool config item.

Depends on https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/96.

### DIFF
--- a/cluster/node-pools/worker-default/userdata.clc.yaml
+++ b/cluster/node-pools/worker-default/userdata.clc.yaml
@@ -20,10 +20,8 @@ systemd:
   - name: locksmithd.service
     mask: true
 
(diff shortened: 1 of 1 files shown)


mikkeloscar:
Cool! We have to remember to migrate existing node pools configured with the `instance_storage` config-item.

aermakov-zalando:
@mikkeloscar I've updated both.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1499, opened 2018-11-07
fix skipper-graceful-start

szuecs:
fix: skipper listens on start and might not have the routes pulled from kubernetes data source

### DIFF
--- a/cluster/manifests/skipper/daemonset.yaml
+++ b/cluster/manifests/skipper/daemonset.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: skipper-ingress
-    version: v0.10.110
+    version: v0.10.112
     component: ingress
 spec:
   selector:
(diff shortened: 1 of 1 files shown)


aryszka:
i think this also requires the flag -wait-first-route-load. https://github.com/zalando/skipper/pull/837/files#diff-9ff66a84471f234f548555972943787fR380

aermakov-zalando:
Is there a reason why this behaviour is not the default?

aermakov-zalando:
That is, I don't see a valid use case in just serving 404s, so why have a flag?

aryszka:
I didn't want to change the default behavior, at least this time, but I tend to agree that 404 cannot make sense in any case, especially as we serve the health check from the routes, too.

szuecs:
:+1:

szuecs:
Let’s merge, the flag removal  can be done later

gargravarr:
:+1:

PR #1505, opened 2018-11-08
 adjust default scaling for zmon-worker

vetinari:
please clear the registry when rolling out this change


mohabusama:
@vetinari hold on on that one, since I want to update the worker as well.

aermakov-zalando:
@vetinari Please sign the commits, DCO check is failing.

aermakov-zalando:
Also, we have a better way of configuring the defaults now, see [config-defaults.yaml](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/config-defaults.yaml). You can remove the conditionals from the deployment files and replace them with just `{{.ConfigItems.zmon_worker_replicas}}`. An additional benefit is that you can have different defaults for test and production clusters, because I'm not sure if massively increasing the memory makes sense for testing.

PR #1510, opened 2018-11-12
Adding memory MaxResourcePolicy to VPAs

muaazsaleem:
Adding memory MaxResourcePolicy to both prometheus and metrics-server
VPAs. This is so that the VPA doesn't scale prometheus and metrics-server
so much that they become unschedulable.

Note that if `promtheus_mem` is set than the vpa can't **scale up** more than
that, so this should be set to the maximum possible value in a cluster


cluster/manifests/metrics-server/metrics-server-vpa.yaml
```diff
@@ -10,4 +10,11 @@ spec:
       application: metrics-server
   updatePolicy:
     updateMode: "Auto"
+  {{if index .ConfigItems "prometheus_mem"}}
```
arjunrn: I think there is a default value for `prometheus_mem` so the if condition is not needed.

arjunrn (changes requested):

cluster/manifests/metrics-server/metrics-server-vpa.yaml
```diff
@@ -10,4 +10,11 @@ spec:
...
+  resourcePolicy:
+    containerPolicies:
+    - containerName: metrics-server
+      maxAllowed:
+        memory: {{.ConfigItems.prometheus_mem}}
```
njuettner: Wouldn't it be better to set it directly:

`memory: {{if index .ConfigItems "prometheus_mem"}}{{.ConfigItems.prometheus_mem}}{{else}}4000Mi{{end}}`

njuettner: same for **prometheus-vpa.yaml**

cluster/manifests/metrics-server/metrics-server-vpa.yaml
njuettner: unfortunately not, we just only set it in the prometheus sts if the config item doesn't exist.
https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/prometheus/statefulset.yaml#L39

cluster/manifests/metrics-server/metrics-server-vpa.yaml
muaazsaleem: Beacause the `MaxResourcePolicy` sets the ceiling for how much the VPA can scale a container **up**, if we setup a default value e.g `4000Mi` than the VPA would never scale beyond that.i.e

```
requests <.....vpa scale up range.....> MaxResourcePolicy
```
I'm just worried that this could hurt us in large clusters ( large node sizes and high no. of nodes). 

I guess the question is, _"what is more likely?"_ The case where the VPA scales prometheus up so much that it becomes unschedulable  or that the `MaxResourcePolicy` would be set too low and the VPA won't be able to scale up `prometheus` enough? 

What do you think?

arjunrn: Either way it won't be any different from the current behavior so there's no harm in doing in that way.

arjunrn (approved):

cluster/manifests/metrics-server/metrics-server-vpa.yaml
muaazsaleem: Sounds good. I'll add a default value

cluster/manifests/metrics-server/metrics-server-vpa.yaml
```diff
@@ -10,4 +10,9 @@ spec:
       application: metrics-server
   updatePolicy:
     updateMode: "Auto"
+  resourcePolicy:
+    containerPolicies:
+    - containerName: metrics-server
+      maxAllowed:
+        memory: {{if index .ConfigItems "prometheus_mem"}}{{.ConfigItems.prometheus_mem}}{{else}}4000Mi{{end}}
```
arjunrn: Is the default for the metrics-server also 4000Mi?

cluster/manifests/metrics-server/metrics-server-vpa.yaml
aermakov-zalando: I'd set the default in `config-defaults.yaml` and just use the config item in here (and set it to a much larger value, e.g. `14Gi`).

cluster/manifests/metrics-server/metrics-server-vpa.yaml
njuettner: it would be better to replace it with "{{.ConfigItems.metrics_service_mem}}" and then just using else whatever it's preferred.

njuettner:
👍

cluster/config-defaults.yaml
```diff
@@ -64,8 +64,14 @@ zmon_worker_cpu: "750m"
 zmon_worker_count: "16"
 {{end}}
 
+prometheus_cpu: "1000m"
+prometheus_mem: "4Gi"
+max_prometheus_mem: "14Gi"
```
arjunrn: If the vpa selects this value the pods would be unschedulable in test clusters.

aermakov-zalando:
👍

arjunrn:
:+1:

PR #1518, opened 2018-11-16
[WIP] Enable rbac

dionb:
Makes it possible to enable RBAC by using a modified version of the webhook which either does a `DENY` e.g. for namespaces which we don't allow or a Pass-through to RBAC.

RBAC roles are defined for some components in `kube-system` but most are still using the special `system` service account handled by the webhook. The idea is to slowly migrate all components to use RBAC.

See also the previous PR which this is based on #467


cluster/manifests/kube2iam/rbac.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: kube-state-metrics
```
mikkeloscar: This looks wrong. This file mentions both `kube-state-metrics` and `skipper-ingress` but is in the `kube2iam` folder.

cluster/manifests/metrics-server/rbac.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: skipper-ingress
```
mikkeloscar: This is wrong: skipper-ingress in the `metrics-server` rbac file.

cluster/manifests/skipper/daemonset.yaml
```diff
@@ -63,7 +64,7 @@ spec:
           - "-max-audit-body=0"
           - "-oauth2-tokeninfo-url={{ .ConfigItems.tokeninfo_url }}"
           - "-histogram-metric-buckets=.01,.025,.05,.075,.1,.2,.3,.4,.5,.75,1,2,3,4,5,7,10,15,20,30,60,120,300,600"
-          - "-opentracing=lightstep component-name=skipper-ingress token=$(LIGHTSTEP_TOKEN) collector=tracing.zmon.zalan.do:8444 cmd-line=skipper-ingress max-buffered-spans=4096"
+          # - "-opentracing=lightstep component-name=skipper-ingress token=$(LIGHTSTEP_TOKEN) collector=tracing.zmon.zalan.do:8444 cmd-line=skipper-ingress max-buffered-spans=4096"
```
mikkeloscar: Note: this should not be merged :)

dionb: That's why it's tagged 'do-not-merge'. But this is required for testing, and you wanted a PR with the current state of everything so other people can look at it.

jrake-revelant: @dionb please bring this PR into a state where it could be merged so it can be review by your peers. Otherwise it is very hard for us to give fast feedback and ensure short cycles.

jrake-revelant: A good practice is also to label it/ call it WIP. For me this is especially important when I go into planning with the other teams.

mikkeloscar:
Closing in favor of #1550

PR #1519, opened 2018-11-16
AuditTrail S3 fallback

aermakov-zalando:
* Create an S3 bucket for AuditTrail fallback
 * Create an IAM role for Emergency Access Service

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -1084,6 +1084,21 @@ Resources:
           PolicyName: root
       RoleName: "{{.Cluster.LocalID}}-kube-metrics-adapter"
     Type: 'AWS::IAM::Role'
+  AuditTrailBucket:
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
👍

mikkeloscar:
Is the intention to use the same bucket for audittrail adapter at some point, or should it be named something with emergency-access-service?

We only need this enabled in production clusters because nothing that uses audittrail is running in test clusters

aermakov-zalando:
I'd use one bucket for all applications. We could also provide a fallback sidecar for the users' applications that'll use this bucket automagically.

It's easier to create it like this (and then use for e.g. staging deployments), and the buckets don't really cost anything, but I can also template it if you think it's worth doing.

mikkeloscar:
:+1:

PR #1521, opened 2018-11-16
setup east west routes

szuecs:
TODO:
- [x] SET skipper-ingress serviceip to 10.3.99.99
- [x] check if it is allocated


cluster/manifests/skipper/daemonset.yaml
```diff
@@ -51,6 +51,7 @@ spec:
           - "-kubernetes-path-mode=path-prefix"
           - "-address=:9999"
           - "-wait-first-route-load"
+          - "-enable-kubernetes-east-west"
```
aermakov-zalando, line 60: Is it too late at this point to ask for a more intuitive name for this?

cluster/manifests/skipper/service.yaml
```diff
@@ -11,6 +11,7 @@ metadata:
     prometheus.io/scrape: "true"
 spec:
   type: ClusterIP
+  clusterIP: 10.3.99.99
```
aermakov-zalando: You can't change `clusterIP` of a service.

cluster/manifests/coredns/configmap.yaml
```diff
@@ -15,6 +15,11 @@ data:
             upstream
             fallthrough in-addr.arpa ip6.arpa
         }
+        template IN A skipper.cluster.local  {
+            match "^.*[.]skipper[.]cluster[.]local"
+            answer "{{ .Name }} 60 IN A 10.3.99.99"
```
aermakov-zalando: This needs to be escaped.

cluster/manifests/skipper/daemonset.yaml
szuecs: yes, because a change would be a breaking change

cluster/manifests/skipper/service.yaml
szuecs: good point

cluster/manifests/coredns/configmap.yaml
szuecs: How can I do this?
Can I do this:

```
{{{  "answer \"{{ .Name }} 60 IN A 10.3.99.99\"" }}}
```

PR #1522, opened 2018-11-19
Apply VPA manifests before anything else

mikkeloscar:
Apply VPA manifests before anything else. This ensures that a VPA defintion e.g. for metrics-server is not applied before the CRD for VPA has been applied.


muaazsaleem:
Makes sense

muaazsaleem:
👍

mikkeloscar:
:+1:

PR #1524, opened 2018-11-19
Allow cluster scale-down to be switched on and off

linki:
Add a config item to turn cluster downscaling on and off. Defaults to the current behaviour: scale down enabled.


linki:
:+1:

gargravarr:
:+1:

PR #1529, opened 2018-11-20
Allow adjusting redis memory via config items

njuettner:
New config item to set zmon redis memory


njuettner:
@mohabusama @vetinari rolling out new cluster updates currently overwrites memory to the default setting. adjusting the memory setting via config item for specific clusters will prevent this.

mohabusama:
👍

mohabusama (approved):

aermakov-zalando:
👍

PR #1530, opened 2018-11-21
fix one cause of 499s

szuecs:
fix context issue causing some 499s with probability around 0.3% under high load

### DIFF
--- a/cluster/manifests/skipper/daemonset.yaml
+++ b/cluster/manifests/skipper/daemonset.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: skipper-ingress
-    version: v0.10.112
+    version: v0.10.121-1-g5f1806d
     component: ingress
 spec:
   selector:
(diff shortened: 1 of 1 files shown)


cluster/manifests/skipper/daemonset.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: skipper-ingress
-    version: v0.10.112
+    version: v0.10.121-1-g5f1806d
```
aermakov-zalando, line 8: Should we cherry-pick it to `.112`, considering that `.121` wasn't tested in the clusters, or is it fine? What other changes are there?

szuecs: Only filters added which we do not enable by default and one fix for kubernetes in GCP. 
I think it's fine.

szuecs:
:+1:

linki:
:+1:

PR #1550, opened 2018-11-29
Rbac squash

dionb:
A squashed commit of all work to enable rbac.
Includes creation of roles, bindings, service-accounts, and config updates

Makes it possible to enable RBAC by using a modified version of the webhook which either does a `DENY` e.g. for namespaces which we don't allow or a Pass-through to RBAC.

RBAC roles are defined for some components in `kube-system` but most are still using the special `system` service account handled by the webhook. The idea is to slowly migrate all components to use RBAC.

See also the previous PR which this is based on #467 

Replaces #1518


mikkeloscar:
:+1:

dionb:
:+1:

PR #1552, opened 2018-11-29
Reset previously removed UIDs which are still in use.

mikkeloscar:
* Resets previously removed UIDs for CLM which are still in use. (Was removed in #1550)
* Use the correct name for `kube-controller-manager` depending on RBAC is enabled or not.


mikkeloscar:
I'm manually testing this in `stups-test` to confirm it solves the problem. Will not wait for e2e as this should fix all `dev` clusters which are currently in a somewhat broken state.

njuettner:
👍

mikkeloscar:
:+1:

mikkeloscar:
Problem was fixed.

PR #1554, opened 2018-11-29
Fix the templating in skipper deployment

mikkeloscar:
Need to use the `$cfg` variable because of how the template scoping works.


mikkeloscar:
:+1:

dionb:
:+1:

PR #1555, opened 2018-11-29
Disable the daemonset when it's not needed anymore

aermakov-zalando:
If a cluster has more nodes than deployment pods, we will still continue running daemonset pods. Disable them automagically once we've rolled out the deployment to all 3 AZs.

### DIFF
--- a/cluster/manifests/skipper/daemonset.yaml
+++ b/cluster/manifests/skipper/daemonset.yaml
@@ -33,6 +33,9 @@ spec:
         operator: Exists
       nodeSelector:
         kubernetes.io/role: worker
+{{ if eq .ConfigItems.zone_suffixes "a,b,c" }}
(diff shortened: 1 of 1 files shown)


cluster/manifests/skipper/daemonset.yaml
```diff
@@ -33,6 +33,9 @@ spec:
         operator: Exists
       nodeSelector:
         kubernetes.io/role: worker
+{{ if eq .ConfigItems.zone_suffixes "a,b,c" }}
+        dummy.zalan.do: no-nodes
```
szuecs: maybe better `dummy.zalando.org` ?

aermakov-zalando: Fixed.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1556, opened 2018-11-29
Add RBAC manifests for ExternalDNS

linki:
Add RBAC manifests for ExternalDNS.

### DIFF
--- a/cluster/manifests/external-dns/rbac.yaml
+++ b/cluster/manifests/external-dns/rbac.yaml
@@ -13,7 +13,6 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: external-dns
-  namespace: kube-system
   labels:
     application: external-dns
 rules:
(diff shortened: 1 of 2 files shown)


linki:
Note, this also wraps the RBAC rules in the `ConfigItem` check to make it consistent but different than the other places so far. Should be fine though.

cluster/manifests/external-dns/rbac.yaml
```diff
@@ -0,0 +1,42 @@
...
+  namespace: kube-system
+  labels:
+    application: external-dns
+---
+# allows to list services and ingresses
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: external-dns
+  namespace: kube-system
```
mikkeloscar: `ClusterRoles` or `ClusterRoleBindings` are not namespaced. I don't know if this makes it break or it just ignores it.

linki: That doesn't make sense 😆 Fixed.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #1562, opened 2018-11-30
ExternalDNS: Add Liveness probes, VPA, Guaranteed QoS

linki:
Cleanup the ExternalDNS Deployment to get familiar with the latest features.

### DIFF
--- a/cluster/manifests/external-dns/deployment.yaml
+++ b/cluster/manifests/external-dns/deployment.yaml
@@ -37,12 +37,13 @@ spec:
           limits:
             cpu: 50m
             memory: 100Mi
-          requests:
-            cpu: 50m
-            memory: 100Mi
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 7979
         securityContext:
           readOnlyRootFilesystem: true
           runAsNonRoot: true
-          runAsUser: 1000
+          runAsUser: 65534
           capabilities:
             drop: ["ALL"]
--- a/cluster/manifests/external-dns/vpa.yaml
+++ b/cluster/manifests/external-dns/vpa.yaml
@@ -1,0 +1,19 @@
+apiVersion: autoscaling.k8s.io/v1beta1
+kind: VerticalPodAutoscaler
+metadata:
+  name: external-dns
+  namespace: kube-system
+  labels:
+    application: external-dns
+spec:
+  selector:
+    matchLabels:
+      application: external-dns
+  updatePolicy:
+    updateMode: Auto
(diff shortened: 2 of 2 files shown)


cluster/manifests/external-dns/deployment.yaml
```diff
@@ -35,13 +35,15 @@ spec:
         - --txt-owner-id={{ .Region }}:{{ .LocalID }}
         resources:
           limits:
-            memory: 100Mi
-          requests:
             cpu: 25m
             memory: 100Mi
+        livenessProbe:
```
szuecs, line 40: any reason why a livenessProbe is used?
I think it needs some more reasoning, not to drop it, but why we use it.

linki: Yes, we should align on when to use what. I see `livenessProbe` as "damn, something is  wrong" and `readinessProbe` as "I'm fine but my caches aren't warm so please let me alone".

Since `/healthz` of ExternalDNS is [dead simple](https://github.com/kubernetes-incubator/external-dns/blob/master/main.go#L254) something clearly wrong must be going on.

linki: @szuecs does it make sense?

szuecs: fine for me

cluster/manifests/external-dns/vpa.yaml
```diff
@@ -0,0 +1,21 @@
+{{ if eq .ConfigItems.vpa_enabled "true" }}
```
szuecs: Please drop this `if`, we switched everything to vpa_enabled "yes"

cluster/manifests/external-dns/deployment.yaml
```diff
@@ -35,14 +35,18 @@ spec:
         - --txt-owner-id={{ .Region }}:{{ .LocalID }}
         resources:
           limits:
-            cpu: 50m
+            cpu: 25m
```
linki: This isn't intended, got changed in the meantime: https://github.com/zalando-incubator/kubernetes-on-aws/commit/a65eb7611610ff1ec344e48e0963a0786b9cdc70#diff-51acfae938d9fe0cc68f5a4ca83a59ea

linki:
Anyone more comments?
* The health check should not harm
* The VPA has not much effect for memory b/c the `recommender` recommends at least `250Mi` for everything.

arjunrn:
:+1:

mikkeloscar:
:+1:

PR #1565, opened 2018-11-30
enable east-west setup

szuecs:
enable east west setup to allow service to service communication through skipper.
Enabled in skipper deployment + internal service IP and coredns daemonset.

### DIFF
--- a/cluster/manifests/coredns-local/configmap.yaml
+++ b/cluster/manifests/coredns-local/configmap.yaml
@@ -15,7 +15,11 @@ data:
             upstream
             fallthrough in-addr.arpa ip6.arpa
         }
-        prometheus :9153
+        template IN A skipper.cluster.local  {
+            match "^.*[.]skipper[.]cluster[.]local"
+            answer "{{"{{"}} .Name {{"}}"}} 60 IN A 10.3.99.99"
+            fallthrough
+        }        prometheus :9153
         proxy . /etc/resolv.conf
         cache 30
         reload
--- a/cluster/manifests/coredns-local/configmap.yaml
+++ b/cluster/manifests/coredns-local/configmap.yaml
@@ -15,8 +15,8 @@ data:
             upstream
             fallthrough in-addr.arpa ip6.arpa
         }
-        template IN A skipper.cluster.local  {
-            match "^.*[.]skipper[.]cluster[.]local"
+        template IN A internal.cluster.local  {
+            match "^.*[.]internal[.]cluster[.]local"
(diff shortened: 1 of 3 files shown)


cluster/manifests/coredns/configmap.yaml
```diff
@@ -15,7 +15,11 @@ data:
             upstream
             fallthrough in-addr.arpa ip6.arpa
         }
-        prometheus :9153
+        template IN A skipper.cluster.local  {
+            match "^.*[.]skipper[.]cluster[.]local"
+            answer "{{"{{"}} .Name {{"}}"}} 60 IN A 10.3.99.99"
+            fallthrough
+        }        prometheus :9153
```
mikkeloscar: nit: newline missing after `}`. I don't know if it matters?

cluster/manifests/coredns/configmap.yaml
```diff
@@ -15,6 +15,11 @@ data:
             upstream
             fallthrough in-addr.arpa ip6.arpa
         }
+        template IN A skipper.cluster.local  {
+            match "^.*[.]skipper[.]cluster[.]local"
```
hjacobs: Should we find a better name than Skipper here? I would like to prevent "leaking" the implementation detail that Skipper is our Ingress proxy to users. I think it would be better to find a meaningful abstract DNS name e.g. `ingress.cluster.local` (or whatever naming idea you would have).

szuecs: You could also consider the „leak“ good, because then they find the right documentation. :)

hjacobs: Having Skipper in the DNS name would expose this to all application configurations (which are using service-to-service in the cluster), please don't do this --- it's not about "branding" or documentation here, but about having meaningful endpoints across potentially hundreds of applications.

szuecs:
Then let’s discuss a better name on Monday.
Maybe `.internal.cluster.local`?
Ingress is already something else in my opinion it is the north-south traffic, that I don’t want to mix.

szuecs:
https://github.com/zalando/skipper/pull/869 was opened to make it possible to set the internal domain to another value.

szuecs:
`.internal.cluster.local` domain suffix was voted by almost 60% of tech employees, that took part in the poll.

szuecs:
One out of 3 skipper-ingress pods was not getting healthy and there are some other errors reported in the logs in the cluster creation process.

szuecs:
rebased on top of dev to get the rbac fix

szuecs:
:+1:

gargravarr:
:+1:

PR #1566, opened 2018-11-30
enable apiMonitoring

### LINKED ISSUE #1516: Activate API usage monitoring in Skipper
To allow usage of the new API monitoring filter we suggest to activate the filter by default.

szuecs:
fix  #1516
enable apiMonitoring via feature toggle enable_apimonitoring, default is not to enable it for production, but enable it for test.

Usage of apiMonitoring filter can lead to memory consumption in skipper. We should close contact to customers, that want to use this feature and monitor the memory consumption more closely.   

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -27,6 +27,12 @@ skipper_ingress_memory: "500Mi"
 # AZ
 zone_suffixes: "a,b,c"
 
+{{if eq .Environment "production"}}
+enable_apimonitoring: "false"
+{{else}}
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

mikkeloscar:
@szuecs could you add to the PR description what the expected impact is by enabling this by default for test clusters? I understand that we need to run it to understand the real impact but will it only have an effect if someone opts-in or will it do something by default as long as the flag is enabled?

mikkeloscar:
:+1:

szuecs:
@mikkeloscar done

PR #1567, opened 2018-11-30
enable clusterRatelimits

szuecs:
add swim protocol based data exchange to enable clusterRatelimits


szuecs:
:+1:

cluster/manifests/skipper/deployment.yaml
```diff
@@ -69,6 +69,13 @@ spec:
           - "-metrics-flavour=codahale,prometheus"
           - "-enable-connection-metrics"
           - "-max-audit-body=0"
+          - "-enable-swarm"
+          - "-swarm-port=9990"
+          - "-swarm-label-selector-key=application"
+          - "-swarm-label-selector-value=skipper-ingress"
+          - "-swarm-leave-timeout=5s"
+          - "-swarm-max-msg-buffer=4194304"
+          - "-swarm-namespace=kube-system"
```
mikkeloscar, line 79: These flags should be wrapped in `{{ if index .ConfigItems "skipper_clusterratelimit" }}`

cluster/cluster.yaml
```diff
@@ -277,6 +277,14 @@ Resources:
...
           IpProtocol: tcp
           ToPort: 9999
+        - CidrIp: "{{.Values.vpc_ipv4_cidr}}"
+          FromPort: 9990
+          IpProtocol: tcp
+          ToPort: 9990
+        - CidrIp: "{{.Values.vpc_ipv4_cidr}}"
+          FromPort: 9990
+          IpProtocol: udp
+          ToPort: 9990
```
mikkeloscar, line 288: This should be conditionally enabled only when clusterratelimit feature is enabled.

szuecs:
@mikkeloscar changed as commented

szuecs:
:+1:

cluster/config-defaults.yaml
```diff
@@ -23,6 +23,10 @@ skipper_ingress_max_replicas: "10"
 skipper_ingress_min_replicas: "1"
 skipper_ingress_cpu: "500m"
 skipper_ingress_memory: "500Mi"
+{{if eq .Environment "production"}}
```
aermakov-zalando, line 26: Please add a default for both environments (set to `false` in production) and then check for `if eq ... "true"`. It's very confusing otherwise, because setting it to `false` wouldn't actually disable anything (and in fact it's impossible to disable in test clusters).

cluster/cluster.yaml
```diff
@@ -277,6 +277,16 @@ Resources:
           FromPort: 9999
           IpProtocol: tcp
           ToPort: 9999
+{{- if index .Cluster.ConfigItems "skipper_clusterratelimit" }}
```
aermakov-zalando: See the comment below about `if eq` vs `if index`.

cluster/config-defaults.yaml
szuecs: ok

szuecs:
@aermakov-zalando fixed as you commented

szuecs:
:+1:

mikkeloscar:
:+1:

PR #1577, opened 2018-12-03
Update to Kubernetes v1.11.5

mikkeloscar:
Update to Kubernetes v1.11.5

PR #1580, opened 2018-12-04
Enable TaintNodesByCondition feature flag, update to v1.11.5

mikkeloscar:
* Enable s`TaintNodesByCondition` feature flag which is needed for when we enable daemonset scheduling with the default scheduler. This must be rolled out in two separate steps.

* Updates nodes to v1.11.5 (This was left out in #1577)

Also removes the `ExperimentalCriticalPodAnnotation` feature flag which we no longer need. We use priorities instead: https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/

### DIFF
--- a/cluster/node-pools/master-default/userdata.clc.yaml
+++ b/cluster/node-pools/master-default/userdata.clc.yaml
@@ -173,7 +173,7 @@ systemd:
       --tls-cert-file=/etc/kubernetes/ssl/worker.pem \
       --tls-private-key-file=/etc/kubernetes/ssl/worker-key.pem \
       --cloud-provider=aws \
-      --feature-gates=ExperimentalCriticalPodAnnotation=true,TaintBasedEvictions=true \
+      --feature-gates=TaintBasedEvictions=true,TaintNodesByCondition=true \
       --pod-infra-container-image=registry.opensource.zalan.do/teapot/pause-amd64:3.1 \
 {{- if not (index .Cluster.ConfigItems "enable_cfs_quota") }}
       --cpu-cfs-quota=false \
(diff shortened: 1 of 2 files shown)


szuecs:
@mikkeloscar is it a wip or not?

mikkeloscar:
@szuecs Sorry for not setting the `do-not-merge` label. I have now updated the description a bit and it's ready to merge now from my PoV.

mikkeloscar:
:+1:

mikkeloscar:
This needs to be tested with the autoscaler before it's ready to merge.

linki:
With this change nodes get tainted with more stuff than before, and that stuff is probably not tolerated by our components. Is that safe?

mikkeloscar:
> With this change nodes get tainted with more stuff than before, and that stuff is probably not tolerated by our components. Is that safe?

We have `NoSchedule` tolerations for all our daemonsets and this will only add `NoSchedule` taints, so it won't kick stuff out of the node, only prevent new pods from scheduling.

The `TaintBasedEviction` which we already have enabled is doing the evictions with `NoExecute` taints.

mikkeloscar:
Discussed with @aermakov-zalando the test cases and we concluded that given this works with the e2e tests it should be fine to roll out to `dev`.

mikkeloscar:
:+1:

mikkeloscar:
I tested this in the following way:

* Spin up a new cluster.
* fill the node on one node and see that it gets the `NoSchedule` taint set.
* schedule pods which result in node scaleup.
* Observe that the autoscaler scales without problems.

So for me this is ready to go into `dev`.

aermakov-zalando:
👍

PR #1573, opened 2018-12-03
Switch CoreDNS Deployment to (node-local) DaemonSet

linki:
Replaces the node-local DNS setup via `dnsmasq` with dedicated and independent CoreDNS instances per node.

Client-side DNS configuration is the same:
* First DNS server is the local node's IP which is backed by a single CoreDNS Pod on the same node
* Second DNS server is a `ServiceIP` which is backed by a set of CoreDNS Pods across the cluster
  * Note, that in this PR we're using the very same DaemonSet's Pods as targets (this could be changed to a dedicated fallback CoreDNS as a Deployment as before)

PR #1581, opened 2018-12-04
Enable CoreDNS's autopath feature

linki:
Enable CoreDNS's [autopath](https://coredns.io/plugins/autopath/) feature for Kubernetes.

So far I didn't find any good indication whether one should also change the `resolv.conf`.

With this setup a lookup to an external name like `google.com` would go first to `google.com.kube-system.svc.cluster.local` which `autopath`-enabled CoreDNS is able to answer with a CNAME to `google.com` instead of an NXDOMAIN, thus not needing the client to do any further requests for the other search domains.

Using `pods verified` is required when using `autopath`.

Interesting read about Kubernetes' DNS issues: https://blog.quentin-machu.fr/2018/06/24/5-15s-dns-lookups-on-kubernetes/

### DIFF
--- a/cluster/manifests/coredns/configmap.yaml
+++ b/cluster/manifests/coredns/configmap.yaml
@@ -11,7 +11,7 @@ data:
         errors
         health
         kubernetes cluster.local in-addr.arpa ip6.arpa {
-        resyncperiod 1m
+            resyncperiod 1m
 {{ if eq .ConfigItems.coredns_autopath "true" }}
             pods verified
 {{ else }}
--- a/cluster/manifests/coredns/configmap.yaml
+++ b/cluster/manifests/coredns/configmap.yaml
@@ -11,10 +11,11 @@ data:
         errors
         health
         kubernetes cluster.local in-addr.arpa ip6.arpa {
-            pods insecure
+            pods verified
             upstream
             fallthrough in-addr.arpa ip6.arpa
         }
+        autopath @kubernetes
         prometheus :9153
         proxy . /etc/resolv.conf
         cache 30
--- a/cluster/manifests/coredns/configmap.yaml
+++ b/cluster/manifests/coredns/configmap.yaml
@@ -11,11 +11,17 @@ data:
         errors
         health
         kubernetes cluster.local in-addr.arpa ip6.arpa {
+{{ if eq .ConfigItems.coredns_autopath "true" }}
             pods verified
+{{ else }}
+            pods insecure
+{{ end }}
             upstream
             fallthrough in-addr.arpa ip6.arpa
         }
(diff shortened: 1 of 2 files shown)


szuecs:
This should be done for coredns as daemonset, not?

njuettner:
Probably we have to adjust the memory.
`This option requires substantially more memory than in insecure mode, since it will maintain a watch on all pods.`

njuettner:
What about having a cluster default config item for `coredns_cpu` and `coredns_mem`?

And change the current deployment to a daemonset like @szuecs suggested. Additionally to that, fixing the QoS from `Burstable` to  `Guaranteed` would make sense to me (`CPU`limit is missing).

/cc @linki @szuecs

linki:
> This should be done for coredns as daemonset, not?

That doesn't matter. It applies to both setups.

linki:
@njuettner Makes sense to me. The DaemonSet/Deployment change is in [another PR](https://github.com/zalando-incubator/kubernetes-on-aws/pull/1573) as it's not directly related. Although, we can think about not using `autopath` with the DaemonSet setup, but I don't see a reason to if autopath works well.

njuettner:
👍

jrake-revelant:
👍

cluster/manifests/coredns/configmap.yaml
```diff
@@ -11,10 +11,18 @@ data:
         errors
         health
         kubernetes cluster.local in-addr.arpa ip6.arpa {
+            resyncperiod 1m
```
linki: @njuettner Should we do this in a separate PR or move into the `if`? This makes this PR essentially contain two changes (1 guarded by a config item and this one). The default sync period is `5m`, right?

njuettner: Correct, the default would be 5 minutes. I'll add an `if`.

linki: Ok. I'm always unsure about this stuff. Just want to mention that's also possible to use a config item and set its default value to `5m` to avoid the `if` or to leave it out entirely for now.

mikkeloscar:
I think for this feature it would be very helpful to do a load test which shows the impact, also in terms of extra resources required on the coredns side.

e.g. similar to what we did for skipper as a deployment: #1457

szuecs:
Please see @mikkeloscar comment, I set this to do-not-merge

linki:
Please note that this feature is guarded by a config item and wouldn't change anything when rolled out.

linki:
We're trying CoreDNS as DaemonSet first. Consider enabling `autopath` if it's not enough.

PR #1543, opened 2018-11-28
Adding manifests for kubernetes-lifecycle-metrics

muaazsaleem:
Deploying a small application called `kubernetes-lifecycle-metrics` in
clusters. It tracks, logs & exposes metrics that can be used to measure
the clusters' SLOs. Currently exposed metrics include:
 - time to schedule pods
 - time to mount the first volume in pods
 - image pull times per container
 - no. of pods waiting to be scheduled and the age of the oldest pod

Signed-off-by: Muhammad Muaaz Saleem <muhammad.muaaz.saleem@zalando.de>

PR #1582, opened 2018-12-05
dev-to-alpha

szuecs:
* **Adding manifests for kubernetes-lifecycle-metrics**
   <sup>Merge pull request #1543 from zalando-incubator/kube-lifecycle-metrics</sup>
* **enable clusterRatelimits**
   <sup>Merge pull request #1567 from zalando-incubator/enable/clusterRatelimit</sup>


cluster/manifests/kubernetes-lifecycle-metrics/deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+      application: kubernetes-lifecycle-metrics
+  template:
+    metadata:
+      labels:
+        application: kubernetes-lifecycle-metrics
+        version: master-1
+      annotations:
+        kubernetes-log-watcher/scalyr-parser: '[{"container": "kubernetes-lifecycle-metrics", "parser": "json"}]'
+    spec:
+      priorityClassName: system-cluster-critical
```
szuecs, line 22: Do we really need this for all components that only provide metrics?
IMHO that should not be done, but more for the important things like metric server that cause an outage of HPA or skipper that cause an outage of http routing.

aermakov-zalando: Why? I don't see why the users' pods should be able to preempt infrastructure stuff.

szuecs: If we have only limited CPU/memory and we have to select the important and the unimportant workloads, then I would say skipper-ingress or Prometheus is more important than this. I think we should use different priorities for different workloads, not treat all the same, does it makes sense?

aermakov-zalando: It does, but I don't think sure it's *that* important. We always have the users' pods to sacrifice, and it won't kick out anything important, so I see no harm in leaving it as is.

szuecs:
:+1:

cluster/manifests/kubernetes-lifecycle-metrics/deployment.yaml
linki: @szuecs @aermakov-zalando good to go?

aermakov-zalando:
👍

PR #1584, opened 2018-12-05
Preparation for migration to CoreDNS DaemonSet

linki:
This prepares the current CoreDNS Deployment to be able to co-exist with both a CoreDNS DaemonSet as well as the dnsmasq DaemonSet.

Main point here is to add `instance: cluster-dns` to the `PodSpec` to differentiate with the upcoming DaemonSet pods. Furthermore, the `coredns` config is already updated to its final version (health check port change due to `hostPort`). This allows us to share the same config for both DS and Deployment and saves a migration step later.

@aermakov-zalando corresponds to
> 1) add instance=cluster-dns to pod template, merge

/cc @njuettner @aermakov-zalando @szuecs

### DIFF
--- a/cluster/manifests/coredns/configmap.yaml
+++ b/cluster/manifests/coredns/configmap.yaml
@@ -9,7 +9,7 @@ data:
   Corefile: |
     .:53 {
         errors
-        health
+        health :9154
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
👍

njuettner:
👍

PR #1593, opened 2018-12-06
Skipper: move to 1 deployment

aermakov-zalando:
Use 1 deployment to avoid downtime when updating.

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -1,11 +1,5 @@
-# example: remove the app-ingress-controller deployment from kube-system
-pre_apply:
-- name: skipper-ingress
-  namespace: kube-system
-  kind: daemonset
-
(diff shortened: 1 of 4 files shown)


cluster/manifests/deletions.yaml
```diff
@@ -1,11 +1,23 @@
...
-  namespace: kube-system
-  kind: daemonset
-
 # everything defined under here will be deleted before applying the manifests
-#pre_apply: []
+pre_apply: []
 
 # everything defined under here will be deleted after applying the manifests
-post_apply: []
+post_apply:
```
mikkeloscar: Probably this should be done in a separate step to be sure the new deployment has enough time to deploy the pods and get them in the ALB?

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1597, opened 2018-12-07
Update admission-controller-deployment.yaml

arjunrn:
Increase the memory given to the VPA recommend to help with the frequent OOM kills. Also specify a CPU limit so that the pods are in the Guaranteed class.

### DIFF
--- a/cluster/manifests/01-vertical-pod-autoscaler/admission-controller-deployment.yaml
+++ b/cluster/manifests/01-vertical-pod-autoscaler/admission-controller-deployment.yaml
@@ -44,10 +44,11 @@ spec:
             readOnly: true
         resources:
           limits:
-            memory: 500Mi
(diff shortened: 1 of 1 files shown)


mikkeloscar:
The admission controller seem to be using very little memory (`stups-test`):

image


I guess you wanted to increase the memory of the `vpa-recommender` but this I would not do because it's leaking memory (we have an internal issue for this):

image

Getting OOM killed once a day is not really a big problem as it can be unavailable for a bit without any impact.

arjunrn:
I messed up. I meant to increase the limits for the recommender. I know that there's a memory leak but @muaazsaleem mentioned that the limits were set to lower than that of whats in the deployment in the original repo and I wanted to check if that would help. Anyway, I guess I can do that without making a change here.

PR #1603, opened 2018-12-10
Remove config item that defined which cluster DNS to use

linki:
Removes the ability to define which cluster-local DNS server is used via a ConfigItem and switches the default to CoreDNS. It rolls all nodes and activates CoreDNS-based cluster-local DNS server on a per-node basis.

This is the state of userdata before we started the rollout of the CoreDNS DaemonSet.

- [ ] do not merge before https://github.com/zalando-incubator/kubernetes-on-aws/pull/1573 is merged to beta
- [ ] do not merge before all clusters switched ClusterDNS to CoreDNS

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -152,5 +152,3 @@ etcd_instance_count: "3"
 {{end}}
 
 dynamodb_service_link_enabled: "false"
-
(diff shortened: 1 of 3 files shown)


mikkeloscar:
Closing for now.

PR #1606, opened 2018-12-11
Disable socket activation for sshd

aermakov-zalando:
According to instructions from [here](https://coreos.com/os/docs/latest/customizing-sshd.html#disabling-socket-activation-for-sshd).


aermakov-zalando:
👍

linki:
:+1:

PR #1604, opened 2018-12-10
dev-to-alpha

linki:
* **Upgrade zmon-worker**
   <sup>Merge pull request #1600 from zalando-incubator/upgrade-zmon-worker</sup>
* **Temporarily switch update policy to reduce dns outages**
   <sup>Merge pull request #1602 from zalando-incubator/dnsmasq-on-update</sup>
* **Skipper: minor fixes**
   <sup>Merge pull request #1599 from zalando-incubator/skipper-minor-fixes</sup>
* **Switch CoreDNS Deployment to (node-local) DaemonSet**
   <sup>Merge pull request #1573 from zalando-incubator/coredns-ds</sup>
* **Add missing RBAC permissions for kube-metrics-adapter**
   <sup>Merge pull request #1598 from zalando-incubator/custom-metrics-rbac</sup>
* **Update kube-downscaler to 0.6 (stackset support)**
   <sup>Merge pull request #1591 from zalando-incubator/downscaler-0.6</sup>
* **Skipper: fix templating**
   <sup>Merge pull request #1594 from zalando-incubator/fix-skipper-2</sup>
* **Skipper: move to 1 deployment**
   <sup>Merge pull request #1593 from zalando-incubator/revert-skipper-one-deployment</sup>
* **Enable TaintNodesByCondition feature flag, update to v1.11.5**
   <sup>Merge pull request #1580 from zalando-incubator/worker-nodes-v1.11.5</sup>
* **Enable VPA by default**
   <sup>Merge pull request #1583 from zalando-incubator/feature/enable-vpa-always</sup>
* **Update ZMON agents**
   <sup>Merge pull request #1586 from zalando-incubator/upgrade-agent</sup>

PR #1612, opened 2018-12-11
Allow prometheus to scrape local coredns pods

linki:
The Service is required to allow Prometheus to find the CoreDNS DaemonSet Pods.

The firewall needs to be opened because the Pod IPs are node IPs and prometheus needs to access them from another node.


szuecs:
:+1:

mikkeloscar:
See my comment here: https://github.com/zalando-incubator/kubernetes-on-aws/pull/1604/files#r240932080 do we actually need the extra service for this or could we re-use the existing coredns service?

szuecs:
@mikkeloscar good question, but in any case we need to create a new dashboard and add/change the selector in zmon checks/alerts. After switch we can cleanup the checks and dashboards.
Do you think it is a blocker to roll out ?

mikkeloscar:
I don't think it's a blocker, it just seems like extra work to roll out an extra service and then remove it at a later point unless there is a technical reason for doing it.

njuettner:
👍

linki:
:+1:

PR #1619, opened 2018-12-13
LaunchTemplates for worker node pools

njuettner:
Master nodes have been using LaunchTemplates for quite some time and we didn't have any issues. We should switch the worker stacks as well and then drop launch configuration related code from CLM.

Benefits:

* LaunchConfigurations are sort of deprecated
* Native support for "spot-at-on-demand-price" in LaunchTemplates, so we can drop the pricing-related code from CLM
* Slightly easier to work with


gargravarr:
:+1:

njuettner:
👍

PR #1624, opened 2018-12-14
Separate LaunchTemplates by using node pool name

njuettner:
LaunchTemplate names have to be unique


njuettner:
There was a problem which I saw when we're using GPU node pools. In general this pops up if we're using the `worker-default` profile 😨. My assumption was we didn't use it anymore.
`worker-splitaz`is not affected.

njuettner:
👍

gargravarr:
:+1:

PR #1625, opened 2018-12-14
Implement HTTP timeout for logging-agent

femueller:
This contains a fix that has been implemented to avoid that avoids stuck connections during rolling updates by using a HTTP timeout.

It has happened in the past that logs weren't shipped to Scalyr properly because of that.

More context can be found here:
https://github.com/zalando-incubator/kubernetes-log-watcher/pull/72

Signed-off-by: Felix Mueller <felix.mueller@zalando.de>


szuecs:
:+1:

mikkeloscar:
:+1:

PR #1629, opened 2018-12-17
Fix CoreDNS PDB

aermakov-zalando:
Modify the PDB for CoreDNS to only match deployment pods. Deleted/renamed because PDBs are immutable for some reason.


szuecs:
:+1:

aermakov-zalando:
👍

PR #1635, opened 2018-12-18
Cadvisor daemonset

njuettner:
The image includes a flag white_listed_container_labels which allows specifying
a comma separated list of container labels that are allowed to be
attached to prometheus metrics. This helps limit the label space of
each data point and limits the memory used by cadvisor compared to just
add all labels from all containers which is the default.

**Why:**

Without this white listing in place we saw the memory footprint of
cadvisor explode in our production environment because users add a lot
of data in pod labels, data which is not useful for metrics but may be
useful for the individual applications.

/cc: @mikkeloscar

### DIFF
--- a/cluster/manifests/cadvisor/daemonset.yaml
+++ b/cluster/manifests/cadvisor/daemonset.yaml
@@ -21,6 +21,8 @@ spec:
       priorityClassName: system-node-critical
 {{ if index .ConfigItems "enable_rbac"}}
       serviceAccountName: cadvisor
+{{ else }}
+      serviceAccountName: system
 {{ end }}
       containers:
       - name: cadvisor
--- a/cluster/manifests/cadvisor/daemonset.yaml
+++ b/cluster/manifests/cadvisor/daemonset.yaml
@@ -24,7 +24,7 @@ spec:
 {{ end }}
       containers:
       - name: cadvisor
-        image: registry.opensource.zalan.do/teapot/cadvisor:v0.30.2
+        image: google/cadvisor:v0.32.0
         args:
         - --housekeeping_interval=10s
         - --max_housekeeping_interval=15s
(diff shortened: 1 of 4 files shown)


mikkeloscar:
It looks like this in stups-test:

image

In the beginning the label whitelisting was not correctly configured, but now when it's enabled it looks ok.

I'll try and compare the metrics output with that of the cadvisor endpoint on kubelet to see if we can reduce it even more, because it exposes a lot of metrics, most of which I don't think we actually use.

mikkeloscar:
So the memory footprint was reduced, but prometheus also somehow stopped successfully scraping the cadvisor pods:

```
level=warn ts=2019-01-02T16:04:51.780154315Z caller=scrape.go:713 component="scrape manager" scrape_pool=cadvisor target=http://10.2.182.166:8080/metrics msg="append failed" err="no token found"
level=warn ts=2019-01-02T16:04:52.966723603Z caller=scrape.go:713 component="scrape manager" scrape_pool=cadvisor target=http://10.2.67.70:8080/metrics msg="append failed" err="no token found"
level=warn ts=2019-01-02T16:04:53.035424529Z caller=scrape.go:713 component="scrape manager" scrape_pool=cadvisor target=http://10.2.180.202:8080/metrics msg="append failed" err="no token found"
level=warn ts=2019-01-02T16:04:54.177277859Z caller=scrape.go:713 component="scrape manager" scrape_pool=cadvisor target=http://10.2.184.170:8080/metrics msg="append failed" err="no token found"
level=warn ts=2019-01-02T16:04:55.399064507Z caller=scrape.go:713 component="scrape manager" scrape_pool=cadvisor target=http://10.2.194.107:8080/metrics msg="append failed" err="no token found"
```

mikkeloscar:
The scraping was fixed by building a cadvisor image from the commit which included my patch. When we built from master we got this bug introduced recently: https://github.com/google/cadvisor/issues/2133

szuecs:
:+1:

mikkeloscar:
Looking like this after running for a while in `stups-test`:

image

mikkeloscar:
:+1:

PR #1636, opened 2018-12-18
Worker nodes: don't enable useless feature gate

aermakov-zalando:
We don't actually need this on the workers, so let's avoid an extra node rotation.


aermakov-zalando:
👍

dionb:
:+1:

PR #1640, opened 2018-12-18
Switch privileged components to system service account

linki:
We have several priviledged components that use the `default` service account. It was a mistake to give that account so many permissions. It got fixed when switching to RBAC but that breaks our deployments. This switches them to `system` so they work with and without RBAC.


dionb:
:+1:

linki:
:+1:

PR #1643, opened 2018-12-18
Update the VPA to version 0.3.0

arjunrn:
A [new version](https://github.com/kubernetes/autoscaler/releases/tag/vertical-pod-autoscaler-0.3.0) was released recently and this PR updates the images to the new version.

### DIFF
--- a/cluster/manifests/01-vertical-pod-autoscaler/admission-controller-deployment.yaml
+++ b/cluster/manifests/01-vertical-pod-autoscaler/admission-controller-deployment.yaml
@@ -20,7 +20,7 @@ metadata:
   namespace: kube-system
   labels:
     application: vpa-admission-controller
-    version: patched-master-7
+    version: patched-0.3.0-master-9
     component: vpa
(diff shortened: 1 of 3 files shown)


hjacobs:
Important: Between versions 0.2.x and 0.3.x there is an alpha to beta switch which includes a change of VPA apiVersion from poc.autoscaling.k8s.io/v1alpha1 to autoscaling.k8s.io/v1beta1.

(https://github.com/kubernetes/autoscaler/tree/vpa-release-0.3/vertical-pod-autoscaler#notice-on-switching-from-alpha-to-beta-030-to-030)

arjunrn:
@hjacobs Since we were building from master we were already using the new [CRD](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/01-vertical-pod-autoscaler/01-crd.yaml).

njuettner:
👍

arjunrn:
:+1:

PR #1649, opened 2018-12-20
Don't use spot_price

aermakov-zalando:
LaunchTemplates allow running spot instances without specifying the price limit. Remove from the templates, drop from CLM later.

### DIFF
--- a/cluster/node-pools/worker-default/stack.yaml
+++ b/cluster/node-pools/worker-default/stack.yaml
@@ -93,10 +93,6 @@ Resources:
         InstanceType: "{{ .NodePool.InstanceType }}"
 {{- if eq .NodePool.DiscountStrategy "spot_max_price"}}
         InstanceMarketOptions:
-          SpotOptions:
-            SpotInstanceType: one-time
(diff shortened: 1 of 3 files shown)


njuettner:
👍

njuettner:
Forgot that we can skip it now, thanks 😄.

I think we can also skip:
* InstanceInterruptionBehavior 
* SpotInstanceType

```
Properties

InstanceInterruptionBehavior The behavior when a Spot Instance is interrupted. The default is terminate.
Valid values include: hibernate, stop, and terminate.
Required: No
Type: String
Update requires: No interruption

MaxPrice The maximum hourly price you're willing to pay for the Spot Instances.
Required: No
Type: String
Update requires: No interruption

SpotInstanceType The Spot Instance request type.
Valid values include: one-time and persistent.
Required: No
Type: String
Update requires: No interruption
```

aermakov-zalando:
I'm not sure what the default is

aermakov-zalando:
👍

njuettner:
👍

PR #1652, opened 2018-12-27
Grant zmon system serviceaccount

dionb:
default service account currently has too much acess, and doesn't have any after move to RBAC, so can cause issues during the migration.

Signed-off-by: Dion Bramley <dion.bramley@zalando.de>


arjunrn:
Can we run the e2e tests on this?

dionb:
:+1:

arjunrn:
:+1:

PR #1661, opened 2019-01-03
zmon: use default SA if RBAC is disabled

aermakov-zalando:
If RBAC is disabled, use the `default` service account since `system` doesn't exist in other namespaces. This means that ZMON will be broken while we disable RBAC, but at least it won't be broken forever.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1666, opened 2019-01-07
Support multiple instance types for worker pools

aermakov-zalando:
* Use `instance_types`, not `instance_type`.
 * Use `MixedInstancesPolicy` for worker pools if more than 1 instance type is defined.

Corresponding CLM PR: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/115.
Depends on 1.12 because I don't want to port CA changes.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -10,6 +10,7 @@ autoscaling_buffer_pods: "1"
 {{else}}
 autoscaling_buffer_pods: "0"
 {{end}}
+experimental_autoscaler_mixed_instance_policy_support: "false"
 
 # skipper daemonset
 skipper_limits_mem: "250Mi"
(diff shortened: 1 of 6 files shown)


mikkeloscar:
LGTM, waiting for CLM merge and e2e before :+1:  :)

aermakov-zalando:
👎

aermakov-zalando:
I still have to fix spot-related issues now that the update to spotness is not atomic.

njuettner:
Really cool 😎 @aermakov-zalando
I'm just wondering if we could also have `OnDemandPercentageAboveBaseCapacity` for spot and on-demand as a separate config item to tweak them like 70% On-Demand and 30% Spot.

aermakov-zalando:
Thanks! We can, I just didn't want to implement the logic to adjust the allocation in CLM after configuring the ASG. AWS doesn't do anything when you change these values, they only control the properties of new instances started afterwards.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1667, opened 2019-01-08
Allow pods to talk to node-local coredns pods via service ip

linki:
The fallback DNS service is a service IP that targets all node-local CoreDNS pods (when `cluster_dns=coredns` config item is set for the cluster). Those pods are reachable **via their host's EC2 IP**. Unfortunately, master and worker nodes are **not** allowed to talk to port 53 on another node. This fixes it.


aermakov-zalando:
👍

njuettner:
👍

PR #1679, opened 2019-01-10
update kube-state-metrics to 1.8.0

### LINKED ISSUE #1678: Update kube-state-metrics to 1.5.0+
There are some valuable new features, e.g. the new `kube_pod_container_status_last_terminated_reason` metric which allows us to easily track termination reasons (like `OOMKill`) as this is currently not conveniently possible (`kube_pod_container_status_terminated_reason` is only a snapshot and depending on scraping interval it would not show all reasons), see https://github.com/kubernetes/kube-state-metrics/issues/344. I saw this new metric field when I went through the Prometheus metrics (scraped from `kube-state-metrics`) with @szuecs recently.

[We currently use `kube-state-metrics` 1.3.0](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/kube-state-metrics/deployment.yaml#L27). Version 1.5.0 is currently beta, but probably available soon: https://github.com/kubernetes/kube-state-metrics/releases

szuecs:
Fixes #1678: this update adds some useful metrics like kube_pod_container_status_last_terminated_reason


linki:
:+1:

linki:
The image hasn't been built yet. We [pull and mirror it to our registry](https://github.bus.zalan.do/teapot/kube-state-metrics/blob/38cc56cd92dda25c332aff7f95b6e68e6846d656/delivery.yaml) but the upstream repo [doesn't have 1.5 yet](https://console.cloud.google.com/gcr/images/google-containers/GLOBAL/kube-state-metrics?gcrImageListsize=30).

szuecs:
then wait until it's build

linki:
Needs https://github.bus.zalan.do/teapot/kube-state-metrics/pull/9

cluster/manifests/kube-state-metrics/deployment.yaml
```diff
@@ -24,17 +24,32 @@ spec:
...
+        readinessProbe:
+          httpGet:
+            path: /
+            port: 8081
+          initialDelaySeconds: 5
+          timeoutSeconds: 5
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
```
mikkeloscar: wrong port?

cluster/manifests/kube-state-metrics/deployment.yaml
```diff
@@ -24,17 +24,32 @@ spec:
...
+          name: http-metrics
+        - containerPort: 8081
+          name: telemetry
+        readinessProbe:
+          httpGet:
+            path: /
+            port: 8081
+          initialDelaySeconds: 5
+          timeoutSeconds: 5
+        livenessProbe:
```
szuecs: Why livenessprobe?
It will lead to metric drops

cluster/manifests/kube-state-metrics/deployment.yaml
```diff
@@ -24,17 +24,32 @@ spec:
...
+          httpGet:
+            path: /healthz
+            port: 8080
+          initialDelaySeconds: 5
+          timeoutSeconds: 5
```
szuecs: Why same configuration as readinessprobe?

szuecs (commented):
Please make sure you have always a reason for livenessprobe

cluster/manifests/kube-state-metrics/deployment.yaml
linki: That's from [the docs](https://github.com/kubernetes/kube-state-metrics/blob/master/examples/standard/deployment.yaml#L22-L27). I guess the idea is to kill it as soon as it's unhealthy.

Should I rather remove it? It might just stay running in unready state forever.

cluster/manifests/kube-state-metrics/deployment.yaml
linki: From the docs. I'll check in the cluster if it works out.

cluster/manifests/kube-state-metrics/deployment.yaml
linki: It's using different endpoints (metrics port vs. telemetry port). So it might make sense even with the same interval configuration. It's from [their repo](https://github.com/kubernetes/kube-state-metrics/blob/master/examples/standard/deployment.yaml#L22-L27). I haven't checked what the telemetry port exposes.

szuecs: might makes sense, thanks for the hints

szuecs:
:+1:

cluster/manifests/kube-state-metrics/deployment.yaml
linki: On second look: It's going to a different endpoint. The [healthz](https://github.com/kubernetes/kube-state-metrics/blob/master/main.go#L246-L249) endpoint indicates the process is running and `/metrics` can be served.

Is there a reason not to kill it in case they don't respond?

cluster/manifests/kube-state-metrics/rbac.yaml
```diff
@@ -3,16 +3,130 @@ kind: ServiceAccount
...
-apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
   name: kube-state-metrics
+  labels:
+    application: kube-state-metrics
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: system:monitor
+  name: kube-state-metrics
```
mikkeloscar, line 18: We need to fix this before we can merge, otherwise we get an error when CLM tries to apply it to existing clusters.

cluster/manifests/kube-state-metrics/deployment.yaml
linki: Actually there's not much difference. One serves the "kubernetes metrics" (number of pods etc.) the other one is metrics of `kube-state-metrics` itself (goroutines).

If we want to follow the "never use the same timings for readiness and liveness" approach I'm happy to change it.

cluster/manifests/kube-state-metrics/rbac.yaml
linki: Added a deletion to fix it. It's not perfect but good enough.

cluster/manifests/kube-state-metrics/deployment.yaml
linki: Dropped the `liveness` probe.

cluster/manifests/kube-state-metrics/deployment.yaml
linki: Resolved by dropping the `liveness` probe.

linki:
:+1:

mikkeloscar:
:+1:

linki:
:+1:

linki:
Follow up https://github.com/zalando-incubator/kubernetes-on-aws/pull/2773. Needs to be merged separately.

PR #1486, opened 2018-10-30
Adjust resource requests for logging

mohabusama:
- Update log-watcher and scalyr-agent

Signed-off-by: Mohab Usama <mohab.abdelhameed@zalando.de>

PR #1680, opened 2019-01-10
Fix/resources

szuecs:
do not overcommit memory and cput settings and use guaranteed prio class for all system components

### DIFF
--- a/cluster/manifests/kube2iam/daemonset.yaml
+++ b/cluster/manifests/kube2iam/daemonset.yaml
@@ -7,6 +7,8 @@ metadata:
     application: kube2iam
     version: master-6
 spec:
+  updateStrategy:
+    type: OnDelete
   selector:
     matchLabels:
       application: kube2iam
--- a/cluster/manifests/kube2iam/daemonset.yaml
+++ b/cluster/manifests/kube2iam/daemonset.yaml
@@ -59,7 +59,7 @@ spec:
         resources:
           limits:
             cpu: 25m
-            memory: 50Mi
+            memory: 100Mi
           requests:
             cpu: 25m
-            memory: 50Mi
+            memory: 100Mi
--- a/cluster/manifests/kube2iam/daemonset.yaml
+++ b/cluster/manifests/kube2iam/daemonset.yaml
@@ -56,8 +56,8 @@ spec:
           timeoutSeconds: 3
         resources:
           limits:
-            cpu: 100m
-            memory: 100Mi
+            cpu: 25m
+            memory: 50Mi
           requests:
             cpu: 25m
-            memory: 25Mi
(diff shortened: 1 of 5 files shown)


cluster/manifests/kube2iam/daemonset.yaml
```diff
@@ -56,8 +56,8 @@ spec:
...
         resources:
           limits:
-            cpu: 100m
-            memory: 100Mi
+            cpu: 25m
+            memory: 50Mi
           requests:
             cpu: 25m
-            memory: 25Mi
+            memory: 50Mi
```
aermakov-zalando: This won't work unless it's `OnDelete`. Nodes will end up having no `kube2iam` pod at all.

aermakov-zalando:
Increasing the requests will just end up in nodes not running the pod at all, unless you change it to `OnDelete`.

aermakov-zalando:
👎

cluster/manifests/kube2iam/daemonset.yaml
szuecs: fixed @aermakov-zalando

aermakov-zalando: Please fix the rest (requests are increased in other daemonsets as well).

szuecs: done, would be nice if you would tell the next time this upfront

aermakov-zalando: sorry!

aermakov-zalando: have you verified that it won't cause issues, btw?

cluster/manifests/logging-agent/daemonset.yaml
```diff
@@ -112,11 +112,11 @@ spec:
 
         resources:
           limits:
-            cpu: 200m
-            memory: 200Mi
+            cpu: 100m
+            memory: 150Mi
```
aermakov-zalando: This definitely won't work, I've seen in crashing in a couple of clusters with the existing `200Mi` limit. There's another PR (https://github.com/zalando-incubator/kubernetes-on-aws/pull/1486) to increase it instead.

aermakov-zalando:
Where do the new limits come from? Are we sure that stuff won't start randomly dying under load?

cluster/manifests/logging-agent/daemonset.yaml
szuecs: I adjusted the resource to the ones from the other PR and closed the other PR

szuecs:
:+1:

aermakov-zalando:
👍

PR #1684, opened 2019-01-11
Make cadvisor resources configurable

mikkeloscar:
Some clusters require more resources for cadvisor. Therefore introduce the following changes to work around the problem:

* Make cadvisor resources configurable per cluster.
* Disable metrics we don't use to save on CPU and a bit on memory.

Graph from `merchant-center` cluster:

image


mikkeloscar:
:+1:

gargravarr:
:+1:

njuettner:
👍

PR #1691, opened 2019-01-11
configure backend connection pool settings via CR

szuecs:

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -29,6 +29,14 @@ skipper_clusterratelimit: "false"
 skipper_clusterratelimit: "true"
 {{end}}
 
+# skipper backend timeout defaults
+skipper_expect_continue_timeout_backend: "30s"
+skipper_keepalive_backend: "30s"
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

cluster/config-defaults.yaml
```diff
@@ -29,6 +29,14 @@ skipper_clusterratelimit: "false"
 skipper_clusterratelimit: "true"
 {{end}}
 
+# skipper backend timeout defaults
+skipper_expect_continue_timeout_backend: "30s"
+skipper_keepalive_backend: "30s"
+skipper_max_idle_connection_backend: 0
```
mikkeloscar: I wonder if this is parsed correctly and doesn't end up as empty string or something like that. I would expect this should be a string to be correct but I'm not 100% sure. @aermakov-zalando do you know?

szuecs: `func Itoa(i int) string` should be fine in Go, there is no error as return value

szuecs:
:+1:

mikkeloscar:
:+1:

PR #1702, opened 2019-01-14
Put CoreDNS DaemonSet in Guaranteed QoS class

linki:
This puts the CoreDNS DaemonSet in Guaranteed QoS class as well.

Since we want to set `limits=requests` I would propose to just specify `limits` which will automatically set `requests` to the same value. This simplifies the yaml a little and avoids putting different numbers accidentally.

### DIFF
--- a/cluster/manifests/coredns-local/daemonset-coredns.yaml
+++ b/cluster/manifests/coredns-local/daemonset-coredns.yaml
@@ -66,10 +66,8 @@ spec:
           successThreshold: 1
           failureThreshold: 3
         resources:
-          requests:
-            cpu: 150m
-            memory: 100Mi
           limits:
+            cpu: 150m
             memory: 100Mi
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
Wouldn't you just get the default requests instead?

linki:
Good point. Not that I know of and not according to the docs but I didn't consider our `default` limit range settings.

https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/
> The output shows that the Container’s memory request is set to match its memory limit. Notice that the Container was not assigned the default memory request value of 256Mi.

I'll check.

linki:
It works like the docs say. The default limit range only applies if you don't specify requests nor limits.

linki:
:+1:

szuecs:
:+1:

PR #1706, opened 2019-01-15
coredns: enable pprof

aermakov-zalando:
This way we can find out what's using so much memory.

### DIFF
--- a/cluster/manifests/coredns-local/configmap.yaml
+++ b/cluster/manifests/coredns-local/configmap.yaml
@@ -17,6 +17,7 @@ data:
         }
         prometheus :9153
         proxy . /etc/resolv.conf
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
👍

linki:
:+1:

linki:
Docs state that [it can be slow](https://github.com/coredns/coredns/tree/master/plugin/pprof). Since it's already used in production should we rather put it behind a config item?

aermakov-zalando:
@linki it's only slow if you use it, just enabling it doesn't do anything. keeping it disabled is also pretty useless, since enabling it would just terminate the pods that you'd want to look at.

linki:
I see. I thought it would be slower merely by enabling pprof.

PR #1708, opened 2019-01-15
Always set the CLC labels

aermakov-zalando:
This makes the switch a lot less annoying since the labels would already be in place.


szuecs:
Why this is not targeting `dev` channel first and use `dev-to-kube-1.12` to propagate?

aermakov-zalando:
@szuecs Because I want to avoid yet another rolling node update so we can just piggy-back on top of 1.12.

aermakov-zalando:
👍

linki:
:+1:

PR #1710, opened 2019-01-15
flannel: init once all nodes are reachable

aermakov-zalando:
Add `experimental_flannel_awaiter` feature flag. If enabled, check that all nodes are reachable and return immediately instead of always waiting for 2 minutes.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1711, opened 2019-01-15
CA: use expander=prefer-spot

aermakov-zalando:
Let's try running it like this and see if it performs good enough.

### DIFF
--- a/cluster/manifests/kube-cluster-autoscaler/daemonset.yaml
+++ b/cluster/manifests/kube-cluster-autoscaler/daemonset.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: kube-cluster-autoscaler
-    version: {{ if eq .ConfigItems.experimental_autoscaler_mixed_instance_policy_support "true" }}v1.12.1-internal-dev17{{ else }}v1.12.1-internal10{{ end }}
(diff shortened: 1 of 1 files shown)


cluster/manifests/kube-cluster-autoscaler/daemonset.yaml
```diff
@@ -41,6 +41,7 @@ spec:
           - --skip-nodes-with-system-pods=false
           - --skip-nodes-with-local-storage=false
           - --scale-up-cloud-provider-template=true
+          - --expander=prefer-spot
```
mikkeloscar, line 45: This flag works in both versions? (You only updated one of them).

aermakov-zalando:
👍

cluster/manifests/kube-cluster-autoscaler/daemonset.yaml
aermakov-zalando: Oops, you're right.

aermakov-zalando:
👍

szuecs:
:+1:

linki:
Related to that: https://github.com/pusher/k8s-spot-rescheduler#deploy-to-kubernetes proposes to set a `PreferNoSchedule` taint on all non-spot workers (https://github.com/pusher/k8s-spot-rescheduler#motivation) so that by default spot instances are filled up first by the scheduler.

@aermakov-zalando @mikkeloscar WDYT?

PR #1713, opened 2019-01-16
Added patched VPA recommender with memory-saver mode enabled

arjunrn:
Added patched version of the VPA recommender with memory-saver mode enabled.

### DIFF
--- a/cluster/manifests/01-vertical-pod-autoscaler/recommender-deployment.yaml
+++ b/cluster/manifests/01-vertical-pod-autoscaler/recommender-deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: vpa-recommender
-    version: patched-0.3.0-master-9
+    version: patched-0.3.0-master-12
     component: vpa
 spec:
   replicas: 1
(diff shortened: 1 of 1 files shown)


mikkeloscar:
It's failing with `OCI runtime create failed: container_linux.go:348: starting container process caused "exec: \"--stderrthresholdinfo=info\": executable file not found in $PATH": unknown
` fwiw.

arjunrn:
:+1:

cluster/manifests/01-vertical-pod-autoscaler/recommender-deployment.yaml
```diff
@@ -16,13 +16,19 @@ spec:
...
       containers:
       - name: recommender
-        image: registry.opensource.zalan.do/teapot/vpa-recommender:patched-0.3.0-master-9
+        image: registry.opensource.zalan.do/teapot/vpa-recommender:patched-0.3.0-master-12
+        args:
+        - --stderrthreshold=info
+        - --v=5
+        - --memory-saver
+        command:
+        - "./recommender"
```
szuecs: is the "." in front correct? 
It now depends on CWD and I would like to have it better absolute path instead.

arjunrn: Made the change

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #1719, opened 2019-01-17
Adding vpa to kubernetes-lifeycle-metrics

muaazsaleem:
The memory requirments can differ on different cluster.
To avoid the container running out of memory adding vpa to the klm


arjunrn:
:+1:

muaazsaleem:
👍

PR #1725, opened 2019-01-17
Kubernetes Dashboard v1.10.1

hjacobs:
Fresh upstream release (27 days old): https://github.com/kubernetes/dashboard/releases/tag/v1.10.1

### DIFF
--- a/cluster/manifests/dashboard/deployment.yaml
+++ b/cluster/manifests/dashboard/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: kubernetes-dashboard
-    version: v1.6.1
+    version: v1.10.1
     kubernetes.io/cluster-service: "true"
 spec:
   replicas: 1
@@ -16,7 +16,7 @@ spec:
     metadata:
       labels:
         application: kubernetes-dashboard
-        version: v1.6.1
+        version: v1.10.1
         kubernetes.io/cluster-service: "true"
     spec:
       dnsConfig:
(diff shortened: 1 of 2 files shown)


mikkeloscar:
I have added RBAC configuration as described in https://github.com/kubernetes/dashboard/wiki/Access-control and also configured the login view feature which is new since the 1.7 version of the dashboard.

The login view allows per user permissions in the dashboard meaning it will no longer use the permissions of the serviceAccount which it's running as but rather replicate the permissions that the user also has via `kubectl`.

With this in place you can use `zkubectl dashboard` and get this screen:

image

here you can either paste an oauth2 token (from `ztoken`), or open your local `kubeconfig` file (`~/.kube/config`) and get access to the dashboard.

It's a bit inconvenient as you can't really pass this automatically via `zkubectl dashboard` because of how the api-server proxy'ing works (It doesn't pass through your Authorization Bearer token). But it would enable users to use the latest version of the dashboard.

In the future we could also put skipper with [oidc](https://opensource.zalando.com/skipper/reference/filters/#oauthoidcanyclaims) filter in front and have a Google login page on `https://dashboard.<cluster>.example.org`. Then users wouldn't need to manually copy a token or open the `kubeconfig` file from disk.

hjacobs:
I'm OK with people pasting the token for now: this is far from optimal, but as not so many people use it I would rather give these "power users" the updated version (some things are broken in our current old dashboard version).

hjacobs:
Does not seem to move forward :-/

PR #1727, opened 2019-01-18
Skipper: delay readiness probe

aermakov-zalando:
Ideally, the readiness check should tie in to the ELB healthiness to avoid situations where the rollout proceeds with pod deletion before new replicas are added back to the ELB. This is a quick and dirty fix to avoid this situation in most cases.


aermakov-zalando:
👍

arjunrn:
:+1:

PR #1732, opened 2019-01-18
API server: increase the ELB timeout

aermakov-zalando:
The ELB timeout causes stuff like `exec` or `port-forward` to stop working after 5 minutes if there's no activity. Let's increase it to the max. value (4k seconds) to avoid it.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -45,7 +45,7 @@ Resources:
         Enabled: true
         Timeout: 60
       ConnectionSettings:
-        IdleTimeout: 300
+        IdleTimeout: 3600
       CrossZone: 'true'
       HealthCheck:
         HealthyThreshold: '2'


aermakov-zalando:
👍

szuecs:
IMHO it's better to help in zkubectl that timeout does not apply and possibly we should write a guide how to use screen to make work it safe to work with DB transactions.

aermakov-zalando:
@szuecs The connection will still be dropped so I'm not sure what you want to change.

aermakov-zalando:
The users also expect stuff like `port-forward` or `exec` to just work, and currently it's pretty broken.

szuecs:
@aermakov-zalando you are right we can't really support something in zkubectl, but IMO users should use screen to do remote shell work.

cluster/cluster.yaml
```diff
@@ -45,7 +45,7 @@ Resources:
         Enabled: true
         Timeout: 60
       ConnectionSettings:
-        IdleTimeout: 300
+        IdleTimeout: 4000
```
szuecs: maybe 3600 would be a better time, because it maps to 1h and would remind us better

aermakov-zalando: Good point, and the difference is not that big anyway.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1733, opened 2019-01-18
CoreDNS: add dnsmasq for caching

aermakov-zalando:
Run dnsmasq as a caching resolver on every node, only forward to CoreDNS if we can't find anything in the cache.

### DIFF
--- a/cluster/manifests/coredns-local/configmap-local.yaml
+++ b/cluster/manifests/coredns-local/configmap-local.yaml
@@ -1,0 +1,30 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: coredns-local
+  namespace: kube-system
+  labels:
+    application: coredns
+data:
+  Corefile: |
+    .:9254 {
+        errors
+        health :9154
+        kubernetes cluster.local in-addr.arpa ip6.arpa {
+            pods insecure
+            upstream
+            fallthrough in-addr.arpa ip6.arpa
+        }
+{{ if eq .ConfigItems.enable_skipper_eastwest "true"}}
+        template IN A ingress.cluster.local  {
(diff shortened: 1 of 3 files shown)


cluster/manifests/coredns-local/configmap-local.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: coredns-local
+  namespace: kube-system
+  labels:
+    application: coredns
+data:
+  Corefile: |
+    .:54 {
```
szuecs: Do we need to run this with root?
I would rather switch it to another port, maybe 9053?

cluster/manifests/coredns-local/service.yaml
```diff
@@ -0,0 +1,21 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: coredns-local-dnsmasq
+  namespace: kube-system
+  labels:
+    application: coredns
```
szuecs, line 7: This overlaps with the current CoreDNS as deployment. Does it interfere on deployment? Do we have hairpin mode set?

aermakov-zalando: I don't think the service labels matter, and the selector includes `instance: node-dns`. In my tests it didn't interfere with anything.

aermakov-zalando: The service is only there for the metrics collection anyway.

szuecs: ok, if it's only metrics collection we can go

mikkeloscar:
:+1:

szuecs:
:+1:

PR #1750, opened 2019-01-22
kube-metrics-adapter-to-kube-1.12-alpha

arjunrn:
Hotfix for `kube-metrics-adapter` with changes for supporting weighted backends.

Also skipper option to disable access logs cluster wide.


aermakov-zalando:
👍

PR #1753, opened 2019-01-23
Update to newer version of kube-metrics-adapter

arjunrn:
This version of `kube-metrics-adapter` does not break compatibility with existing HPAs which use the `request-per-second` metric.


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #1763, opened 2019-01-25
Preparation steps for TaintNodesByCondition

aermakov-zalando:
* CLC: add a toleration for `not-ready`. This will make sure that we can safely enable `TaintNodesByCondition` even in clusters running CLC.
 * Use the correct feature flag in `default-limits`.
 * Update the admission controller to include the node admitter.
 * Admission controller: handle node `CREATE`s


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #1779, opened 2019-01-31
Reduce kube-proxy CPU requests/limits

aermakov-zalando:
It doesn't use that much, and this will free up 1 more whole CPU.


aermakov-zalando:
👍

PR #1788, opened 2019-02-01
Fail if fetching the IP address doesn't work

aermakov-zalando:
Otherwise we end up with an empty IP address which breaks the next steps.

### DIFF
--- a/cluster/node-pools/master-default/userdata.clc.yaml
+++ b/cluster/node-pools/master-default/userdata.clc.yaml
@@ -38,7 +38,7 @@ systemd:
       Type=simple
       Restart=on-failure
       RestartSec=1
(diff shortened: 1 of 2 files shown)


mikkeloscar:
Added `-euo pipefail` in all the places!

mikkeloscar:
:+1:

mikkeloscar:
Let's just roll this out despite the failing statefulset test. It's because of an issue with CLM where it looks up a node already deleted:

```
time="2019-02-01T11:57:18Z" level=info msg="Terminating node" node=ip-172-31-16-89.eu-central-1.compute.internal 
time="2019-02-01T11:58:53Z" level=debug msg="Found 1 nodes to cordon" strategy=rolling 
time="2019-02-01T11:58:53Z" level=fatal msg="Fail to provision: nodes "ip-172-31-16-89.eu-central-1.compute.internal" not found" 
```

aermakov-zalando:
👍

PR #1793, opened 2019-02-04
Prometheus: store container_memory_{rss/cache}

aermakov-zalando:
This should provide better results than just working_set_bytes.


aermakov-zalando:
👍

szuecs:
:+1:

PR #1799, opened 2019-02-04
Use the correct size for buffer pods

aermakov-zalando:
We use ~1032m CPU now, let's bump it to 1250m (and automate it!).


aermakov-zalando:
👍

szuecs:
:+1:

PR #1810, opened 2019-02-08
PCS validation: don't validate token

aermakov-zalando:
Apparently `type: array` doesn't validate `null`. I was testing it with `kubectl apply` but it looks like it was not doing anything in the background so I didn't notice. Unfortunately there's no clean way to validate `null|["a","b"]`, and _a lot_ of PCSes have `privileges: null`. Let's relax the validation for now and then slowly migrate the users to something more sane.


aermakov-zalando:
👍

gargravarr:
:+1:

PR #1822, opened 2019-02-12
enabling default encryption of EBS

jrake-revelant:
AWS allows us to encrypt data at rest by default without performance penalty. this PR enables it.

### DIFF
--- a/cluster/node-pools/master-default/stack.yaml
+++ b/cluster/node-pools/master-default/stack.yaml
@@ -54,6 +54,7 @@ Resources:
           Ebs:
             VolumeSize: 20
             VolumeType: standard
+            Encrypted: true
         EbsOptimized: false
         IamInstanceProfile:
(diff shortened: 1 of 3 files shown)


jrake-revelant:
Failing probably because of [this](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSEncryption.html):

```
There is no direct way to encrypt an existing unencrypted volume, or to remove encryption from an encrypted volume. However, you can migrate data between encrypted and unencrypted volumes.
```

jrake-revelant:
@mikkeloscar this was never merged anywhere else, was it?

mikkeloscar:
@jrake-revelant no it wasn't, but we can probably enable this globally per account instead of just for the cluster components.

mikkeloscar:
Closing this since we will enable EBS encryption at the account level.

PR #1825, opened 2019-02-13
Kubernetes v1.12.6 + Use Docker images for individual components instead of hyperkube

aermakov-zalando:
This includes two changes:

* Update to Kubernetes `v1.12.6` (https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.12.md#downloads-for-v1126)
* Use individual images for Kubernetes components. We are making this change away from `hyperkube` because it doesn't seem to be widely used for running Kubernetes and we run into issues like: https://github.com/kubernetes/kubernetes/issues/73587. There's also talk about removing support for `hyperkube`: https://github.com/kubernetes/kubernetes/issues/74148

### DIFF
--- a/cluster/manifests/kube-proxy/daemonset.yaml
+++ b/cluster/manifests/kube-proxy/daemonset.yaml
@@ -31,10 +31,8 @@ spec:
       hostNetwork: true
       containers:
       - name: kube-proxy
-        image: registry.opensource.zalan.do/teapot/hyperkube:v1.12.5
-        command:
-        - /hyperkube
-        - proxy
+        image: registry.opensource.zalan.do/teapot/kube-proxy:v1.12.5-1
+        args:
         - --config=/config/kube-proxy.yaml
         - --v=2
         securityContext:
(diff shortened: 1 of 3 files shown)


mikkeloscar:
> Apparently nobody uses hyperkube

Could you add a reference for this statement? :) I mean I don't really care, but if I look at this PR in 6 months it will not be clear why we changed away from hyperkube.

hjacobs:
Please also reference the issues we had with Hyperkube (if any).

mikkeloscar:
:+1:

mikkeloscar:
Reference to why we go away from hyperkube is in the PR comment

szuecs:
Can you summarize what `kube-component` is https://github.com/zalando-incubator/kubernetes-on-aws/pull/1825/files#diff-8f1eda593049fa31a03c0238a0135008R165 ?

mikkeloscar:
@szuecs `kube-component` is the generic name for the binary inside the docker image that we build. (you can see our internal build pipeline here: https://github.bus.zalan.do/teapot/kubernetes-pipeline/blob/9d3803c93cf47a903b447f2aaddf356160546432/Dockerfile#L7)

szuecs:
ok thanks

szuecs:
:+1:

PR #1829, opened 2019-02-14
Prometheus: record etcd metrics

aermakov-zalando:
They're pretty useful to investigate latency issues and we don't record them at all. Depends on https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/127.


mikkeloscar:
:+1:

njuettner:
👍

PR #1833, opened 2019-02-18
disallow ssh password login

szuecs:
Adds `AuthenticationMethods publickey` to the default `/etc/ssh/sshd_config` to only allow logins with `publickey`.

### DIFF
--- a/cluster/node-pools/master-default/userdata.clc.yaml
+++ b/cluster/node-pools/master-default/userdata.clc.yaml
@@ -985,7 +985,7 @@ storage:
     contents:
       inline: |
         Subsystem sftp internal-sftp
-        ClientAliveInterval 600
+        ClientAliveInterval 180
         AuthenticationMethods publickey
         UseDNS no
         UsePAM yes
--- a/cluster/node-pools/master-default/userdata.clc.yaml
+++ b/cluster/node-pools/master-default/userdata.clc.yaml
@@ -980,6 +980,19 @@ storage:
         NTP=169.254.169.123
 
   - filesystem: root
+    path: /etc/ssh/sshd_config
+    mode: 0400
+    contents:
+      inline: |
+        Subsystem sftp internal-sftp
+        ClientAliveInterval 600
+        AuthenticationMethods publickey
+        UseDNS no
+        UsePAM yes
+        PrintLastLog no # handled by PAM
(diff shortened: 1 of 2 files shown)


cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -979,6 +979,19 @@ storage:
...
+  - filesystem: root
+    path: /etc/ssh/sshd_config
+    mode: 0400
+    contents:
+      inline: |
+        Subsystem sftp internal-sftp
+        ClientAliveInterval 600
+        AuthenticationMethods publickey
+        UseDNS no
+        UsePAM yes
```
mikkeloscar, line 991: This should be `no`?

szuecs: I tested that the proposed solution from https://github.com/coreos/bugs/issues/2114 works correctly even with PAM enabled.

szuecs:
:+1:

cluster/node-pools/master-default/userdata.clc.yaml
mikkeloscar: Ok, I didn't understand what was changed from the default config. But I see it's the line: `AuthenticationMethods publickey` that is added. Was `ClientAliveInterval` also changed? I see it has the value `180` on the node I have checked:

```
$ sudo cat /etc/ssh/sshd_config 
# Use most defaults for sshd configuration.
Subsystem sftp internal-sftp
ClientAliveInterval 180
UseDNS no
UsePAM yes
PrintLastLog no # handled by PAM
PrintMotd no # handled by PAM
```

szuecs: yes, I thought it makes sense, but let's revert this value, because change less the better.

szuecs:
:+1:

mikkeloscar:
:+1:

njuettner:
I'm not sure if anything from this is relevant: https://infosec.mozilla.org/guidelines/openssh#modern-openssh-67

szuecs:
@njuettner If we need to harden openssh we can do this, too. The PR was more about deleting attack surface for brute forcing password login, which is not really possible in our setup, but like this we do not have to argue about that. :)

PR #1834, opened 2019-02-19
Enable cadvsior disk metrics

njuettner:
By enabling disk metrics, we can get the current disk usage from containers and disk read and write operations.

* Allow prometheus to scrape
   - `container_fs_usage_bytes` # Number of bytes that are consumed by the container on this filesystem.
   - `container_fs_limit_bytes` # Number of bytes that can be consumed by the container on this filesystem.
   - `container_fs_reads_bytes_total` # Cumulative count of bytes read
   - `container_fs_writes_bytes_total` # Cumulative count of bytes written

`container_fs_io_current` does return 0, see issue
https://github.com/google/cadvisor/issues/1403

I've enabled it one cluster and it seems it slightly needs more CPU:

image

Memory consumption hasn't changed so far:

image

### DIFF
--- a/cluster/manifests/cadvisor/daemonset.yaml
+++ b/cluster/manifests/cadvisor/daemonset.yaml
@@ -29,7 +29,7 @@ spec:
         - --max_housekeeping_interval=15s
         - --event_storage_event_limit=default=0
         - --event_storage_age_limit=default=0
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

szuecs:
:+1:

PR #1835, opened 2019-02-19
Set Fluentd flush thread settings back to default

christianberg:
Flush thread settings were customized before in an attempt to go
through a backlog of logs more quickly. We now avoid having a large
backlog by discarding logs older than 4 hours.

We go back to the default settings to reduce memory usage and CPU
spikes.

Signed-off-by: Christian Berg <christian.berg@zalando.de>


gargravarr:
:+1:

njuettner:
👍

PR #1836, opened 2019-02-19
Add config item to control which coredns pods are in the coredns service (10.3.0.11)

mikkeloscar:
Add temporary flag for the DNS rollout, when set to true, only the Deployment based CoreDNS pods will be added to the central CoreDNS service (10.3.0.11).

This will prevent new daemonset based CoreDNS pods from being added to the central service before the node is ready. This is assumed to be a problem only during the rolling update because with the new DNS setup going to the central service endpoint will only happen if the local CoreDNS is down.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #1838, opened 2019-02-19
Kubernetes Janitor

hjacobs:
Run [kube-janitor](https://github.com/hjacobs/kube-janitor) in non-prod clusters.

This is a first iteration to try out different use cases, e.g. to add annotations for temporary PR deployments (from CDP). The deployment itself will not do anything right now unless resources are annotated with `janitor/ttl`. I added an `if` to make sure it does not accidentally run in production (yet).

`kube-janitor` is already running in our platform test cluster and I observed the resource consumption (it's stable below 100Mi memory). We can consider moving to a CronJob and a less frequent schedule to limit the impact on the Kubernetes API server (as `kube-janitor` lists nearly all resources periodically), but I don't see any problems right now.

The ConfigMap is currently only an example and will not do anything (no active rules defined).
I included it to make clear how this would look like and make it easy to start adding rules if we later want to.

### DIFF
--- a/cluster/manifests/kube-janitor/rules-config.yaml
+++ b/cluster/manifests/kube-janitor/rules-config.yaml
@@ -1,0 +1,30 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kube-janitor
+  namespace: kube-system
+data:
+  rules.yaml: |-
+    # example rules configuration to set TTL for arbitrary objects
+    # see https://github.com/hjacobs/kube-janitor for details
+    rules:
+      - id: require-application-label
+        # remove deployments and statefulsets without a label "application"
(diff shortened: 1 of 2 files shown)


cluster/manifests/kube-janitor/rules-config.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: v1
```
mikkeloscar, line 2: Should also only be deployed to non-production environments IMO.

hjacobs: done

jrake-revelant:
👍

hjacobs:
:+1:

PR #1843, opened 2019-02-20
Cleanup old DNS setup (step 1)

mikkeloscar:
This is step 1/2 for cleaning up resources from the old DNS setup which are no longer needed.

It does the following:

* Removes `coredns` deployment
* Removes `coredns` configmap which was only used by the deployment. (there is another called `coredns-local` used by the daemonset pods.
* Removes `dnsmasq-node` daemonset
* Removes `dnsmasq-node` service (used for gathering metrics form the old daemonset)
* Removes coredns related config items used by the deployment.
* Changes the default for config item `cluster_dns` to the value `coredns` which is what we have set for all clusters in the cluster registry.
* Remove the anti-affinity on coredns daemonset pods

Next step will be to remove the `cluster-dns` label from the nodes, but this we can always do (has to be done after this is completely rolled out).


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1845, opened 2019-02-20
Use a dedicated config item for service account key

aermakov-zalando:
This allows us to rotate the apiserver key.


cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -900,6 +900,20 @@ storage:
...
+    contents:
+      remote:
+        url: "data:text/plain;base64,{{ .Cluster.ConfigItems.service_account_private_key }}"
+
+  - filesystem: root
+    path: /etc/kubernetes/ssl/service-account-public-key.pem
+    mode: 0664
+    contents:
+      remote:
+        url: "data:text/plain;base64,{{ .Cluster.ConfigItems.service_account_private_key | base64Decode | publicKey | base64 }}"
```
mikkeloscar, line 915: My understanding is that this would result in a new key every time CLM is invoked? If so, then we shouldn't do it this way IMO.

aermakov-zalando: No, it just writes a public key into a separate file. Not that it's really doing anything for security, but we could always improve it.

aermakov-zalando: That is, the output of this is stable so there won't be any random node rotations.

mikkeloscar: > That is, the output of this is stable so there won't be any random node rotations.

Thanks, that is what I meant.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1847, opened 2019-02-21
dev-to-alpha

mikkeloscar:
* **Cleanup old DNS setup (step 1)**
   <sup>Merge pull request #1843 from zalando-incubator/cleanup-old-dns</sup>


mikkeloscar:
:+1:

cluster/manifests/coredns-local/daemonset-coredns.yaml
```diff
@@ -24,15 +24,6 @@ spec:
         version: v1.2.0
         component: cluster-dns
     spec:
-      affinity:
```
jrake-revelant, line 27: why did this affinity exist? so that the old setup did not get scheduled on a single node?

mikkeloscar: Yes, it was in place to control whether pods from the new setup or the old setup would land on a node depending on the node label `cluster-dns`.

szuecs:
:+1:

PR #1854, opened 2019-02-25
Remove cluster-dns label from nodes

mikkeloscar:
Cleanup `cluster-dns` label from nodes as it is not used anymore.

Note: this will require a rolling node update and could be done along with other changes, it's not super important.


njuettner:
👍

mikkeloscar:
:+1:

PR #1867, opened 2019-02-27
Inject ndots in each pod by using admission controller flag `--pod-dns-ndots`

njuettner:
The purpose of this PR is to downsize the amount of DNS requests
by changing the default ndots 5 to ndots 2 for each pod.

* Mutate the Pod Spec DNSConfig if not set by adding ndots 2

Further documentation: [Pod's DNS Config](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-config)

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -161,7 +161,6 @@ experimental_cluster_lifecycle_controller: "false"
 teapot_admission_controller_default_cpu_request: "25m"
 teapot_admission_controller_default_memory_request: "100Mi"
 teapot_admission_controller_process_resources: "true"
-teapot_admission_controller_ndots: "2"
 
 {{if eq .Environment "e2e"}}
 teapot_admission_controller_enabled: "true"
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
👍

avaczi:
:+1:

cluster/config-defaults.yaml
```diff
@@ -161,6 +161,7 @@ experimental_cluster_lifecycle_controller: "false"
 teapot_admission_controller_default_cpu_request: "25m"
 teapot_admission_controller_default_memory_request: "100Mi"
 teapot_admission_controller_process_resources: "true"
+teapot_admission_controller_ndots: "2"
```
aermakov-zalando: Let's keep it disabled for now, or at least enable only in test clusters first?

njuettner: Agreed, the reason why I added this was to run e2e tests to see if everything passes.

mikkeloscar:
Closing in favor of #1875

PR #1868, opened 2019-02-27
Increasing the limit of inotify watches on worker nodes

muaazsaleem:
Increasing the limit of inotify watches from 8192 to 100.000 on worker nodes in
light of issues with user pods keeping Heapster from starting up

**Note** this will cause nodes to roll. 

Signed-off-by: Muhammad Muaaz Saleem <muhammad.muaaz.saleem@zalando.de>

### DIFF
--- a/cluster/node-pools/worker-default/userdata.clc.yaml
+++ b/cluster/node-pools/worker-default/userdata.clc.yaml
@@ -415,6 +415,13 @@ storage:
         PrintMotd no    # handled by PAM
 
   - filesystem: root
+    path: /etc/sysctl.d/01-max-user-watches.conf
+    mode: 0644
+    contents:
+      inline: |
+        fs.inotify.max_user_watches=100000
+
+  - filesystem: root
     path: /opt/bin/drain-node
     mode: 0755
     contents:


mikkeloscar:
@muaazsaleem just for reference, what is the default limit that we have?

muaazsaleem:
> @muaazsaleem just for reference, what is the default limit that we have?

It's like `8192`

mikkeloscar:
```
# sysctl fs.inotify | grep wat
fs.inotify.max_user_watches = 8192
```

arjunrn:
Did you test that the change work by creating a pet cluster?

njuettner:
Does it make sense to have it configurable? Assuming it was happening in only one cluster and not all of them?

mikkeloscar:
Why do we choose `100000` as the new value, what is the decision based on?

muaazsaleem:
> Did you test that the change work by creating a pet cluster?

Yeah I set up a test cluster and the limit was increased once the nodes were rolled.

> Does it make sense to have it configurable? Assuming it was happening in only one cluster and not all of them?

We keep this under config items if we want to avoid rolling nodes in some clusters.  The change is so small that it shouldn't matter and since we're only merging to alpha/beta once a week the rolling nodes shouldn't be an overhead. WDYT?

cc: @njuettner

muaazsaleem:
> Why do we choose `100000` as the new value, what is the decision based on?

Basically guestimation. I didn't want the no. to be so big that it affects the kernel and not so small that we have to increase it soon after. So I looked at how much people were increasing it on the internet and picked something.

source: https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers

njuettner:
The reason why I also think it make sense to a make it more configurable is that by increasing the number of notify watches, we will also allow the system to consume more memory if they got used.

> Assuming you set the max at 524288 and all were used (improbable), you'd be using approximately 256MB/512MB of 32-bit/64-bit kernel memory.

So in the end this will have an effect on the instances. It seems they are still pretty lightweight but I'm just saying we could potentially use more memory on system side and having less resources for K8s.

njuettner:
👍

muaazsaleem:
👍

muaazsaleem:
Merging into `dev` so at least _a_ fix will get rolled out. I'll take a look today at wether the memory for the watches is counter per `cgroup`. If it's counted as per cgroup than it won't affect system memory and if it does than we'll make it more configurable

PR #1874, opened 2019-03-02
Enable logging of .svc DNS names in CoreDNS

mikkeloscar:
This enables logging of names that end on `.svc` from the client side.
This is to be able to find all the occurances of clients calling names
like `name.namespace.svc` which would stop working if we enforce
`ndots:2` by default for all pods.

We can detect the usage of `.svc` names by logging any occurance of
`svc.svc.cluster.local` lookups in CoreDNS.


mikkeloscar:
:+1:

njuettner:
👍

PR #1875, opened 2019-03-04
Ndots kube system

njuettner:
Only for testing purposes right now

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -161,7 +161,6 @@ experimental_cluster_lifecycle_controller: "false"
 teapot_admission_controller_default_cpu_request: "25m"
 teapot_admission_controller_default_memory_request: "100Mi"
 teapot_admission_controller_process_resources: "true"
-teapot_admission_controller_ndots: "2"
 
 {{if eq .Environment "e2e"}}
 teapot_admission_controller_enabled: "true"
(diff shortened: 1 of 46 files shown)


njuettner:
@mikkeloscar please review, I omitted VPA and CoreDNS.

mikkeloscar:
@njuettner anything which uses `hostNetwork` should use the default `/etc/resolv.conf` from the node, so the custom setting shouldn't be needed (This includes CoreDNS). I'm actually curious what happens when you set it, I would expect that it just ignores it, but maybe it injects the ndots setting?

Otherwise it looks good, we should try to run e2e and see if everything works with this in place!

njuettner:
> @njuettner anything which uses `hostNetwork` should use the default `/etc/resolv.conf` from the node, so the custom setting shouldn't be needed (This includes CoreDNS). I'm actually curious what happens when you set it, I would expect that it just ignores it, but maybe it injects the ndots setting?
> 
> Otherwise it looks good, we should try to run e2e and see if everything works with this in place!

Indeed, I totally missed it 😄. Thanks! 
I've just removed the ndots for the hostNetwork files.

szuecs:
> > @njuettner anything which uses `hostNetwork` should use the default `/etc/resolv.conf` from the node, so the custom setting shouldn't be needed (This includes CoreDNS). I'm actually curious what happens when you set it, I would expect that it just ignores it, but maybe it injects the ndots setting?
> > Otherwise it looks good, we should try to run e2e and see if everything works with this in place!
> 
> Indeed, I totally missed it smile. Thanks!
> I've just removed the ndots for the hostNetwork files.

It depends.
hostNetwork with `dnsPolicy: ClusterFirstWithHostNet` should have them.

njuettner:
👍

mikkeloscar:
:+1:

PR #1880, opened 2019-03-05
Disable automatic transparent hugepages

aermakov-zalando:
Transparent hugepages are [problematic](https://www.google.com/search?client=safari&rls=en&q=transparent+hugepage+issues&ie=UTF-8&oe=UTF-8) and the common recommendation is to disable them. Setting the corresponding flag to `madvise` disables the fully-automatic behaviour but allows applications to opt-in by using `mmap`.

### DIFF
--- a/cluster/node-pools/master-default/userdata.clc.yaml
+++ b/cluster/node-pools/master-default/userdata.clc.yaml
@@ -43,6 +43,21 @@ systemd:
       [Install]
       WantedBy=multi-user.target
 
+  - name: disable-thp.service
+    enable: true
+    contents: |
+      [Unit]
+      Before=docker.service
+
+      [Service]
+      Type=simple
+      Restart=on-failure
+      RestartSec=1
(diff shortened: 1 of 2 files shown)


mdedetrich:
For reference, I became aware of this when experiencing occasional severe latency spikes in redis, `latency doctor` in redis gave this as a first message

> I have a few advices for you:

> I detected a non zero amount of anonymous huge pages used by your process. This creates very serious latency events in different conditions, especially when Redis is persisting on disk. To disable THP support use the command 'echo never > /sys/kernel/mm/transparent_hugepage/enabled', make sure to also add it into /etc/rc.local so that the command will be executed again after a reboot. Note that even if you have already disabled THP, you still need to restart the Redis process to get rid of the huge pages already created.

aermakov-zalando:
👍

szuecs:
:+1:

szuecs:
Thanks @mdedetrich

PR #1896, opened 2019-03-13
Update CLC

aermakov-zalando:
* CLC: update to `master-3` (support `decommission-priority`)
* Master nodes: set `decommission-priority` so the nodes will always be updated first


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1899, opened 2019-03-14
Put ingress-template-controller behind feature toggle to gradually decommission it

mikkeloscar:
This puts the `ingress-template-controller` resources behind a toggle such that we can gradually decommission it from clusters which are not using it anymore.

The motivation for this is that stacksets solves the same problem much better and should be the only thing we support for traffic switching, since it doesn't make sense to support two implementations.

I will add the `enable_ingress_template_controller: "true"` config-item to all clusters which currently has `IngressTemplates` deployed before we merge this.

[Internal issue](https://github.bus.zalan.do/teapot/issues/issues/1715)


mikkeloscar:
:+1:

szuecs:
:+1:

PR #1900, opened 2019-03-14
Fix/redis based csluterratelimits

szuecs:
- remove swim based swarm port
- enable redis based cluster ratelimits with 2 node redis ring sharding

### DIFF
--- a/cluster/manifests/skipper/skipper-redis.yaml
+++ b/cluster/manifests/skipper/skipper-redis.yaml
@@ -1,0 +1,43 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  labels:
+    application: skipper-redis
+    version: v4.0.9
+  name: skipper-redis
+  namespace: kube-system
+spec:
+  replicas: 2
+  selector:
+    matchLabels:
+      application: skipper-redis
+  serviceName: skipper-redis
+  template:
+    metadata:
+      labels:
+        application: skipper-redis
+        version: v4.0.9
+    spec:
+      containers:
+      - image: registry.opensource.zalan.do/zmon/redis:4.0.9-master-6
+        name: skipper-redis
+        ports:
+        - containerPort: 6379
+          protocol: TCP
+        readinessProbe:
+          exec:
+            command:
+            - redis-cli
+            - ping
+          failureThreshold: 3
+          initialDelaySeconds: 10
+          periodSeconds: 60
+          successThreshold: 1
+          timeoutSeconds: 1
+        resources:
+          limits:
+            cpu: 100m
+            memory: 100Mi
+      dnsPolicy: ClusterFirst
+      restartPolicy: Always
+      schedulerName: default-scheduler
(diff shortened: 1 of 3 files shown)


cluster/manifests/skipper/skipper-redis.yaml
```diff
@@ -0,0 +1,45 @@
+{{ if eq .ConfigItems.skipper_clusterratelimit "true"}}
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  labels:
+    application: skipper-redis
```
mikkeloscar: maybe call it `skipper-ingress-redis` just to be consistent with how we name it? :)

szuecs: done

cluster/manifests/skipper/skipper-redis.yaml
```diff
@@ -0,0 +1,45 @@
...
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  labels:
+    application: skipper-ingress-redis
+    version: v4.0.9
+  name: skipper-ingress-redis
+  namespace: kube-system
+spec:
+  replicas: 2
```
aermakov-zalando: I suggest to always create the statefulset (and the service), and set the replicas to either 2 or 0 based on `skipper_clusterratelimit`. Otherwise you would have to manually cleanup after disabling it.

szuecs: I think this feature should go GA and it's not as easy to jsut change the number of replicas, but you need to modify the skipper-ingress params, too.
So we would need to build the string we pass to skipper with the same arg.
Feel free to contribute to this PR, otherwise I would like ignore it for now and enhance it later.

szuecs:
:+1:

cluster/manifests/skipper/skipper-redis.yaml
aermakov-zalando: I'm not talking about making the number of replicas configurable. What I meant was moving the `if` to `replicas`:

    replicas: {{if eq .ConfigItems.skipper_clusterratelimit "true"}}2{{else}}0{{end}}

Which allows us to enable or disable the flag by changing just one config item and then waiting. In the PR's version, disabling it means we have to manually login to the cluster and delete the statefulset as well.

cluster/manifests/skipper/skipper-redis.yaml
```diff
@@ -0,0 +1,45 @@
...
+            command:
+            - redis-cli
+            - ping
+          failureThreshold: 3
+          initialDelaySeconds: 10
+          periodSeconds: 60
+          successThreshold: 1
+          timeoutSeconds: 1
+        resources:
+          limits:
```
aermakov-zalando: Please add requests as well.

szuecs: requests are the same as limits if unspecified or do we break some internal admission hook?

cluster/manifests/skipper/skipper-redis.yaml
szuecs: done

szuecs:
:+1:

aermakov-zalando:
👍

PR #1902, opened 2019-03-14
Update CoreOS to 2023.5.0

aermakov-zalando:
This should fix the EBS issues on fifth generation instances.


aermakov-zalando:
Apparently not, the timeout is still 30 seconds.

aermakov-zalando:
Folded into #1903.

PR #1903, opened 2019-03-14
nvme: set timeout to max

aermakov-zalando:
This should fix issues with EBS on fifth generation instances, and AWS explicitly [recommends doing it](https://forums.aws.amazon.com/message.jspa?messageID=820271#jive-message-834572).


mikkeloscar:
:+1:

szuecs:
:+1:

PR #1907, opened 2019-03-15
Fix daemonset requests

aermakov-zalando:
* Update CA to v1.12.2-internal49: correctly handle daemonsets with limits but no requests
 * Add explicit requests to all our daemonsets


mikkeloscar:
:+1:

njuettner:
👍

PR #1910, opened 2019-03-15
skipper prometheus bucket config down to 100us instead of 10ms

szuecs:
change prometheus bucket configuration, such that we can monitor filters and route lookups, because min of 10ms is not enough


mikkeloscar:
@szuecs do you have any numbers on the impact on memory of either skipper or prometheus, if any?

szuecs:
@mikkeloscar nothing measurable. I posted in chat some numbers, but these might be different in the fabric GW test cluster with >2000 ingress. 
In the end https://github.com/zalando-incubator/kubernetes-on-aws/pull/1911 will reduce the memory much more than we increase with this one.

szuecs:
:+1:

aermakov-zalando:
👍

PR #1915, opened 2019-03-19
Admission controller: application label check

aermakov-zalando:
* Admission controller: update to the latest version
* Production deployments: deploy PCS & daemonset for the Kio proxy
* If `teapot_admission_controller_validate_application_label` is set to `true`, enable application label validation

### DIFF
--- a/cluster/manifests/admission-control/daemonset.yaml
+++ b/cluster/manifests/admission-control/daemonset.yaml
@@ -1,0 +1,56 @@
+{{ if eq .Environment "production" }}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: admission-controller-proxy
+  namespace: kube-system
+  labels:
+    application: admission-controller-proxy
+spec:
+  selector:
+    matchLabels:
+      application: admission-controller-proxy
+  updateStrategy:
+    type: RollingUpdate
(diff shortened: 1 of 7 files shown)


njuettner:
👍

cluster/manifests/admission-control/daemonset.yaml
```diff
@@ -0,0 +1,56 @@
...
+          - --address=127.0.0.1:8285
+          - --token-secret=/meta/credentials/admission-controller-token-secret
+          - --token-type=/meta/credentials/admission-controller-token-type
+        resources:
+          limits:
+            cpu: 25m
+            memory: 100Mi
+          requests:
+            cpu: 25m
+            memory: 100Mi
```
mikkeloscar: probably `10-20Mi` is enough? This is ofc. on the master nodes, so not so bad, but I don't think we need `100Mi`

aermakov-zalando: I don't want to go lower than `50Mi` because we already had issues with a 3 line shell script, due to the cache memory being counted as well. Will update to 50, though!

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #1932, opened 2019-03-22
Make it possible to configure nat CIDR blocks per cluster

mikkeloscar:
This will make it possible to use Egress feature in clusters with custom subnets for VPN.

Discussed with @szuecs that it makes sense to just do the manual definition of CIDRs for now. 

~This PR should also include the fixes from: https://github.com/szuecs/kube-static-egress-controller/pull/21 I'll update it when the image is built.~ done


mikkeloscar:
:+1:

aermakov-zalando:
:+1:

PR #1942, opened 2019-03-25
Delete the default limits if process_resources is enabled

aermakov-zalando:
This matches the logic for when they are deployed.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1951, opened 2019-03-28
Downscaler: use a better timezone

aermakov-zalando:
We have a change to DST soon, let's handle this automatically.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #1971, opened 2019-04-04
Updated the version of kube-metrics-adapter

arjunrn:
This version of kube-metrics-adapter is based on the newer autoscaling/v2beta2 package and has uses an updated version of the custom-metrics-server.

Also increased the memory for the e2e pods because it kept getting oom-killed.


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #1974, opened 2019-04-05
Enable Egress e2e tests

mikkeloscar:
Enables the Egress e2e tests which were disabled long ago. Bugs were fixed in the egress controller so the tests should work better now.


mikkeloscar:
The image used in the tests is not compliant, we need to build a new image with ping.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #1994, opened 2019-04-10
Updated the VPAs to v1beta2

arjunrn:
This is the right version of the VPAs because feature for object reference was only introduced in v1beta2.


arjunrn:
:+1:

muaazsaleem:
👍

PR #2011, opened 2019-04-15
Adjust and parametrize ZMON components

mohabusama:
Adjust and parametrize ZMON components


cluster/config-defaults.yaml
```diff
@@ -86,7 +86,11 @@ zmon_worker_count: "0"
 {{else if eq .Environment "production"}}
 logging_agent_enabled: "true"
 zmon_agent_replicas: "1"
+zmon_agent_mem: "500Mi"
```
aermakov-zalando: You can just define it once outside the `if`, it'd be easier to keep the settings in sync.

mohabusama: done

aermakov-zalando:
👍

szuecs:
:+1:

PR #2019, opened 2019-04-17
PCS: validate application name

aermakov-zalando:
For some reason, Kio returns a `400` instead of a `404` when you GET an application with an invalid name. This is considered a fatal error in credentials-provider and triggers a 24x7 alert. Validate the application name to prevent this.


aermakov-zalando:
👍

arjunrn:
:+1:

PR #2022, opened 2019-04-18
Don't allow downscaling prometheus too much

aermakov-zalando:
Since the VPA still can't deal with OOMs correctly, just forbid it from downscaling Prometheus to the minimum. This should allow our E2E tests to run.


aermakov-zalando:
👍

szuecs:
:+1:

PR #2025, opened 2019-04-18
Add support for customizable AZs

aermakov-zalando:
Use `.Values.availability_zones` instead of using `subnets` or just hardcoding. Depends on https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/140.


hjacobs:
What's the background for this configuration option? Why do we need it?

aermakov-zalando:
@hjacobs A few instance types are not available in some of the AZs. LaunchTemplates perform extensive validation, so this prevents us from even creating the CF stack for the cluster, because it's hardcoded to use all 3.

njuettner:
👍

patandrick:
:+1:

mikkeloscar:
:+1:

njuettner:
👍

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2048, opened 2019-04-25
Audit pod events

njuettner:
This needs to be activated for compliance reports.


njuettner:
👍

cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -789,7 +789,16 @@ storage:
           # don't audit events from the system:apiserver user
           - level: None
             users: ["system:apiserver"]
-          # don't audit any kube-controller-manager events
+          # audit pod events from kube-controller-manager
+          - level: Request
+            users: ["system:kube-controller-manager"]
+            omitStages:
+              - "RequestReceived"
+            verbs: ["create". "delete", "update", "patch", "deletecollection"]
```
mikkeloscar: Do we want to add `update` and `patch`, are they relevant? @aermakov-zalando had an opinion on this?

njuettner: By looking at the amount of events we would send (Pod events without GET,LIST,WATCH) it would be ok. But I'm also fine dropping them if they not necessary. I have no opinion on that.

aermakov-zalando: I would say yes, but it's not really necessary in the first iteration. You can't change compliance-relevant things because pods are immutable.

njuettner:
I just realized that by allowing verbs `create` for Pods it will also log events for executing containers because `exec` is a subresource of `create`.

njuettner:
> I just realized that by allowing verbs `create` for Pods it will also log events for executing containers because `exec` is a subresource of `create`.

Nevermind those events were seen already before

PR #2050, opened 2019-04-26
Update kube-metrics-adapter to fix ZMON collector and Skipper metrics with some missing hosts

mikkeloscar:
PRs:
1. https://github.com/zalando-incubator/kube-metrics-adapter/pull/48
2. https://github.com/zalando-incubator/kube-metrics-adapter/pull/47
3. https://github.com/zalando-incubator/kube-metrics-adapter/pull/50

### DIFF
--- a/cluster/manifests/kube-metrics-adapter/deployment.yaml
+++ b/cluster/manifests/kube-metrics-adapter/deployment.yaml
@@ -25,7 +25,7 @@ spec:
       serviceAccountName: custom-metrics-apiserver
       containers:
       - name: kube-metrics-adapter
-        image: registry.opensource.zalan.do/teapot/kube-metrics-adapter:master-28
+        image: registry.opensource.zalan.do/teapot/kube-metrics-adapter:master-29
         args:
         - --prometheus-server=http://prometheus.kube-system.svc.cluster.local
         - --skipper-ingress-metrics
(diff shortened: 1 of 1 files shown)


arjunrn:
:+1:

mikkeloscar:
:+1:

mikkeloscar:
@arjunrn I have one question/concern about zalando-incubator/kube-metrics-adapter#47. Before it would fail on any error, but now, what would happen if we get an error from what would normally be the max and we then choose a lower metric? Say an ingress has two hostnames and the main hostname fails because of a connection issue to prometheus, but the second hostname with `0 rps` succeeds because the connection works when that call is made?

arjunrn:
@mikkeloscar The fix was meant for the opposite situation. What was happening is that the metrics are only created on the fly. So there won't be metrics for hosts which are not being called. But there will be for ones which are but the adapter would return an error anyway. I've not fixed the adapter and created a [PR](https://github.com/zalando-incubator/kube-metrics-adapter/pull/50) to only ignore invalid result errors.

mikkeloscar:
@arjunrn awesome!

mikkeloscar:
@arjunrn updated the image

arjunrn:
👍

mikkeloscar:
:+1:

PR #2051, opened 2019-04-26
Always enable teapot admission controller

aermakov-zalando:
It hasn't caused any issues so far, and we have to enable it for compliance checks anyway.


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #2053, opened 2019-04-27
Add VPA to kube-state-metrics

mikkeloscar:
Adds VPA to [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics) which is crashing with the default resource setting in big clusters like `search`.


mikkeloscar:
:+1:

arjunrn:
👍

PR #2059, opened 2019-04-29
Drop pod quotas in 'default' and 'kube-system'

mikkeloscar:
Drops the `1500` limit pod quotas from `default` and `kube-system` namespace. These quotas was put in place long ago to protect the API server against broken cronjobs. This is no longer an issue as cronjob behavior has been fixed upstream and the therefore don't provide a meaningful value.
The quotas are rather causing problems for big clusters with more that `1500` pods in `kube-system` namespace so drop them to avoid that.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2065, opened 2019-04-29
Allow updating kube-node-ready/kube-proxy

aermakov-zalando:
We're probably not going to change their resource requirements any time soon, so `OnDelete` is no longer necessary.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2066, opened 2019-04-29
Enable pod events and E2E tests for auditing pods

njuettner:
Enable pod audit events for compliance reports and add E2E tests for auditing.

### DIFF
--- a/test/e2e/cluster_config.sh
+++ b/test/e2e/cluster_config.sh
@@ -8,6 +8,7 @@ clusters:
   api_server_url: ${API_SERVER_URL}
   channel: ${1}
   config_items:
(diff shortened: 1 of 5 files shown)


test/e2e/cluster_config.sh
```diff
@@ -8,6 +8,7 @@ clusters:
   api_server_url: ${API_SERVER_URL}
   channel: ${1}
   config_items:
+    audit_pod_events: "true"
```
mikkeloscar: This is not needed when set in the `cluster/config-defaults.yaml`

mikkeloscar:
:+1:

njuettner:
👍

PR #2071, opened 2019-04-30
Update admission-controller & proxy

aermakov-zalando:
Update the application list incrementally so we don't kill Kio.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2077, opened 2019-04-30
Disable --configure-cloud-routes

aermakov-zalando:
We don't use them anyway, and this performs a lot of calls so we get rate limited.


mikkeloscar:
:+1:

njuettner:
👍

PR #2079, opened 2019-04-30
Reduce terminated-pod-gc-threshold to 500

aermakov-zalando:
The default is 12500 which is way too high. Keeping just 500 recently terminated pods is more than enough.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2090, opened 2019-05-03
Make it possible to configure max-pods and node-cidr-mask-size per cluster

mikkeloscar:
This enables us to change the configuration per cluster.

We want to change the configuration because the current settings limits us to `256` nodes when each node has a `/24`. Setting this to `/25` would allow around `~510` nodes at the cost of limiting the number of pods per node to `~62`.

Note I didn't pass the `max-pods` flag for the worker kubelet to prevent another rolling upgrade. Will add this to the v1.13 branch.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -206,3 +206,17 @@ experimental_schedule_daemonset_pods: "false"
 
 # Feature toggle for auditing events
 audit_pod_events: "true"
+
+# CIDR configuration for nodes and pods
+# Changing this will change the number of nodes and pods we can schedule in the
+# cluster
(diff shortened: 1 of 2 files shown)


arjunrn:
:+1:

mikkeloscar:
:+1:

hjacobs:
Observed error message (for context):

```
Error while processing Node Add/Delete: failed to allocate cidr: CIDR allocation failed; there are no remaining CIDRs left to allocate in the accepted range'
```

Relevant controller code: https://github.com/kubernetes/kubernetes/blob/a89f8c11a5f4f132503edbc4918c98518fd504e3/pkg/controller/nodeipam/ipam/cidrset/cidr_set.go#L136-L156

PR #2097, opened 2019-05-06
add lifo filters to all routes 

szuecs:
Add preconfigured default lifo filters to all routes. This will omit memory spikes and cluster ingress downtime, caused by one or a few failing routes.

## load test data

- 1 ingress that targets a deployment with high latency
- 1 ingress that targets a deployment without additional latency

The 2 data paths should not affect the shared infrastructure and should not be visible in the latency of the "good application" ingress metrics.

### skipper-ingress cluster overview

image
image

### application targets

traffic without latency should not be affected
image

high latency, leading to timeouts
image

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -26,7 +26,7 @@ skipper_ingress_memory: "500Mi"
 skipper_ingress_tracing_buffer: "16384"
 
 # skipper default filters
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

mikkeloscar:
:+1:

PR #2098, opened 2019-05-07
Max pods workers

mikkeloscar:
Follow up to #2090 for worker nodes. Rolling out with v1.13 to reduce number of times we roll nodes.


mikkeloscar:
Wrong target branch... Closing.

PR #2115, opened 2019-05-08
cleanup using nginx in all ingress tests as backends

szuecs:


mikkeloscar:
I don't think this is relevant anymore since we did a bunch of fixes for improving the flakiness of e2e tests.

Please reopen if you don't agree :)

PR #2117, opened 2019-05-09
Update pdb-controller to v0.0.9

mikkeloscar:
Updates the pdb-controller to v0.0.9 which allows setting an annotation on deployments to remove the PDB if the pods of a deployment/statefulset is not getting ready.

Set the value for `5m` for prometheus as I think this is a reasonable time for when the pdb should be removed, such that the VPA can recover Prometheus.

[All changes since v0.0.8](https://github.com/mikkeloscar/pdb-controller/compare/v0.0.8...v0.0.9)


mikkeloscar:
:+1:

njuettner:
👍

PR #2125, opened 2019-05-12
kube-metrics-adapter: Reduce default CPU and add VPA

mikkeloscar:
The `kube-metrics-adapter` consistenly uses less than the requested resources across all clusters.

* Reduce default CPU to `10m` (avg usage last 7 days is `1m` with a few spikes to `~60m`).
* Add VPA to allow using less Memory.

CPU last 7 days:

image

Memory last 7 days:

image


mikkeloscar:
:+1:

njuettner:
👍

PR #2130, opened 2019-05-13
skipper: disable lifo by default

aermakov-zalando:
There are issues with downscaled services, so let's disable it for now.


aermakov-zalando:
👍

linki:
👍

PR #2134, opened 2019-05-14
Secure kubelet port with authentication/authorization

mikkeloscar:
Enables authentication for the secured kubelet port such that not every use can control the kubelet. It provisions a `kubelet-client` cert on the apiserver which should provide the identity `kube-apiserver-kubelet-client` to allow the apiserver access to kubelet for `kubectl logs,exec` etc.

https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-authentication-authorization/

This also moves the kubelet flags to a config file which is becoming the standard for configuring kubelet and allows easier migration to Dynamic kubelet config in the future.

https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/

https://kubernetes.io/docs/tasks/administer-cluster/reconfigure-kubelet/

## TODO:

* [x] Provision `kubelet-client` cert/key for e2e (internal PR)
* [x] Provision `kubelet-client` cert/key for all clusters
* [x] ~Figure out how to handle rollout for `metrics-server` (How can it work for both old and new kubelets?)~ During the rollout it will work fine going to the readOnly port, and after everything is upgraded it will go to the secured port.
* [x] Roll out valid worker certificates

### DIFF
--- a/cluster/node-pools/worker-default/userdata.clc.yaml
+++ b/cluster/node-pools/worker-default/userdata.clc.yaml
@@ -340,14 +340,14 @@ storage:
     mode: 0664
     contents:
       remote:
-        url: "data:text/plain;base64,{{ .Cluster.ConfigItems.worker_cert_decompressed }}"
+        url: "data:text/plain;base64,{{ .Cluster.ConfigItems.worker_cert }}"
 
   - filesystem: root
     path: /etc/kubernetes/ssl/worker-key.pem
     mode: 0664
     contents:
       remote:
-        url: "data:text/plain;base64,{{ .Cluster.ConfigItems.worker_key_decompressed }}"
+        url: "data:text/plain;base64,{{ .Cluster.ConfigItems.worker_key }}"
 
   - filesystem: root
     path: /etc/kubernetes/ssl/ca.pem
--- a/cluster/node-pools/worker-default/userdata.clc.yaml
+++ b/cluster/node-pools/worker-default/userdata.clc.yaml
@@ -229,7 +229,11 @@ systemd:
       --cpu-cfs-quota=false \
 {{- end }}
       --system-reserved=cpu=100m,memory=164Mi \
-      --kube-reserved=cpu=100m,memory=282Mi
+      --kube-reserved=cpu=100m,memory=282Mi \
+      --client-ca-file=/etc/kubernetes/ssl/ca.pem \
+      --anonymous-auth=false \
+      --authentication-token-webhook=true \
+      --authorization-mode=Webhook
       ExecStop=-/usr/bin/rkt stop --uuid-file=/var/run/kubelet-pod.uuid
       Restart=always
       RestartSec=10
--- a/cluster/node-pools/worker-default/userdata.clc.yaml
+++ b/cluster/node-pools/worker-default/userdata.clc.yaml
@@ -215,25 +215,10 @@ systemd:
 {{- end }}
       --node-labels={{ .Values.node_labels }} \
       --cluster-dns=${PRIVATE_EC2_IPV4} \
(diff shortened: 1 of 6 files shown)


mikkeloscar:
This doesn't work right now because of the worker certs we use:

```
Error from server: Get https://ip-10-120-30-40.eu-central-1.compute.internal:10250/containerLogs/kube-system/skipper-ingress-1234-abc/skipper-ingress?follow=true: x509: certificate is valid for *.*.compute.internal, *.ec2.internal, not ip-10-120-30-40.eu-central-1.compute.internal
```
I believe `*.*.compute.internal` is not valid and we should instead generate certs with `*.eu-central-1.compute.internal`

arjunrn:
Thanks  for taking this up. Is it possible to move these options to the configuration file as described [here](https://github.com/kubernetes/kubernetes/blob/v1.13.6/staging/src/k8s.io/kubelet/config/v1beta1/types.go)

cluster/node-pools/worker-default/userdata.clc.yaml
```diff
@@ -229,7 +229,11 @@ systemd:
       --cpu-cfs-quota=false \
 {{- end }}
       --system-reserved=cpu=100m,memory=164Mi \
-      --kube-reserved=cpu=100m,memory=282Mi
+      --kube-reserved=cpu=100m,memory=282Mi \
+      --client-ca-file=/etc/kubernetes/ssl/ca.pem \
+      --anonymous-auth=false \
+      --authentication-token-webhook=true \
+      --authorization-mode=Webhook
```
szuecs: maybe you do not need the webhook flags on worker node kubelets?

mikkeloscar: This is needed for kubelet to make SubjectAccessReview calls to the API server authorize requests. See: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-authentication-authorization/#kubelet-authorization

mikkeloscar:
> Thanks for taking this up. Is it possible to move these options to the configuration file as described here

@arjunrn sure, we can also move this (and other configs) to the config file while we are at it anyway.

cluster/node-pools/worker-default/userdata.clc.yaml
szuecs: Interesting, as far as I understand it's used for the token authentication/authorization. So requests that go to kubelet with a token auth, will hit the apiserver webhook passed by the kubelet.
It would be nice if a flow chart or something like this would exist in the k8s docs for this. It would make it much more clear. I added https://github.com/kubernetes/website/issues/14330.

mikkeloscar:
@arjunrn I switched most of the kubelet flags to a config file. I left `--cluster-dns` as a flag because we need to inject the `PRIVATE_EC2_IPV4` and I didn't want to deal with this in the config for now :P

arjunrn:
@mikkeloscar thanks. Yes, that flag cannot be moved as is. But this already simplifies things a lot as well for the subsequent Ubuntu move.

cluster/node-pools/worker-default/userdata.clc.yaml
mikkeloscar: > Interesting, as far as I understand it's used for the token authentication/authorization. So requests that go to kubelet with a token auth, will hit the apiserver webhook passed by the kubelet.

Correct! Thanks for creating the issue upstream :)

mikkeloscar:
We could merge this together with #2141

szuecs:
:+1:

mikkeloscar:
:+1:

szuecs:
:+1:

PR #2143, opened 2019-05-16
Enable DAX permissions

john-coleman:
* https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DAX.access-control.html


szuecs:
Please request to enable it via internal channels. thanks

PR #2155, opened 2019-05-22
Set ndots 2 by default

njuettner:
As far as I understood we can simply set ndots to 5 for the clusters
which have s problem currently in the registry because it won't overwrite to the default setting.


njuettner:
👍

cluster/config-defaults.yaml
```diff
@@ -177,12 +177,13 @@ teapot_admission_controller_default_memory_request: "100Mi"
 teapot_admission_controller_process_resources: "true"
 teapot_admission_controller_validate_application_label: "false"
 teapot_admission_controller_application_min_creation_time: "2999-12-31T23:59:59Z"
-teapot_admission_controller_ndots: "2"
 
 {{if eq .Environment "e2e"}}
 teapot_admission_controller_ignore_namespaces: "^kube-system|(e2e-tests-(downward-api|kubectl|projected|statefulset|pod-network)-.*)$"
+teapot_admission_controller_ndots: "5"
```
mikkeloscar: Note to self: document on which features we are no longer conforming to upstream Kubernetes in our docs :)

test/e2e/run_e2e.sh
```diff
@@ -116,7 +116,7 @@ export KUBECONFIG="$KUBECONFIG"
 ginkgo -nodes=25 -flakeAttempts=2 \
     -focus="(\[Conformance\]|\[StatefulSetBasic\]|\[Feature:StatefulSet\]\s\[Slow\].*mysql|\[Zalando\])" \
     -skip="(\[Serial\])" \
-    -skip="(should.support.retrieving.logs.from.the.container.over.websockets|should.support.remote.command.execution.over.websockets|\[Serial\])" \
+    -skip="(should.provide.DNS.for.the.cluster|should.provide.DNS.for.services|should.support.retrieving.logs.from.the.container.over.websockets|should.support.remote.command.execution.over.websockets|\[Serial\])" \
```
mikkeloscar, line 127: Let's add a comment above why we disable these two tests

njuettner:
👍

mikkeloscar:
:+1:

PR #2160, opened 2019-05-23
Add ubuntu based node pool

linki:
This is an initial attempt to add a default ubuntu-based worker node pool called `default-ubuntu-worker`. It's supposed to make use of the changes in https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/142.

Note that it's a minimum viable version and comments are appreciated. Goal of this PR is to integrate early and test changes quickly with the help of CLM etc.

Here are a couple of notes:
* separated into its own node pool profile to not mess with exiting node pool configs so that we can merge this earlier.
  * the CF stacks of CoreOS and Ubuntu are practicly identical. However, no symlink was used to allow them to evolve independently
* The config item for the used AMI was generalised into ~~`worker_ami`~~`kuberuntu_image` ~~to better describle what it actually does and~~ to allow CoreOS pools to co-exist. If we intend to have node pools with different AMIs (such as for master/worker or testing different Kuberuntu AMIs in a single cluster) this could move into the node pool config items.
* ~~No `split-az` config to keep it simple for now~~ ✅ 
* The Ubuntu AMI config item will change more frequently than the CoreOS AMI (especially now during the testing phase), so setting it hard like this might be too tedious.

We could switch teapot to this channel to try out the new node pool profiles there first. ✅

### DIFF
--- a/cluster/node-pools/master-ubuntu-default/userdata.yaml
+++ b/cluster/node-pools/master-ubuntu-default/userdata.yaml
@@ -7,7 +7,7 @@ write_files:
   - owner: root:root
     path: /etc/etcd-member/environment
     content: |
-      ETCD_ENDPOINT=etcd-server.etcd.teapot.zalan.do:2379
+      ETCD_ENDPOINT={{ .Cluster.ConfigItems.etcd_endpoints }}
 
   - owner: root:root
     path: /etc/kubernetes/secrets.env
--- a/cluster/node-pools/master-ubuntu-default/userdata.yaml
+++ b/cluster/node-pools/master-ubuntu-default/userdata.yaml
@@ -13,8 +13,8 @@ write_files:
     path: /etc/kubernetes/secrets.env
     content: |
       NODEPOOL_TAINTS=node-role.kubernetes.io/master=:NoSchedule{{if index .NodePool.ConfigItems "taints"}},{{.NodePool.ConfigItems.taints}}{{end}}
+      NODE_LABELS={{ .Values.node_labels }}{{if index .NodePool.ConfigItems "labels"}},{{.NodePool.ConfigItems.labels}}{{end}}
       NODEPOOL_NAME={{ .NodePool.Name }}
-      NODEPOOL_LABEL_STATIC=kuberuntu-master
       KUBELET_ROLE=master
 
   - owner: root:root
--- a/cluster/node-pools/master-ubuntu-default/userdata.yaml
+++ b/cluster/node-pools/master-ubuntu-default/userdata.yaml
@@ -13,7 +13,7 @@ write_files:
     path: /etc/kubernetes/secrets.env
     content: |
       NODEPOOL_TAINTS=node-role.kubernetes.io/master=:NoSchedule{{if index .NodePool.ConfigItems "taints"}},{{.NodePool.ConfigItems.taints}}{{end}}
-      NODE_LABELS={{ .Values.node_labels }}{{if index .NodePool.ConfigItems "labels"}},{{.NodePool.ConfigItems.labels}}{{end}}
+      NODE_LABELS=node-role.kubernetes.io/master,kubernetes.io/role=master,master=true,cluster-lifecycle-controller.zalan.do/decommission-priority=999,cluster-lifecycle-controller.zalan.do/replacement-strategy=ensure-minimum-healthy-nodes,{{ .Values.node_labels }}{{if index .NodePool.ConfigItems "labels"}},{{.NodePool.ConfigItems.labels}}{{end}}
       NODEPOOL_NAME={{ .NodePool.Name }}
       KUBELET_ROLE=master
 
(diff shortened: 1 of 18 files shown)


cluster/node-pools/master-ubuntu-default/userdata.yaml
```diff
@@ -0,0 +1,571 @@
+#cloud-config
+runcmd:
+  - [ systemctl, start, gen-controller-manager-config.service ]
+  - [ systemctl, start, etcd-member.service ]
+
+write_files:
+  - owner: root:root
+    path: /etc/etcd-member/environment
+    content: |
+      ETCD_ENDPOINT=etcd-server.etcd.teapot.zalan.do:2379
```
linki: That's the wrong etcd endpoint. Seems like the dynamic config from [here](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/node-pools/master-default/userdata.clc.yaml#L75) got lost in the change.

cluster/node-pools/master-ubuntu-default/userdata.yaml
```diff
@@ -0,0 +1,571 @@
...
+      clusters:
+      - name: local
+        cluster:
+          server: http://127.0.0.1:8080
+      users:
+      - name: kubelet
+      contexts:
+      - context:
+          cluster: local
+          user: kubelet
```
linki, line 39: Can this file be part of the ami?

mikkeloscar: Has to be different on master and worker nodes

cluster/node-pools/master-ubuntu-default/userdata.yaml
```diff
@@ -0,0 +1,571 @@
...
+
+  - owner: root:root
+    path: /etc/kubernetes/config/image-policy-webhook.yaml
+    content: |
+      imagePolicy:
+        kubeConfigFile: /etc/kubernetes/config/image-policy-webhook-kubeconfig.yaml
+        allowTTL: 2
+        denyTTL: 2
+        retryBackoff: 500
+        defaultAllow: false
```
linki, line 434: Can/should this be moved to the image?

mikkeloscar: I would really like to keep the config that we are most likely to touch to be in this repository otherwise I'm afraid we create a weird split of configuration between the AMI repository and kubernetes-on-aws which can cause confusion and make changes harder to perform (changes in two places instead of one).

For me a logical split would be to say, systemd units have to be on the AMI to achieve the desired booting sequence, but anything else config related that is not generic between master and worker should be in "userdata" in `kubenertes-on-aws`. WDYT?

cluster/node-pools/worker-default/userdata.clc.yaml
```diff
@@ -214,6 +214,7 @@ systemd:
       --node-labels=${REPLACEMENT_STRATEGY_LABEL} \
 {{- end }}
       --node-labels={{ .Values.node_labels }} \
+      --node-labels=beta.kubernetes.io/distro=coreos \
```
mikkeloscar: Just a note: Depending on when we merge this we may want to avoid rolling the old nodes at the same time. Could be solved by using `NotIn` on the daemonset.

cluster/node-pools/master-ubuntu-default/userdata.yaml
linki: I agree. By baking our own AMI we opened a can of worms in that regard. We should define and agree upon what kind of config goes into the AMI vs. Userdata.

cluster/node-pools/worker-default/userdata.clc.yaml
linki: Done

cluster/node-pools/master-ubuntu-default/userdata.yaml
linki: Let's leave it then

test/e2e/run_e2e.sh
```diff
@@ -30,6 +32,17 @@ export ETCD_ENDPOINTS="$ETCD_ENDPOINTS"
...
+if [[ "${E2E_NODE_OS}" == "coreos" ]]; then
+    export MASTER_PROFILE="master"
+    export WORKER_PROFILE="worker"
+elif [[ "${E2E_NODE_OS}" == "ubuntu" ]]; then
+    export MASTER_PROFILE="master-ubuntu"
+    export WORKER_PROFILE="worker-ubuntu"
+else
+    echo "Unsupported E2E_NODE_OS: ${E2E_NODE_OS}"
+    exit 1
+fi
```
linki, line 44: Just as an idea: The cluster YAML for e2e tests above is already a list of clusters (but contains only one cluster at the moment). Rather than templating the config like this we could also add a new entry to the list to say "please run the e2e tests against the following clusters".

linki:
lgtm

test/e2e/run_e2e.sh
arjunrn: I thought about that but this way we can see each cluster as it's own CDP step.

arjunrn:
:+1:

linki:
👍

PR #2153, opened 2019-05-22
Kubelet: switch to config instead of arguments, use the new certificate [1/3]

aermakov-zalando:
This mostly unreverts #2134. Authentication is still not enabled, though, it'll follow up in a separate PR.

PR #2167, opened 2019-05-24
Kubelet: use authenticated API by default [2/3]

aermakov-zalando:
Update API server & metrics-server to use the secure Kubelet endpoint.

**NOTE**: do not merge until #2153 is fully rolled out.

### DIFF
--- a/cluster/manifests/metrics-server/deployment.yaml
+++ b/cluster/manifests/metrics-server/deployment.yaml
@@ -27,12 +27,6 @@ spec:
       containers:
       - name: metrics-server
         image: registry.opensource.zalan.do/teapot/metrics-server:v0.3.2
-        args:
-        # Connect to kubelet on 'completely insecure' port.
-        # We need to configure kubelet differently to be able to use the secure
-        # port 10250.
-        - --deprecated-kubelet-completely-insecure
-        - --kubelet-port=10255
         resources:
           limits:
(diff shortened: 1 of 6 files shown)


njuettner:
Suggestion to disable it on kubelet is to set it to `-read-only-port 0` or is there anything else which needs this readonly port to be open?

mikkeloscar:
> Suggestion to disable it on kubelet is to set it to -read-only-port 0 or is there anything else which needs this readonly port to be open?

We just need to do it in two steps. We disable it in #2170 : https://github.com/zalando-incubator/kubernetes-on-aws/pull/2170/files#diff-8f1eda593049fa31a03c0238a0135008L293

mikkeloscar:
@aermakov-zalando I updated prometheus to use the secure kubelet port and also changed the apiserver scraping to use 443 instead of the readonly nginx address.

szuecs:
@mikkeloscar do we have zmon checks that go directly through the readonly nginx ?

mikkeloscar:
:+1:

linki:
👍

aermakov-zalando:
@mikkeloscar Thanks!

mikkeloscar:
@szuecs yes, I think we still have some ZMON checks using the nginx endpoint. @aermakov-zalando and I talked about getting rid of the nginx proxy, but we didn't really look into how much depend on it yet. But at least prometheus no longer depends on it :)

PR #2168, opened 2019-05-24
Get rid of nat_cidr_blocks

aermakov-zalando:
We shouldn't have to set it manually, computers can do that!


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2172, opened 2019-05-27
Make it possible to enable CustomResourceWebhookConversion

mikkeloscar:
Makes it possible to enable `CustomResourceWebhookConversion` which some users wants to test in their clusters.

https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#webhook-conversion


mikkeloscar:
:+1:

njuettner:
👍

PR #2174, opened 2019-05-27
skipper default resource adjustments and cpu hpa

szuecs:
With introduction of the lifo queue, we want to be able to absorb trafficon failing backends that can harm other routes.
In this case we need to have more memory and cpu resources.
At the same time it makes sense to scale up via hpa faster and reduce the cpu scaling to 90%


szuecs:
:+1:

muaazsaleem:
👍

PR #2183, opened 2019-05-31
fix ALB 5xx caused by same idle timeouts ALB vs skipper

szuecs:
fix ALB 5xx caused by same server idle timeouts of skipper to client idle timeouts from alb


njuettner:
👍

szuecs:
:+1:

PR #2188, opened 2019-06-03
Label nodes with their linux distro

linki:
adds a distro node label and limits gpu device installer to onyl run on coreos instances. /cc @njuettner

### DIFF
--- a/cluster/manifests/nvidia/nvidia-driver-installer.yaml
+++ b/cluster/manifests/nvidia/nvidia-driver-installer.yaml
@@ -34,6 +34,10 @@ spec:
             - matchExpressions:
               - key: zalando.org/nvidia-gpu
                 operator: Exists
+              - key: beta.kubernetes.io/distro
+                operator: In
+                values:
+                - coreos
       priorityClassName: system-node-critical
       hostNetwork: true
       hostPID: true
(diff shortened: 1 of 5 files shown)


linki:
👍

njuettner:
👍

cluster/manifests/nvidia/nvidia-driver-installer.yaml
```diff
@@ -34,6 +34,10 @@ spec:
             - matchExpressions:
               - key: zalando.org/nvidia-gpu
                 operator: Exists
+              - key: beta.kubernetes.io/distro
```
mikkeloscar, line 37: Let's pick a "valid" node label which the kubelet will not complain about.

Logs from kubelet:

```
in 1.15, --node-labels in the 'kubernetes.io' namespace must begin with an allowed prefix (kubelet.kubernetes.io, node.kubernetes.io) or be in the specifically allowed set (beta.kubernetes.io/arch, beta.kubernetes.io/instance-type, beta.kubernetes.io/os, failure-domain.beta.kubernetes.io/region, failure-domain.beta.kubernetes.io/zone, failure-domain.kubernetes.io/region, failure-domain.kubernetes.io/zone, kubernetes.io/arch, kubernetes.io/hostname, kubernetes.io/instance-type, kubernetes.io/os)
```

There is `kubernetes.io/os` but I'm not 100% sure it's meant to be used for this. Otherwise we should use one of the valid prefixes: `kubelet.kubernetes.io, node.kubernetes.io`

linki: I'll change it

aermakov-zalando: The list of prefixes is a lot bigger (https://github.com/kubernetes/kubernetes/pull/68267). If we can find anything that makes sense, we could just migrate our internal labels to something like `zalando.node.kubernetes.io/...`.

linki: I've put it under `node.kubernetes.io/distro`. wdyt?

mikkeloscar: fine by me

PR #2193, opened 2019-06-04
Run etcd-proxy as apiserver sidecar

linki:
Refactors the `etcd-member` systemd unit into an `apiserver` sidecar.

Currently `etcd-proxy` runs co-located with the API server Pods managed via a host-level `systemd` unit and `rkt`. This brought some challenges in our new Ubuntu setup, mainly how to retrieve the binary from a remote source which was previously done through rkt and docker image discovery.

This PR avoids using a systemd unit and runs `etcd-proxy` as a docker container in a sidecar ultimately managed by the kubelet. We lose the ability to use systemd to express dependencies between the etcd-member unit and other units, but it looks like only `node-drainer` made use of it and it doesn't seem to be required.

The approach is inspired by self-hosted Kubernetes: `kubeadm` and `kind` also run etcd as Pods on the master nodes. Since our etcd-Pod is just a stateless proxy a sidecar should suffice.


cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -558,6 +539,18 @@ storage:
             volumeMounts:
             - name: config-volume
               mountPath: /etc/nginx
+          - name: etcd-proxy
+            image: registry.opensource.zalan.do/teapot/etcd-proxy:master-2
+            - {{ .Cluster.ConfigItems.etcd_endpoints }}
```
aermakov-zalando, line 605: `arguments:` seems to be missing

cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -558,6 +539,18 @@ storage:
...
+            ports:
+            - containerPort: 2379
+            resources:
+              requests:
+                cpu: 25m
+                memory: 25Mi
+              limits:
```
aermakov-zalando: I would not have any explicit limits here. We don't particularly care about them on the master nodes, and I we don't want to debug OOMs.

cluster/node-pools/master-default/userdata.clc.yaml
linki: 🤦‍♂

linki:
I made the use of the sidecar approach configurable via config items. The default is our current way via the systemd unit.

szuecs:
@linki how do you access from masters the etcd if you need to do it manually?
Does etcdctl just works or will this change?
I don't really see the point of having now a sidecar approach, because it seems less reliable, because you create an unresolved dependency, which we now have to hope it's "fast enough".

aermakov-zalando:
Sidecar seems fine to me, no other components should need it. If we really need to do something directly on etcd nodes, we could either connect directly (without going to the proxy), or ssh into one of the etcd nodes.

aermakov-zalando:
I also don't see a lot of value in making this conditional. If the sidecar approach works and is easier for us, just go for it.

szuecs:
I agree with not having it conditional, but I disagree with not having a dependency. We can easily have apiservers which have not yet the sidecar started to connect to etcd, how do we solve this?
We already had dependency issues in the past and we can now solve it, why not solve it now?

aermakov-zalando:
The apiservers go to localhost, so they wouldn't be able to connect until the sidecar starts. It's exactly the same as with the current approach.

mikkeloscar:
Since we run with hostNetwork `etcdctl` on the node should still work fine (assuming apiserver pod is up). I see the etcd-proxy as a dependency for the apiserver, similar to how it's used as a dependency for flannel, and therefore it makes sense to me to have it as a sidecar.

linki:
Thanks for looking at this PR and your comments.

`etcdctl` on the host works as before because the `etcd-proxy` container runs with `hostNetwork`.

**Regarding the seemingly reduced reliability**
`systemd` might be more reliable when it comes to managing the `etcd-proxy` process than `kubelet`. However, we already trust `kubelet` with running, watching and restarting the `apiserver` process and all of its sidecars. Therefore, I would say the `etcd-proxy` container is as reliable as the `apiserver` itself.

**Regarding the etcd-apiserver race condition**
`apiserver` indeed crashes when `etcd-proxy` isn't running. It gets restarted until `etcd-proxy` is running. If `etcd-proxy` crashes later `apiserver` keeps running but reports unhealthy and a failed `etcd` under `/healthz`. From the outside the `apiserver` is just down (EOF).

========================================================================

I would highly prefer the ConfigItem option. Rolling it out behind a feature flag allows to merge this without an acutal change to running clusters, avoiding long-lived open PRs. Once rolled out it's easier to enable/disable this for selected clusters. Once an option (old or new) becomes the de-facto value the switch should be removed.

I know that this change isn't necessary but looking at it with fresh eyes the approach seems natural. We use this pattern in [other components](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/flannel/daemonset.yaml#L46-L59). It would allow to avoid fetching a binary [through docker cp](https://github.bus.zalan.do/teapot/kubernetes-on-ubuntu/blob/master/scripts/provision.sh#L4-L12) in our Ubuntu AMI.

**General Question**
Apart from the reliability questions which should have a high weight, in general, do you agree on the goal of pushing as much manifests into the Kubernetes layer when possible? I think having less scripts and binaries in the AMI and more containers in Kubernetes-land is a good thing.

*For instance, if `node-drainer` could be a DaemonSet would you still prefer it on the host like now?*

This approach could directly counter other goals, like a fast startup time for worker nodes, so we should clarify that first.

szuecs:
For the first part I am fine, because it works similar to a dependency. 
The second part depends IMO on the deployment model. Is the pod part of a kubernetes version?
My opinion: If yes would be an argument to put it into the image, if no daemonset is the way to go.

linki:
@szuecs Thanks for your opinion.

aermakov-zalando:
👍

szuecs:
@linki should we merge this, now?

linki:
Not yet, I forgot the Ubuntu pool, which actually triggered this :laughing:

linki:
@arjunrn I added the feature to the ubuntu pool as well.

I'm going to run e2e with enabled proxy now to see if that works. Afterwards I switch it back to `false` by default.

@arjunrn When the sidecar is enabled the `etcd-member` is not started. I don't see any hard dependency in the Ubuntu image so in theory this should work with the current AMIs already.

mikkeloscar:
I think the feature flag should be enabled by default and only have it as a failsafe to roll back?

linki:
@mikkeloscar Then I leave the default on `etcd_proxy_as_sidecar: true`. (Had to change the commit msg though).

Tests were green so it's +1 from me.

linki:
:+1:

szuecs:
:+1:

PR #2197, opened 2019-06-05
Enable application label validation by default

aermakov-zalando:
This should be enabled in new clusters as well.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2200, opened 2019-06-05
Add RBAC for Zalando IAM users [1/2]

mikkeloscar:
Replace the static mapping of Zalando IAM user to group with finer grained permissions possible with RBAC. I.e. instead of giving `credentials-provider` `Administrator` access, just give it access to Platformcredentialssets and secrets.

## Status

* [x] CLM
* [x] Credentials Provider
* [x] Reduce CDP Controller permissions from `PowerUser` to what is actually needed.
* [x] `deployment-discovery`
* [x] `stups_zmon-zmon`
* [x] `stups_kube-resource-report`
* [x] `stups_oxygen`

This change is 1/2. First we need to roll out all the RBAC roles/bindings, and then we need to change the authnz webhook to not set groups manually. I'll make another PR with the second step.

### DIFF
--- a/cluster/manifests/roles/cdp-controller-rbac.yaml
+++ b/cluster/manifests/roles/cdp-controller-rbac.yaml
@@ -1,0 +1,49 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cdp-controller
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  - namespaces
+  - serviceaccounts
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
+- apiGroups:
+  - ""
+  resources:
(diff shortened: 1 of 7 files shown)


tkrop:
@mikkeloscar *deployment-discovery* currently only accesses the path `/api/v1/pods`. However, we have an issue pending to evaluate the service URLs. In this case we would need in addition to `pods` the access to `endpoints`, `services`, and `ingresses`. I'm not sure whether we require `get` and `list` access on all endpoints whether `get` is enough. Please enter provide us all permission for all above resources.

mikkeloscar:
@tkrop Thanks I added the role!

mikkeloscar:
:-1:

mikkeloscar:
Seems to be an issue for CDP-controller to read pod logs

mikkeloscar:
:+1:

szuecs:
:+1:

PR #2201, opened 2019-06-06
Symlink equal cloudformation stack definitions

linki:
CloudFormation Stack descriptions between CoreOS- and Ubuntu-based node pools are almost identical. They only differ in the config item used to detect the correct AMI. By making the config item conditional inside the Stack definition itself we can share the very same file for both pool types.

The underlying problem is that the configuration for the used AMI is a global config item and therefore cluster-wide. With node pools using different AMIs this would be more correctly defined as a NodePool config item. However, we currently miss support for NodePool config item defaults, hence this approach.


linki:
This PR avoids any divergence between the CloudFormation Stacks for the CoreOS and Ubuntu pools.

Though, I'm not sure if the added complexity is worth it. Let me know what you prefer.

mikkeloscar:
IMO it's fine like this. There's no sub string search function in the templates? :)

aermakov-zalando:
CLM templating is very easy to extend, so I'd just expose `strings.Contains` to the template code. However, since this is transient, we could also just keep it like this.

linki:
We could also add `sprig` to get a bunch of useful string functions: https://masterminds.github.io/sprig/strings.html

mikkeloscar:
:+1:

linki:
👍

PR #2202, opened 2019-06-06
RBAC zalando iam 2 [2/2]

mikkeloscar:
Follow up to #2200 (Needs to be rebased after it has been merged)

Updates the Authnz webhook to provide correctly formatted user names for the RBAC role bindings.

### DIFF
--- a/cluster/manifests/roles/cockpit-skipper-rbac.yaml
+++ b/cluster/manifests/roles/cockpit-skipper-rbac.yaml
@@ -1,0 +1,54 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cockpit-skipper
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  - namespaces
+  - deployments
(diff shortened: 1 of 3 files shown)


cluster/manifests/roles/cockpit-skipper-rbac.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
mikkeloscar, line 1: @tkrop We need to roll this out separately to ensure that we don't drop access to `cockpit-skipper` during the rollout.

tkrop: They currently do not even have access. This request was issued to reflect the roll out #2252. So if something breaks we do not care here ;).

mikkeloscar: ok

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2211, opened 2019-06-07
Keep files.yaml out of non-ubuntu based node pools

linki:
It's not used in CoreOS pools and messes with the tool that determines whether a pool needs to be updated.

/cc @aermakov-zalando


aermakov-zalando:
Thanks!

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2219, opened 2019-06-12
Introduce kube-aws-iam-controller

mikkeloscar:
This introduces [kube-aws-iam-controller](https://github.com/mikkeloscar/kube-aws-iam-controller) along side kube2iam and changes some of our application to use that instead of kube2iam.

## TODO

* [x] kube-node-ready
* [x] kube-static-egress-controller
* [x] emergency-access-service
* [x] ingress-controller
* [ ] kube-cluster-autoscaler (waiting for latest version with updated `aws-sdk-go`)
* [ ] logging-agent (ruby, support not validated yet)
* [x] zmon-aws-agent
* [x] etcd-backup
* [ ] external-dns (waiting for next release)
* [x] kube-metrics-adapter

### Only needed for "gerry"
* [ ] zmon-worker
* [ ] zmon-scheduler
* [ ] zmon-agent

### others

* [x] Add e2e test for testing the controller. Both the positive test and the negative test.


mikkeloscar:
Added an e2e test which assigns an IAM role to a pod that adds a file to an s3 bucket.

I'm not sure how to actually write a "negative" e2e test that makes sense. If the IAM role is not provisioned then the pod would just be pending, which we could try to test for, but does it make sense? If the role is provisioned but doesn't give permission to the right thing, then we are testing that our IAM role is configured correctly instead, which doesn't make so much sense for kube-aws-iam-controller.

mikkeloscar:
Added the negative test which validates that a pod not requesting an IAM role doesn't get the worker IAM role.

mikkeloscar:
:+1:

mikkeloscar:
The components not migrated to `kube-aws-iam-controller` will be migrated in another PR. For now we can use it to validate both `kube2iam` and `kube-aws-iam-controller`.

aermakov-zalando:
👍

PR #2220, opened 2019-06-12
Update CLC

aermakov-zalando:
* Correctly handle pools with non-zero min. size and no prescaling
 * Don't terminate unowned pods (same behaviour as CA)


aermakov-zalando:
👍

linki:
👍

PR #2221, opened 2019-06-12
Route external requests to API server via skipper proxy sidecar

linki:
This adds a sidecar to the API server Pod that becomes the primary entrypoint for cluster-external requests. This way all requests that come from the ELB will be run through skipper for some sort of processing.

The provided skipper configuration just logs all external traffic that's not from Zalando users but doesn't change anything.

The feature can be enabled/disabled via a config item called: `apiserver_proxy`

There's a race condition between changing the CloudFormation related resources (ELB target port and security groups) and actually rolling the nodes until there's something listening on that port. Therefore, the skipper sidecar is deployed all the time and stays unused. Once the config item is flipped ELB and FW rules are updated and requests will find an already running sidecar on the nodes. Flipping the config item back works as well: in this case the sidecar keeps running but becomes unused again.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -54,7 +54,7 @@ Resources:
         Timeout: '5'
         UnhealthyThreshold: '2'
       Listeners:
-        - InstancePort: 443
+        - InstancePort: {{ if eq .ConfigItems.apiserver_proxy "true" }}8443{{ else }}443{{ end }}
           InstanceProtocol: SSL
           LoadBalancerPort: 443
           PolicyNames: []
@@ -182,14 +182,14 @@ Resources:
     Type: 'AWS::EC2::SecurityGroupIngress'
   MasterSecurityGroupIngressFromLoadBalancer:
     Properties:
-      FromPort: 443
+      FromPort: {{ if eq .ConfigItems.apiserver_proxy "true" }}8443{{ else }}443{{ end }}
       GroupId: !Ref MasterSecurityGroup
       IpProtocol: tcp
       SourceSecurityGroupId: !Ref MasterLoadBalancerSecurityGroup
       Tags:
         - Key: 'kubernetes.io/cluster/{{.Cluster.ID}}'
           Value: owned
-      ToPort: 443
+      ToPort: {{ if eq .ConfigItems.apiserver_proxy "true" }}8443{{ else }}443{{ end }}
     Type: 'AWS::EC2::SecurityGroupIngress'
   MasterSecurityGroupIngressFromLoadBalancerHealthCheck:
     Properties:
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -224,17 +224,6 @@ Resources:
           Value: owned
       ToPort: 443
     Type: 'AWS::EC2::SecurityGroupIngress'
-  MasterSecurityGroupIngressFromWorkerToMasterKubeletReadOnly:
-    Properties:
-      FromPort: 10255
-      GroupId: !Ref MasterSecurityGroup
-      IpProtocol: tcp
-      SourceSecurityGroupId: !Ref WorkerSecurityGroup
-      Tags:
-        - Key: 'kubernetes.io/cluster/{{.Cluster.ID}}'
-          Value: owned
-      ToPort: 10255
-    Type: 'AWS::EC2::SecurityGroupIngress'
   MasterSecurityGroupIngressFromWorkerToMasterKubelet:
     Properties:
       FromPort: 10250
@@ -341,10 +330,6 @@ Resources:
           IpProtocol: tcp
           ToPort: 9153
         - CidrIp: "{{.Values.vpc_ipv4_cidr}}"
-          FromPort: 10255
-          IpProtocol: tcp
-          ToPort: 10255
-        - CidrIp: "{{.Values.vpc_ipv4_cidr}}"
           FromPort: 10250
           IpProtocol: tcp
           ToPort: 10250
@@ -408,17 +393,6 @@ Resources:
         - Key: 'kubernetes.io/cluster/{{.Cluster.ID}}'
           Value: owned
       ToPort: 8472
-    Type: 'AWS::EC2::SecurityGroupIngress'
-  WorkerSecurityGroupIngressFromWorkerToWorkerKubeletReadOnly:
-    Properties:
-      FromPort: 10255
-      GroupId: !Ref WorkerSecurityGroup
-      IpProtocol: tcp
-      SourceSecurityGroupId: !Ref WorkerSecurityGroup
-      Tags:
-        - Key: 'kubernetes.io/cluster/{{.Cluster.ID}}'
-          Value: owned
-      ToPort: 10255
     Type: 'AWS::EC2::SecurityGroupIngress'
   WorkerSecurityGroupIngressFromWorkerToWorkerKubelet:
     Properties:
@@ -1291,7 +1265,43 @@ Resources:
       Description: "AWS service role for application autoscaling DynamoDB"
     Type: "AWS::IAM::ServiceLinkedRole"
 {{- end }}
-
+  RemoteFilesEncryptionKey:
+    Type: "AWS::KMS::Key"
+    Properties:
+      Description: The master key for encryption of remote files
+      EnableKeyRotation: true
+      KeyPolicy:
+        Version: "2012-10-17"
+        Id: "remote-files-encryption-key"
+        Statement:
+          - Sid: "Enable IAM User Permissions"
+            Effect: "Allow"
+            Principal:
+              AWS: "arn:aws:iam::{{ .Cluster.InfrastructureAccount | getAWSAccountID }}:root"
+            Action: "kms:*"
+            Resource: "*"
+          - Sid: "Allow encryption with the key"
+            Effect: "Allow"
+            Principal:
+              AWS: "arn:aws:iam::{{.Cluster.InfrastructureAccount | getAWSAccountID}}:role/cluster-lifecycle-manager-entrypoint"
+            Action:
+            - "kms:Describe"
+            - "kms:Get*"
+            - "kms:Encrypt"
+            Resource: "*"
+          - Sid: "Enable master and worker nodes to decrypt the remote files"
+            Effect: "Allow"
+            Principal:
+              AWS:
+                - "Fn::GetAtt":
+                  - "MasterIAMRole"
+                  - "Arn"
+                - "Fn::GetAtt":
+                  - "WorkerIAMRole"
+                  - "Arn"
+            Resource: "*"
+            Action:
+            - "kms:Decrypt"
 Outputs:
   MasterIAMRole:
     Export:
@@ -1313,3 +1323,7 @@ Outputs:
     Export:
       Name: '{{.Cluster.ID}}:worker-security-group'
     Value: !Ref WorkerSecurityGroup
+  RemoteFilesEncryptionKey:
+    Export:
+      Name: '{{ .Cluster.ID}}:remote-files-encryption-key'
+    Value: !Ref RemoteFilesEncryptionKey
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -204,7 +204,7 @@ cluster_dns: "coredns"
 coredns_log_svc_names: "true"
 
 coreos_image: "ami-06142be4afc3d5b1b" # Container Linux 2079.4.0 (HVM, eu-central-1)
-kuberuntu_image: "ami-06c4e377669c3cb33" # Kuberuntu (dev) (HVM, eu-central-1)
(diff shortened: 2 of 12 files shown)


cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -558,6 +558,35 @@ storage:
...
+            readinessProbe:
+              httpGet:
+                path: /metrics
+                port: 9911
+              initialDelaySeconds: 60
+              timeoutSeconds: 5
+            resources:
+              requests:
+                cpu: 100m
+                memory: 250Mi
```
linki, line 585: @szuecs Can you recommend a suitable resource request here?

cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -558,6 +558,35 @@ storage:
...
+            - -tls-cert=/etc/kubernetes/ssl/apiserver.pem
+            - -tls-key=/etc/kubernetes/ssl/apiserver-key.pem
+            - -inline-routes
+            - 'z: JWTPayloadAnyKV("iss", "https://identity.zalando.com") -> disableAccessLog() -> "https://127.0.0.1:443"; all: * -> enableAccessLog() -> "https://127.0.0.1:443";'
+            ports:
+            - containerPort: 8443
+            readinessProbe:
+              httpGet:
+                path: /metrics
+                port: 9911
```
linki: @szuecs Do you know of a better endpoint to use for readinessProbe?

cluster/config-defaults.yaml
```diff
@@ -236,3 +236,6 @@ custom_resource_webhook_conversion: "false"
 node_max_pods: "110" # Default: 110
 # flag passed to controller-manager
 node_cidr_mask_size: "24" # Default: 24
+
+# when set to true, routes external traffic to the apiserver through a skipper sidecar
+apiserver_proxy: "false"
```
mikkeloscar, line 241: I would suggest enabling this by default. We can start by opting out in all clusters by adding `apiserver_proxy: "false"` to all existing clusters.

test/e2e/cluster_config.sh
```diff
@@ -49,6 +49,7 @@ clusters:
     enable_rbac: "true"
     dynamodb_service_link_enabled: "false"
     skipper_ingress_cpu: 100m
+    apiserver_proxy: "true"
```
linki: TODO: This should be turned off for the final merge.

cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -558,6 +558,36 @@ storage:
...
             - name: config-volume
               mountPath: /etc/nginx
+          - name: skipper-proxy
+            image: registry.opensource.zalan.do/pathfinder/skipper:v0.10.233
+            args:
+            - skipper
+            - -address=:8443
+            - -tls-cert=/etc/kubernetes/ssl/apiserver.pem
+            - -tls-key=/etc/kubernetes/ssl/apiserver-key.pem
+            - -insecure
```
linki, line 567: @szuecs @mikkeloscar Do you mind having `-insecure` here?

Otherwise we'd need to pass the `ca` to skipper. I tried with comma-separating `-tls-cert` (like the docs say) but skipper failed. Should that work? I guess that's the wrong flag?

cluster/node-pools/master-default/userdata.clc.yaml
aryszka: how about injecting a route for the readiness? This would have the semantic meaning of "routing is successfully configured ergo ready to serve".

cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -558,6 +558,36 @@ storage:
...
+            - -inline-routes
+            - 'z: JWTPayloadAnyKV("iss", "https://identity.zalando.com") -> disableAccessLog() -> "https://127.0.0.1:443"; all: * -> enableAccessLog() -> "https://127.0.0.1:443";'
```
linki: We might need `-experimental-upgrade` for `exec` to work. @szuecs WDYT?

cluster/cluster.yaml
```diff
@@ -54,7 +54,7 @@ Resources:
         Timeout: '5'
         UnhealthyThreshold: '2'
       Listeners:
-        - InstancePort: 443
+        - InstancePort: {{ if eq .ConfigItems.apiserver_proxy "true" }}8443{{ else }}443{{ end }}
```
aermakov-zalando, line 57: This will cause the API servers to be unavailable for the whole duration of the rollout because the port will be changed, but the existing nodes don't run Skipper.

cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -558,6 +558,36 @@ storage:
...
+            ports:
+            - containerPort: 8443
+            readinessProbe:
+              httpGet:
+                path: /metrics
+                port: 9911
+              initialDelaySeconds: 60
```
aermakov-zalando: That's quite a lot. Why a minute?

cluster/cluster.yaml
linki: That's why I didn't guard the sidecar itself with this condition. This way going from `apiserver_proxy: false` to `apiserver_proxy: true` during a migration shouldn't have this problem, does it?

cluster/node-pools/master-default/userdata.clc.yaml
linki: Taken from [how we run skipper-ingress](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/skipper/deployment.yaml#L114). But I'm open for suggestions.

mikkeloscar: I would think it's even too much for skipper ingress, but there we need skipper to connect to the apiserver and configure all the routes. Here we don't need that, so should be much faster.

cluster/node-pools/master-default/userdata.clc.yaml
mikkeloscar: Is it complicated to pass the ca? It's already on the node. :)

cluster/node-pools/master-default/userdata.clc.yaml
szuecs: SGTM, we have to learn anyways from that

cluster/node-pools/master-default/userdata.clc.yaml
szuecs: readinessProbe should be pass through, such that the pod is ready to serve traffic. Skipper alone might not be what we want.
So create a route through skipper that will hit the former readinessProbe of the apiserver and use this in the readinessProbe. Like this you know that the path through the proxy is ready to server apiserver traffic.

cluster/node-pools/master-default/userdata.clc.yaml
aermakov-zalando: The 60 seconds in skipper-ingress is a hack to slow down the rotation of Skippers during upgrade IIRC. The readiness probe doesn't even do much here, since the ELB will contact Skipper anyway, so we could safely reduce it.

cluster/cluster.yaml
aermakov-zalando: Ah, I've missed that, sorry! Then the only potential concern is accidentally enabling this before the PR is rolled out, but we could just not do that.

cluster/node-pools/master-default/userdata.clc.yaml
szuecs: You probably can also cat ca.pem and apiserver.pem into a apiserver-combined.pem, which is very common setup in webservers. On the other hand localhost is also fine to ignore, because if you can wiretap localhost you most likely can also read the apiserver-key.pem....
If we need to change the CA, and the workaround above does not work we need a little PR to skipper to be able to replace the RootCAs.

cluster/node-pools/master-default/userdata.clc.yaml
szuecs: yes, for: exec, attach, logs, port-forward

cluster/cluster.yaml
szuecs: IIRC clm, will first update CF stacks and then start the rolling master nodes, so this needs 2 steps, not?

cluster/node-pools/master-default/userdata.clc.yaml
szuecs: Skipper starts in `<5` seconds if you load from routes file in inline routes. You do not need to set initialDelaySeconds.

cluster/config-defaults.yaml
linki: Introducing this at the same time as enabling it won't work and e2e will fail. I suggest we keep it like that and switch the default in a separate PR. This way we can use e2e to test the update paths as well.

cluster/node-pools/master-default/userdata.clc.yaml
linki: Alright, then I suggest we keep `insecure`.

cluster/node-pools/master-default/userdata.clc.yaml
linki: @szuecs thanks, would you recommend any other flags?

szuecs: -runtime-metrics
-enable-connection-metrics
-enable-prometheus-metrics

You should also test if port-forwarding and exec works as expected and how long the connection is working. Maybe we should adjust some timeouts and idle connection handling.

cluster/node-pools/master-default/userdata.clc.yaml
linki: I used a new route `/kube-system/healthz` that points to the `/healthz` of the API server. I didn't want to use the real `/healthz` because it needs to be forwarded to the insecure endpoint  (readiness prober doesn't have credentials) and that could leak tokens.

cluster/node-pools/master-default/userdata.clc.yaml
linki: Added, thanks.

cluster/node-pools/master-default/userdata.clc.yaml
linki: set to 10.

cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -557,6 +557,40 @@ storage:
...
+            - -enable-prometheus-metrics
+            - -inline-routes
+            - 'z: JWTPayloadAnyKV("iss", "https://identity.zalando.com") -> disableAccessLog() -> "https://127.0.0.1:443"; h: Path("/kube-system/healthz") -> setPath("/healthz") -> "http://127.0.0.1:8080"; all: * -> enableAccessLog() -> "https://127.0.0.1:443";'
+            ports:
+            - containerPort: 8443
+            readinessProbe:
+              httpGet:
+                path: /kube-system/healthz
+                port: 8443
+              initialDelaySeconds: 10
```
szuecs: I would drop it

linki: done

cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -557,6 +557,40 @@ storage:
             volumeMounts:
             - name: config-volume
               mountPath: /etc/nginx
+          - name: skipper-proxy
+            image: registry.opensource.zalan.do/pathfinder/skipper:v0.10.233
```
linki: TODO: Needs to be updated to a skipper that allows POST requests to do protocol upgrades, see https://github.com/zalando/skipper/pull/1091.

linki:
I reversed the skipper filter from "allow all zalandos" to "reject all service accounts" instead. This way things like worker kubelet tokens that go through the ELB will be allowed by skipper.

This should be fine regarding the problem we want to fix, let me know if you think otherwise.

linki:
👍

cluster/config-defaults.yaml
```diff
@@ -204,7 +204,7 @@ cluster_dns: "coredns"
 coredns_log_svc_names: "true"
 
 coreos_image: "ami-06142be4afc3d5b1b" # Container Linux 2079.4.0 (HVM, eu-central-1)
-kuberuntu_image: "ami-07922c3ba50e085fb" # Kuberuntu (dev) (HVM, eu-central-1)
+kuberuntu_image: "ami-064dbf2c30178e807" # Kuberuntu (dev) (HVM, eu-central-1)
```
njuettner, line 207: Could we please use the latest AMI > ami-0d856c4c2daf9b569, so see if the boot process has improved?

linki: I would defer this to later as it would take another 1h to run e2e 😞

njuettner:
:+1:

cluster/config-defaults.yaml
njuettner: Then let's merge it

PR #2223, opened 2019-06-13
Platform Injected Environment Variables

zeitlinger:
This is the configuration to support the new version of the admission controller that can inject environment variables.


cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -472,6 +473,14 @@ storage:
...
               - --application-registry-url=http://127.0.0.1:8285/
 {{- end }}
+              - --pod-inject-environment-variables
+              - --pod-environment-variable=_PLATFORM_ACCOUNT={{ .Cluster.Alias }}
+              - --pod-environment-variable=_PLATFORM_TOKENINFO_URL={{ .ConfigItems.tokeninfo_url }}
+
+              - --pod-environment-variable=_PLATFORM_OPENTRACING_TAG_ACCOUNT={{ .Cluster.Alias }}
+              - --pod-environment-variable=_PLATFORM_OPENTRACING_LIGHTSTEP_COLLECTOR_PORT=8443
+              - --pod-environment-variable=_PLATFORM_OPENTRACING_LIGHTSTEP_COLLECTOR_HOST=tracing.stups.zalan.do
+              - --pod-environment-variable=_PLATFORM_OPENTRACING_LIGHTSTEP_ACCESS_TOKEN={{ .ConfigItems.lightstep_token }}
```
mikkeloscar: This will expose the token to everyone with read access in the cluster. We should think of a way to configure this such that normal users can't just read the token. Unfortunately we can't use secrets afaik because this is a mirror pod.

Any ideas how to solve this?

zeitlinger:
This secret is visible in the spec of each pod in the cluster.

The problem with secrets is that they are namespaced, meaning we'd have to
add a secret whenever a namespace gets created.

cluster/node-pools/master-default/userdata.clc.yaml
zeitlinger: as discussed elsewhere:
- token can only be used for writing
- token is not security relevant

cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -469,6 +470,14 @@ storage:
               - --pod-dns-ndots={{ .Cluster.ConfigItems.teapot_admission_controller_ndots }}
 {{- if eq .Cluster.ConfigItems.teapot_admission_controller_process_resources "true" }}
               - --pod-process-resources
+              - --pod-enable-env-inject
```
aermakov-zalando, line 472: This needs to use a separate config item because it's not related to resources at all.

cluster/node-pools/master-ubuntu-default/userdata.yaml
```diff
@@ -206,6 +207,14 @@ write_files:
             - --pod-dns-ndots={{ .Cluster.ConfigItems.teapot_admission_controller_ndots }}
 {{- if eq .Cluster.ConfigItems.teapot_admission_controller_process_resources "true" }}
             - --pod-process-resources
+            - --pod-enable-env-inject
```
aermakov-zalando, line 209: Same here, please use a different config item instead.

test/e2e/admission_controller.go
```diff
@@ -0,0 +1,143 @@
...
+		platformEnvVar := 0
+		for _, v := range pod.Spec.Containers[0].Env {
+			if strings.HasPrefix(v.Name, "_PLATFORM_") {
+				platformEnvVar++
+			}
+		}
+
+		Expect(platformEnvVar >= 14).To(BeTrue())
+
+		bytes, err := cs.CoreV1().Pods(ns).GetLogs(pod.Name, &v1.PodLogOptions{}).DoRaw()
```
aermakov-zalando: I wouldn't bother checking that since we're only interested in the injection part. If environment variables in general are broken, we'll have bigger issues.

test/e2e/admission_controller.go
```diff
@@ -0,0 +1,143 @@
...
+		Expect(zone).To(HavePrefix("eu-central-"))
+
+		platformEnvVar := 0
+		for _, v := range pod.Spec.Containers[0].Env {
+			if strings.HasPrefix(v.Name, "_PLATFORM_") {
+				platformEnvVar++
+			}
+		}
+
+		Expect(platformEnvVar >= 14).To(BeTrue())
```
aermakov-zalando: I would rather have an explicit test that fetches environment variables from the pod and compares them to known-good values. For the zone variables you can fetch the zone from the node object and verify that it's set correctly, otherwise we never know if there are any issues.

test/e2e/admission_controller.go
```diff
@@ -0,0 +1,143 @@
...
+
+var _ = framework.KubeDescribe("Admission controller tests", func() {
+	f := framework.NewDefaultFramework("zalando-kube-admission-controller")
+	var cs kubernetes.Interface
+
+	BeforeEach(func() {
+		cs = f.ClientSet
+	})
+
+	It("Pods should get deployment info from deployment and zone from node [Zalando]", func() {
```
aermakov-zalando: I'd add a test that ensures that the users can override the variable and we wouldn't overwrite it with the default values.

cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -469,6 +470,14 @@ storage:
...
+              - --pod-inject-env=_PLATFORM_ACCOUNT={{ .Cluster.Alias }}
+              - --pod-inject-env=_PLATFORM_TOKENINFO_URL={{ .ConfigItems.tokeninfo_url }}
```
aermakov-zalando: Do we need this? Apps pick it up from `TOKENINFO_URL` instead.

zeitlinger:
There is already a test where this is checked, I think _PLATFORM_ACCOUNT.

.gitignore
```diff
@@ -3,6 +3,8 @@ __pycache__
 *.swp
 .tox/
 *.egg*
+*.iml
+.idea/
```
miry: Is it not a duplication of the line https://github.com/zalando-incubator/kubernetes-on-aws/pull/2223/files#diff-a084b794bc0759e7a6b77810e01874f2R10 ?

aermakov-zalando: Thanks! It does look like a duplicate.

hjacobs:
Can we add a summary of the changes in this PR's description? I.e. which environment variables are injected etc? Thanks!

aermakov-zalando:
Here's the current list:
* `_PLATFORM_ACCOUNT`, `_PLATFORM_OPENTRACING_TAG_ACCOUNT`: cluster name
* `_PLATFORM_OPENTRACING_LIGHTSTEP_COLLECTOR_HOST`, `_PLATFORM_OPENTRACING_LIGHTSTEP_COLLECTOR_PORT`, `_PLATFORM_OPENTRACING_LIGHTSTEP_ACCESS_TOKEN`: lightstep address
* `_PLATFORM_APPLICATION`: application name (from the corresponding label)
* `_PLATFORM_DEPLOYMENT_ID`, `_PLATFORM_OPENTRACING_TAG_DEPLOYMENT_ID`: deployment ID (from the corresponding annotation on the owning resource)
* `_PLATFORM_PIPELINE_ID`: CDP pipeline ID (from the corresponding annotation on the owning resource)
* `_PLATFORM_DOCKER_IMAGE`, `_PLATFORM_OPENTRACING_TAG_ARTIFACT`: docker image (from the container spec)
* `_PLATFORM_ZONE`, `_PLATFORM_OPENTRACING_TAG_ZONE`: AZ name (from the node where the pod was scheduled)

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2228, opened 2019-06-14
Fixed image tags for kubelet containers

arjunrn:
Change image tags for containers run by kubelet to `fixed` so that they run cached images. 

Signed-off-by: Arjun Naik <arjun.rn@gmail.com>

### DIFF
--- a/cluster/node-pools/master-ubuntu-default/userdata.yaml
+++ b/cluster/node-pools/master-ubuntu-default/userdata.yaml
@@ -86,7 +86,6 @@ write_files:
         namespace: kube-system
         labels:
           application: kube-apiserver
-          version: v1.13.7
         annotations:
           kubernetes-log-watcher/scalyr-parser: |
             [{"container": "webhook", "parser": "json-structured-log"}]
@@ -98,7 +97,7 @@ write_files:
         hostNetwork: true
         containers:
         - name: kube-apiserver
-          image: registry.opensource.zalan.do/teapot/kube-apiserver:v1.13.7
+          image: nonexistent.zalan.do/teapot/kube-apiserver:fixed
           args:
           - --apiserver-count={{ .Values.apiserver_count }}
           - --bind-address=0.0.0.0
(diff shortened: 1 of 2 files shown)


mikkeloscar:
For kubernetes components this is fine. But now we need to change the ami and make a PR here just to update e.g. the webhook. It seems like a little too much premature optimization at the cost of more human effort, especially considering this is master nodes, where startup time is not as important as workers.

cluster/node-pools/master-ubuntu-default/userdata.yaml
```diff
@@ -98,7 +98,7 @@ write_files:
         hostNetwork: true
         containers:
         - name: kube-apiserver
-          image: registry.opensource.zalan.do/teapot/kube-apiserver:v1.13.7
+          image: nonexistent.zalan.do/teapot/kube-apiserver:fixed
```
mikkeloscar, line 100: Would it make sense to use something like `local-cache` instead of `nonexistent` as a more descriptive name?

arjunrn:
Since we don't really update the webhook or any of these images very often I don't think there is more work. It's this versus having some stuff in the AMI and some stuff here. Better to be consistent right?

cluster/node-pools/master-ubuntu-default/userdata.yaml
arjunrn: I think `nonexistent` is more descriptive because the reason for it is tagged with a registry which doesn't exist so that it cannot be pulled.

mikkeloscar:
In my opinion it's about defining where the split should be between what's in the AMI and what's here. IMO if it's not shared between master and worker, it does not belong in a shared AMI unless we really want to optimize boot time with this which should require some sort of test to prove that it's work it (again my opinion).

Problem with this right now, is that whenever we change e.g. the webhook we would also rotate the workers because the AMI is updated.

arjunrn:
:+1:

mikkeloscar:
:+1:

PR #2229, opened 2019-06-14
Enable apiserver skipper proxy by default

linki:
This enables the API server proxy by default.
* reconfigures apiserver ELB to go to another port
* changes security groups according to the new ports being used

Also:
* update skipper-proxy to a version that correctly handles watches
  * allows to watch initially empty lists often used in e2e tests
* increase timeout from skipper to clients to 60m
  * allows to `kubectl logs` up to 60 minutes. Seems like apiserver's timeout is slighly less than that.
* do not access log kube-probe requests
  * avoids distracting output

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -241,6 +241,6 @@ node_max_pods: "110" # Default: 110
 node_cidr_mask_size: "24" # Default: 24
 
 # when set to true, routes external traffic to the apiserver through a skipper sidecar
-apiserver_proxy: "false"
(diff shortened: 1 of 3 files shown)


linki:
:+1:

mikkeloscar:
Is it safe to enable by default now that the working skipper is not already rolled out? Seems to me we still need to do it in two steps? Or if you set all to false in the cluster-registry it's also fine :)

linki:
@mikkeloscar It's now disabled in all clusters.

mikkeloscar:
:+1:

PR #2231, opened 2019-06-17
Setting prometheus max memory to max of the Node

muaazsaleem:
We often have problems of the VPA not being able to scale prometheus
up even though memory is available on the Node so set max possible memory
to max memory available on the Node.

ref: https://github.bus.zalan.do/teapot/issues/issues/1933
Signed-off-by: Muhammad Muaaz Saleem <muhammad.muaaz.saleem@zalando.de>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ __pycache__
 *.egg*
 test/e2e/e2e.test
 test/e2e_modules
+.idea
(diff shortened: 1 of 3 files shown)


muaazsaleem:
👍

arjunrn:
:+1:

PR #2233, opened 2019-06-18
Disallow SSH from the outside

aermakov-zalando:
This still allows connecting via the Odd host, but closes access from the outside.


aermakov-zalando:
👍

jrake-revelant:
👍

PR #2241, opened 2019-06-18
Explicitly set the root EBS volume to delete on termination

arjunrn:
This property is inherited from the EBS volume which was used to build the AMI. If it is false then the volumes are left behind after the instances are terminated.


aermakov-zalando:
👍

arjunrn:
:+1:

PR #2244, opened 2019-06-19
Override kube-proxy hostname from downward API

arjunrn:
Since we got rid of the hack which makes the local hostname match the AWS hostname kube-proxy cannot create the correct routes _(affects Ubuntu only)_.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2247, opened 2019-06-20
Increase QPS for kubelet, controller-manager

aermakov-zalando:
The default for kubelet is ridiculously low, causing 5+ minute long delays when 20 pods are started at the same time.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2248, opened 2019-06-20
Fixed parens for env var in kube-proxy

arjunrn:
Parenthesis should be used to interpolate environment variable into the command instead of curly braces.

### DIFF
--- a/cluster/manifests/kube-proxy/daemonset.yaml
+++ b/cluster/manifests/kube-proxy/daemonset.yaml
@@ -33,7 +33,7 @@ spec:
       - name: kube-proxy
         image: registry.opensource.zalan.do/teapot/kube-proxy:v1.13.7
         args:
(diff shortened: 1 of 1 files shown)


mikkeloscar:
Wonder why e2e didn't catch this? Was kube-proxy still able to start?

mikkeloscar:
:+1:

mikkeloscar:
Ok I see this:

```
node.go:103] Failed to retrieve node info: nodes "${hostname_override}" not found
```

So I'm guessing "some" functionality is just not working, which we don't catch in the e2e.

arjunrn:
Because the host didn't match and because of the `externalTrafficPolicy` which is set on the `kube-node-ready` service the IP tables rules are not created. We can add an E2E test which checks that the service is available.

arjunrn:
👍

PR #2251, opened 2019-06-21
Updated VPA to version 0.5.1

arjunrn:
Release notes are [here](https://github.com/kubernetes/autoscaler/releases/tag/vertical-pod-autoscaler-0.5.1). No changes to the CRD seem to be required.

### DIFF
--- a/cluster/manifests/01-vertical-pod-autoscaler/admission-controller-deployment.yaml
+++ b/cluster/manifests/01-vertical-pod-autoscaler/admission-controller-deployment.yaml
@@ -19,7 +19,7 @@ metadata:
   namespace: kube-system
   labels:
     application: vpa-admission-controller
-    version: patched-0.4.0-master-14
+    version: patched-0.5.1-master-15
     component: vpa
 spec:
   replicas: 1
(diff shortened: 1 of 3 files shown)


szuecs:
but: https://github.com/kubernetes/autoscaler/releases/tag/vertical-pod-autoscaler-0.5.0
not supported anymore: `autoscaling.k8s.io/v1beta1`

arjunrn:
> but: https://github.com/kubernetes/autoscaler/releases/tag/vertical-pod-autoscaler-0.5.0
> not supported anymore: `autoscaling.k8s.io/v1beta1`

@szuecs that's correct but like the release notes say the VPAs with the version `v1beta1` can be created but won't receive recommendations. Like you can see [here](https://github.com/kubernetes/autoscaler/blob/vertical-pod-autoscaler-0.5.1/vertical-pod-autoscaler/deploy/vpa-beta2-crd.yaml) they keep the old version in the CRD but don't use it while generating recommendations.

mikkeloscar:
:+1:

arjunrn:
:+1:

PR #2256, opened 2019-06-24
Fixed the VPA recommender for missing keys

arjunrn:
The VPA recommender runs with a _memory saver_ mode which only stores metrics for pods with matching VPAs. But the new version prints an error whenever there metrics for which there are no corresponding pods, which is a majority of the pods. So the logs are cluttered with error logs. This fix suppressed that logs whenever the recommender is running memory-saver mode.


mikkeloscar:
:+1:

arjunrn:
👍

PR #2258, opened 2019-06-25
Give CDP controller access to CRDs that it needs to manage

mikkeloscar:
cdp-controller now needs these extra permissions.

* Create AWSIAMRoles for deployment pods
* Manage GradualDeployment resources for canary deployments.

### DIFF
--- a/cluster/manifests/roles/cdp-controller-rbac.yaml
+++ b/cluster/manifests/roles/cdp-controller-rbac.yaml
@@ -34,6 +34,27 @@ rules:
   - list
   - watch
   - patch
+- apiGroups:
+  - "zalando.org"
+  resources:
+  - awsiamroles
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+- apiGroups:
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

cluster/manifests/roles/cdp-controller-rbac.yaml
```diff
@@ -34,6 +34,27 @@ rules:
...
   - patch
+- apiGroups:
+  - "zalando.org"
+  resources:
+  - awsiamroles
+  verbs:
+  - get
+  - list
+  - watch
+  - create
```
aermakov-zalando, line 45: `update` is missing?

mikkeloscar: cdp-controller doesn't do update of the resource, so this is currently not needed.

arjunrn:
:+1:

njuettner:
:+1:

PR #2262, opened 2019-06-26
Make external service account token usage configurable

linki:
This allows to disable external service account token per cluster.

The skipper rule works like that:
* intercept requests with service account tokens
  * toggleable via config item
  * token validation happens at apiserver if not intercepted
* forward custom health check to apiserver's `/healthz`
* forward everything else to apiserver
  * token validation happens at apiserver


linki:
:+1:

mikkeloscar:
:+1:

PR #2270, opened 2019-07-01
adjust lifo timeout from 10s to 3s

szuecs:
I measured in our internal issue tracker, in issue 1960, that 10s -> 3s fix the stdlib based buffio memory usage on huge connection spikes caused by retries

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -33,7 +33,7 @@ skipper_ingress_memory: "1Gi"
 skipper_ingress_tracing_buffer: "8192"
 
 # skipper default filters
-skipper_default_filters: 'enableAccessLog(4,5) -> lifo(2000,20000,"10s")'
+skipper_default_filters: 'enableAccessLog(4,5) -> lifo(2000,20000,"3s")'
 
 # skipper backend timeout defaults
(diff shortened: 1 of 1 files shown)


gargravarr:
:+1:

cluster/config-defaults.yaml
```diff
@@ -33,7 +33,7 @@ skipper_ingress_memory: "1Gi"
 skipper_ingress_tracing_buffer: "8192"
 
 # skipper default filters
-skipper_default_filters: 'enableAccessLog(4,5) -> lifo(2000,20000,"10s")'
+skipper_default_filters: 'enableAccessLog(4,5) -> lifo(2000,20000,"3s")'
```
sarslanhan, line 36: @szuecs does this mean that skipper drops requests after 3s regardless of the stack size? If yes, I think we should first check what requests will fail as it imposes a hidden latency limit for every service running on K8s.

szuecs: @sarslanhan no it means only that skipper is not able to pop from stack within 3s.
10s would be also bad for long blocking calls, so don't worry with that.

szuecs:
:+1:

PR #2271, opened 2019-07-01
Enable kube-aws-iam-controller by default for kube-system components

mikkeloscar:
Enable in all clusters by default, with the option of still disabling it per cluster via cluster-registry.

Note: So far only the components checked off in #2219 are using this.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #2274, opened 2019-07-01
Egress controller should be allowed to delete tags

mikkeloscar:
New version of egress controller uses tags for storing Egress IPs. Allow deleting tags when resources are cleaned up.

Follow up to #2273


mikkeloscar:
Replaced by the revert in #2275 will add this when the fixed version of egress controller is ready

PR #2277, opened 2019-07-02
Added more instance storage types

arjunrn:
With more instance types available for the spot pool the cluster should be ready faster. Should cut down some of the time needed to finish the E2E.


njuettner:
👍

arjunrn:
:+1:

PR #2279, opened 2019-07-02
Allow Test tokens to access test clusters

linki:
Introduces the `sandbox-bridge` for test clusters.

This allows Zalando PlatformIAM tokens of `test` services to access the Kubernetes API server of `test` clusters.

### DIFF
--- a/cluster/node-pools/master-default/userdata.clc.yaml
+++ b/cluster/node-pools/master-default/userdata.clc.yaml
@@ -511,7 +511,7 @@ storage:
               - name: CLUSTER_ID
                 value: {{if index .Cluster.ConfigItems "webhook_id"}}{{ .Cluster.ConfigItems.webhook_id }}{{else}}{{ .Cluster.ID }}{{end}}
               - name: TOKEN_INTROSPECTION_URL
-                value: http://127.0.0.1:9021/oauth2/introspect
+                value: http://127.0.0.1:{{ if eq .Cluster.Environment "production" }}9021{{else}}9023{{end}}/oauth2/introspect
               - name: USER_GROUPS
                 value: credentials-provider=Administrator,credprov-kube-ops-view-read-only-token=ReadOnly,credprov-cluster-lifecycle-manager-cluster-rw-token=Administrator,stups_cluster-lifecycle-manager=Administrator,credprov-cluster-lifecycle-manager-test-cluster-rw-token=Administrator,credprov-cdp-controller-cluster-token=PowerUser,stups_deployment-discovery=ReadOnly,stups_cockpit-skipper=ReadOnly,stups_zmon-zmon=ReadOnly,stups_kube-resource-report=ReadOnly,stups_oxygen=ReadOnly,kubelet=system:masters
               - name: BUSINESS_PARTNER_IDS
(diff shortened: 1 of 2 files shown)


cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -545,6 +545,68 @@ storage:
...
+                -> setRequestHeader("X-Checking-JWT", "true")
+                -> <loopback>;
+              prod: Path("/oauth2/introspect") && Header("X-Checking-JWT", "true") && JWTPayloadAllKV("iss", "https://identity.zalando.com")
+                -> dropRequestHeader("Authorization") -> dropRequestHeader("X-Checking-JWT")
+                -> "http://127.0.0.1:9021/oauth2/introspect";
+              sandbox: Path("/oauth2/introspect") && Header("X-Checking-JWT", "true") && JWTPayloadAllKV("iss", "https://sandbox.identity.zalando.com")
+                -> dropRequestHeader("Authorization") -> dropRequestHeader("X-Checking-JWT")
+                -> "http://127.0.0.1:9022/oauth2/introspect";
+              fallback: Path("/oauth2/introspect") && QueryParam("token") && Header("X-Checking-JWT", "true")
+                -> inlineContent("{\"active\":false}") -> <shunt>;
```
linki, line 600: `/introspect` works different than `/tokeninfo` in that it expects a token as part of the query instead of the `Authorization` header. In order to use `JWTPayloadAllKV` we need to `queryToHeader` the token first but also keep the `token` in the query for the upstream tokeninfo server.

This can lead to an infinite route matching loop as we don't remove the `query` as seen in other examples. Adding the `X-Checking-JWT` header kind of solves it. However, we're missing the rule for when `token` isn't present at all. Currently this returns a `404` when it should result in the same as `fallback`. Adding a rule for that resulted in other rules not matching correctly anymore so I leave it like that for now.

szuecs:
Why do we need introspection?
Tokeninfo should be enough.

linki:
It was needed to support `Cloudyn` back in the days. They needed a different Business Partner ID that was only supported via introspection.

Maybe we can transition back to `/tokeninfo`? /cc @mikkeloscar

cluster/node-pools/master-default/userdata.clc.yaml
linki: Readded the final rule as e2e passed before.

cluster/node-pools/master-default/userdata.clc.yaml
```diff
@@ -546,6 +546,70 @@ storage:
...
+                -> queryToHeader("token", "Authorization", "Bearer %s")
+                -> setRequestHeader("X-Checking-JWT", "true")
+                -> <loopback>;
+              prod: Path("/oauth2/introspect") && Header("X-Checking-JWT", "true") && JWTPayloadAllKV("iss", "https://identity.zalando.com")
+                -> dropRequestHeader("Authorization") -> dropRequestHeader("X-Checking-JWT")
+                -> "http://127.0.0.1:9021/oauth2/introspect";
+              sandbox: Path("/oauth2/introspect") && Header("X-Checking-JWT", "true") && JWTPayloadAllKV("iss", "https://sandbox.identity.zalando.com")
+                -> dropRequestHeader("Authorization") -> dropRequestHeader("X-Checking-JWT")
+                -> "http://127.0.0.1:9022/oauth2/introspect";
+              fallback1: Path("/oauth2/introspect") && QueryParam("token") && Header("X-Checking-JWT", "true")
```
aryszka, line 597: to avoid conflict between prod<->fallback1 and sandbox<->fallback1, either remove `QueryParam("token")` from fallback1, or add to both prod and sandbox. If we expect to add more routes here in the future, then probably the latter way is better.

linki: Done, thanks.

cluster/node-pools/master-default/userdata.clc.yaml
linki: Made rules more specific to avoid conflicts.

mikkeloscar:
> Maybe we can transition back to /tokeninfo? /cc @mikkeloscar

If we can make it work as-is in a reasonable way then I would just keep it with introspection because otherwise we have to revert things in the webhook to handle tokeninfo repsonse instead and also think about the assumptions we have made e.g. for Zalando IAM service identities as used for RBAC (It currently includes the business partner i.e. `zalando`). But if it's much simpler to make this work with `/tokeninfo` then we can also switch back and just deal with the changes needed if we ever need multiple business partner support in the future (which is maybe not super likely).

linki:
Rebased on https://github.com/zalando-incubator/kubernetes-on-aws/pull/2202 because it requires its changes to the auth webhook. (Don't go to team service as non-human user)

linki:
@mikkeloscar It works now. Could you give it a quick review?

mikkeloscar:
:+1:

njuettner (approved):
👍

njuettner:
👍

PR #2291, opened 2019-07-07
Reduce CPU requests for master components

mikkeloscar:
This reduces CPU requests for `tokeninfo` and `image-policy-webhook`  based on metrics from the last 7 days.

The reason for this is that the components don't use as much CPU as requested and we now have to many containers on the master that all the things can not fit on smaller instance types.

tokeninfo CPU 7 days

image

image-policy-webhook 7 days

image


njuettner:
👍

mikkeloscar:
:+1:

PR #2296, opened 2019-07-08
Disable external service accounts by default

linki:
All clusters override this config so we can safely enable this by default.
New clusters will therefore be created with the enforcement in place from the beginning.


njuettner:
👍

linki:
:+1:

PR #2305, opened 2019-07-10
Don't mount host certs. Certs are available in the containers

mikkeloscar:
This avoids mounting certs from the host which is not needed since the certs are already in the containers. 

Mounting the certs has the downside that the host certs must be in a special location, which is not always the same from system to system. In Ubuntu nodes we do a hack to move the certs to the expected location but this then breaks the assumptions about where certs are on the ubuntu node for other components (I was playing around with [ec2-instance-connect](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-connect-set-up.html) which didn't work because of this).

Once this is rolled out we can drop the special relocation of certs in the Ubuntu AMI.

### DIFF
--- a/cluster/manifests/kube-proxy/daemonset.yaml
+++ b/cluster/manifests/kube-proxy/daemonset.yaml
@@ -51,9 +51,6 @@ spec:
             cpu: 50m
             memory: 200Mi
         volumeMounts:
-        - name: ssl-certs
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2255, opened 2019-06-24
SSL policy >= 1.2

szuecs:
roll out restrictive TLS policy to only support >= TLS1.2 to test clusters
see also https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html#describe-ssl-policies 
Signed-off-by: Sandor Szücs <sandor.szuecs@zalando.de>

PR #2312, opened 2019-07-15
Enable SSL Policy >= 1.2 for test

mikkeloscar:
Follow up to: https://github.com/zalando-incubator/kubernetes-on-aws/pull/2255 which was reverted when we needed users to fix CNAMEs pointing directly to ALB addresses. The CNAMEs are now fixed so we can restart the rollout.

Only enables new SSL policy for test clusters and for new clusters. (existing production clusters have the policy: `ELBSecurityPolicy-2016-08` set in cluster registry config items).


mikkeloscar:
:+1:

szuecs:
:+1:

PR #2313, opened 2019-07-15
Ensure pre-apply steps are only done for existing clusters

mikkeloscar:
Without this in place CLM will try to delete resources before it has any access rights, because the `ClusterRoleBinding` for CLM has not yet been setup when a new cluster is created.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2320, opened 2019-07-17
vpa-updater: fix argument passing

aermakov-zalando:
We need to provide both `command` and `args`, otherwise the defaults from the Docker image override everything.


aryszka:
:+1:

szuecs:
:+1:

PR #2323, opened 2019-07-18
vpa: fix logging

aermakov-zalando:
Logging was completely broken in VPA and only was fixed 4+ months after the switch to `klog` (https://github.com/kubernetes/autoscaler/pull/1994). Cherry-pick the patch as well so we can figure out why it doesn't work.


mikkeloscar:
:+1:

PR #2331, opened 2019-07-22
Allow to specify per-pod pid limit

linki:
Enable PID limit enforcement and introduce a config item to specify that limit per pod. Currently Kubernetes doesn't enforce any limit which corresponds to setting the config item to "-1".

The limit of 4096 has been chosen to roughly match the internally proposed "Something like 5k would probably be a good initial estimate".

Merging this PR enforces the limit.

The FeatureGate is enabled by default in Kubernetes 1.14 and can be dropped from the config then.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #2341, opened 2019-07-25
Hardcode cluster-lifecycle-manager to prevent chicken/egg problem

mikkeloscar:
In #2200 and #2202 we moved permissions from a simple mapping of UID to Role to a more limited permission set per service user using RBAC.

This however introduce a chicken/egg problem for CLM because it needs permission to create the initial rolebinding for itself..

The quick fix is to still map CLM to a specified group (`system:masters`) such that it always have admin permissions.


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #2343, opened 2019-07-25
Removed stack weights to determine traffic in kube-metrics-adapter

arjunrn:
Using the `stackset-weight` to determine the traffic on an ingress causes the calculated RPS to be higher than it actually is.


arjunrn:
:+1:

szuecs:
:+1:

PR #2349, opened 2019-07-25
Allow to configure PID limit per node pool

linki:
Allows to specify `pod_max_pids` config item on a node pool. If it's not defined on the node pool the cluster-wide `pod_max_pids` config item is used. Defaults to 4096 if not defined in cluster registry.


aermakov-zalando:
Do we want to maybe automagically do it in CLM? I.e. copy all cluster config items to the node pool config items, unless they're overridden?

linki:
Refactored so that a PID limit of `-1` doesn't change userdata at all. /cc @mikkeloscar

mikkeloscar:
:+1:

szuecs:
:+1:

PR #2350, opened 2019-07-25
Updated the Ubuntu AMI with Nvidia driver 410

arjunrn:
Updated the NVIDIA driver.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -215,7 +215,7 @@ cluster_dns: "coredns"
 coredns_log_svc_names: "true"
 
 coreos_image: "ami-0d1579b60bb706fb7" # Container Linux 2079.6.0 (HVM, eu-central-1)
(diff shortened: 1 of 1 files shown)


cluster/config-defaults.yaml
```diff
@@ -215,7 +215,7 @@ cluster_dns: "coredns"
 coredns_log_svc_names: "true"
 
 coreos_image: "ami-0d1579b60bb706fb7" # Container Linux 2079.6.0 (HVM, eu-central-1)
-kuberuntu_image: "ami-087d64da45b8052cd" # zalando-ubuntu-kubernetes-production-v1.13.7-master-47 (HVM, eu-central-1)
+kuberuntu_image: "ami-036931dfa3f2f3272" # zalando-ubuntu-kubernetes-production-v1.13.7-master-47 (HVM, eu-central-1)
```
mikkeloscar: Would be nice if we could also update the comment with the AMI name IMO, then it's easier to understand which version it is without first resolving the ami ID

mikkeloscar:
:+1:

arjunrn:
:+1:

PR #2358, opened 2019-07-26
Bind Emergency and Manual Groups same as PowerUser

mikkeloscar:
Binds `Emergency` and `Manual` Groups same way as `PowerUser` since they are defined to have the same permissions.

This is currently done in the auth webhook, but we should migrate to RBAC to fix blacklist/whitelist bugs in the auth webhook.

### DIFF
--- a/test/e2e/authorisation_test.go
+++ b/test/e2e/authorisation_test.go
@@ -2157,6 +2157,34 @@ var _ = framework.KubeDescribe("Authorization tests", func() {
 				},
 			},
 			{
+				msg: "cdp service account can't escalate permissions",
+				reqBody: `{
+					"apiVersion": "authorization.k8s.io/v1beta1",
+					"kind": "SubjectAccessReview",
+					"spec": {
+					"resourceAttributes": {
+						"namespace": "",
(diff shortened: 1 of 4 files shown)


mikkeloscar:
:+1:

test/e2e/authorisation_test.go
```diff
@@ -2156,6 +2156,34 @@ var _ = framework.KubeDescribe("Authorization tests", func() {
...
+						"verb": "escalate",
+						"group": "*",
+						"resource": "clusterroles"
+					},
+					"user": "system:serviceaccount:default:cdp",
+					"group": []
+					}
+				}`,
+				expect: expect{
+					status: http.StatusCreated,
```
aryszka, line 2176: just curious: why are we using 201, when the request was rejected?

aryszka:
:+1:

test/e2e/authorisation_test.go
mikkeloscar: This is how the AccessReview API works. You "create" a review request `201` and then the response is whether the user would have access to perform that action or not.

PR #2360, opened 2019-07-26
Simplify template logic based on config item propagation

linki:
Simplifies NodePool config item lookup by using pending CLM feature: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/153


linki:
:+1:

aermakov-zalando:
👍

PR #2361, opened 2019-07-26
cadvisor: explicitly publish Kubelet metrics

aermakov-zalando:
v0.33.0 doesn't publish these by default if `--docker-only` is enabled.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2365, opened 2019-07-29
UX: additionalPrinterColumns for PlatformCredentialsSet

hjacobs:
Adding `additionalPrinterColumns` makes `kubectl` and [kube-web-view](https://codeberg.org/hjacobs/kube-web-view/) much more useful:
Screenshot_2019-07-29 platformcredentialssets - Kubernetes Web UI(1)

Sadly the CRD JSONPath only supports simple expression, i.e. there is no way to show the number of clients/tokens (would require some `length()` function) or the keys of maps (to list token names or similar). Related K8s issue to support complex stuff in CRD JSONPath: https://github.com/kubernetes/kubernetes/issues/67268

### DIFF
--- a/cluster/manifests/01-platformcredentialsset/customresourcedefinition.yaml
+++ b/cluster/manifests/01-platformcredentialsset/customresourcedefinition.yaml
@@ -16,6 +16,23 @@ spec:
       - pcs
     categories:
       - all
+  additionalPrinterColumns:
+  - JSONPath: .spec.application
+    description: ID of application registered in application registry
+    name: Application
+    type: string
+  - JSONPath: .status.errors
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
Application is fine, but I disagree with adding errors/problems. We could easily add a `status` field in credentials-provider to show whether the PCS is healthy or not, but printing raw errors makes the output of `zkubectl get pod` pretty much unreadable.

hjacobs:
@aermakov-zalando OK, agreed, will change accordingly and add a new status field.

aermakov-zalando:
👍

hjacobs:
:+1:

PR #2370, opened 2019-07-30
Updated the webhook with policies for cache

arjunrn:
This version of the webhook trusts cached images in the Ubuntu AMIs.


aermakov-zalando:
Let's drop `compliance_checker_enabled` then? We've never ended up disabling it, and it'll be impossible in the future anyway (because the old webhook won't support cached images).

mikkeloscar:
:+1:

arjunrn:
👍

PR #2372, opened 2019-07-30
Allow cdp service account to escalate

mikkeloscar:
Partly "revert" of #2358 to allow only `cdp` service account to `escalate`. This is to prevent breaking deployments which currently creates RBAC roles/bindings.


aryszka:
:+1:

mikkeloscar:
Merging eventhough coreos e2e fails. The failure is because of unstable egress addressed in #2376

mikkeloscar:
:+1:

PR #2376, opened 2019-07-30
Add egress route

mikkeloscar:
Our Egress tests are unstable because `readmyip.appspot.com` is not always served from the expected IP range.

This is a quick hack to at least fail fast if the IP is not in an expected range. A longer term solution would be to have a more reliable endpoint with a static IP we control.

### DIFF
--- a/test/e2e/egress.go
+++ b/test/e2e/egress.go
@@ -15,6 +15,7 @@ package e2e
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


szuecs (approved):

szuecs:
👍

mikkeloscar:
:+1:

PR #2384, opened 2019-07-31
Dev to kube 1.14

aermakov-zalando:
Merge the recent changes from `dev`. Note that `setup_modules.sh` was changed to checkout the 1.14 branch instead of a tag because of the stupid `github.com/renstrom/dedent` rename.


mikkeloscar:
:+1:

linki:
Looks like this fails because the e2e tests start by creating the previous cluster and then do an update. However the previous cluster doesn't come up because the old AMI doesn't exist.

linki:
:+1:

aermakov-zalando:
👍

PR #2385, opened 2019-07-31
egress test: fix race conditions, add an explicit timeout

aermakov-zalando:
* create the pod with RestartPolicy: Never, otherwise the test might never capture the exit status
 * time out after 10 minutes if we failed

### DIFF
--- a/test/e2e/egress.go
+++ b/test/e2e/egress.go
@@ -14,6 +14,7 @@ limitations under the License.
 package e2e
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"time"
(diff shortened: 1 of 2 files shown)


test/e2e/egress.go
```diff
@@ -81,37 +82,37 @@ var _ = framework.KubeDescribe("Static Egress creation", func() {
...
-			switch n := p.Status.ContainerStatuses[0].State.Terminated.ExitCode; n {
-			case 0:
-				return
-			case 2:
-				// set error
-				Expect(fmt.Errorf("failed to change public IP")).NotTo(HaveOccurred())
-				return
-			}
-		}
+		Eventually(func() error {
```
szuecs, line 92: Does `Eventually` retry?

I think the problem was that it takes a while to actually have everything setup before you get the routing change applied.

aermakov-zalando: Yes, it'll retry every 10 seconds and time out after 10 minutes (see https://github.com/zalando-incubator/kubernetes-on-aws/pull/2385/files#diff-b9184f93ce39a9ec0e9abba3438a9650R94).

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2386, opened 2019-07-31
ingress test: use assertions instead of log.Fatal

aermakov-zalando:
This could maybe explain the tests randomly failing?


linki:
:+1:

aermakov-zalando:
👍

PR #2389, opened 2019-08-01
Consume scalyr-agent-log also via fluentd

pheanex:
Signed-off-by: Konstantin Manna <konstantin.manna@zalando.de>


mikkeloscar:
:+1:

pheanex:
@christianberg @anton-ryzhov @vwiessner @astarnell please also review

cluster/manifests/logging-agent/config.yaml
```diff
@@ -42,6 +55,18 @@ data:
       @type token
     </filter>
 
+    <filter scalyr-agent.**>
+      @type record_transformer
+      <record>
+        # s3_path is: ${application}/${component}/${namespace}/${environment}/${version}/${cont_name}
+        s3_path     "logging-agent/logging/visibility/{{ .Environment }}/v0.30/scalyr-agent"
+        application "logging-agent"
+        version     "v0.30"
```
anton-ryzhov: Is it possible to extract version from `deamonset.yaml`, or declare it in another place and use in this two?

cluster/manifests/logging-agent/daemonset.yaml
```diff
@@ -244,4 +250,7 @@ spec:
       - name: fluentd-posfiles
         hostPath:
           path: /var/lib/fluentd
+
+      - name: scalyr-agent-logs
```
anton-ryzhov, line 254: When this directory is [re]created? On pod restart? Or on container restart?

cluster/manifests/logging-agent/config.yaml
```diff
@@ -42,6 +55,18 @@ data:
       @type token
     </filter>
 
+    <filter scalyr-agent.**>
+      @type record_transformer
+      <record>
+        # s3_path is: ${application}/${component}/${namespace}/${environment}/${version}/${cont_name}
+        s3_path     "logging-agent/logging/visibility/{{ .Environment }}/v0.30/scalyr-agent"
```
anton-ryzhov: `{{ .Environment }}` is not expanded

pheanex: As far as I understand this is something that is expanded when it is deployed/applied

cluster/manifests/logging-agent/daemonset.yaml
pheanex: https://kubernetes.io/docs/concepts/storage/volumes/#emptydir

cluster/manifests/logging-agent/config.yaml
mikkeloscar: Yes, this is expanded by the [cluster-lifecycle-manager](https://github.com/zalando-incubator/cluster-lifecycle-manager) when applying the manifests.

cluster/manifests/logging-agent/config.yaml
pheanex: We also hardcode it here https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/logging-agent/daemonset.yaml#L9
I do not know. Can we introduce some kind of variable in another manifest and reuse it here?
Do you know @mikkeloscar ?

anton-ryzhov: Hardcode in one place is declaration. In several places — unmaintainable thing

mikkeloscar: I can't think of any way to do this now.

mikkeloscar: I mean you could set the variable in `cluster/config-defaults.yaml` and use that, but I don't know how obvious it is when you need to change it.

aermakov-zalando: You could also move both resources into one yaml and then use template variables to define it once. Not sure if you think it'll be uglier or not. :)

pheanex:
:+1:

anton-ryzhov (approved):

anton-ryzhov:
:+1:

christianberg:
Should we also use `logging-agent-version` variable here?
https://github.com/zalando-incubator/kubernetes-on-aws/blob/57990653d51d508f638f89ed41bfe045aec94456/cluster/manifests/logging-agent/daemonset.yaml#L86
Would need to republish the docker image with tag `v0.30` instead of `0.30`, though.
Even better would be to decouple the version of the `logging-agent` from the version of the `kubernetes-log-watcher`. (This can/should happen in a separate PR.)

pheanex:
@christianberg IIRC we agreed if we do this, we will do it in another PR

arjunrn:
:+1:

jrake-revelant:
👍

PR #2402, opened 2019-08-08
Changed Ubuntu AMI to use ID from name and account ID

arjunrn:
The AMI id is derived from the AMI name and the account owner of the image.


arjunrn:
:+1:

jrake-revelant:
👍  although I didn't check if this one exists ;)

jrake-revelant:
👍

PR #2410, opened 2019-08-12
Fixed e2e tests to be compatible to other channels

arjunrn:
The revert from #2406 prevents merges with higher channels because the structure of end-to-end tests are different in different branches. This PR add only the change of the e2e test back so that changes can be merged.


aermakov-zalando:
👍

njuettner:
👍

PR #2413, opened 2019-08-13
Make node_exporter resources configurable

aermakov-zalando:
We don't want to use VPA for daemonsets, so we have to do it like this. :(


aermakov-zalando:
👍

jrake-revelant:
👍

PR #2416, opened 2019-08-13
Ubuntu nodes: drop the extra volume

aermakov-zalando:
This causes a creation of an extra EBS volume for every node which is completely unused.


aermakov-zalando:
👍

arjunrn:
👍

PR #2439, opened 2019-08-22
Prometheus: fix the realm check

aermakov-zalando:
Tokeninfo keys are different from the token keys. :(


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2443, opened 2019-08-23
deployment-config: add cloudformation-available

aermakov-zalando:
DC clusters support secret encryption, only CF is unavailable.


cluster/manifests/infrastructure-configs/configmap.yaml
```diff
@@ -8,3 +8,4 @@ data:
   scalyr-team-token: "{{.Cluster.ConfigItems.scalyr_team_token}}"
   create-namespaces: "true"
   aws-available: "true"
```
mikkeloscar, line 10: We can drop this or?

aermakov-zalando: Not yet, it's required for the current version. :(

PR #2446, opened 2019-08-23
Allow static cpu manager policy to be enabled on nodepools

arjunrn:
The `static` CPU management policies allow entire CPUs to be allocated to pods which might be useful for workloads which benefit from CPU affinity. The `static` policy is only enabled when the config item is present and static to the value `static`.

### DIFF
--- a/cluster/node-pools/worker-ubuntu-default/userdata.yaml
+++ b/cluster/node-pools/worker-ubuntu-default/userdata.yaml
@@ -53,6 +53,7 @@ write_files:
         SupportPodPidsLimit: true
       podPidsLimit: {{ .NodePool.ConfigItems.pod_max_pids }}
 {{- end }}
+      cpuManagerPolicy: {{ .NodePool.ConfigItems.cpu_manager_policy }}
       maxPods: {{ .Cluster.ConfigItems.node_max_pods }}
       healthzPort: 10248
       healthzBindAddress: "0.0.0.0"
--- a/test/e2e/cluster_config.sh
+++ b/test/e2e/cluster_config.sh
@@ -69,6 +69,8 @@ clusters:
     profile: ${WORKER_PROFILE}-splitaz
     min_size: 3
     max_size: 21
+    config_items:
+      cpu_manager_policy: static
   - discount_strategy: spot_max_price
     instance_types: ["m4.large", "m5.large", "m5.xlarge", "m4.xlarge"]
     name: default-worker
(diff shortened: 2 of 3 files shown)


test/e2e/cluster_config.sh
```diff
@@ -69,6 +69,8 @@ clusters:
     profile: ${WORKER_PROFILE}-splitaz
     min_size: 3
     max_size: 21
+    config_items:
+      cpu_manager_policy: static
   - discount_strategy: spot_max_price
     instance_types: ["m4.large", "m5.large", "m5.xlarge", "m4.xlarge"]
```
SAitharaju: Does it mean that the cpu_manager_policy can be set only in the node pools with these type of instances?

SAitharaju, line 75: Does it mean that the cpu_manager_policy can be set only in the node pools with these type of instances?

arjunrn: No, those are just our end to end tests where we create nodepools.

mikkeloscar:
:+1:

test/e2e/cluster_config.sh
```diff
@@ -69,6 +69,8 @@ clusters:
     profile: ${WORKER_PROFILE}-splitaz
     min_size: 3
     max_size: 21
+    config_items:
+      cpu_manager_policy: static
```
linki, line 73: wrong name: `cpu_manager_policy != cpu_management_policy`

cluster/node-pools/worker-ubuntu-default/userdata.yaml
```diff
@@ -52,6 +52,9 @@ write_files:
       featureGates:
         SupportPodPidsLimit: true
       podPidsLimit: {{ .NodePool.ConfigItems.pod_max_pids }}
+{{- end }}
+{{- if eq .NodePool.ConfigItems.cpu_management_policy "static" }}
```
linki: Are there more resonable options? If yes it would make sense to do:
```
{{- if ... }}
  cpuManagerPolicy: {{ .NodePool.ConfigItems.cpu_management_policy }}
{{- end }}
```

linki: Or even drop the `if` if you have a default.

mikkeloscar: There's only `none` and `static`, but it might make sense to just drop the if since we are going to roll the nodes for v1.14 anyway, so the if doesn't provide so much in this case :)

arjunrn: I have dropped the `if` and used the same keyword everywhere.

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #2447, opened 2019-08-23
Only test Ubuntu based clusters

arjunrn:
We no longer want to maintain CoreOS based clusters. They can be dropped from the tests.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #2449, opened 2019-08-26
Use linux-aws kernel, increase arp cache threshold

mikkeloscar:
Reverts the Ubuntu AMI to use `linux-aws` (default) kernel. Increases the arp cache thresholds:

```
net.ipv4.neigh.default.gc_thresh1 = 1024
net.ipv4.neigh.default.gc_thresh2 = 4096
net.ipv4.neigh.default.gc_thresh3 = 16384
```

This is done to avoid arp_cache overflows as seen with previous AMI.

PR #2450, opened 2019-08-26
Add daemonset for setting arp_cache thresholds

mikkeloscar:
This adds a daemonset which increases the arp_cache threshold settings on each node.

This is a mitigation for the `arp_cache` overflow issue we have seen since last week. We can roll it out if we continue to see production clusters with this issue.

The real fix is updating the AMI: https://github.com/zalando-incubator/kubernetes-on-aws/pull/2449


mikkeloscar:
Closing at this change was not needed.

PR #2454, opened 2019-08-27
Drop CoreOS master node profile

aermakov-zalando:
We don't run these anymore, so let's just drop the configuration.


cluster/node-pools/master-default/stack.yaml
```diff
@@ -54,12 +49,6 @@ Resources:
           AssociatePublicIpAddress: true
           Groups:
           - !ImportValue '{{ .Cluster.ID }}:master-security-group'
-        BlockDeviceMappings:
```
mikkeloscar, line 57: Why is this not needed?

aermakov-zalando: This is now managed by the AMI. We could still customise it, but IIRC we decided not to initially since the CoreOS and Ubuntu pools used the same template but different device names.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2456, opened 2019-08-27
RBAC: rollout [2/2]

mikkeloscar:
This is the first step to remove all explicit allows from the authorization webhook.

The goal is to only use the webhook for explicit denies in cases where we can't express this via RBAC e.g.:
> give read access to all secrets except in kube-system namepase

Before the webhook had a mix of allow and deny rules which made it sometime overlap with the corresponding RBAC rules defined in kubernetes-on-aws. This is both confusing and dangerous because it makes it easy to change something in one place and not the other.
Now the webhook will only do explicit denies e.g. by blocking any access to secrets in kube-system e.g. for `PowerUser` role. everything else is passed through to RBAC and either explicitly whitelisted via the RBAC role+binding or just not allowed if RBAC doesn't allow it.

~Note that `cdp` service account, `operator` serviceaccount and `default` serviceaccount still has explicit allow rules implemented in the webhook. This is because we can't express those rules in RBAC e.g. give read access to `default` serviceaccount for all namespaces.
We should fix those remaining cases by migrating users away from these special "legacy" serviceaccounts, but this is a bigger task not directly part of this PR but part of the RBAC epic.~

Depends on: https://github.com/zalando-incubator/kubernetes-on-aws/pull/2507

## TODO

* [x] Update webhook image
* [x] Update legacy SA config items for all clusters!

### DIFF
--- a/cluster/manifests/roles/poweruser-role.yaml
+++ b/cluster/manifests/roles/poweruser-role.yaml
@@ -838,3 +838,18 @@ rules:
   - create
   - update
   - patch
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - roles
+  - rolebindings
+  - clusterroles
+  - clusterrolebindings
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
--- a/test/e2e/authorisation_test.go
+++ b/test/e2e/authorisation_test.go
@@ -520,38 +520,39 @@ var _ = framework.KubeDescribe("Authorization tests", func() {
 				},
 			},
 
-			//- poweruser can not use privileged PSP
-			{
-				msg: "access to use privileged PodSecurityPolicy for PowerUser should not be allowed",
-				reqBody: fmt.Sprintf(`{
-					"apiVersion": "authorization.k8s.io/v1beta1",
-					"kind": "SubjectAccessReview",
-					"spec": {
-					"resourceAttributes": {
-						"name": "privileged",
-						"namespace": "",
-						"verb": "use",
-						"group": "extensions",
-						"resource": "podsecuritypolicies"
-					},
-					"user": "sszuecs",
-					"group": [
-						"%s"
-					]
-					}
-				}`, powerUserGroup),
-				expect: expect{
-					status: http.StatusCreated,
-					body: fmt.Sprintf(`{
-					"apiVersion": "authorization.k8s.io/v1beta1",
-					"kind": "SubjectAccessReview",
-					"status": {
-						"allowed": false,
-						"reason":"unauthorized access sszuecs/[%s]"
-					}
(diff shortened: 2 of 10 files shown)


mikkeloscar:
:+1:

cluster/manifests/roles/poweruser-role.yaml
```diff
@@ -838,3 +838,18 @@ rules:
...
+  - clusterroles
+  - clusterrolebindings
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
```
njuettner, line 855: OOC, in which case do you need the delete those resources as a poweruser?

mikkeloscar: Deleting would be needed if you want to cleanup resources that you no longer need. e.g. you created a `clusterrole` to try out an operator, but then later decide to remove it.

I thought about also removing `delete` but if you can `update`/`patch` you can cause the same "harm" anyway so I don't see that it brings much value to prevent `delete`.

njuettner:
👍

cluster/manifests/roles/poweruser-role.yaml
njuettner: thanks, feel free to merge

szuecs: cleanup your deployment,svc,ingress if you move to a different namespace, f.e.

test/e2e/authorisation_test.go
```diff
@@ -552,38 +495,39 @@ var _ = framework.KubeDescribe("Authorization tests", func() {
...
-					"kind": "SubjectAccessReview",
-					"status": {
-						"allowed": false,
-						"reason":"unauthorized access sszuecs/[%s]"
-					}
-				}}`, powerUserGroup),
-				},
-			},
+			////- poweruser can not use privileged PSP
+			//// TODO: disable privileged PSP access for PowerUsers.
```
szuecs, line 499: why do we need to enable this?

mikkeloscar: We still allow PowerUers to use privileged PSP in order to be able to create pods which uses sysctl commands e.g. for setting `vm.max_map_count`. Once we find away to allow this setting without privileged PSP (or remove the need for users to set this) we can disable privileged PSP for PowerUser.

szuecs: Some sysctl  values make sense to change for different workloads. Maybe we can enable users to only specify container/process based sysctl settings.

mikkeloscar: I agree we can think of this, but I would really like to do this as part of another task to not increase the scope of this too much.

This change doesn't give more permissions than what already exists.

szuecs: fine for me

PR #2377, opened 2019-07-30
Add node-problem-detector to cluster setup

linki:
Adds the node problem detector config files.

PR #2458, opened 2019-08-28
Cut off resources from cadvisor in favour of node-problem-detector

linki:
Follow up to https://github.com/zalando-incubator/kubernetes-on-aws/pull/2377.

Take some resources off `cadvisor` and account them for `node-problem-detector`. This keeps the overall usage of our components like before. We currently reserve more resources than we need for `cadvisor` so it seems safe to keep it from there.

Values for `node-problem-detector` detector are taken from current usage in `dev` channel but probably need to be adjusted for other clusters.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -79,8 +79,12 @@ image_policy: "dev"
 {{end}}
 
 # cadvisor settings
-cadvisor_cpu: "150m"
-cadvisor_memory: "150Mi"
+cadvisor_cpu: "125m"
+cadvisor_memory: "125Mi"
+
+# node-problem-detector settings
+node_problem_detector_cpu: "25m"
+node_problem_detector_memory: "25Mi"
 
 # node exporter settings
 node_exporter_cpu: "20m"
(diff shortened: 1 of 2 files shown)


cluster/config-defaults.yaml
```diff
@@ -79,8 +79,12 @@ image_policy: "dev"
...
 
 # cadvisor settings
-cadvisor_cpu: "150m"
-cadvisor_memory: "150Mi"
+cadvisor_cpu: "125m"
+cadvisor_memory: "125Mi"
+
+# node-problem-detector settings
+node_problem_detector_cpu: "25m"
+node_problem_detector_memory: "25Mi"
```
mikkeloscar, line 87: I think `50Mi` is the minimum enforced by our admission controller in other namespaces because it mitigates issues in docker/kubelet. It may be ok to set it lower than `50Mi` @aermakov-zalando WDYT?

aermakov-zalando: It should be fine as long as it runs, but we did have problems in the past even with the most basic things (a container running shell+cat) due to how caches are accounted for. Let's see if it works.

linki:
One collector als needs `lsblk` which is not in the container image.

```console
apt-get install util-linux
```

EDIT: follow up: https://github.com/zalando-incubator/kubernetes-on-aws/pull/2462

arjunrn:
:+1:

mikkeloscar:
:+1:

PR #2459, opened 2019-08-28
Add 'PowerUser' binding for cdp service account

mikkeloscar:
Adds a role binding to give `cdp` service account of the default namespace `PowerUser` permissions. These permissions are used by cdp deployment pods.

We roll this out via kubernetes-on-aws because we only need the binding for the `cdp` service account in the `default` namespace and because having it here (instead of created by cdp-controller) means the `cdp-controller` can run with fewer permissions (it can escalate permissions by creating pods, but this may change in the future if cdp deployments are implemented differently).


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2460, opened 2019-08-28
Updated the Ubuntu Image to use persistent paths for S3 downloads

arjunrn:
The configuration to download files from s3 should be persisted across reboots. The `/var/run/` directory is temporary and it's contents are lost after a reboot.

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -22,7 +22,7 @@ write_files:
       KUBELET_ROLE=master
 
   - owner: root:root
-    path: /var/run/s3-certs.env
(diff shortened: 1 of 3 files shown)


cluster/node-pools/master-default/userdata.yaml
```diff
@@ -22,7 +22,7 @@ write_files:
       KUBELET_ROLE=master
 
   - owner: root:root
-    path: /var/run/s3-certs.env
+    path: /var/kuberuntu/s3-certs.env
```
mikkeloscar: The path is `/etc/kuberuntu/s3-certs.env` in the AMI

arjunrn:
:+1:

aermakov-zalando:
👍

PR #2465, opened 2019-08-29
Node exporter should read node stats from mounted host paths

arjunrn:
Currently node exporter is not reading the node information correctly from the `sysfs` and `procfs` correctly. Mounting them from the host will expose correct metrics.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2471, opened 2019-08-30
Fixed which mounts are monitored by the node exporter

arjunrn:
Mount the host root as `readOnly` to correctly find the mounted disks. Also exclude all docker and kubelet mounts.


arjunrn:
👍

mikkeloscar:
:+1:

PR #2481, opened 2019-09-03
Add per-cluster admin role back

aermakov-zalando:
A number of teams are apparently using this. Add back for now, we'll discuss how we want to approach it later.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2483, opened 2019-09-03
Update to Kubernetes v1.15

mikkeloscar:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#v1150

Introduces Kubernetes `v1.15.6`.

## TODO

* [x] Update Ubuntu AMI to `v1.15.x`
* [x] Update e2e tests to be based on `v1.15.x`
* [x] The system:aws-cloud-provider cluster role, deprecated in v1.13, is no longer auto-created.
* [x] The deprecated flag --conntrack-max has been removed from kube-proxy. Users of this flag should switch to --conntrack-min and --conntrack-max-per-core instead. (https://github.com/kubernetes/kubernetes/pull/78399)
* [x] The StorageObjectInUseProtection admission plugin is enabled by default, so the default enabled admission plugins are now NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,DefaultTolerationSeconds,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota,StorageObjectInUseProtection. Please note that if you previously had not set the --admission-control flag, your cluster behavior may change (to be more standard).
* [x] Fix deprecated kubelet node labels which will be removed in v1.16
* [x] Ingress resources will no longer be served from extensions/v1beta1 in v1.19. Migrate use to the networking.k8s.io/v1beta1 API, available since v1.14. Existing persisted data can be retrieved via the networking.k8s.io/v1beta1 API.
* [x] PodSecurityPolicy resources will no longer be served from extensions/v1beta1 in v1.16. Migrate to the policy/v1beta1 API, available since v1.10. Existing persisted data can be retrieved via the policy/v1beta1 API.
* [x] PriorityClass resources will no longer be served from scheduling.k8s.io/v1beta1 and scheduling.k8s.io/v1alpha1 in v1.17. Migrate use to the scheduling.k8s.io/v1 API, available since v1.14. Existing persisted data can be retrieved via the scheduling.k8s.io/v1 API.


arjunrn:
After the last version update we talked about having multiple fields for the AMI so that AMI changes to base `dev` branch do not affect the version upgrade branch.

mikkeloscar:
> After the last version update we talked about having multiple fields for the AMI so that AMI changes to base `dev` branch do not affect the version upgrade branch.

Fine by me, I just think I was not in this discussion and I don't fully understand how it should work? Could you give an example?

mikkeloscar:
@arjunrn I changed it like we discussed.

hjacobs:
What's the ETA for merge?

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2484, opened 2019-09-04
kube-downscaler: add VPA

aermakov-zalando:
It scales with the number of resources, and we don't want to babysit it.

### DIFF
--- a/cluster/manifests/kube-downscaler/vpa.yaml
+++ b/cluster/manifests/kube-downscaler/vpa.yaml
@@ -1,0 +1,20 @@
+apiVersion: autoscaling.k8s.io/v1beta2
+kind: VerticalPodAutoscaler
+metadata:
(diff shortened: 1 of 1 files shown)


cluster/manifests/kube-downscaler/vpa.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: autoscaling.k8s.io/v1beta2
+kind: VerticalPodAutoscaler
+metadata:
+  name: kube-downscaler
+  namespace: visibility
```
hjacobs: visibility?

aermakov-zalando: Oops :(

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2485, opened 2019-09-04
add log-slo-heartbeat sidecar to logging-agent ds

pheanex:
Add "log-slo-heartbeat"-sidecar in order to produce data to monitor fulfillment of logging SLO
@zalando-incubator/team-kubernetes-log-watcher 
Please review


cluster/config-defaults.yaml
```diff
@@ -117,10 +117,12 @@ zmon_worker_cpu: "750m"
 zmon_worker_count: "16"
 {{end}}
 logging_watcher_mem: "200Mi"
-logging_scalyr_mem: "200Mi"
+logging_scalyr_mem: "190Mi"
+logging_slo_heartbeat_mem: "10Mi"
```
aermakov-zalando: The minimum currently is `25Mi`, so this will be an effective increase of `15Mi`. Please check if it could be rolled out to existing clusters.

pheanex: Works with these settings in playground.
Btw.: Is 25Mi the limit for pods or containers? (I was assuming the former)

aermakov-zalando: It works, but the container will be started with a 25Mi request/limit. This might be too much in some of the clusters, rendering user pods unschedulable. The minimum request is per-container since pods don't have memory requests/limits.

pheanex:
:+1:

cluster/config-defaults.yaml
```diff
@@ -117,10 +117,12 @@ zmon_worker_cpu: "750m"
 zmon_worker_count: "16"
 {{end}}
 logging_watcher_mem: "200Mi"
-logging_scalyr_mem: "200Mi"
+logging_scalyr_mem: "175Mi"
+logging_slo_heartbeat_mem: "10Mi"
```
mikkeloscar: could you also bump it to `25Mi` to not add confusion later on? :)

pheanex: Sure can do. Just wanted to avoid any further work in case you decrease the minimum to sth lower than 25Mi :smile:

pheanex:
:+1:

mikkeloscar:
:+1:

aermakov-zalando:
@pheanex Have you checked if the increase is safe?

mikkeloscar:
Right the CPU increased, but the Memory doesn't increase anymore.

pheanex:
@aermakov-zalando  what do you mean by safe?

pheanex:
@mikkeloscar do you suggest to also decrease the cpu-request by 10m from one of the other 3 containers?

astarnell:
@pheanex please increase the ds version

pheanex:
:+1:

pheanex:
:+1:

astarnell:
:+1:

aermakov-zalando:
👍

szuecs:
:+1:

PR #2486, opened 2019-09-04
Give collaborators access to nodes and daemonsets

aermakov-zalando:
Production: require both global 24x7 Emergency and per-cluster Emergency/Manual. Test: just require global Deployer.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2490, opened 2019-09-05
Rename kuberuntu_image var to prevent merge conflicts

mikkeloscar:
This just renamed the `kuberuntu_image` variable such that when we prepare for version upgrades we have less change of merge conflicts in case were we update both the AMI for version upgrade in a new branch and also update the AMI for fixes to the old version.

Ref: https://github.com/zalando-incubator/kubernetes-on-aws/pull/2483#issuecomment-527829603


arjunrn:
👍

aermakov-zalando:
👍

PR #2504, opened 2019-09-10
Make e2e tests build with go1.13

mikkeloscar:
Moves the flag handling away from `init()` such that the e2e tests can build with go1.13. It's annoying to build locally if it only works with go1.12 :)


mikkeloscar:
:+1:

mikkeloscar:
:+1:

szuecs:
:+1:

PR #2506, opened 2019-09-11
Revert "Update auth webhook to v0.6.2"

aermakov-zalando:
This breaks a few normal deployments that are running in `visibility`. Reverting temporarily until they're moved out.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2508, opened 2019-09-11
Added VPA for kube-janitor only for non-prod clusters

arjunrn:
`kube-janitor` only runs in non-production clusters.


aermakov-zalando:
👍

arjunrn:
👍

PR #2510, opened 2019-09-12
RBAC: rollout [1/2]

mikkeloscar:
This PR is part of the complete RBAC migration: #2456 It must be rolled out first to ensure we don't drop permissions during a cluster update.

**NOTE: these changes doesn't actually take effect until #2456 is rolled out**

It does the following:

* Cleans up old/legacy roles/bindings
* Defines a `poweruser` role with all the general permissions needed by groups: `PowerUser`, `Manual`, `Emergency`.
* Defines a `readonly` role with all the general read permissions needed by groups: `ReadOnly`.
* Updates the admission-controller to create RBAC roles whenever a CRD is created in the cluster. The roles are aggregated to `poweruser` and `readonly` respectively.
* Grants `daemonset` access to `Collaborator` groups for `visibility` namespace only.

## TODO
* [x] update admission-controller image.


mikkeloscar:
:+1:

cluster/manifests/roles/poweruser-role.yaml
```diff
@@ -0,0 +1,846 @@
...
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - impersonate
```
szuecs: why?

cluster/manifests/roles/poweruser-role.yaml
```diff
@@ -0,0 +1,846 @@
...
+  resources:
+  - poddisruptionbudgets
+  verbs:
+  - list
+- apiGroups:
+  - policy
+  resources:
+  - poddisruptionbudgets
+  verbs:
+  - patch
```
szuecs: Why do they need to change a PSP?

cluster/manifests/roles/poweruser-role.yaml
mikkeloscar: Good catch, this list was generated from the official `edit` role which is default in Kubernetes. I missed this, and will remove it now after a better review of the permissions in this list.

mikkeloscar: (It would have been blocked by our webhook though) :)

szuecs:
:+1:

mikkeloscar:
:+1:

PR #2511, opened 2019-09-12
Always run CLC

aermakov-zalando:
`rolling` strategy will be dropped from CLM at some point, so CLC always needs to run.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2515, opened 2019-09-13
logging-agent: support new node configuration [1/2]

aermakov-zalando:
In the new AMI, Docker root directory was moved to a different location. This needs to be merged before we rotate the nodes so the logging agent will work for both node types. I've created a separate daemonset for old nodes, which we can drop after the migration is complete.


aermakov-zalando:
👍

anton-ryzhov:
This change shouldn't trigger reingestion of logs into scalyr nor into S3. But definitely worth to check this after first deployments

christianberg (approved):

christianberg:
👍

arjunrn:
:+1:

PR #2522, opened 2019-09-17
Update kubernetes-lifecycle-metrics

aermakov-zalando:
Fix the bug with pods that were created in the failed state treated as active.


linki:
:+1:

aermakov-zalando:
👍

PR #2533, opened 2019-09-18
Fix RBAC for dashboard/admission controller proxy

aermakov-zalando:
* dashboard: don't overwrite audittrail adapter's role definitions
* admission-controller: the daemonset is only for the oauth injecting proxy, which doesn't access any Kubernetes APIs (but needs to be able to use host networking). drop the cluster role for it and replace with a binding for the privileged PSP role.


szuecs:
:+1:

aermakov-zalando:
👍

arjunrn:
👍

PR #2541, opened 2019-09-19
Add additional capabilities to privileged psp

linki:
This adds additional Linux capabilities to the allowed set for the privileged psp.

I want to discuss if this is a feasible change.

### DIFF
--- a/cluster/manifests/psp/pod_security_policy.yaml
+++ b/cluster/manifests/psp/pod_security_policy.yaml
@@ -7,6 +7,9 @@ spec:
     rule: RunAsAny
   privileged: true
   allowPrivilegeEscalation: true
+  allowedCapabilities:
+  - IPC_LOCK
+  - SYS_RESOURCE
   hostPID: true
   hostNetwork: true
   hostIPC: true
--- a/cluster/manifests/psp/pod_security_policy.yaml
+++ b/cluster/manifests/psp/pod_security_policy.yaml
@@ -7,9 +7,6 @@ spec:
     rule: RunAsAny
   privileged: true
   allowPrivilegeEscalation: true
-  allowedCapabilities:
-  - IPC_LOCK
-  - SYS_RESOURCE
   hostPID: true
   hostNetwork: true
   hostIPC: true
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
What are the use cases for these?

cluster/manifests/psp/pod_security_policy.yaml
```diff
@@ -7,6 +7,9 @@ spec:
     rule: RunAsAny
   privileged: true
   allowPrivilegeEscalation: true
+  allowedCapabilities:
+  - IPC_LOCK
+  - SYS_RESOURCE
```
szuecs: Does this mean /sys access?

cluster/manifests/psp/pod_security_policy.yaml
```diff
@@ -7,6 +7,9 @@ spec:
     rule: RunAsAny
   privileged: true
   allowPrivilegeEscalation: true
+  allowedCapabilities:
+  - IPC_LOCK
```
szuecs: lgtm

cluster/manifests/psp/pod_security_policy.yaml
szuecs: this could be harmful, because the process is allowed to increase RLIMIT_NPROC and RLIMIT_NOFILE

What is the reason you suggest this?

linki:
These are common requirements for running Cassandra:
* https://github.com/instaclustr/cassandra-operator/blob/294b16080c841404843cffbb6fc03badb95d816e/java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java#L151-L152
* https://github.com/Orange-OpenSource/cassandra-k8s-operator/blob/6747d42f73bc390615d34d0ddafd161a14184f05/pkg/controller/cassandracluster/testdata/cassandracluster-2DC-dc1-rack1-sts.yaml#L181

This can stay here for documentation purposes until we're sure what do to. Just didn't want to forget it.

linki:
I removed the changes from the existing PSPs and created an `elevated` PSP that is like the restricted one but with the additional CAPs. Currently it'll only be created in test clusters if this gets merged.

linki:
In our setup `restricted` is for normal Pods (user's apps) and `privileged` is for everything else that really requires more permissions, like cluster components, certain high-prio customers of us.

Given that and looking at the very broad permissions of `privileged` already I thought giving it more capabilities is just fine.

Of course, I'm in favour of having a more fine grained solution in the future.

linki:
`v2.0` of instaclustr's cassandra operator allows to specify a config without needing the capabilities specified here (by accepting a performance hit).

I'm closing this for now as it's not required anymore.

PR #2542, opened 2019-09-19
Remove the VPA for zmon-scheduler

arjunrn:
Remove the VPA for zmon-scheduler


hjacobs:
What's the reason for this? I'm interested in the VPA experience :smile:

arjunrn:
@hjacobs this is what happens.
Screen Shot 2019-09-19 at 17 52 30

As seen in the graph the VPA learnt the correct memory usage of the application and then it reduced the memory. But after reducing the memory the scheduler used much more. As a result it kept getting OOM killed. There is a flag in the vpa which can be used to evict VPA managed pods which are being OOM killed but it is only for pods with single container(the scheduler uses 2 containers). So the pod keeps getting OOM killed till the VPA is "confident" that the memory usage is in fact higher and evicts/recreates it again with higher memory. This is not a problem in the application itself but due to how Java manages memory. It might be using much more than it actually needs in order to optimize performance.

szuecs:
:+1:

pitr:
I wouldn't say VPA learned the "correct memory usage". Scheduler uses close to 3G of memory during the first day, then it stabilizes at 1G. 1.8G is not the "correct memory usage".

VPA not being aware of multi container setups seems like a huge limitations. Also this introduces potential issues if VPA enabled single-container pod get changed into multi-container setup.

arjunrn:
@pitr The VPA algorithm is not magical but basis its recommendations on recent usage patterns. And it is a safe assumption to make that given the same workload, the CPU and memory usage should remain the same(with the exception of JVM applications apparently).

> VPA not being aware of multi container setups seems like a huge limitations.

That is a misunderstanding. With a multicontainer pod the VPA cannot make a call if the OOM kill should trigger an eviction because the OOMing container might be non-critical. Hence it does not intervene.

mikkeloscar:
:+1:

PR #2548, opened 2019-09-24
PCS schema: define the schema for tokens

aermakov-zalando:
We can use `nullable: true` now, which means we can define the schema for the tokens.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2549, opened 2019-09-24
E2E: don't run on Spot

aermakov-zalando:
They take up to an hour, and we don't want them aborted halfway due to AWS taking an instance back.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2554, opened 2019-09-25
New directories for docker/kubelet

aermakov-zalando:
Right now they use the bind mount, but we'd like to get rid of it.


szuecs:
:+1:

aermakov-zalando:
👍

PR #2564, opened 2019-09-27
Keep the system service account for safe RBAC rollout

mikkeloscar:
Keep the system service account for safe RBAC rollout


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2569, opened 2019-09-30
problem-detector: Don't match on version label, e2e test

linki:
uses the correct labels selectors and deploys it during e2e to detect errors with deployment files.


aermakov-zalando:
👍

linki:
:+1:

PR #2570, opened 2019-09-30
Adapt problem detector docker paths to ubuntu setup

linki:
Problem detector default configs use `/var/lib/docker` to detect pattern in logs. This adapts the configs to use our path.

### DIFF
--- a/cluster/manifests/node-problem-detector/ds-node-problem-detector.yaml
+++ b/cluster/manifests/node-problem-detector/ds-node-problem-detector.yaml
@@ -7,6 +7,8 @@ metadata:
   labels:
     application: node-problem-detector
     version: v0.7.1
+  annotations:
+    config/hash: {{ "cm-node-problem-detector.yaml" | manifestHash }}
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
Do you mind adding the configmap hash to the pod template (see [here](https://github.com/zalando-incubator/kubernetes-on-aws/blob/3789f70617a301dfd205496bc06ff837bd063d7d/cluster/manifests/prometheus/statefulset.yaml#L22) for example)? This is a hack we have to make the pods update if the configmap is changed.

linki:
@aermakov-zalando Thanks, added it.

cluster/manifests/node-problem-detector/ds-node-problem-detector.yaml
```diff
@@ -7,6 +7,8 @@ metadata:
   labels:
     application: node-problem-detector
     version: v0.7.1
+  annotations:
```
aermakov-zalando: This needs to be in the pod template, otherwise nothing will actually be rotated.

linki: :man_facepalming:

linki: fixed

aermakov-zalando:
👍

szuecs:
:+1:

PR #2571, opened 2019-10-01
Update kube-metrics-adapter

aermakov-zalando:
Changes:
 * Return an error for `request-per-second` metric without a backend name for ingresses that use traffic switching.


aermakov-zalando:
👍

szuecs:
:+1:

PR #2586, opened 2019-10-08
Make autoscaler resource usage configurable

aermakov-zalando:
Not sure if I'd like VPA to constantly restart it, so let's still run with static resources for now.


szuecs:
@aermakov-zalando something missing to use the parameters?
https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/kube-cluster-autoscaler/daemonset.yaml#L52-L58 does not show that we use it already

szuecs:
:+1:

aermakov-zalando:
👍

PR #2592, opened 2019-10-09
skipper: restart when the lightstep token changes

aermakov-zalando:
Otherwise we'll have to wait until the next node rotation.


aermakov-zalando:
👍

szuecs:
:+1:

PR #2595, opened 2019-10-09
[WIP] Encrypt secrets in etcd using KMS

linki:
**work in progress, e2e test run**

Pass a simple encryption configuration to apiserver where the encryption key is stored in cluster registry.

This protects against an attacker that gained access to an etcd instance to read all data. It doesn't protect against a compromised Kubernetes master. ~~A fully fledged solution would be to [integrate with AWS KMS](https://kubernetes.io/docs/tasks/administer-cluster/kms-provider/), e.g. https://github.com/kubernetes-sigs/aws-encryption-provider~~

AWS KMS based encryption has been added to this PR.

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -106,6 +106,9 @@ write_files:
           - --etcd-prefix={{ .Cluster.ConfigItems.apiserver_etcd_prefix }}
           - --storage-backend=etcd3
           - --storage-media-type=application/vnd.kubernetes.protobuf
+          {{ if eq .Cluster.ConfigItems.enable_encryption "true" }}
+          - --encryption-provider-config=/etc/kubernetes/config/encryption-config.yaml
+          {{ end }}
           - --allow-privileged=true
           - --service-cluster-ip-range=10.3.0.0/16
           - --secure-port=443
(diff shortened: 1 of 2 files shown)


cluster/node-pools/master-default/userdata.yaml
```diff
@@ -711,3 +714,21 @@ write_files:
...
+    content: |
+      apiVersion: apiserver.config.k8s.io/v1
+      kind: EncryptionConfiguration
+      resources:
+        - resources:
+          - secrets
+          providers:
+          - aescbc:
+              keys:
+              - name: key1
```
njuettner, line 731: Do we want to change the key name?

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -711,3 +714,21 @@ write_files:
...
+    path: /etc/kubernetes/config/encryption-config.yaml
+    permissions: '0400'
+    content: |
+      apiVersion: apiserver.config.k8s.io/v1
+      kind: EncryptionConfiguration
+      resources:
+        - resources:
+          - secrets
+          providers:
+          - aescbc:
```
szuecs, line 729: Can we use GCM instead of CBC?

aermakov-zalando: According to [the docs](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/#providers), it won't be secure enough unless we rotate the keys often. I'd just go with KMS, even though that'll require a bit more setup.

linki:
We can postpone this PR. I wanted to see how the basic setup works and whether e2e passes. It's surprisingly simple and I'm happy to demo it.

KMS is the natural choice since we're on AWS. The configs here are half-way to what's needed for KMS. However, it does require some more research.

cluster/node-pools/master-default/userdata.yaml
linki: Agreed, the docs are quite clear about that. We should keep it in mind in case we see performance issues.

cluster/node-pools/master-default/userdata.yaml
linki: Key rotation works by adding a key entry, running it, re-creating all secrets and then removing the old key. Hence, I wanted to go with a "cattle-like" key scheme. I'll address that for the final PR.

linki:
Follow up https://github.com/zalando-incubator/kubernetes-on-aws/pull/2956

PR #2601, opened 2019-10-10
CA/CLC: fix race condition during eviction

aermakov-zalando:
Include pod uid in eviction requests to avoid evicting an already moved pod.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2602, opened 2019-10-10
Proxy plugin was dropped in a previous release from CoreDNS

njuettner:
Instead we need to proceed with the forward plugin


aermakov-zalando:
👍

njuettner:
👍

PR #2534, opened 2019-09-19
Update to v1.14.8

aermakov-zalando:
[Changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md/#v1147)

PR #2604, opened 2019-10-11
Update AMI

njuettner:
Use the latest AMI for Kubernetes 1.14.7:

Changes:

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#changelog-since-v1146


szuecs:
We can't use 1.14.7, because of the ELB/service type loadbalancer issue.
See also https://github.com/zalando-incubator/kubernetes-on-aws/pull/2534#issuecomment-534006128

PR #2605, opened 2019-10-15
cdp-controller-rbac: Read+Patch access to StackSet

thilp:
With https://github.com/zalando-incubator/stackset-controller/issues/159, CDP needs to read and change the traffic map on StackSet resources.

CDP needs that access even before the new StackSet format is rolled out, because it is planned to support both formats at the same time during the transition period. See https://github.bus.zalan.do/automata/cdp/pull/1302 for details.

### DIFF
--- a/cluster/manifests/roles/cdp-controller-rbac.yaml
+++ b/cluster/manifests/roles/cdp-controller-rbac.yaml
@@ -46,6 +46,15 @@ rules:
 - apiGroups:
   - "zalando.org"
   resources:
+  - stacksets
+  verbs:
+  - get
+  - list
+  - watch
+  - patch
+- apiGroups:
+  - "zalando.org"
+  resources:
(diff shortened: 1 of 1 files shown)


cluster/manifests/roles/cdp-controller-rbac.yaml
```diff
@@ -43,6 +43,15 @@ rules:
...
   - create
+- apiGroups:
+  - "zalando.org"
+  resources:
+  - stacksets
+  verbs:
+  - get
+  - list
+  - watch
+  - patch
```
aermakov-zalando, line 54: You probably want `update` as well, since it might be easier to use?

aermakov-zalando:
👍

thilp:
:+1:

mikkeloscar:
:+1:

PR #2607, opened 2019-10-15
Automatically calculate max pods/max nodes

aermakov-zalando:
* Max pods per node: calculate based on node CIDR
* CA: limit the total number of nodes based on node CIDR


aermakov-zalando:
👍

gargravarr:
:+1:

PR #2615, opened 2019-10-18
CA: parameterize the version temporarily

aermakov-zalando:
This will make it easier to test the new version. I'll make it static again after we're comfortable enough.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2619, opened 2019-10-21
Make CIDRs for pod and service network configurable

linki:
Replace hard-coded values for `PodCIDR` and `ServiceCIDR` with config items. This would allow us to test different configurations.

The goal is to change the current `PodCIDR` to a larger subnet such that we can support more nodes per cluster without reducing the pod count per node and therefore going beyond the arbitrary 1k/2k cluster limit.

## Background

The default setting:
```yaml
pod_cidr: "10.2.0.0/16"
node_cidr_mask_size: "24"
```

allows for 251 nodes with 110 pods per node (theoretically 256 nodes but we reserve 5. Also theoretically 128 pods per node (2 IPs are reserved per pod) but Kubernetes limits at 110).

The default production setting:
```yaml
pod_cidr: "10.2.0.0/16"
node_cidr_mask_size: "25"
```

allows for 507 nodes with 64 pods per node.

Our biggest clusters use the following setting:
```yaml
pod_cidr: "10.2.0.0/16"
node_cidr_mask_size: "26"
```

which allows for 1019 nodes with 32 pods per node.

From there it becomes unfeasible to continue with this approach (i.e. `node_cidr_mask_size: "27"`) because there would only be a few pods running on each node.

## Solution

By increasing the size of the Pod network itself we get access to a vast amount of pod IPs to support very large clusters. We also don't need to fiddle with the node_cidr_mask_size because there's just no need for it anymore.

The first step is to make these settings customisable such that we can play around with different combinations in pet clusters.

For instance, we could run the following configuration:
```yaml
pod_cidr: "10.32.0.0/12"
node_cidr_mask_size: "24"
```

which allows for 4091 nodes with 110 pods per node.

This way we can run more than 1000 nodes per cluster. We also don't have to limit the number of pods per node in any way and thus can fully utilize very large instance types.


linki:
Closing for now

linki:
Re-opening from the archives.

linki:
Closing for now again.

Ticket for tracking support for larger clusters: https://github.bus.zalan.do/teapot/issues/issues/3160

PR #2628, opened 2019-10-30
Drop quota from visibility namespace

mikkeloscar:
We decided to drop the quotas a while ago. We forgot about visibility.


mikkeloscar:
:+1:

aermakov-zalando:
:+1:

PR #2630, opened 2019-10-30
remove resource quota on visibility namespace

linki:
We removed quotas for `default` and `kube-system`, but forgot `visibility`. We're running into this limit now.

/cc @hjacobs


njuettner:
👍

mikkeloscar:
Was done in #2628

PR #2631, opened 2019-10-30
Prometheus config item for VPA min CPU

gargravarr:
Signed-off-by: Rodrigo Reis <rodrigo@zalando.de>


mikkeloscar:
The point of this is to be able to bump the CPU suggestions manually in case the VPA is not fast enough.

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #2632, opened 2019-10-30
enable route lifo metrics

aryszka:
This enables metrics for the lifo queues per route. This means that we would have 2 additional metrics for each route, which can blow up the number of prometheus metrics. The reason why we need this change is that we need the queue size from the lifo as an input for scaling the cluster, since it can prevent the scaling based on the memory in extreme cases.


aryszka:
closing this and adding this change to a shared pull request

PR #2633, opened 2019-10-30
enable debug listener

aryszka:
with the debug listener, we can check which route would match a certain request and how would it alter the request flow


aryszka:
closing this and adding this change to a shared pull request

PR #2640, opened 2019-10-31
Remove egress tests for now

aermakov-zalando:
The egress test constantly breaks e2e runs because neither the test itself, nor the egress controller are designed to run side by side with other instances. Let's just drop the test for now.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2641, opened 2019-10-31
Update to the new CA

aermakov-zalando:
* Workers: define scaling priority
* Remove `cluster_autoscaler_version`/`cluster_autoscaler_expander`
* Add `experimental_autoscaler_1_14` instead
* Make sure nodes have ephemeral storage capacity

### DIFF
--- a/cluster/manifests/kube-cluster-autoscaler/daemonset.yaml
+++ b/cluster/manifests/kube-cluster-autoscaler/daemonset.yaml
@@ -5,7 +5,6 @@ metadata:
   namespace: kube-system
   labels:
     application: kube-cluster-autoscaler
-    version: {{.Cluster.ConfigItems.cluster_autoscaler_version}}
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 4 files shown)


linki:
:+1:

aermakov-zalando:
👍

cluster/manifests/kube-cluster-autoscaler/daemonset.yaml
```diff
@@ -5,7 +5,6 @@ metadata:
   namespace: kube-system
   labels:
     application: kube-cluster-autoscaler
-    version: {{.Cluster.ConfigItems.cluster_autoscaler_version}}
```
linki, line 8: afaik, the version label is something we ought to set.

aermakov-zalando: Is it okay if I put it back when we roll it out to all clusters? I could also do a follow-up PR if you'd prefer it the other way around.

linki: Sure, that's fine.

PR #2642, opened 2019-11-01
enable tokeninfo via sidecar in ingress, if set by cluster-registry

szuecs:

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -153,6 +153,15 @@ spec:
           value: "{{ .ConfigItems.opendid_provider_cfg_url }}"
         - name: ISSUER
           value: "{{ .ConfigItems.openid_issuer }}"
+        - name: ENABLE_OPENTRACING
+          value: "true"
+        - name: OPENTRACING_LIGHTSTEP_COMPONENT_NAME
+          value: "tokeninfo-skipper-ingress"
+        - name: OPENTRACING_LIGHTSTEP_ACCESS_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: skipper-ingress
+              key: lightstep-token
 {{ end }}
 {{ if eq .ConfigItems.enable_apimonitoring "true"}}
       volumes:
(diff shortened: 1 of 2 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -128,6 +128,41 @@ spec:
           - name: filters
             mountPath: /etc/config/default-filters
 {{ end }}
+{{ if eq .ConfigItems.tokeninfo_url "http://127.0.0.1:9021/oauth2/tokeninfo" }}
```
mikkeloscar, line 131: nit: I think it would be simpler if we just had a config item: `enable_local_tokeninfo` otherwise you have to remember that correct value of `http://127.0.0.1:9021/oauth2/tokeninfo` to configure it.

cluster/manifests/skipper/deployment.yaml
```diff
@@ -128,6 +128,41 @@ spec:
...
+            port: 9021
+          periodSeconds: 3
+          failureThreshold: 2
+        resources:
+          limits:
+            cpu: 1000m
+            memory: 64Mi
+          requests:
+            cpu: 1000m
+            memory: 64Mi
```
aermakov-zalando: Should we increase the memory requirements a bit, since we're already provisioning a ton of CPU?

szuecs: I just used values from their configuration, wdyt?
Do you have any suggestion?
I think the skipper auth connection pool does not allow connection spikes, so I thought this could be fine.

cluster/manifests/skipper/deployment.yaml
szuecs: But this will need to create a lot more configuration overhead, because you would need to have it in multiple places.
For me this was more DRY, so I disagree with being a nit.

cluster/manifests/skipper/deployment.yaml
aermakov-zalando: Why not? A CPU reservation of `1` already reserves 25% of our default instance, so the memory would be unused anyway.

szuecs: default now 256Mi but configurable via CR

szuecs:
:+1:

mikkeloscar:
:+1:

PR #2644, opened 2019-11-01
Use parallel management policy for prometheus

linki:
Currently our two Prometheus replicas use the `OrderedReady` management strategy. This leads to `prometheus-1` not being started if `prometheus-0` isn't ready. Since there's no dependency between the two it's perfectly fine to use the parallel strategy which starts both instances at the same time.

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -1,5 +1,8 @@
 # everything defined under here will be deleted before applying the manifests
-pre_apply: []
+pre_apply:
+- name: prometheus
+  namespace: kube-system
+  kind: StatefulSet
 
 # everything defined under here will be deleted after applying the manifests
 post_apply:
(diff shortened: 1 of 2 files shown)


linki:
This fails because `podManagementPolicy` is an immutable field.

The Prometheus StatefulSet needs to be deleted in order to update that field. Deleting a StatefulsSet retains the volumes so there's no data lost.

szuecs:
@linki we do not want the behavior on tear down https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#parallel-pod-management
As documented this will be a full outage of prometheus on every deployment of Prometheus

linki:
@szuecs yikes, thanks. That would be bad.

Though it also states:
> This option only affects the behavior for scaling operations. Updates are not affected.

I'll test that out later.

linki:
@szuecs It does a normal rolling update when the manifest is changed. I think that's good to go.

szuecs:
:+1:

linki:
:+1:

PR #2649, opened 2019-11-01
Updated etcd backup cronjob to run every minute.

gargravarr:
Signed-off-by: Rodrigo Reis <rodrigo@zalando.de>

### DIFF
--- a/cluster/manifests/etcd-backup/cronjob.yaml
+++ b/cluster/manifests/etcd-backup/cronjob.yaml
@@ -7,7 +7,7 @@ metadata:
     application: etcd-backup
     version: "master-9"
 spec:
-  schedule: "23 * * * *"
+  schedule: "*/1 * * * *"
   concurrencyPolicy: Forbid
   failedJobsHistoryLimit: 1
   jobTemplate:


hjacobs:
Do we backup every minute then? Is there any potential negative impact?

gargravarr:
Yes, it takes about 20 seconds to complete the job.

gargravarr:
Let's try it in dev channel and see what load it puts on etcd cluster. We can then adjust it based on that.

gargravarr:
@hjacobs actually we've just set stups-test channel to run based on this branch over the weekend. Monday we'll see how it went.

mikkeloscar:
Note: with the current retention policy on the s3 bucket (7 days) this will change the storage size from `~6Gi` -> `~350Gi` based on one example (search cluster).

hjacobs:
I think it would make sense to use something more reasonable like 10 minutes as interval?

gargravarr:
@mikkeloscar what is the value of having a 7 days retention policy? Is it safe to reduce this to 1 or 2 days?

gargravarr:
@hjacobs etcd-backup job takes 20s to complete. Backing up every minute will increase the probability of restoring etcd to a consistent state, in case of failure.

mikkeloscar:
> @mikkeloscar what is the value of having a 7 days retention policy? Is it safe to reduce this to 1 or 2 days?

@gargravarr considering we don't trust backups that are 1 hour old there's no real value in keeping the backups for that long. It's only for safety IMO i.e. if we don't notice that an etcd cluster is down for some hours, then we still have "some" backups available which is better than nothing.

It's IMO reasonable to lower the retention to 48 hours.

gargravarr:
:+1:

aermakov-zalando:
👍

PR #2658, opened 2019-11-06
Update logging-agent

aermakov-zalando:
Changes:
 * Reduce the load on the API server by doing GET requests instead of LIST.
 * Send a better User-Agent.


aermakov-zalando:
👍

njuettner:
👍

PR #2663, opened 2019-11-07
Update `kubernetes-log-watcher`

anton-ryzhov:
Update `kubernetes-log-watcher` (https://github.com/zalando-incubator/kubernetes-log-watcher/pull/101)
Add `log-watcher-config`


anton-ryzhov:
:+1:

astarnell:
:+1:

cluster/manifests/logging-agent/daemonset.yaml
```diff
@@ -237,6 +241,10 @@ spec:
         hostPath:
           path: /var/log/journal
 
+      - name: log-watcher-config
+        configMap:
+          name: log-watcher-config
```
aermakov-zalando, line 247: This should be marked as optional and the configmap itself not deployed via CLM. Otherwise any changes that you make will get overwritten every time we roll out something to the clusters, which is probably not something you want.

cluster/manifests/logging-agent/daemonset.yaml
```diff
@@ -132,6 +134,8 @@ spec:
         - name: containerlogs
           mountPath: /mnt/containers
           readOnly: true
+        - name: log-watcher-config
+          mountPath: /mnt/log-watcher-config
```
aermakov-zalando, line 138: `readOnly: true`

cluster/manifests/logging-agent/daemonset.yaml
anton-ryzhov: We decided we don't want to have this change persisted after cluster redeployment, at least now. This is more emergency option to mitigate unexpected things.

anton-ryzhov:
:+1:

cluster/manifests/logging-agent/daemonset.yaml
aermakov-zalando: Cluster redeployments happen very often, sometimes multiple times per day. We also don't notify anyone for most of them. Are you really sure you don't want to do it with an optional configmap?

anton-ryzhov: "This is fine" 😀

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2666, opened 2019-11-07
Drop limits in cluster autoscaler

mikkeloscar:
Drop limits in cluster autoscaler to prevent OOMkill in big clusters.

We decided to not run with limits at it's on the master nodes which we control and we don't see spikes/leaks in the cluster-autoscaler.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2675, opened 2019-11-11
Use issuer as defined in tokens for production

mikkeloscar:
Only the `provider_cfg_url` should be updated. We still have the same issuer as before.

This fixes an issue in the tokeninfo skipper sidecar configuration.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #2682, opened 2019-11-13
vpa-admission-controller: restart when the certificate changes

aermakov-zalando:
Otherwise it might still be running with an expired certificate.


aermakov-zalando:
👍

szuecs:
:+1:

PR #2685, opened 2019-11-13
Make experimental autoscaler version configurable

linki:
Makes the experimental autoscaler version configurable. I believe we'll need change this more often in the coming weeks.


linki:
:+1:

mikkeloscar:
:+1:

PR #2687, opened 2019-11-13
Default storage class: allow volume expansion

aermakov-zalando:
This works fine, and can help us recover from incidents.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2689, opened 2019-11-14
Use old version of autoscaler, make expander configurable

linki:
switches back to autoscaler v1.12 which comes with the priority expander included.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -13,7 +13,7 @@ autoscaling_buffer_pods: "0"
 cluster_autoscaler_cpu: "100m"
 cluster_autoscaler_memory: "300Mi"
 
-cluster_autoscaler_version: v1.12.2-internal.5
+# defines which expander the autoscaler should use
 cluster_autoscaler_expander: highest-priority
 
 # ALB config created by kube-aws-ingress-controller
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -13,8 +13,8 @@ autoscaling_buffer_pods: "0"
 cluster_autoscaler_cpu: "100m"
 cluster_autoscaler_memory: "300Mi"
 
-cluster_autoscaler_version: v1.12.2-internal.4
-cluster_autoscaler_expander: prefer-spot
+cluster_autoscaler_version: v1.12.2-internal.5
+cluster_autoscaler_expander: highest-priority
 
 # ALB config created by kube-aws-ingress-controller
 kube_aws_ingress_controller_ssl_policy: "ELBSecurityPolicy-TLS-1-2-2017-01"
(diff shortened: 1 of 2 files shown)


cluster/manifests/kube-cluster-autoscaler/daemonset.yaml
```diff
@@ -31,7 +33,7 @@ spec:
         effect: NoSchedule
       containers:
       - name: cluster-autoscaler
-        image: registry.opensource.zalan.do/teapot/kube-cluster-autoscaler:{{if eq .Cluster.ConfigItems.experimental_autoscaler_1_14 "true" }}{{ .Cluster.ConfigItems.experimental_autoscaler_1_14_version }}{{else}}v1.12.2-internal.4{{end}}
+        image: registry.opensource.zalan.do/teapot/kube-cluster-autoscaler:{{ .Cluster.ConfigItems.cluster_autoscaler_version }}
```
aermakov-zalando: Do we still want to keep it customisable if we don't have the settings for the 1.14 version?

cluster/config-defaults.yaml
```diff
@@ -13,8 +13,8 @@ autoscaling_buffer_pods: "0"
 cluster_autoscaler_cpu: "100m"
 cluster_autoscaler_memory: "300Mi"
 
-experimental_autoscaler_1_14: "false"
-experimental_autoscaler_1_14_version: v1.14.5-internal.10
+cluster_autoscaler_version: v1.12.2-internal.4
+cluster_autoscaler_expander: prefer-spot
```
aermakov-zalando: Let's change it to priority by default, or do you see any problems with this?

linki: no, enabled by default now.

cluster/manifests/kube-cluster-autoscaler/daemonset.yaml
linki: removed the config item

cluster/manifests/kube-cluster-autoscaler/daemonset.yaml
```diff
@@ -43,19 +45,11 @@ spec:
           - --skip-nodes-with-system-pods=false
           - --skip-nodes-with-local-storage=false
           - --scale-up-cloud-provider-template=true
+          - --expander={{ .Cluster.ConfigItems.cluster_autoscaler_expander }}
           - --balance-similar-node-groups
+          - --max-node-provision-time=10m
```
aermakov-zalando: Let's reduce it a bit, to e.g. 7m?

linki: changed to `7m`

linki:
:+1:

aermakov-zalando:
👍

PR #2690, opened 2019-11-14
Tune prometheus for faster startup

aermakov-zalando:
It takes _ages_ for Prometheus to restart, since it doesn't flush anything during shutdown so it has to replay all the WAL files. This is mostly IO bound, so let's try to tune it by enabling WAL compression and reducing the interval from the default hardcoded 2 hours to 30 minutes.


aermakov-zalando:
I'm not sure about the settings (because upstream doesn't recommend running with anything other than the hardcoded 2 hours), but let's just see how it works. I don't think anyone is even considering startup time, presumably because no one runs it with a VPA. :(

aermakov-zalando:
👍

gargravarr:
:+1:

PR #2691, opened 2019-11-14
disable VPA for zmon-agent

pitr:
**INCOMPLETE!! DO NOT MERGE!!!**

VPA occasionally will decrease memory beyond what zmon-agent uses, which results in OOMkills and subsequently crash loops.

**INCOMPLETE!! DO NOT MERGE!!!**

It's missing change in `cluster/manifests/deletions.yaml` to delete existing vpa


aermakov-zalando:
Is this still needed?

pitr:
doesn't look like it

PR #2692, opened 2019-11-14
Drop the prometheus deletion

aermakov-zalando:
We don't want to delete it every time we roll any cluster update. Easier to just delete it manually once to let CLM proceed.


linki:
:+1:

aermakov-zalando:
👍

PR #2714, opened 2019-11-21
Update autoscaler to support newer instance types

mikkeloscar:
Update autoscaler to support newer instance types

Ref: https://github.com/zalando-incubator/autoscaler/pull/17


linki:
:+1:

mikkeloscar:
:+1:

PR #2716, opened 2019-11-22
set zmon_scheduler_mem via CR

vetinari:
some clusters need an override

Signed-off-by: Hanno Hecker <hanno@zalando.de>


mikkeloscar:
:+1:

vetinari:
do we need extra permissions in the CR to set this for a specific cluster?

cluster/config-defaults.yaml
```diff
@@ -121,6 +121,7 @@ zmon_agent_replicas: "0"
 zmon_aws_agent_replicas: "0"
 zmon_redis_replicas: "0"
 zmon_scheduler_replicas: "0"
+zmon_scheduler_mem: "0Gi"
```
pitr, line 124: is this correct, zero gigs

mikkeloscar: This is for e2e tests. It doesn't really matter since replicas is set to `0` anyway

gargravarr:
:+1:

PR #2725, opened 2019-11-26
Improve CoreDNS configuration

mikkeloscar:
**THIS IS WIP - needs testing**

This introduces improvements to the CoreDNS configuration as suggested
in https://github.com/coredns/coredns/issues/2593#issuecomment-466888239
The change is to use multiple server directives to avoid expensive
lookup from Kubernetes plugin in terms of reverse DNS lookup or
expensive regex matching for `ingress.cluster.local` names.

* Use the `ready` plugin for readinessProbe
  https://github.com/coredns/coredns/tree/master/plugin/ready
* Enable splitting metrics by zone to see difference between internal and external names

image

### DIFF
--- a/cluster/manifests/coredns-local/configmap-local.yaml
+++ b/cluster/manifests/coredns-local/configmap-local.yaml
@@ -7,27 +7,41 @@ metadata:
     application: coredns
 data:
   Corefile: |
-    .:9254 {
-        errors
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

mikkeloscar:
:+1:

PR #2726, opened 2019-11-26
Add aws-account-id, worker-role-arn to deployment-config

aermakov-zalando:
This will allow our deployment step to pick it up automatically so the users don't need to configure it every time.


aermakov-zalando:
👍

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2728, opened 2019-11-27
Prometheus: increase readiness delay to 2 minutes

aermakov-zalando:
`rate[1m]` might not behave correctly if we only have data for 1 minute, so increase the readiness delay a bit. Let's see if this helps.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2735, opened 2019-12-02
Drop node admitter

aermakov-zalando:
It was added to work around a bug in Kubernetes that's been fixed a long time ago, let's just drop it.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2742, opened 2019-12-03
Remove node-problem-detector from cluster setup

linki:
Removes node problem detector as it hasn't provided any value so far.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2748, opened 2019-12-04
Admission controller: better deployment defaults

aermakov-zalando:
* Update admission controller to `master-43`
 * Set better defaults for rolling update strategy of deployments (1 max unavailable, 5% max surge). Note that max unavailable is a static value since it rounds down, so 1% would be 0 for most of our stuff. Unfortunately, API server applies defaults _before_ passing the request to the admission controller, so 25%/25% set by the users would also be replaced.


aermakov-zalando:
👍

aermakov-zalando:
👍

muaazsaleem:
👍

PR #2749, opened 2019-12-04
Limit nodes considered for ingress load balancer based on taints [1/2]

mikkeloscar:
This is a proposal for how we can reduce the amount of targets in the ingress load balancers to only the relevant nodes.

Right now the problem is that all nodes are considered as target (but only those running skipper will get traffic) even if the node is a master node or a special node pool which wouldn't be able to run skipper because of `taints`.

The idea is to first only limit the possible targets to worker nodes and then tag all node pools which DON'T have any taints defined to be considered as target. Later we can be more specific and allow certain type of nodes e.g. if we have a dedicated system/skipper node pool in place.

The `tag:zalando.org/ingress-enabled=true` is marked as TODO for now because we need to do the rollout in two steps to be safe.

### DIFF
--- a/cluster/manifests/ingress-controller/deployment.yaml
+++ b/cluster/manifests/ingress-controller/deployment.yaml
@@ -38,6 +38,8 @@ spec:
         - --nlb-cross-zone
         {{ end }}
         env:
+        - name: CUSTOM_FILTERS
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

mikkeloscar:
Depends on full rollout of v1.15 before we can merge because it depends on the `node.kubernetes.io/role` label.

mikkeloscar:
This can now be merged!

mikkeloscar:
This is now ready for review!

mikkeloscar:
:+1:

szuecs:
:+1:

PR #2755, opened 2019-12-05
Update admission controller to prevent scheduling on masters

linki:
Prevent user pods from running on master nodes.


aermakov-zalando:
We need to update the webhook as well (so `visibility` is protected).

linki:
Which webhook?

aermakov-zalando:
👍

linki:
:+1:

PR #2757, opened 2019-12-06
Switch to capacity-optimized spot

aermakov-zalando:
* Support both `spot_max_price` and `spot` discount strategies (so we could migrate to the latter)
 * Change spot pools to use the capacity optimized allocation strategy


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2760, opened 2019-12-09
Make it possible to configure one custom DNS zone

mikkeloscar:
This is a simple way to allow an additional custom nameserver to be configured in CoreDNS. The use case right now is to enable DC DNS in certain DC connected clusters.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #2761, opened 2019-12-09
Audittrail adapter: S3 fallback

aermakov-zalando:
* Audittrail: add S3 fallback (similar to EAS)
 * Change the configuration to always use the same config item


aermakov-zalando:
👍

muaazsaleem:
👍

PR #2775, opened 2019-12-12
Update node selector for skipper

mikkeloscar:
Update node selector for skipper. The legacy `kubernetes.io/role` selector will not be supported by v1.16, so we need to roll this out before.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #2782, opened 2019-12-13
Add remote write to prometheus with config item

arjunrn:
The `prometheus_remote_write` config item will allow prometheus to push metrics data to a central location.

### DIFF
--- a/cluster/manifests/prometheus/configmap.yaml
+++ b/cluster/manifests/prometheus/configmap.yaml
@@ -10,7 +10,8 @@ data:
     global:
       scrape_interval: 15s
       external_labels:
-        monitor: 'skipper-ingress'
+        prometheus_replica: @@POD_NAME@@
+        cluster: {{ .Cluster.Alias }}
     scrape_configs:
     # scrape from kube-apiserver pods
     - job_name: 'kubernetes-apiservers'
@@ -172,3 +173,8 @@ data:
        - source_labels: [ __name__ ]
          regex: 'reflector.*'
          action: drop
+{{- if ne .ConfigItems.prometheus_remote_write "disabled" }}
+    remote_write:
+    - url: {{ .ConfigItems.prometheus_remote_write }}
+      bearer_token_file: /meta/credentials/remote-write
+{{- end }}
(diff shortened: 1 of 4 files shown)


cluster/manifests/prometheus/credentialset.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: zalando.org/v1
+kind: PlatformCredentialsSet
+metadata:
+  name: remote-write-credentials
```
aermakov-zalando: Rename to `prometheus-remote-write-credentials` maybe?

arjunrn:
:+1:

cluster/manifests/prometheus/configmap.yaml
```diff
@@ -172,3 +172,8 @@ data:
        - source_labels: [ __name__ ]
          regex: 'reflector.*'
          action: drop
+{{- if ne .ConfigItems.prometheus_remote_write "disabled" }}
+    remote_write:
+    - url: {{ .ConfigItems.prometheus_remote_write }}
+      bearer_token_file: /meta/credentials/remote-write
+{{- end }}
```
mikkeloscar, line 180: I learned yesterday from the Prometheus people that we should define two external labels (currently we have `monitor: 'skipper-ingress'` in the top which doesn't seem like the right label anymore). One label should be e.g. `cluster` and the other should be `replica` such that you can de-duplicate the metrics pushed from the two instances. Would make sense to use `POD_NAME` as replica value, but not sure if this can be passed via a flag or how it's easiest to have a different value for the two pods of the StatefulSet.

szuecs:
@arjunrn what will happen if remote location is down all day?
Does the Prometheus will store it in backlog and crash later?

arjunrn:
> Does the Prometheus will store it in backlog and crash later?

There is a queue of limited size where blocks are stored. I've not configured the size of the queues yet because I have no experience with it.

cluster/manifests/prometheus/configmap.yaml
```diff
@@ -10,7 +10,8 @@ data:
     global:
       scrape_interval: 15s
       external_labels:
-        monitor: 'skipper-ingress'
+        pod_name: @@POD_NAME@@
```
mikkeloscar: Maybe call this a more generic name like `replica` as `pod_name` would be a valid label name in metrics?

mikkeloscar:
:+1:

arjunrn:
:+1:

PR #2783, opened 2019-12-13
Audittrail bucket: allow access from a central role

aermakov-zalando:
Setup a policy on the audit events bucket. If `audittrail_root_account_role` is set, additionally give access to the bucket to a central role.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2774, opened 2019-12-12
Update to Kubernetes v1.16

mikkeloscar:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#v1160

## Changes

* [x] Drop legacy node labels no longer supported by kubelet (e.g. `kubernetes.io/role`)
* [x] Remove `CustomResourceWebhookConversion` feature flag, must be set to `true` in `v1.16` (Introduced in https://github.com/zalando-incubator/kubernetes-on-aws/pull/2172)
* [x] Remove `CustomResourcePublishOpenAPI` feature flag, must be set to `true` in `v1.16` (https://github.com/zalando-incubator/kubernetes-on-aws/pull/2621)
  * [ ] Fix issue which made it break CRDs: https://github.com/zalando-incubator/kubernetes-on-aws/pull/2627
    * [ ] ~Add `status` validation for `PlatformCredentialsSets`~
    * [x] Ensure `status` validation for `AWSIAMRole`
    * [ ] ~Ensure `status` validation for VPA CRDs~
    * [ ] ~Ensure `status` validation for Stackset/Stack CRDs~
* [x] Make it possible to enable `EndpointSlice` (https://github.com/robscott/k8s-website/blob/4370360fc1c46b4e530dec59f5bb45b2c8b24840/content/en/docs/tasks/administer-cluster/enabling-endpoint-slices.md#enabling-endpoint-slices)
* [x] Update flannel/cni to use new config format (conflist) (https://github.com/zalando-incubator/kubernetes-on-aws/pull/2796)
* [x] Update clc to look for `node.kubernetes.io/node-pool` label instead of `kubernetes.io/node-pool` (https://github.com/zalando-incubator/kubernetes-on-aws/pull/2800)
* [x] Fix/workaround e2e test depending on old node label (https://github.com/kubernetes/kubernetes/blob/224be7bdce5a9dd0c2fd0d46b83865648e2fe0ba/test/e2e/framework/util.go#L1976)
  * Fixed by using the hack: `-allowed-not-ready-nodes=1` when running e2e. Can be improved in `v1.17`.
* [x] Support `apiextensions.k8s.io/v1` in admission-controller
* [x] Update CRDs to `apiextensions.k8s.io/v1`
* [x] Update components which use old APIs like daemonset v1beta1
  * [x] Update cluster autoscaler to not use daemonset v1beta1 api.
[...]

PR #2786, opened 2019-12-14
Change master nodeSelector to new value

mikkeloscar:
Use `node.kubernetes.io/role=master` as nodeSelector since `node-role.kubernetes/master` won't be supported in v1.16.
Add a matching taint to align and migrate the naming with v1.16 rollout.

Required before #2774


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2789, opened 2019-12-16
Allow to configure ExternalDNS TXT prefix

linki:
Allows people to use custom TXT records on the same DNS names as their Ingresses, e.g. for SPF etc.

For example see https://github.com/kubernetes-sigs/external-dns/issues/930#issuecomment-474427868

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -324,3 +324,6 @@ ssh_vpc_only: "false"
 # configure custom dns zone
 custom_dns_zone: "" # zone name e.g. example.org
 custom_dns_zone_nameservers: "" # space seperated list of nameserver IP addresses
+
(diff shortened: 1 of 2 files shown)


linki:
:+1:

cluster/config-defaults.yaml
```diff
@@ -324,3 +324,6 @@ ssh_vpc_only: "false"
 # configure custom dns zone
 custom_dns_zone: "" # zone name e.g. example.org
 custom_dns_zone_nameservers: "" # space seperated list of nameserver IP addresses
+
+# prefix prepended to ownership TXT records for external-dns
+external_dns_ownership_prefix:
```
aermakov-zalando: Please set it to an empty string explicitly.

linki:
:+1:

mikkeloscar:
:+1:

PR #2796, opened 2019-12-16
Switch to conflist cni configuration for flannel

mikkeloscar:
Switches to use the newer `conflist` cni configuration. This is required
because Kubernetes v1.16.0 introduces config validation:
https://github.com/kubernetes/kubernetes/pull/80482 which is no longer
able to validate the old configuration. The behavior is the same with
the new configuration.

The configuration comes from the documented deployment from flannel: https://github.com/coreos/flannel/pull/888. I did not include the portMapping config has we don't have pods with hostports (running in pod network).

Unfortunately this is not really covered by our e2e as it will only apply to new nodes. I have tested this by setting up a new cluster based on this configuration.

Required for #2774

### DIFF
--- a/cluster/manifests/flannel/daemonset.yaml
+++ b/cluster/manifests/flannel/daemonset.yaml
@@ -35,7 +35,7 @@ spec:
           - name: CNI_CONFIG_SOURCE
             value: /etc/kube-flannel/cni-conf.json
           - name: CNI_CONFIG_TARGET
-            value: /etc/cni/net.d/10-flannel.conf
+            value: /etc/cni/net.d/10-flannel.conflist
         resources:
           requests:
             cpu: 25m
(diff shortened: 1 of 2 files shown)


cluster/manifests/flannel/daemonset.yaml
```diff
@@ -35,7 +35,7 @@ spec:
           - name: CNI_CONFIG_SOURCE
             value: /etc/kube-flannel/cni-conf.json
           - name: CNI_CONFIG_TARGET
-            value: /etc/cni/net.d/10-flannel.conf
+            value: /etc/cni/net.d/10-flannel.conflist
```
linki, line 38: Does it do the filename change by itself? Because I don't see it anywhere else in this PR.

linki:
:+1:

cluster/manifests/flannel/daemonset.yaml
mikkeloscar: It's just the name of the target file, Our init container copies from `CNI_CONFIG_SOURCE` to `CNI_CONFIG_TARGET`

mikkeloscar:
:+1:

PR #2800, opened 2019-12-16
Update clc to support new node-pool node label

mikkeloscar:
CLC now looks for node label `node.kubernetes.io/node-pool` instead of `kubernetes.io/node-pool`.

Required for #2774


mikkeloscar:
:+1:

linki:
:+1:

PR #2805, opened 2019-12-17
DaemonSet pods: add an ephemeral-storage request

aermakov-zalando:
Kubernetes will still try to evict daemonset pods even though they're critical. Let's give them ephemeral storage that they'll never consume to prevent that.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2807, opened 2019-12-18
Allow VPA to delete pods

aermakov-zalando:
It now deletes pods when they OOM crash, so it must have the permission to do that.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2812, opened 2019-12-18
Always evict on OOMs, even if the pod is not old enough

aermakov-zalando:
Currently, if the pod fails after `--evict-after-oom-threshold` (10 minutes) but before `--pod-lifetime-update-threshold`, it'll just keep running. There's no point in this, because we'd rather fix the pod instead.

### DIFF
--- a/cluster/manifests/01-vertical-pod-autoscaler/updater-deployment.yaml
+++ b/cluster/manifests/01-vertical-pod-autoscaler/updater-deployment.yaml
@@ -30,6 +30,8 @@ spec:
           - --v=4
           - --stderrthreshold=info
           - --min-replicas=1
+          - --pod-lifetime-update-threshold=12h
+          - --evict-after-oom-threshold=12h
         resources:
           limits:
             cpu: 200m


mikkeloscar:
Can you explain why setting it to `12h` fixes the issue? I don't understand this enough to see why this would help.

aermakov-zalando:
VPA usually doesn't evict the pod unless it's older than `pod-lifetime-update-threshold`. However, if it terminated due to an oomkill, and the total runtime (based on start time/end time in `lastState`) was less `evict-after-oom-threshold`, eviction is allowed. The defaults are 12 hours for the total lifetime and 10 minutes for the quick OOM threshold. This basically means that if your pod is reliably crashing 11 minutes after startup, VPA won't bother doing anything with it until it's at least 12 hours old. Since it doesn't make a lot of sense, we can just increase the quick OOM window so that VPA is allowed to evict no matter what.

mikkeloscar:
Got it, thanks!

mikkeloscar:
:+1:

arjunrn:
:+1:

PR #2828, opened 2019-12-27
Drop broken fluentd metrics from etcd servers

arjunrn:
Because of this [bug](https://github.com/prometheus/node_exporter/issues/704) there is a lot of noise generated in the remote write end points.


gargravarr:
:+1:

arjunrn:
:+1:

PR #2831, opened 2020-01-06
Skipper: drop query string from access logs

aermakov-zalando:
We don't want to leak OAuth2 tokens to the logs.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2839, opened 2020-01-07
Enable rotation of service account tokens

linki:
Adds timestamps to `ServiceAccountTokens` so that they can be rejected by the API server when they have been expired (via `iat`, `exp` and `nbf` fields). Tokens will also be rejected once the corresponding Pod terminates.

`exp`: expires at
`iat`: issued at
`nbf`: not valid before (equals issued at)

```json
# JWT header
# ...
# JWT payload
{
  "aud": [
    "kubernetes/serviceaccount"
  ],
  "exp": 1578415862,
  "iat": 1578412262,
  "iss": "kubernetes/serviceaccount",
  "kubernetes.io": {
    "namespace": "kapp",
    "pod": {
      "name": "chaoskube-56967cd969-bcmht",
      "uid": "feaf2cc8-ff62-4681-a92a-f25f96a6beff"
    },
    "serviceaccount": {
      "name": "chaoskube",
      "uid": "d1a29604-36f2-4c90-94e8-d7886353e2f9"
    }
  },
  "nbf": 1578412262,
  "sub": "system:serviceaccount:kapp:chaoskube"
}
# JWT signature
# ...
```

Like before Kubernetes injects ServiceAccountToken mounts into the Pod but they look different now.

```yaml
  # PodSpec
  # ...
  # following is injected into each pod
  volumes:
  - name: kube-api-access-7nxqz
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          expirationSeconds: 3600
          path: token
      - configMap:
          items:
          - key: ca.crt
            path: ca.crt
          name: kube-root-ca.crt
      - downwardAPI:
          items:
          - fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
            path: namespace
```

The feature uses `projected` volumes under the hood. With this it's possible to retain the original filesystem structure which is three files that are now gathered from different sources.

```console
ls /var/run/secrets/kubernetes.io/serviceaccount/
ca.crt     namespace  token
```

Instead of a ServiceAccount secret with three entries the following is used:

`namespace`: via downward API
`ca.crt`: a shared configmap containing only the root cert of the cluster
`token`: new volume type `serviceAccountToken` that provides an expiring token

Currently, the "old" secret still exists and contains a non-expiring ServiceAccountToken which still gets accepted by the API server.

The only different for clients is the requirement to reload refreshed tokens from the filesystem before they expire.

Part of: https://github.bus.zalan.do/teapot/issues/issues/1993

Todo:
- [ ] prevent issue of "old" service account secrets (`--controllers` flag on controller manager)
- [ ] test how clients behave when the token gets refreshed (recent `client-go` works fine)
- [ ] cleanup
- [ ] probably: find a way to opt-out individual pods to ease migration efforts
- [ ] optional: open-up external access for service account tokens again

In order for non-root applications to read the credentials you might need to add the following to your pod spec. This requirement might go away in future versions. Also see [here](https://github.com/kubernetes-sigs/external-dns/pull/1185#issuecomment-531678363).

```yaml
      securityContext:
        fsGroup: 65534
```

Original feature issue: https://github.com/kubernetes/kubernetes/issues/70679

### DIFF
--- a/cluster/manifests/psp/pod_security_policy.yaml
+++ b/cluster/manifests/psp/pod_security_policy.yaml
@@ -43,3 +43,4 @@ spec:
   - persistentVolumeClaim
   - downwardAPI
   - configMap
+  - projected
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -44,6 +44,7 @@ write_files:
 {{- if ne .NodePool.ConfigItems.pod_max_pids "-1" }}
       featureGates:
         SupportPodPidsLimit: true
+        BoundServiceAccountTokenVolume: true
       podPidsLimit: {{ .NodePool.ConfigItems.pod_max_pids }}
 {{- end }}
       maxPods: {{ nodeCIDRMaxPods (parseInt64 .Cluster.ConfigItems.node_cidr_mask_size) 8 }}
@@ -120,7 +121,7 @@ write_files:
           - --authorization-mode=Webhook,RBAC
           - --authorization-webhook-config-file=/etc/kubernetes/config/authz.yaml
           - --admission-control-config-file=/etc/kubernetes/config/image-policy-webhook.yaml
-          - --feature-gates=TaintNodesByCondition={{.Cluster.ConfigItems.experimental_schedule_daemonset_pods}},ScheduleDaemonSetPods={{.Cluster.ConfigItems.experimental_schedule_daemonset_pods}},TTLAfterFinished=true,CustomResourceWebhookConversion={{.Cluster.ConfigItems.custom_resource_webhook_conversion}},CustomResourcePublishOpenAPI={{.Cluster.ConfigItems.custom_resource_publish_openapi}}
+          - --feature-gates=TaintNodesByCondition={{.Cluster.ConfigItems.experimental_schedule_daemonset_pods}},ScheduleDaemonSetPods={{.Cluster.ConfigItems.experimental_schedule_daemonset_pods}},TTLAfterFinished=true,CustomResourceWebhookConversion={{.Cluster.ConfigItems.custom_resource_webhook_conversion}},CustomResourcePublishOpenAPI={{.Cluster.ConfigItems.custom_resource_publish_openapi}},BoundServiceAccountTokenVolume=true
           - --anonymous-auth=false
           {{ if ne .Cluster.ConfigItems.audittrail_url "" }}
           - --audit-webhook-config-file=/etc/kubernetes/config/audit.yaml
@@ -147,6 +148,8 @@ write_files:
           - --kubelet-certificate-authority=/etc/kubernetes/ssl/ca.pem
           - --kubelet-client-certificate=/etc/kubernetes/ssl/kubelet-client.pem
           - --kubelet-client-key=/etc/kubernetes/ssl/kubelet-client-key.pem
+          - --service-account-signing-key-file=/etc/kubernetes/ssl/service-account-private-key.pem
+          - --service-account-issuer=kubernetes/serviceaccount
           livenessProbe:
             httpGet:
               host: 127.0.0.1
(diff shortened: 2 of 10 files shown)


cluster/node-pools/master-default/userdata.yaml
```diff
@@ -44,6 +44,7 @@ write_files:
 {{- if ne .NodePool.ConfigItems.pod_max_pids "-1" }}
       featureGates:
         SupportPodPidsLimit: true
+        BoundServiceAccountTokenVolume: true
```
linki: TODO: move out of unrelated `if` block

cluster/node-pools/worker-default/userdata.yaml
```diff
@@ -54,6 +54,7 @@ write_files:
 {{- if ne .NodePool.ConfigItems.pod_max_pids "-1" }}
       featureGates:
         SupportPodPidsLimit: true
+        BoundServiceAccountTokenVolume: true
```
linki: TODO: move out of unrelated `if` block

cluster/manifests/psp/pod_security_policy.yaml
```diff
@@ -43,3 +43,4 @@ spec:
   - persistentVolumeClaim
   - downwardAPI
   - configMap
+  - projected
```
linki, line 46: TODO: check which other implications this has

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -120,7 +121,7 @@ write_files:
           - --authorization-mode=Webhook,RBAC
           - --authorization-webhook-config-file=/etc/kubernetes/config/authz.yaml
           - --admission-control-config-file=/etc/kubernetes/config/image-policy-webhook.yaml
-          - --feature-gates=TaintNodesByCondition={{.Cluster.ConfigItems.experimental_schedule_daemonset_pods}},ScheduleDaemonSetPods={{.Cluster.ConfigItems.experimental_schedule_daemonset_pods}},TTLAfterFinished=true,CustomResourceWebhookConversion={{.Cluster.ConfigItems.custom_resource_webhook_conversion}},CustomResourcePublishOpenAPI={{.Cluster.ConfigItems.custom_resource_publish_openapi}}
+          - --feature-gates=TaintNodesByCondition={{.Cluster.ConfigItems.experimental_schedule_daemonset_pods}},ScheduleDaemonSetPods={{.Cluster.ConfigItems.experimental_schedule_daemonset_pods}},TTLAfterFinished=true,CustomResourceWebhookConversion={{.Cluster.ConfigItems.custom_resource_webhook_conversion}},CustomResourcePublishOpenAPI={{.Cluster.ConfigItems.custom_resource_publish_openapi}},BoundServiceAccountTokenVolume=true
```
linki: TODO: make configurable via config items

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -41,11 +41,9 @@ write_files:
 {{- if not (index .Cluster.ConfigItems "enable_cfs_quota") }}
       cpuCFSQuota: false
 {{- end }}
-{{- if ne .NodePool.ConfigItems.pod_max_pids "-1" }}
```
linki, line 44: I dropped this we don't use `-1` anywhere anymore and it simplified the configuration here.

linki:
Ultimately, this has to be split into two PRs to make the rollout safe:
- PR 1: apply manifest changes for `securityContext` and `projected` volume
- PR 2: roll the nodes and switch to rotated service account tokens

cluster/manifests/psp/pod_security_policy.yaml
linki: [projected volumes](https://kubernetes.io/docs/concepts/storage/volumes/#projected) allow to mount other resources under the same path: configMaps, secrets, downwardAPI and serviceAccount tokens.

Nothing that the user cannot already mount individually, so that should be safe to allow.

linki:
Merging this doesn't enable it yet. It needs to be enabled with a config item.

linki:
@mikkeloscar @szuecs I would like to get your opinion on adding `securityContext.fsGroup` to several deployments in this PR (e.g. skipper, external-dns).

This is necessary because the new tokens are mounted with slightly different file permissions (more secure).

Previously, inside the container it looked like this (note the file permissions, owner and group):
```console
$ ls -la /var/run/secrets/kubernetes.io/serviceaccount/..data/
-rw-r--r--    1 root     root           872 Jan 13 10:17 token
```
So, the token was readable by everybody, including non-root users in the container (e.g. 1000 for skipper would be able to read it)

With the new token system token mounts looks like the following (note the slightly different file permission):
```console
$ ls -la /var/run/secrets/kubernetes.io/serviceaccount/..data/
-rw-r-----    1 root     root         908 Jan 13 09:58 token
```
This means non-root users cannot read the token anymore (e.g. skipper).

One way to fix it is to use `securityContext.fsGroup: 1000` which changes the file permissions of any mounted volume (note the changed group file permission to the running process' user):
```console
$ ls -la /meta/aws-iam/..data/
-rw-r--r--    1 root     1000          69 Jan 13 10:34 role-arn
```
This should not matter since the files were world-readable before.

However, the service account token mounts now look like this:
```console
$ ls -la /var/run/secrets/kubernetes.io/serviceaccount/..data/
-rw-r-----    1 root     1000         908 Jan 13 09:58 token
```
which means they can now be read by the main process (e.g. skipper, external-dns, etc.)

Do you see any issues with this?

mikkeloscar:
@linki I don't see any issues with changing the file permissions.

szuecs:
If you change all securityContext.fsgroups it seems to be fine for me.
It's progress in security, so let's do it!

linki:
:+1:

szuecs:
:+1:

PR #2842, opened 2020-01-08
Update k8s-authnz-webhook

aermakov-zalando:
Auth webhook: move the token to request body so it doesn't get logged on accident.


linki:
:+1:

szuecs:
:+1:

PR #2849, opened 2020-01-09
Add support listener port to skipper-ingress service

mikkeloscar:
This allows proxying to a skipper support endpoint e.g. `/routes` via the generic `skipper-ingress` service

This way users without access to `kube-system` namespace can still get skipper routes for debugging via:

```
zkubectl tunnel skipper-ingress.kube-system.svc.cluster.local 9911
```

```
curl localhost:9911/routes
```

### DIFF
--- a/cluster/manifests/skipper/service.yaml
+++ b/cluster/manifests/skipper/service.yaml
@@ -12,8 +12,13 @@ metadata:
 spec:
   type: ClusterIP
   ports:
-    - port: 80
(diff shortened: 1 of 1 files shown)


szuecs:
The general service is fine, but I don't like that it's in the same service.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #2850, opened 2020-01-09
Adding ephemeral storage support to autoscaler

muaazsaleem:
Update the autoscaler version to include ephemeral storage support to autoscaler. This makes it possible for the autoscaler to scale from 0 when based on pods requesting ephemeral storage.

ref: https://github.com/zalando-incubator/autoscaler/pull/25


muaazsaleem:
👍

arjunrn:
:+1:

PR #2851, opened 2020-01-09
Use the correct numbers for ephemeral storage in template nodes

aermakov-zalando:
Instead of just hardcoding 30Gi for `ephemeral-storage`, use a reasonable estimate based on the instance types used in the pool. Related CLM change: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/231.

Additionally, simplify the templates by using `IsSpot()` function instead of duplicating the logic.

### DIFF
--- a/cluster/node-pools/worker-default/stack.yaml
+++ b/cluster/node-pools/worker-default/stack.yaml
@@ -77,6 +77,7 @@ Resources:
         Value: {{if .NodePool.IsSpot }}true{{else}}false{{end}}
       - Key: k8s.io/cluster-autoscaler/node-template/resources/ephemeral-storage
         PropagateAtLaunch: false
+        # 40Gi for EBS storage or 95% of instance storage
         Value: "$data.NodePool.AvailableStorage 42949672960 0.95"
       - Key: k8s.io/cluster-autoscaler/node-template/label/zalando.org/scaling-priority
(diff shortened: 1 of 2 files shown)


cluster/node-pools/worker-default/stack.yaml
```diff
@@ -74,15 +74,15 @@ Resources:
         Value: ready
       - Key: k8s.io/cluster-autoscaler/node-template/label/aws.amazon.com/spot
         PropagateAtLaunch: true
-        Value: {{if or (eq .NodePool.DiscountStrategy "spot") (eq .NodePool.DiscountStrategy "spot_max_price")}}true{{else}}false{{end}}
+        Value: {{if .NodePool.IsSpot }}true{{else}}false{{end}}
       - Key: k8s.io/cluster-autoscaler/node-template/resources/ephemeral-storage
         PropagateAtLaunch: false
-        Value: "30Gi"
+        Value: "$data.NodePool.AvailableStorage 42949672960 0.95"
```
mikkeloscar, line 81: nit: maybe add a comment what this number is? :)

aermakov-zalando: Good point, done.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2854, opened 2020-01-10
Get rid of -ubuntu- profiles

aermakov-zalando:
We've switched a long time ago, let's simplify our setup.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2858, opened 2020-01-13
Admission controller: new policy for pod-managing resources

aermakov-zalando:
* Update admission controller to the latest version:
  * Support for `--resource-cutoff-timestamp`, `--resource-validate-redeployments` removed. Resources are always checked unless `--validate-pod-template-resources` is disabled.
  * Application labels are validated if the application is being redeployed as well, drop support for grandfathering.
* Drop support for `teapot_admission_controller_resource_validate_redeployments` config item, add `teapot_admission_controller_validate_pod_template_resources` instead. The only cluster where this was used was updated.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2869, opened 2020-01-15
Give PowerUser/ReadOnly access to persistentvolumes via RBAC

mikkeloscar:
Give `PowerUser/ReadOnly` access to persistentvolumes via RBAC.

* Read access should generally be fine.
* Adds write access for `update` and `patch` which are needed by postgres-operator (and thereby must be in the poweruser role for cdp to deploy the operator roles).


mikkeloscar:
:+1:

linki:
:+1:

PR #2877, opened 2020-01-20
Make it possible to allow collaborator admin access in pet clusters

aermakov-zalando:
Currently we only allow access to e2e clusters, but pet clusters should also work.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -271,6 +271,13 @@ custom_dns_zone_nameservers: "" # space seperated list of nameserver IP addresse
 # prefix prepended to ownership TXT records for external-dns
 external_dns_ownership_prefix: ""
 
+# special roles for test/pet clusters
+{{if eq .Cluster.Environment "e2e"}}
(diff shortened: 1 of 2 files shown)


cluster/config-defaults.yaml
```diff
@@ -271,6 +271,13 @@ custom_dns_zone_nameservers: "" # space seperated list of nameserver IP addresse
 # prefix prepended to ownership TXT records for external-dns
 external_dns_ownership_prefix: ""
 
+# special roles for test/pet clusters
+{{if eq .Cluster.Environment "e2e"}}
+collaborator_administrator_access: "false"
+{{else}}
+collaborator_administrator_access: "true"
+{{end}}
```
mikkeloscar, line 279: Is the logic not reversed here? Should be `true` in case of e2e?

aermakov-zalando: Oops.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2879, opened 2020-01-20
Set various configurations for virtual memory

arjunrn:
When both the config items  `vm_dirty_background_bytes` and `vm_dirty_bytes` are set a `sysctl` override file is created. Also override the maximum number of `mmaps` allowed  by default.
Also based on the information from this [article](https://www.suse.com/support/kb/doc/?id=7000830) it would be safe to increase the allowed number of mmaps for processes. It should not have an impact on the performance of virtual memory nor would it increase resource usage. So we have increased the value to 8x the default value.

### DIFF
--- a/cluster/node-pools/worker-default/userdata.yaml
+++ b/cluster/node-pools/worker-default/userdata.yaml
@@ -87,3 +87,19 @@ write_files:
           cacheAuthorizedTTL: "5m"
           cacheUnauthorizedTTL: "30s"
       readOnlyPort: 10255 # TODO: remove after complete rollout of webhook auth
+
+  - owner: root:root
+    path: /etc/sysctl.d/02-vm-max-mmaps.conf
+    content: |
+      vm.max_map_count=524288
+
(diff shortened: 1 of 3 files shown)


jrake-revelant:
👍

cluster/node-pools/worker-default/userdata.yaml
```diff
@@ -87,3 +87,19 @@ write_files:
           cacheAuthorizedTTL: "5m"
           cacheUnauthorizedTTL: "30s"
       readOnlyPort: 10255 # TODO: remove after complete rollout of webhook auth
+
+  - owner: root:root
+    path: /etc/sysctl.d/02-vm-max-mmaps.conf
+    content: |
+      vm.max_map_count=524288
```
mikkeloscar, line 94: could we add a comment or at least extend the PR description with the reason we choose this value? I'm not arguing that we should NOT have this value, but it should just be clear how/why the decision was made when we go back to this change in 6 months :)

arjunrn: Updated the PR description.

arjunrn:
:+1:

PR #2881, opened 2020-01-20
Remove proxy permissions from PowerUser role

linki:
Addresses an issue that could lead to private resources being exposed.


linki:
This needs some further discussion. Closing for now.

PR #2884, opened 2020-01-21
kubelet: drop readOnlyPort

aermakov-zalando:
We forgot to drop this, and it's not used anymore.


mikkeloscar:
Could we include this in v1.16 rollout? No need to roll the nodes just for this IMO.

aermakov-zalando:
Sure, but we have an AMI update coming soon anyway, no? For 1.16 we still need to fix CA.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2885, opened 2020-01-21
Fix mirror pods

aermakov-zalando:
* Update authnz webhook to not reject the kubelet's user when checking PSPs
 * Add an e2e test to check that mirror pods for main components are created properly


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2888, opened 2020-01-22
Fix deployment role

aermakov-zalando:
Update the deployment IAM role to allow access to whitelisted services.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2887, opened 2020-01-22
Update to 1.15.9

aermakov-zalando:
[Changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#v1159).

PR #2889, opened 2020-01-22
Remove the runcmd directive

arjunrn:
Because `cloud-final.service` does not start before `multi-user.target` anything that requires `cloud-final.service` causes a circular dependency in SystemD. More information [here](https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/1623868).


aermakov-zalando:
👍

aermakov-zalando:
I've cherry-picked to https://github.com/zalando-incubator/kubernetes-on-aws/pull/2887.

PR #2890, opened 2020-01-22
Restore the kubectl node role label

aermakov-zalando:
This was dropped from the AMI because 1.16 kubelet forbids setting these, but kubectl wasn't updated to the new label. We should migrate to doing that in the admission controller in the future, but for now let's just make it work.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2891, opened 2020-01-23
[1/2] CDP controller role: allow the use of v2 tokens

aermakov-zalando:
Let's get rid of V1 tokens, especially for things that have access to all clusters. It looks like CDP controller is the only remaining app.


aermakov-zalando:
👍

aryszka:
:+1:

PR #2896, opened 2020-01-24
Hotfix RBAC

mikkeloscar:
With all the RBAC changes we managed to merge the `poweruser` role changes together with webhook rollout which effectively means CDP deployments of RBAC roles will not work during the next cluster update.

To avoid this situation, let's hotfix the role changes in stable, before the big rollout of #2895


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2897, opened 2020-01-24
Restore the old kubernetes.io/node-pool label

aermakov-zalando:
Users are still relying on it apparently.


aermakov-zalando:
:+

aermakov-zalando:
👍

PR #2901, opened 2020-01-24
ReadOnly/PowerUser: allow access to storageclasses

aermakov-zalando:
Users should be able to modify these.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2902, opened 2020-01-24
Adding ClusterRoles for the three Fabric operators:

cjbooms:
Fabric Gateway
Fabric Event Streams
Sloppy

### DIFF
--- a/cluster/manifests/roles/fabric-roles.yaml
+++ b/cluster/manifests/roles/fabric-roles.yaml
@@ -1,0 +1,101 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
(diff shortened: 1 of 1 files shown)


mikkeloscar:
FYI: this is being done to avoid chicken and egg problems for Fabric tools which are deployed to every cluster anyway.

Without these roles they need to create the CRD before creating the role which is problematic as they have two projects depending on each other (metacontroller and their own controller).

mikkeloscar:
:+1:

cjbooms:
👍

herojan:
👍

mikkeloscar:
:+1:

herojan:
👍

szuecs:
:+1:

PR #2903, opened 2020-01-24
Allow restoring the legacy node pool label

aermakov-zalando:
Turns out some of the clusters users still rely on it, so we can't just drop it. Additionally, remove the autoscaler template label so it doesn't try to scale up since the nodes won't match anyway.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2917, opened 2020-01-27
Apply admission-control configuration first

mikkeloscar:
This fixes an issue where we didn't create poweruser and readonly roles for some CRDs because the CRDs were applied before the admission-controller was configured to create roles for CRDs.

This is relevant when a cluster is first created.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2918, opened 2020-01-28
[WIP] Add aws pod identity components

arjunrn:
Add aws pod identity components


cluster/node-pools/master-default/files.yaml
```diff
@@ -51,3 +51,11 @@ files:
     data: "{{ .Cluster.ConfigItems.service_account_private_key | base64Decode | publicKey | base64 }}"
     permissions: 0400
     encrypted: false
+  - path: /var/www/openid-configuration/discovery.json
+    data: "{{ generateOidcDiscoveryDocument .Cluster.LocalID .Values.hosted_zone | base64 }}"
```
mikkeloscar: nit: function name should be `generateOIDCDiscoveryDocument` to follow the convention we use from go.

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -124,6 +124,9 @@ write_files:
           - --service-account-signing-key-file=/etc/kubernetes/ssl/service-account-private-key.pem
           - --service-account-issuer=kubernetes/serviceaccount
           {{- end }}
+          - --service-account-key-file=/etc/kubernetes/ssl/pod-identity-service-account-public-key.pem
+          - --service-account-signing-key-file=/etc/kubernetes/ssl/pod-identity-service-account-key.pem
+          - --service-account-issuer=https://{{.Cluster.LocalID}}.{{.Values.hosted_zone}}/
```
mikkeloscar: Probably makes sense to simply use the `APIServerURL` here instead of constructing it manually. And I guess the same could be done the other place you pass `hosted_zone`.

linki:
@arjunrn @aermakov-zalando I am a bit confused now. We added the injection of AWS IAM related stuff (e.g. `AWS_WEB_IDENTITY_TOKEN_FILE` env var) to our own admission controller, but AWS' pod-identity-webhook [already does that as well](https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/pkg/handler/handler.go#L159-L162). Why did we move this our own admission controller and is the one from AWS still needed now?

cluster/node-pools/master-default/userdata.yaml
linki: Will be done.

linki:
Clarified: We indeed rebuilt the entire thing in our own admission controller. I missed that even while working on it. :astonished:

linki:
Apart from the changes to CA this has been fully absorbed through other PRs.

Pending PR for CA is here: https://github.com/zalando-incubator/kubernetes-on-aws/pull/3131

PR #2921, opened 2020-01-28
Adding roles for status subresource for all 3 Fabric operators.

cjbooms:
These are also required as metacontoller updates the status subresources directly for each of the 3 operators.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2875, opened 2020-01-17
[WIP] Enable rotation of service account tokens

linki:
Enables rotated service account tokens feature.

PR #2924, opened 2020-01-28
Switch to dashboard v2 and remove heapster

linki:
Update to v2 of the [Kubernetes dashboard](https://kubernetes.io/docs/tasks/access-application-cluster/web-ui-dashboard/)

In v2 the dashboard doesn't rely on heapster which isn't compatible with https://github.com/zalando-incubator/kubernetes-on-aws/pull/2875.

The dashboard changed some auth related things. To get the original behaviour back we serve on the insecure port.

Users can to `kubectl proxy` and then open the dashboard in their browsers at:

http://localhost:8001/api/v1/namespaces/kube-system/services/kubernetes-dashboard/proxy/#/overview?namespace=default

or via `kubectl -n kube-system port-forward svc/kubernetes-dashboard 9090:80` at:

http://localhost:9090/#/overview?namespace=default


cluster/manifests/dashboard/rbac.yaml
```diff
@@ -3,11 +3,92 @@ kind: ServiceAccount
...
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: kubernetes-dashboard
+  labels:
+    application: kubernetes-dashboard
+rules:
+# Allow Metrics Scraper to get metrics from the Metrics server
```
aermakov-zalando: That looks like an extremely limited role. How will the users be able to interact with deployments, services, etc?

linki:
Double check: `kubectl top pods`

mikkeloscar:
:+1:

szuecs:
:+1:

PR #2927, opened 2020-01-29
MID-2409 remove fabric ClusterRoles, these are no longer needed

cjbooms:
On suggest from @mikkeloscar Team Fabric have defined the required `ClusterRole`'s  in each of their operator repositories, along with a `ClusterRoleBinding` that binds these `ClusterRole`'s to the `ServiceAccount` used by their `StatefulSet` `metacontroller`


mikkeloscar:
:+1:

arjunrn:
@mikkeloscar we need to add these to the delete list right?

mikkeloscar:
They are named the same as what we automatically create from the admission-controller, so they will just be overwritten by that process.

szuecs:
:+1:

PR #2935, opened 2020-01-31
Enable node admitter for node labeling

mikkeloscar:
Enables the admitter which will put the `kubernetes.io/role` labels onto newly created nodes.

This way we can see node roles with `kubectl`.

This is needed for Kubernetes v1.16 #2774


mikkeloscar:
:+1:

szuecs:
:+1:

PR #2937, opened 2020-01-31
check local tokeninfo is ready on startup

szuecs:

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: skipper-ingress
-    version: v0.11.31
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

ruiaraujo:
what happens if local token info fails to starts?

szuecs:
@ruiaraujo skipper will not start the Listener, such that ALB/NLB target group will not see it as healthy and kubernetes does not see this as ready, too. 
https://github.com/zalando/skipper/pull/1292 is the change

ruiaraujo:
Yes, but could a permanent failure of the side car prevent the ingress from scaling up?

szuecs:
Answered internally

gargravarr:
:+1:

PR #2950, opened 2020-02-04
Add HPA admitter

mikkeloscar:
This adds an HPA admitter which enforces that HPAs targeting `Deployment` or `StatefulSet` resources always use the `apps/v1` `apiVersion` in the `scaleTargetRef` definition. This is needed because it's possible to deploy an HPA targeting e.g. `extensions/v1beta1` in Kubernetes v1.16, but this does not work as `extensions/v1beta1` has been disabled for `Deployments` and `StatefulSets` in Kubernetes v1.16.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2952, opened 2020-02-04
Allow disabling legacy apiVersions

mikkeloscar:
This adds a config-item `disable_legacy_api_versions` which when set to `true` (default) will disable legacy `apiVersions` like `extensions/v1beta1/deployments`, `apps/v1beta1`, `apps/v1beta2`.

The idea is that we set it manually to `false` in all clusters, and then we can disable once we are sure nothing will break for the users.


mikkeloscar:
:+1:

hjacobs:
:+1:

PR #2953, opened 2020-02-05
Allow PowerUsers to deploy LimitRanges

aermakov-zalando:
These are used by a couple of teams to provide better defaults for their pods (apparently the operator they use makes it very hard to customise the defaults). Let's allow for them now.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2956, opened 2020-02-05
Encrypt secrets in etcd using KMS

linki:
Use KMS to encrypt secrets in etcd. Follow up of https://github.com/zalando-incubator/kubernetes-on-aws/pull/2595.

It mitigates the effects of unauthorized access to an etcd endpoint by preventing reading clear-text secrets. It doesn't protect against a compromised Kubernetes master because it has access to etcd and is allowed to decrypt with the KMS key.

All providers are used for reading but only the first is used for writing. This way one can seamlessly migrate from non-encrypted to encrypted, encrypted to non-encrypted and from one key to another.

The only thing to do is to swap the order which can be done via the `enable_encryption: true|false` config item. When enabled, secrets get encrypted when they are written, so it's lazy. Same for decryption in case it's turned off later. Both encrypted and non-encrypted secrets can always be read.

To be sure that all secrets are stored encrypted one has to force a rewrite of all secrets after turning the feature on (and waiting for all masters to update) with `kubectl get secrets --all-namespaces -o json | kubectl replace -f -`.

Note that a compromised etcd is quite terrible still. Using secrets encryption prevents an attacker to immediately extract secrets from etcd but it doesn't prevent her from scheduling a pod that mounts the secret and exposes it as part of the scheduled container image.

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -430,7 +430,6 @@ write_files:
             requests:
               cpu: 25m
               memory: 25Mi
-        {{- if index .Values "ClusterStackOutputs" }}
         - name: aws-encryption-provider
           image: registry.opensource.zalan.do/teapot/aws-encryption-provider:master-1
           command:
@@ -453,7 +452,6 @@ write_files:
           volumeMounts:
           - mountPath: /var/run/kmsplugin
             name: var-run-kmsplugin
-        {{- end }}
         volumes:
         - hostPath:
             path: /etc/kubernetes/ssl
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -432,8 +432,7 @@ write_files:
               memory: 25Mi
         {{- if index .Values "ClusterStackOutputs" }}
         - name: aws-encryption-provider
-          image: quay.io/linki/aws-encryption-provider:latest
-          imagePullPolicy: Always
+          image: registry.opensource.zalan.do/teapot/aws-encryption-provider:master-1
           command:
           - /aws-encryption-provider
           - --key=arn:aws:kms:{{ .Cluster.Region }}:{{ .Cluster.InfrastructureAccount | getAWSAccountID }}:key/{{ .Values.ClusterStackOutputs.EtcdEncryptionKey }}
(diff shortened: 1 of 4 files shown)


cluster/node-pools/master-default/userdata.yaml
```diff
@@ -427,6 +430,31 @@ write_files:
             requests:
               cpu: 25m
               memory: 25Mi
+        {{- if index .Values "EtcdEncryptionKey" }}
```
linki: This is a guard to work with older versions of CLM that doesn't pass Stack Outputs in the values.

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -693,3 +725,25 @@ write_files:
...
+          - identity: {}
+      {{- end }}
+          - kms:
+              name: aws-encryption-provider
+              endpoint: unix:///var/run/kmsplugin/socket.sock
+              cachesize: 1000
+              timeout: 3s
+      {{- if eq .Cluster.ConfigItems.enable_encryption "true" }}
+          - identity: {}
+      {{- end }}
```
linki, line 755: This allows turning encryption on and off at will.

aermakov-zalando: We can't really turn it off once we've enabled it, can we?

linki: We can. All providers are used for reading but only the first is used for writing. This way one can seamlessly migrate from non-encrypted to encrypted, encrypted to non-encrypted and from one key to another.

The only thing to do is to swap the order and force a rewrite of every secret via `kubectl get secrets --all-namespaces -o json | kubectl replace -f -`.

mikkeloscar:
:+1:

aermakov-zalando:
👍

linki:
Added another config item that disables all components that are necessary to support encryption. The supporting components are required to seamlessly switched between encrypted and non-encrypted secrets. This should only be relevant during the migration to secret encryption by default.

These are harmless when encryption is turned off but nevertheless this config item leaves them out completely in order to not block this from being merged to dev. Specifically, this includes the encryption key which I'm not sure how to protect yet so it's easier to not create it in the first place.

linki:
:+1:

szuecs:
Thanks for the good explanation and the roll out plan!

szuecs:
:+1:

PR #2958, opened 2020-02-06
Admission controller: prevent deletion of in-use SAs

aermakov-zalando:
This will hopefully prevent some incidents.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2960, opened 2020-02-06
DNS: make the resources configurable

aermakov-zalando:
We already have issues with CoreDNS in at least one of the clusters.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2965, opened 2020-02-07
Improve CRD deletion

aermakov-zalando:
* Fix the role cleanup when the CRD is deleted
 * Forbid deleting a CRD unless all resources are gone
 * Allow the users to delete CRDs now that it's safe


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #2968, opened 2020-02-10
Prometheus: fix the QoS class

aermakov-zalando:
Init container was missing the resources, so the whole pod was downgraded to `Burstable`.


arjunrn:
:+1:

aermakov-zalando:
👍

PR #2972, opened 2020-02-10
Shutdown masters more gracefully

muaazsaleem:
### Wait to serve current requests before terminating api-server

To avoid spiking error rates for clients, take api-server out of
the kubernetes.default endpoint and then wait while the current
requests are being served

ref: https://github.com/kubernetes/kubernetes/pull/74416

Signed-off-by: Muaaz Saleem <muhammad.muaaz.saleem@zalando.de>

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -152,6 +152,7 @@ write_files:
           - --kubelet-certificate-authority=/etc/kubernetes/ssl/ca.pem
           - --kubelet-client-certificate=/etc/kubernetes/ssl/kubelet-client.pem
           - --kubelet-client-key=/etc/kubernetes/ssl/kubelet-client-key.pem
+          - --shutdown-delay-duration=60s
           livenessProbe:
             httpGet:
               host: 127.0.0.1


cluster/node-pools/master-default/userdata.yaml
```diff
@@ -97,6 +97,10 @@ write_files:
           image: nonexistent.zalan.do/teapot/kube-apiserver:fixed
           args:
           - --apiserver-count={{ .Values.apiserver_count }}
+          {{- if eq .Cluster.ConfigItems.endpoint_reconciliation "false" }}
```
aermakov-zalando: This will still roll the nodes I think because of the extra spaced.

cluster/manifests/kube-api-server/kuberntes-service.yaml
```diff
@@ -0,0 +1,40 @@
...
+  ports:
+  - name: https
+    port: 443
+    protocol: TCP
+    targetPort: 443
+  sessionAffinity: None
+  type: ExternalName
+---
+{{end}}
+apiVersion: v1
```
aermakov-zalando: Split into another file maybe?

cluster/manifests/kube-api-server/kuberntes-service.yaml
```diff
@@ -0,0 +1,40 @@
...
+  namespace: kube-system
+spec:
+  ports:
+  - name: https
+    port: 443
+    protocol: TCP
+    targetPort: 443
+  selector:
+    application: kube-apiserver
+  sessionAffinity: None
```
aermakov-zalando: Should we copy the settings from `kubernetes.default`?

cluster/manifests/kube-api-server/kuberntes-service.yaml
```diff
@@ -0,0 +1,40 @@
+# Since even the "lease" based endpoint reconciler does not shutdown
```
aermakov-zalando: Typo in the filename (`kuberntes`).

cluster/config-defaults.yaml
```diff
@@ -1,3 +1,6 @@
+# temporary feature toggle for the "kubernetes" endpoint reconciler, default is "lease", otherwise "none"
+endpoint_reconciliation: "true"
```
aermakov-zalando: You need two feature toggles (one to rollout the updated service, another one to disable the reconciler). This will disable the reconciler first but rollout the service later.

cluster/manifests/kube-api-server/kuberntes-service.yaml
```diff
@@ -0,0 +1,40 @@
...
+  name: kubernetes
+  namespace: default
+spec:
+  externalName: kubernetes.kube-system.svc.cluster.local
+  ports:
+  - name: https
+    port: 443
+    protocol: TCP
+    targetPort: 443
+  sessionAffinity: None
```
aermakov-zalando: Same question here: do we maybe need it?

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -155,6 +159,11 @@ write_files:
               path: /healthz
             initialDelaySeconds: 120
             timeoutSeconds: 15
+          # wait before terminating the api server so old connections get a chance to drain while the pod status is "terminating"
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-c",  " sleep 60"]
```
linki: There's also the `--shutdown-delay-duration` flag which might help here:

> Time to delay the termination. During that time the server keeps serving requests normally and /healthz returns success, but /readzy immediately returns failure. Graceful termination starts after this delay has elapsed. This can be used to allow load balancer to stop sending traffic to this server.

muaazsaleem: `--shutdown-delay-duration` doesn't help here because it's designed to give applications more time to shutdown. Our problem is that the api-server itself doesn't handle shutdown properly. So we're stopping new connections from coming in and waiting for connections to drain manually.

linki: I think that's what it does: I started apiserver with with `--shutdown-delay-duration=60s` and when I send it a `kill` it immediately takes the IP out of the endpoints list while the apiserver itself keeps running for another minute.

Only available from 1.16.

linki: For reference: https://github.com/kubernetes/kubernetes/pull/74416

aermakov-zalando: Nice! If this works, then we don't even need the rest of it.

muaazsaleem: Just tested `--shutdown-delay-duration` and it works! I think I was confusing it with `terminationGracePeriodSeconds ` in [PodSpec]( https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#pod-v1-core)

muaazsaleem:
👍

mikkeloscar:
:+1:

PR #2975, opened 2020-02-11
New Rule to Cleanup Resources from Pull Requests

### LINKED ISSUE #2930: Standardized Kube-Janitor Rules
In general, the deployment of [_kube-janitor_](https://github.com/hjacobs/kube-janitor) into every non-production is what were looking for to clean up our deployment artifacts in pull requests (PR).
But actually the convention in the default configuration in [`rules-config.yaml`](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/kube-janitor/rules-config.yaml) are quite strange for development teams:
- `id: require-application-label`: this one is fine
- `id: temporary-cdp-namespaces` is kind of what we need, but to specific to CDP and namespace names matching `d-*`.

*What we are aiming for?*

Imagine we are developing an application named `foobar` and have two clusters `proddi` for production and `devvi` for development. Of course, the production application from the master branch will run under `foobar.proddi.org` (`namespace: foobar`).

As kind of staging system for us and others, we would provide `foobar.devvi.org` (`namespace: foobar`) deployed from the _master_-branch and long running.
But if we are working on a PR like PR 5, we would like to get temporary resources for `foobar-pr-5.devvi.org` (namespace: pr-5-foobar`). 

_Proposed solution:_ Either we are able to overwrite the default configuration for _kube-janitor_ for our cluster or introduce kind of a company-wide standard to put temporary resources into namespaces matching `pr-*`. 

If you agree, I can create a PR for this.

heroldus:
The rule cleans up namespaces (including their resources) matching `*.-pr-.*` like `my-app-pr-123` after 24h to fix #2930. 

The rule allows to put resources build from pull requests into special namespaces which get cleaned up automatically. They don't mess up the K8s cluster anymore and disappear after 24h.

**TODO**:  We need to inform the developers as we cannot check whether the part after `-pr-` is a number. `my-pr-app` would match, too and would get cleaned up!


aermakov-zalando:
I've added a comment in #2930. A day sounds way too small for me, so I'd opt for less developer frustration, especially considering how little money this will save.

cluster/manifests/kube-janitor/rules-config.yaml
```diff
@@ -25,4 +25,13 @@ data:
         # see http://jmespath.org/specification.html#starts-with
         jmespath: "starts_with(metadata.name, 'd-')"
         ttl: 24h
+      - id: cleanup-resources-from-pull-requests
+        # After 24h, delete all resources in namespaces matching .*-pr-.* :
```
mikkeloscar: The comment is no longer correct since it configurable :)

cluster/config-defaults.yaml
```diff
@@ -279,3 +279,6 @@ vm_dirty_bytes: ""
 
 # temporary flag for kubernetes.io/node-pool node label
 legacy_node_pool_label_enabled: "false"
+
+# default ttl for kube janitor for resources build from PRs in namespaces matching .*-pr-.*
+default_pr_ttl: "1w"  # 1 week
```
mikkeloscar: could you prefix this config key with: `kube_janitor` so it's easier to understand what it's used for.

mikkeloscar:
:+1:

heroldus:
👍

hjacobs:
:+1:

PR #2980, opened 2020-02-12
Restrict access to etcd encryption key to Administrator

linki:
This denies PowerUser any kind of write and usage access to the etcd secrets encryption key. It still allows Administrator and the root account full access to the key.

Not the best solution but certainly works as expected.

New future roles that have `kms:*` would still be able to access the key but how likely is that at the moment? Unfortunately, a wild-card deny on `Shibboleth*` isn't supported in this syntax.

I would say for now it's totally sufficient to move forward.


cluster/cluster.yaml
```diff
@@ -1614,6 +1614,47 @@ Resources:
             Action:
             - "kms:Encrypt"
             - "kms:Decrypt"
+          - Sid: "Deny PowerUser to manage or use this key"
+            Effect: "Deny"
+            Principal:
+              AWS: "arn:aws:iam::{{ .Cluster.InfrastructureAccount | getAWSAccountID }}:role/Shibboleth-PowerUser"
```
mikkeloscar: Why not just Allow only Administrator e.g. white list? This currently wouldn't deny `Shibboleth-Manual` for instance.

linki: Thanks for the hint. I didn't think of `Manual` :disappointed: 

The problem is that PowerUser gets `Allow: kms:*` from the IAM policy and I don't intend to change that just yet. That leaves the black list like that.

There's a chance to define a white list here as well and have the final permissions be the intersection of both but I'm not sure. I'll check that but if you have any hints I'm happy to hear them.

linki: @mikkeloscar This works, thanks!

linki:
:+1:

gargravarr:
:+1:

PR #2984, opened 2020-02-13
Allow CLM to keep managing this key

linki:
I accidentally dropped permission to this key for CLM in https://github.com/zalando-incubator/kubernetes-on-aws/pull/2980. It needs it though.


linki:
:+1:

gargravarr:
:+1:

PR #2995, opened 2020-02-18
Collect vertical pod autoscaler object metrics

arjunrn:
`kube-state-metrics` supports VPA object metrics as well but it is not enabled by default.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #2996, opened 2020-02-18
Hack: Opt-in to using rotated service account tokens

linki:
This is how one can use rotating service account tokens without enabling it cluster-wide via the feature flag.


linki:
Closing for now. Keeping it locally for reference later.

PR #2999, opened 2020-02-18
Added a feature flag to restrict prometheus TSDB by size

arjunrn:
The below image shows the size of the TSDBs across our clusters. The smallest are less than a GB in size and the largest are around 10GBs. By setting the the max size for the TSDB to be `0.8` of the minimum memory for prometheus we can ensure that the TSDB always fits into memory when the container is started. This is important to prevent the constant restarts when the prometheus containers get a very small recommendation from the VPA.

prometheus_tsdb_sizes

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -107,9 +107,10 @@ scalyr_team_token: ""
 
 prometheus_cpu: "1000m"
 prometheus_mem: "4Gi"
-prometheus_mem_min: "1Gi"
+prometheus_mem_min: "2Gi"
 prometheus_cpu_min: "0"
 prometheus_remote_write: "disabled"
+prometheus_tsdb_retention_size: "disabled"
 
 metrics_service_cpu: "100m"
 metrics_service_mem: "200Mi"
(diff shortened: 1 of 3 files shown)


cluster/config-defaults.yaml
```diff
@@ -107,9 +107,10 @@ scalyr_team_token: ""
 
 prometheus_cpu: "1000m"
 prometheus_mem: "4Gi"
-prometheus_mem_min: "1Gi"
+prometheus_mem_min: "2Gi"
```
mikkeloscar, line 110: What's the motivation for increasing the min? It seems like this should work fine from your description?

> The smallest are less than a GB in size

arjunrn: The min size of the memory determines how large the TSDB can be. This means that for large clusters `1Gi` might be too small resulting in only a few hours of retention. We can keep it at `1Gi` and then customize this on a cluster-by-cluster basis but it's just more work.

arjunrn:
:+1:

cluster/config-defaults.yaml
mikkeloscar: Was there a metric we can use to determine the retention time for different clusters? This way we can evaluate if it later makes sense to lower it.

If I look over the last hour only 10% of Prometheis are using more than `2Gi` memory.

image

mikkeloscar: We're now tracking the retention time for each Prometheus, to understand how this change impacts it!

mikkeloscar:
:+1:

cluster/config-defaults.yaml
mikkeloscar: image

PR #3002, opened 2020-02-19
Define VPA as single app with components

mikkeloscar:
We labelled the vpa "backwards" when we introduced it. This means we have 3 apps defined in our application registry, when we only need one.

Change all components to have ~`application: vpa`~ `application: vertical-pod-autoscaler` and a custom `component` label.

To roll this out in a "safe" way I added pre-apply delete for all components. For `updater` and `recommender` this is ok to happen all the time, for admission-controller I renamed it to have `-new` suffix which I will rename later with another PR once this is all rolled out.

### DIFF
--- a/cluster/manifests/01-vertical-pod-autoscaler/02-secret.yaml
+++ b/cluster/manifests/01-vertical-pod-autoscaler/02-secret.yaml
@@ -4,8 +4,8 @@ metadata:
   name: vpa-tls-certs
   namespace: kube-system
   labels:
-    application: vpa-admission-controller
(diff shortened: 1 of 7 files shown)


hjacobs:
`vertical-pod-autoscaler` might be a better application ID --- more meaningful and less confusing (other colleagues should not have to guess when they see "vpa")

mikkeloscar:
@hjacobs good point. I have changed it.

mikkeloscar:
:+1:

hjacobs:
:+1:

PR #3007, opened 2020-02-20
roll back to version v0.11.44

aryszka:
rolling back to version v0.11.44 due to discovered panics in the compress filter


aryszka:
:+1:

aryszka:
not rolling back, as we need some of the other changes

PR #3019, opened 2020-02-24
Add a cluster-local file to the secrets

aermakov-zalando:
Our e2e clusters run with the same set of certificates, so the generated 'secrets' file is the same for all of them. However, they're encrypted with a per-run KMS key, so it's possible that an e2e will fail because its secrets are reencrypted with a key from a different run. Add a dummy file with the cluster's local ID to prevent this as a quick hack.


aryszka:
:+1:

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3020, opened 2020-02-24
Webhook drop non resource urls

mikkeloscar:
Updates the webhook to drop support for allowing non-resource URLs. Instead those will be handled by RBAC which correctly blocks non GET requests to the non-resource URLs.
Before we allowed e.g. POST to `/api` which would trigger non valid audit events.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3021, opened 2020-02-24
Disable scale down delay

aermakov-zalando:
It's not really clear what problem it's supposed to solve, and it can prevent clusters from being scaled down for a very long time (in the worst case days). Let's disable it.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3027, opened 2020-02-26
Align beta with kube-1.16-beta

mikkeloscar:
Promote `beta` channel to Kubernetes v1.16


aermakov-zalando:
Why? Let's just merge via `dev` the normal way.

PR #3033, opened 2020-02-26
auth-webhook: log default sa reads

mikkeloscar:
Updates the auth webhook to log whenever a default service account is reading from the APIserver. This change is made in order to discover all clusters where the default service account is used for reading from the Kubernetes API which is something we want to prevent in the future.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3034, opened 2020-02-27
Set vm dirty bytes by default

mikkeloscar:
We have enabled it everywhere (except `search`) so let's configure it by default.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3036, opened 2020-02-27
Ingress LB: Only target node pools where skipper can run

mikkeloscar:
This enables only putting node pools as Ingress LB target if skipper can actually run on the node pool. The motivation for this is that AWS does "connection draining" on nodes that are part of an Autoscaling Group which is a target of a Load Balancer, even if the node was never actually a healthy target that got any traffic.
The impact of this "connection draining" is that node decommissioning takes ~10 min. longer than it has to.

## How it works

1. All node pools without taint (such that skipper couldn't run) gets a tag `zalando.org/ingress-enabled=true`
2. The ingress-controller uses this tag as a filter to only consider node pools to be attached in the Load Balancers.

With support for dedicated skipper node pool, this setup becomes a bit more complicated as we have to be careful to have the right node pools in the Load Balancer at the right time when we enable/disable dedicated skipper node pool.

To enable a safe configuration I introduced a config item: `disable_default_nodepool_lb_attach=true|false`.
The idea is that if you want to enable dedicated skipper node pool support for a cluster you do the following:

1. Add node pool with `taints=dedicated=skipper-ingress:NoSchedule`
2. Move skipper to dedicated nodes: set `disable_default_nodepool_lb_attach=true`
3. IMPORTANT: Wait for all skippers to be running on the dedicated node pool
4. Remove normal node pools from LB target: set `disable_default_nodepool_lb_attach=true`

And if you want to roll back to non dedicated node pool for skipper, it's the reverse order:

1. Add normal node pools as LB targets: set `disable_default_nodepool_lb_attach=false`
2. WAIT to ensure the change is applied, you should see the "unhealthy" targets in the LBs.
3. Move skipper to non-dedicated nodes: set `disable_default_nodepool_lb_attach=false`
4. Optional: remove the node pool with `taints=dedicated=skipper-ingress:NoSchedule`

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -29,6 +29,9 @@ skipper_ingress_cpu: "1000m"
 skipper_ingress_memory: "1Gi"
 skipper_ingress_tracing_buffer: "8192"
 enable_dedicate_nodepool_skipper: "false"
+# This setting should only be enabled after `enable_dedicate_nodepool_skipper`
+# has been enabled and is successfully switched
+disable_default_nodepool_lb_attach: "false"
 
 # skipper default filters
 skipper_default_filters: 'enableAccessLog(4,5) -> lifo(2000,20000,"3s")'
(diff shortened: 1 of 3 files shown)


mikkeloscar:
Closing in favor of #3074

PR #3037, opened 2020-02-28
Admission: add support for aws-credentials-waiter injection

aermakov-zalando:
If `teapot_admission_controller_experimental_inject_aws_waiter` is set to `true`, automatically inject the AWS credentials waiter initContainer to pods that need it. Will test in a couple of clusters and then enable unconditionally.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3038, opened 2020-02-28
Enable secrets encryption by default in test clusters

linki:
Enable secrets encryption by default in test clusters.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -289,11 +289,19 @@ enable_hpa_scale_to_zero: "true"
 
 # setup supporting components to enable encryption
 # this flag must only be switched from true to false when enable_encryption is false and all secrets were decrypted
+{{if eq .Environment "test"}}
+support_encryption: "true"
+{{else}}
 support_encryption: "false"
+{{end}}
 # enable encryption of secrets in etcd, requires support_encryption: true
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
We can't roll it out in one go, since API servers that have not been updated won't be able to read the secrets. Let's split into two.

cluster/config-defaults.yaml
```diff
@@ -289,11 +289,19 @@ enable_hpa_scale_to_zero: "true"
 
 # setup supporting components to enable encryption
 # this flag must only be switched from true to false when enable_encryption is false and all secrets were decrypted
+{{if eq .Environment "test"}}
+support_encryption: "true"
```
aermakov-zalando, line 293: I'd also rename the config item to something more specific (encryption of what?).

linki:
I'm currently updating them. This is in preparation to drop the cluster-registry overrides in the future and to use encryption from the start for new clusters.

Do you still see issues with it?

aermakov-zalando:
If they're manually updated then it should be fine.

cluster/config-defaults.yaml
linki: I would defer this. The config item hopefully won't be here for long.

linki:
:+1:

aermakov-zalando:
👍

PR #3040, opened 2020-02-28
Secrets encryption: increase cache size to 50000

aermakov-zalando:
This pretty much *has* to be greater than the number of secrets we have in the clusters, otherwise performance degrades massively. We'll add monitoring later, for now let's just stop the bleeding. Memory impact is pretty much nil since the cached entries are very small.


linki:
:+1:

szuecs:
:+1:

PR #3044, opened 2020-03-02
PCS: Add schema validation for status

mikkeloscar:
This adds status validation for the `PlatformCredentialsSets` CRD. Without this the status is not returned when reading a resource from the API which results in errors for e.g. our deployment system where the status is expected.

This issue is only observed in clusters created AFTER Kubernetes v1.16 was merged, but it is unclear why it happens as there are no errors logged in the APIServer or anywhere else.

~Additionally it removes `nullable: true` as this is not supported and results in the following error in APIServer logs:~

```
'ERROR $root.definitions.org.zalando.v1.PlatformCredentialsSet.properties.spec.properties.tokens.additionalProperties.schema.properties.privileges has invalid property: nullable'
```

### DIFF
--- a/cluster/manifests/01-platformcredentialsset/customresourcedefinition.yaml
+++ b/cluster/manifests/01-platformcredentialsset/customresourcedefinition.yaml
@@ -82,5 +82,34 @@ spec:
                 enum:
                   - v1
                   - v2
+          status:
+            type: object
+            properties:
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3054, opened 2020-03-09
Fix duplicate keys in manifest YAMLs

aermakov-zalando:
We've got a bunch of issues that we never noticed. https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/272 uses more strict validation, so let's fix these before applying.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3065, opened 2020-03-11
Add support for internal skipper API Server proxy

mikkeloscar:
This adds an internal skipper proxy in front of the API Server (`10.3.0.1`) when the flag `apiserver_internal_proxy` is `true` (default `false`).

The purpose is to route all traffic going to `10.3.0.1` through this proxy in clusters where the `default` service account is being used to read from the APIServer and log any requests from pods with the default service account, such that we can track the source IP and identify the application which needs to be updated.

Unfortunately this only shows the node where the pod is running as the source IP so we still can't 100% identify which pod is the source. But it should help us get a better understanding than simply knowing the `default` service account is being used which is all the visibility we currently have.

Since we never ran with a skipper in front of `10.3.0.1` the intention is to ONLY enable this in test clusters where we see the `default` service account being used and verify that this does not cause issues for other workloads before we decide if it should be enabled for production clusters as well.
This proxy should be completely removed once we have migrated away from legacy service accounts.

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -110,7 +110,7 @@ write_files:
           {{- end }}
           - --allow-privileged=true
           - --service-cluster-ip-range=10.3.0.0/16
-          - --secure-port=443
+          - --secure-port={{ if eq .Cluster.ConfigItems.apiserver_internal_proxy "true" }}6443{{else}}443{{end}}
           - --enable-admission-plugins=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,DefaultTolerationSeconds,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota,StorageObjectInUseProtection,PodSecurityPolicy,ImagePolicyWebhook,Priority,ExtendedResourceToleration
           - --tls-cert-file=/etc/kubernetes/ssl/apiserver.pem
           - --tls-private-key-file=/etc/kubernetes/ssl/apiserver-key.pem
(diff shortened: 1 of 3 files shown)


cluster/node-pools/master-default/userdata.yaml
```diff
@@ -433,6 +433,51 @@ write_files:
...
+          - -insecure
+          - -experimental-upgrade
+          - -runtime-metrics
+          - -enable-connection-metrics
+          - -enable-prometheus-metrics
+          - -write-timeout-server=60m
+          - -inline-routes
+          - |
+            s: JWTPayloadAllKV("kubernetes.io/serviceaccount/service-account.name", "default")
+              -> enableAccessLog()
```
szuecs, line 455: Maybe add the auditLog filter https://opensource.zalando.com/skipper/reference/filters/#auditlog to log data from the body in order to get more information about the client. I don't really know if this is valuable, but might be interesting to test.

aryszka: we could try

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -433,6 +433,51 @@ write_files:
           - mountPath: /etc/kubernetes/ssl
             name: ssl-certs-kubernetes
             readOnly: true
+{{ if eq .Cluster.ConfigItems.apiserver_internal_proxy "true" }}
+        - name: skipper-proxy-internal
+          image: registry.opensource.zalan.do/pathfinder/skipper:v0.11.35
```
aryszka, line 438: why not a more recent version of skipper? latest is v0.11.66

cluster/node-pools/master-default/userdata.yaml
mikkeloscar: Tried `auditlog` filter but it doesn't add additional information in our case.

mikkeloscar:
This is most likely being replaced by #3066

mikkeloscar:
Closing in favor of #3066

PR #3066, opened 2020-03-11
Enable read auditlog for test with enable_default_sa=true

mikkeloscar:
This achieves the same as #3065 but uses Kubernetes audit logs instead of a proxy.

The purpose of this is to log (to stdout from `kube-apiserver`) all read events from service accounts not in `kube-system` namespace such that we can better identify pods using the `default` service account to read from the API server, something we want to migrate away from.

This logs more events than we need as we can't easily match "`default` service account in ANY namespace" via the audit configuration, but it will only be enabled for clusters where `enable_default_sa=true` and only in `test` as a start to see how much it helps us before we start to add extra logging in production clusters.

The events we will get looks like this:

```
{"kind":"Event","apiVersion":"audit.k8s.io/v1","level":"Request","auditID":"x","stage":"ResponseComplete","requestURI":"/apis","verb":"get","user":{"username":"system:serviceaccount:default:default","uid":"x","groups":["system:serviceaccounts","system:serviceaccounts:default","system:authenticated"]},"sourceIPs":["10.x.x.x"],"userAgent":"curl/7.64.0","responseStatus":{"metadata":{},"code":200},"requestReceivedTimestamp":"x","stageTimestamp":"x","annotations":{"authorization.k8s.io/decision":"allow","authorization.k8s.io/reason":"RBAC: allowed by ClusterRoleBinding \"system:discovery\" of ClusterRole \"system:discovery\" to Group \"system:authenticated\""}}
```

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -139,6 +139,11 @@ write_files:
           - --audit-log-maxage=0
           - --audit-log-maxbackup=0
           - --audit-policy-file=/etc/kubernetes/config/audit-policy.yaml
+          {{ else if and (eq .Cluster.ConfigItems.enable_default_sa "true") (eq .Cluster.ConfigItems.auditlog_read_access "true") }}
+          - --audit-log-path=/dev/stdout
+          - --audit-log-maxage=0
+          - --audit-log-maxbackup=0
+          - --audit-policy-file=/etc/kubernetes/config/audit-policy-ro.yaml
           {{ end }}
           # enable aggregated apiservers
           - --client-ca-file=/etc/kubernetes/ssl/ca.pem
@@ -744,6 +749,24 @@ write_files:
         - level: Request
           omitStages:
             - "RequestReceived"
+
+{{ if and (eq .Cluster.ConfigItems.enable_default_sa "true") (eq .Cluster.ConfigItems.auditlog_read_access "true") }}
+  - owner: root:root
+    path: /etc/kubernetes/config/audit-policy-ro.yaml
+    content: |
+      apiVersion: audit.k8s.io/v1beta1
+      kind: Policy
+      rules:
+      - level: None
+        verbs: ["create", "update", "patch", "delete", "deletecollection"]
+      - level: None
+        userGroups: ["system:serviceaccounts:kube-system"]
+      - level: Request
+        verbs: ["watch", "list", "get"]
+        userGroups: ["system:serviceaccounts"]
+        omitStages:
+        - "RequestReceived"
+{{ end }}
 
 {{- if eq .Cluster.ConfigItems.support_encryption "true" }}
   - owner: root:root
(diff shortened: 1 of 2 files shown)


cluster/node-pools/master-default/userdata.yaml
```diff
@@ -139,6 +139,11 @@ write_files:
           - --audit-log-maxage=0
           - --audit-log-maxbackup=0
           - --audit-policy-file=/etc/kubernetes/config/audit-policy.yaml
+          {{ else if and (eq .Cluster.ConfigItems.enable_default_sa "true") (eq .Cluster.ConfigItems.auditlog_read_access "true") }}
+          - --audit-log-path=/dev/stdout
+          - --audit-log-maxage=0
+          - --audit-log-maxbackup=0
+          - --audit-policy-file=/etc/kubernetes/config/audit-policy-ro.yaml
```
aermakov-zalando, line 147: Can this coexist with the other auditlog configurations?

aermakov-zalando: We have `audittrail_url` set in one of the test clusters, for example, just to test the whole thing (adapter + resender).

mikkeloscar: Ah right. I missed that. Let me change it so it supports both if possible.

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -745,6 +750,24 @@ write_files:
...
 
+{{ if and (eq .Cluster.ConfigItems.enable_default_sa "true") (eq .Cluster.ConfigItems.auditlog_read_access "true") }}
+  - owner: root:root
+    path: /etc/kubernetes/config/audit-policy-ro.yaml
+    content: |
+      apiVersion: audit.k8s.io/v1beta1
+      kind: Policy
+      rules:
+      - level: None
+        verbs: ["create", "update", "patch", "delete", "deletecollection"]
```
muaazsaleem: I understand why the following `None` rule is necessary:
```
- level: None
   userGroups: ["system:serviceaccounts:kube-system"]
```
Could you say why we need to explicitly disable the `verbs` we never enabled?

cluster/node-pools/master-default/userdata.yaml
aermakov-zalando: If that's too complicated, we could also make them mutually exclusive and just chase people in the test cluster to migrate off. We don't change the audittrail adapter very often anyway.

mikkeloscar: I changed it so it can work together, only thing is we would now log all audit events to the apiserver logs when `auditlog_read_access: "true"`. Considering we only need this for a limited time to identify the users I think this is ok and I wouldn't make it more complicated than we need to.

cluster/node-pools/master-default/userdata.yaml
mikkeloscar: Good point, this is not actually needed. I have dropped it.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3073, opened 2020-03-17
Add missing RBAC permission to kube-state-metrics

linki:
It now looks for VPAs and needs permission.

### DIFF
--- a/cluster/manifests/kube-state-metrics/rbac.yaml
+++ b/cluster/manifests/kube-state-metrics/rbac.yaml
@@ -78,6 +78,7 @@ rules:
   - autoscaling
   resources:
   - horizontalpodautoscalers
+  - verticalpodautoscalers
   verbs:
   - list
   - watch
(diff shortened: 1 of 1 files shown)


cluster/manifests/kube-state-metrics/rbac.yaml
```diff
@@ -78,6 +78,7 @@ rules:
   - autoscaling
   resources:
   - horizontalpodautoscalers
+  - verticalpodautoscalers
```
mikkeloscar: Does this work? I would imagine `autoscaling.k8s.io` is a different `apiGroup` than `autoscaling`?

linki: That's what it claimed it needs in the error message.

linki: Ah, no, `autoscaling.k8s.io` is not equal to `autoscaling` I guess :face_with_head_bandage:

linki: Hopefully fixed.

linki:
:+1:

mikkeloscar:
:+1:

PR #3081, opened 2020-03-18
Add rbac for scalyr-key-rotation

pheanex:
Change to allow the service scalyr-key-rotation access to the two scalyr secrets.
Please check if this makes sense to you with the 4 yaml documents.
Maybe there is a simpler way of doing this?

### DIFF
--- a/cluster/manifests/roles/scalyr-key-rotation-rbac.yaml
+++ b/cluster/manifests/roles/scalyr-key-rotation-rbac.yaml
@@ -1,0 +1,63 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: scalyr-key-rotation
+  namespace: visibility
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  resourceNames:
+  - scalyr-keys
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: scalyr-key-rotation
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  resourceNames:
+  - scalyr-delegation-token
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: scalyr-key-rotation
+  namespace: visibility
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: scalyr-key-rotation
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: User
+  name: zalando-iam:zalando:service:scalyr-key-rotation
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: scalyr-key-rotation
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: scalyr-key-rotation
+subjects:
+- apiGroup: rbac.authorization.k8s.io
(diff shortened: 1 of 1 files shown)


cluster/manifests/roles/scalyr-key-rotation-rbac.yaml
```diff
@@ -0,0 +1,63 @@
...
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  resourceNames:
+  - scalyr-delegation-token
+  verbs:
+  - get
+  - list
```
aermakov-zalando: `list` can't be used with `resourceNames` unfortunately, you need a separate entry for it.

cluster/manifests/roles/scalyr-key-rotation-rbac.yaml
```diff
@@ -0,0 +1,63 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
aermakov-zalando: Why is this a ClusterRole? It's only used in one namespace, so it should be a Role.

cluster/manifests/roles/scalyr-key-rotation-rbac.yaml
```diff
@@ -0,0 +1,63 @@
...
+  name: scalyr-key-rotation
+  namespace: visibility
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: scalyr-key-rotation
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: User
+  name: zalando-iam:zalando:service:scalyr-key-rotation
```
aermakov-zalando, line 48: If you intend to use PlatformCredentialsSets, this would be `stups_scalyr-key-rotation`.

cluster/manifests/roles/scalyr-key-rotation-rbac.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - scalyr-delegation-token
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
```
aermakov-zalando: Again, this could be a RoleBinding.

cluster/manifests/roles/scalyr-key-rotation-rbac.yaml
pheanex: I am not even sure we need list. Let's remove it and see if we can work without it.

cluster/manifests/roles/scalyr-key-rotation-rbac.yaml
pheanex: With the deprecation of `stups` will this also be renamed at some point? I've tripped over this issue a couple of times in different places now

cluster/manifests/roles/scalyr-key-rotation-rbac.yaml
```diff
@@ -0,0 +1,61 @@
...
+kind: Role
+metadata:
+  name: scalyr-key-rotation
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  resourceNames:
+  - scalyr-delegation-token
```
aermakov-zalando, line 29: Is it supposed to be in `visibility` as well? In this case please just add the second name to the role instead of creating another one.

cluster/manifests/roles/scalyr-key-rotation-rbac.yaml
aermakov-zalando: Very unlikely, because it's used everywhere.

pheanex:
Closed as suggested by @aermakov-zalando as we can deploy this via the service repository

cluster/manifests/roles/scalyr-key-rotation-rbac.yaml
pheanex: no it is in default namespace

PR #3082, opened 2020-03-18
Enable the EBS CSI driver

arjunrn:
This PR adds [CSI components](https://kubernetes-csi.github.io/docs/) for Kubernetes with the [EBS driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver).
 Also changed in this PR is the `prometheus` statefulset. Another statefulset needs to be created because the `persistentVolumeClaims` cannot be modified. Instead a new `statefulset` is created with CSI volumes.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -1340,8 +1340,61 @@ Resources:
           PolicyName: root
       RoleName: "{{.Cluster.LocalID}}-kube-metrics-adapter"
     Type: 'AWS::IAM::Role'
-{{ if eq .Cluster.Environment "e2e" }}
-  # This is a hack to easily create an aws iam role and s3 bucket for testing
+
+  EBSCSIControllerIAMRole:
+    Properties:
+      AssumeRolePolicyDocument:
+        Statement:
+          - Action:
+              - 'sts:AssumeRole'
+            Effect: Allow
+            Principal:
+              Service:
+                - ec2.amazonaws.com
+          - Action:
+              - 'sts:AssumeRole'
+            Effect: Allow
+            Principal:
+              AWS: !Join
+                - ''
+                - - 'arn:aws:iam::{{.Cluster.InfrastructureAccount | getAWSAccountID}}:role/'
+                  - !Ref MasterIAMRole
+          - Action:
+              - 'sts:AssumeRole'
+            Effect: Allow
+            Principal:
+              AWS: !Join
+                - ''
+                - - 'arn:aws:iam::{{.Cluster.InfrastructureAccount | getAWSAccountID}}:role/'
+                  - !Ref WorkerIAMRole
+        Version: 2012-10-17
+      Path: /
+      Policies:
+        - PolicyDocument:
+            Statement:
+              - Effect: Allow # All the policies needed for the EBS CSI driver
+                Action:
+                  - 'ec2:AttachVolume'
+                  - 'ec2:CreateSnapshot'
+                  - 'ec2:CreateTags'
+                  - 'ec2:CreateVolume'
+                  - 'ec2:DeleteSnapshot'
+                  - 'ec2:DeleteTags'
+                  - 'ec2:DeleteVolume'
+                  - 'ec2:DescribeInstances'
+                  - 'ec2:DescribeSnapshots'
+                  - 'ec2:DescribeTags'
+                  - 'ec2:DescribeVolumes'
+                  - 'ec2:DetachVolume'
+                  - 'ec2:ModifyVolume'
+                Resource: '*'
+            Version: 2012-10-17
+          PolicyName: root
+      RoleName: "{{.Cluster.LocalID}}-ebs-csi-controller"
+    Type: 'AWS::IAM::Role'
+
+  {{ if eq .Cluster.Environment "e2e" }}
+# This is a hack to easily create an aws iam role and s3 bucket for testing
   # AWS IAM intergration in e2e tests.
   E2EAWSIAMTestRole:
     Properties:
--- a/cluster/manifests/03-ebs-csi/02-csi-driver.yaml
+++ b/cluster/manifests/03-ebs-csi/02-csi-driver.yaml
@@ -1,0 +1,143 @@
+{{- if eq .ConfigItems.prometheus_csi_ebs "true" }}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: csi-node-driver
+  namespace: kube-system
+spec:
+  selector:
+    matchLabels:
+      application: container-storage-interface
+      component: driver
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        application: container-storage-interface
+        component: driver
+    spec:
+      initContainers:
+        - name: cleanup
+          image: registry.opensource.zalan.do/stups/alpine:3.10.3-5
+          command:
+             - /bin/sh
+          args:
+             - -c
+             - rm -rf /registration/ebs.csi.aws.com-reg.sock /csi/csi.sock
+          resources:
+            requests:
+              cpu: 10m
+              memory: 30Mi
+            limits:
+              cpu: 10m
+              memory: 30Mi
+          volumeMounts:
+            - mountPath: /csi
+              name: plugin-dir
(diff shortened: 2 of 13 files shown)


hjacobs:
Can you link to more information on what this is about?

cluster/cluster.yaml
```diff
@@ -1340,8 +1340,61 @@ Resources:
...
+                - - 'arn:aws:iam::{{.Cluster.InfrastructureAccount | getAWSAccountID}}:role/'
+                  - !Ref MasterIAMRole
+          - Action:
+              - 'sts:AssumeRole'
+            Effect: Allow
+            Principal:
+              AWS: !Join
+                - ''
+                - - 'arn:aws:iam::{{.Cluster.InfrastructureAccount | getAWSAccountID}}:role/'
+                  - !Ref WorkerIAMRole
```
aermakov-zalando, line 1369: Do we really need it? We'll run the controller on the masters, so the workers shouldn't be able to delete volumes.

cluster/manifests/03-ebs-csi/02-csi-driver.yaml
```diff
@@ -0,0 +1,127 @@
...
+            periodSeconds: 10
+            timeoutSeconds: 3
+          name: ebs-plugin
+          ports:
+            - containerPort: 9808
+              name: healthz
+              protocol: TCP
+          resources:
+            requests:
+              cpu: 50m
```
aermakov-zalando: Does it really need `50m`? This might be out of our budget in some of the clusters, and most of the time it shouldn't even do anything.

cluster/manifests/03-ebs-csi/controller-aws-iam-role.yaml
```diff
@@ -0,0 +1,9 @@
+{{ if eq .ConfigItems.kube_aws_iam_controller_kube_system_enable "true"}}
```
aermakov-zalando: We haven't had any issues with it, so I'd drop the config item. We can do this in a separate PR, but I wouldn't bother with it for new things.

cluster/manifests/03-ebs-csi/ebs-controller.yaml
```diff
@@ -0,0 +1,152 @@
...
+      containers:
+        - args:
+            - controller
+            - --endpoint=$(CSI_ENDPOINT)
+            - --logtostderr
+            - --v=5
+          env:
+            - name: CSI_ENDPOINT
+              value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
+          {{ if eq .ConfigItems.kube_aws_iam_controller_kube_system_enable "true"}}
```
aermakov-zalando: Same here. Let's just use one mechanism.

cluster/manifests/03-ebs-csi/02-csi-driver.yaml
```diff
@@ -0,0 +1,127 @@
+apiVersion: apps/v1
```
aermakov-zalando, line 2: Since we're just testing it, let's make it conditional? Deploying it to all clusters might not be feasible because we might not have enough space on the nodes.

cluster/manifests/03-ebs-csi/02-csi-driver.yaml
```diff
@@ -0,0 +1,127 @@
...
+  template:
+    metadata:
+      labels:
+        application: container-storage-interface
+        component: driver
+    spec:
+      containers:
+        - args:
+            - node
+            - --endpoint=$(CSI_ENDPOINT)
```
aermakov-zalando, line 43: Didn't know you could do that!

cluster/manifests/03-ebs-csi/02-csi-driver.yaml
```diff
@@ -0,0 +1,127 @@
...
+            limits:
+              cpu: 50m
+              memory: 50Mi
+          securityContext:
+            privileged: true
+          volumeMounts:
+            - mountPath: /opt/podruntime/kubelet
+              mountPropagation: Bidirectional
+              name: kubelet-dir
+            - mountPath: /csi
```
aermakov-zalando, line 76: `/var/run/csi`? This way it'll be cleaned up on reboots.

cluster/manifests/03-ebs-csi/02-csi-driver.yaml
```diff
@@ -0,0 +1,127 @@
...
+            - name: DRIVER_REG_SOCK_PATH
+              value: /opt/podruntime/kubelet/plugins/ebs.csi.aws.com/csi.sock
+          image: registry.opensource.zalan.do/teapot/csi-node-driver-registrar:v1.1.0-master-1
+          lifecycle:
+            preStop:
+              exec:
+                command:
+                  - /bin/sh
+                  - -c
+                  - rm -rf /registration/ebs.csi.aws.com-reg.sock /csi/csi.sock
```
aermakov-zalando: See above, if we only cleanup in `preStop` it'll be in a broken state after a node reboot.

cluster/manifests/03-ebs-csi/ebs-controller.yaml
```diff
@@ -0,0 +1,152 @@
...
+              protocol: TCP
+          resources:
+            requests:
+              cpu: 50m
+              memory: 50Mi
+            limits:
+              cpu: 50m
+              memory: 50Mi
+          volumeMounts:
+            - mountPath: /var/lib/csi/sockets/pluginproxy/
```
aermakov-zalando, line 51: Same here, `.sock` files are usually in `/var/run`.

cluster/cluster.yaml
arjunrn: So the worker role is necessary right now because of how `kube-aws-iam-controller` gets credentials right now. It assumes the worker's role first to keep compatibility with kube2iam. So the `WorkerIAMRole` needs to be present for now. Once we move to pod identity credentials then this can be removed.

cluster/manifests/03-ebs-csi/02-csi-driver.yaml
arjunrn: I will tie it to the same config item which enables the CSI backed PVCs for prometheus.

cluster/manifests/03-ebs-csi/02-csi-driver.yaml
arjunrn: Can be used for when the application cannot accept a secret through a mounted environment variable.

cluster/manifests/03-ebs-csi/02-csi-driver.yaml
arjunrn: What if we change it to a `preStart` hook?

cluster/manifests/03-ebs-csi/ebs-controller.yaml
arjunrn: This socket is not mounted from the host but is an `emptyDir` so node restarts should not be an issue.

cluster/manifests/03-ebs-csi/02-csi-driver.yaml
aermakov-zalando: That should work. We can also use an initcontainer for this.

arjunrn: I will change it to an `initContainer`. Also the socket file needs to be that particular directory for kubelet to talk to it.

cluster/cluster.yaml
aermakov-zalando: I see, thanks!

cluster/manifests/03-ebs-csi/02-csi-driver.yaml
arjunrn: This was addressed by having an `initContainer`. See other comment.

arjunrn:
:+1:

cluster/cluster.yaml
linki: Can you add a note that this can be removed as a comment somewhere (Also fine in the pod identity issue)? Otherwise I think it'll just stick around b/c no one knows what is used for what.

jrake-revelant:
👍

PR #3083, opened 2020-03-19
Only handle CRD v1 in admission-controller

mikkeloscar:
Updates the admission-controller to only handle CRDs in v1. This ensures that our automatic RBAC rules for CRDs works both with `v1beta1` (as before) and also with `v1`.

Admission controller will only accept CRD of `v1` and any `v1beta1` CRDs will be converted by the APIserver before being sent to the admission controller.

Before it didn't work for `v1` CRDs because the status and scale subresources were not correctly discovered in the admission controller.

### DIFF
--- a/cluster/manifests/01-admission-control/teapot.yaml
+++ b/cluster/manifests/01-admission-control/teapot.yaml
@@ -110,10 +110,11 @@ webhooks:
       url: "https://localhost:8085/crd"
       caBundle: "{{ .ConfigItems.ca_cert_decompressed }}"
     failurePolicy: Fail
+    matchPolicy: Equivalent
     rules:
       - operations: ["CREATE", "UPDATE", "DELETE"]
         apiGroups: ["apiextensions.k8s.io"]
-        apiVersions: ["v1", "v1beta1"]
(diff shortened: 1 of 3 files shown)


cluster/manifests/01-admission-control/teapot.yaml
```diff
@@ -113,7 +113,7 @@ webhooks:
     rules:
       - operations: ["CREATE", "UPDATE", "DELETE"]
         apiGroups: ["apiextensions.k8s.io"]
-        apiVersions: ["v1", "v1beta1"]
+        apiVersions: ["v1"]
```
aermakov-zalando, line 117: What happens if someone still applies using `v1beta1`?

mikkeloscar: Good point I was missing: `matchPolicy: Equivalent`, added now.

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchpolicy

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3084, opened 2020-03-20
AWS OIDC based IAM credentials (Part 1)

linki:
Incremental rollout of https://github.com/zalando-incubator/kubernetes-on-aws/pull/2918

This deploys the pod identity webhook and registers it with the API. It's behind a feature flag to make it easier to merge. Even with the config item set to enable it wouldn't use the new credentials as applications need to opt-in via an annotation on the service account. This will be part of the following PRs.


cluster/cluster.yaml
```diff
@@ -124,6 +124,12 @@ Resources:
           FromPort: 8085
           IpProtocol: tcp
           ToPort: 8085
+{{- if eq .Cluster.ConfigItems.teapot_admission_controller_service_account_iam "true" }}
```
arjunrn: This should be dropped. I think it's from when the webhook as deployed as it's own sidecar.

cluster/manifests/prometheus/configmap.yaml
```diff
@@ -52,6 +52,22 @@ data:
       - action: replace
         source_labels: ['__meta_kubernetes_pod_label_application']
         target_label: application
+{{- if eq .Cluster.ConfigItems.teapot_admission_controller_service_account_iam "true" }}
```
arjunrn: Dropped as well

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -128,6 +128,11 @@ write_files:
           - --service-account-signing-key-file=/etc/kubernetes/ssl/service-account-private-key.pem
           - --service-account-issuer=kubernetes/serviceaccount
           {{- end }}
+          {{- if eq .Cluster.ConfigItems.teapot_admission_controller_service_account_iam "true" }}
```
arjunrn, line 131: I was not sure about this so I created a new key. I think this is not required. Only one `service-account-signing-key-file` and corresponding `service-account-key-file` should be specified.

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -243,13 +248,49 @@ write_files:
...
+{{- if eq .Cluster.ConfigItems.teapot_admission_controller_service_account_iam "true" }}
+            - --service-account-iam
 {{- end }}
           ports:
             - containerPort: 8085
           volumeMounts:
             - mountPath: /etc/kubernetes/ssl
               name: ssl-certs-kubernetes
               readOnly: true
+{{- if eq .Cluster.ConfigItems.teapot_admission_controller_service_account_iam "true" }}
```
arjunrn: Dropped as well

cluster/cluster.yaml
linki: I totally missed that. Removed the related stuff.

cluster/node-pools/master-default/userdata.yaml
linki: You're right. Dropped that as well.

linki:
:+1:

muaazsaleem:
👍

PR #3091, opened 2020-03-25
Update ExternalDNS to use pod identity credentials

linki:
This enables ExternalDNS to use the new IAM credentials mechanism.

Looks a bit cumbersome but that's due to the **three** different ways we can get AWS IAM credentials at the moment and I didn't want to change that for now.

Before continuing with the next services I'll intend to reduce it to avoid the many conditionals.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -715,6 +715,16 @@ Resources:
     Properties:
       AssumeRolePolicyDocument:
         Statement:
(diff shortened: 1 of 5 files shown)


linki:
:+1:

arjunrn:
Technically this change is not reversible by changing the value of the config item because the IAMRole is deleted.

linki:
@arjunrn :man_facepalming: You're right. I fixed it.

linki:
:+1:

arjunrn:
👍

PR #3092, opened 2020-03-27
Test clusters: add rules for StackSets and CronJobs without application label

hjacobs:
`kube-janitor` was only deleting Deployments and StatefulSets without the `application` label after 7 days. Do the same for StackSets and CronJobs. I consider this a bug fix and not a new feature.
Note that the 7 days are counted from the last deployment time (CDP `deployment-time` annotation), i.e. some resources without the `application` might never get deleted as long as they are deployed every week.

### DIFF
--- a/cluster/manifests/kube-janitor/rules-config.yaml
+++ b/cluster/manifests/kube-janitor/rules-config.yaml
@@ -21,8 +21,8 @@ data:
         # remove StackSets without a pod label "application"
         resources:
           - stacksets
-        # see http://jmespath.org/specification.html
-        jmespath: "!(spec.stackTemplate.spec.podTemplate.metadata.labels.application) && metadata.creationTimestamp > '2019-04-15'"
+        # StackSet propagates top-level labels to pods, so check podTemplate and StackSet itself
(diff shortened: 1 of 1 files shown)


hjacobs:
I will provide more details on affected resources internally.

hjacobs:
This only affects 78 StackSets in test clusters and 25 of them are "user-based" PR deployments (i.e. one StackSet per developer).

hjacobs:
:+1:

hjacobs:
Looks like StackSetController does some magic and propagates top-level labels to Pods: https://github.com/zalando-incubator/stackset-controller/blob/38e32198c2bbcd7cbe27e622b130ceb3826815e8/pkg/core/stackset.go#L83

Will change the config.

hjacobs:
UPDATE: no StackSets would be deleted by the new configuration :tada:

hjacobs:
:+1:

hjacobs:
Can we merge this? It does not actually delete anything right now, but it prevents new deployments without `application` label to appear.

mikkeloscar:
:+1:

PR #3094, opened 2020-03-27
Update CoreDNS to version 1.6.9

arjunrn:
For @szuecs 
Update the CoreDNS version to 1.6.9

This version has the fix for this [issue](https://github.com/coredns/coredns/pull/3640).
This change does not change the current behavior because the default value for concurrency is 0 which means there is no concurrency control. This can be changed by setting the config item `coredns_max_concurrency`.


aryszka:
:+1:

mikkeloscar:
:+1:

PR #3095, opened 2020-03-28
kube-downscaler v20.3.1: downscale StatefulSets, Stacks with HPA

hjacobs:
New upstream version v20.3.1: https://github.com/hjacobs/kube-downscaler/releases/tag/20.3.1

* scale down Stacks with HPA (new feature in v20.3.1: https://github.com/hjacobs/kube-downscaler/pull/93)
* also downscale StatefulSets
* suspend CronJobs (https://github.com/hjacobs/kube-downscaler/issues/78)
* configure CDP's `deployment-time` annotation (https://github.com/hjacobs/kube-downscaler/issues/87)

This reduces the number of worker nodes in some test clusters during off-hours dramatically (e.g. from 13 to 6).

### DIFF
--- a/cluster/manifests/kube-downscaler/deployment.yaml
+++ b/cluster/manifests/kube-downscaler/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: kube-downscaler
-    version: v0.18
+    version: v20.3.1
 spec:
   replicas: 1
   selector:
@@ -15,7 +15,7 @@ spec:
     metadata:
       labels:
         application: kube-downscaler
-        version: v0.18
+        version: v20.3.1
     spec:
       dnsConfig:
         options:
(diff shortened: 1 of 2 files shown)


hjacobs:
This is pending an internal merge for kube-downscaler v20.3.1.

cluster/manifests/kube-downscaler/deployment.yaml
```diff
@@ -26,13 +26,14 @@ spec:
       containers:
       - name: downscaler
         # see https://github.com/hjacobs/kube-downscaler/releases
-        image: registry.opensource.zalan.do/teapot/kube-downscaler:0.18
+        image: registry.opensource.zalan.do/teapot/kube-downscaler:20.3.1
         args:
           - --interval=30
           - --exclude-namespaces=kube-system,visibility
-          - --exclude-deployments=kube-downscaler,downscaler,postgres-operator
+          - --exclude-deployments=kube-downscaler
```
mikkeloscar: What if a postgres statefulset is opt-out to scaledown, shouldn't the `postgres-operator` still run in that case?

hjacobs: Yes, but I would rather annotate the Postgres Operator deployment itself than adding all deployments as exclusions here. It already has the annotation:

Screenshot_2020-04-01 postgres-operator (Deployment in default) - Kubernetes Web View

hjacobs: Or scratch that, it was only in one cluster. I will add the exclusion back for now.

mikkeloscar:
:+1:

hjacobs:
:+1:

PR #3096, opened 2020-03-30
Allow tcp_keepalive sysctls

aermakov-zalando:
These are fairly benign so there's no reason to not allow customising them. We also want to inject better defaults via the admission controller, but we need to update kubelets/PSPs first.

### DIFF
--- a/cluster/manifests/psp/pod_security_policy.yaml
+++ b/cluster/manifests/psp/pod_security_policy.yaml
@@ -19,6 +19,10 @@ spec:
     rule: RunAsAny
   supplementalGroups:
     rule: RunAsAny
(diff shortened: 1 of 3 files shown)


aryszka:
:+1:

aermakov-zalando:
This works in a pet cluster:
```
root@test-76f49cbff4-nrktg:/# sysctl -a 2>/dev/null | grep keepalive
net.ipv4.tcp_keepalive_intvl = 1
net.ipv4.tcp_keepalive_probes = 10
net.ipv4.tcp_keepalive_time = 5
```

aermakov-zalando:
👍

gargravarr:
:+1:

PR #3103, opened 2020-03-30
Remove ScheduleDaemonSetPods experimental flag

mikkeloscar:
`TaintNodesByCondition` and `ScheduleDaemonSetPods` is true in `v1.17` and can't be disabled so remove the feature flags.

We have an internal ticket for making this safe such that we won't accidentally change requests of daemonsets which would evict everything else on a node. We agreed that it's ok to roll as far as beta channel without the safety in place, but for stable channel we also need the safety which will be handled in the admission-controller.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3106, opened 2020-03-31
upgrade to ingress controller version v0.10.6

aryszka:
Upgrading ingress-controller to v0.10.6.


aryszka:
the feature is not ready yet

PR #3110, opened 2020-04-01
external-dns v0.7.1

njuettner:
* Prometheus metric to see when external-dns had the last successful sync with the DNS provider


njuettner:
👍

jrake-revelant:
👍

PR #3111, opened 2020-04-02
Give kube-cost-allocator watch access for pods/nodes

mikkeloscar:
As part of the Cloud Cost Efficiency initiative for doing better per application reporting we have built a small tool that watches for pods/nodes in all clusters and records their start and end time in order to later calculate their usage hours and thus cost.

For this we need read permissions in all clusters.


mikkeloscar:
:+1:

hjacobs:
:+1:

PR #3120, opened 2020-04-06
New "delete_vpc_resources" config item to delete vpc dependant resources

muaazsaleem:
While re-configuring a VPC we must delete all VPC dependent resources,
while keeping others. Introducing a config item to automate that.

Signed-off-by: Muaaz Saleem <muhammad.muaaz.saleem@zalando.de>

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -1,7 +1,7 @@
 AWSTemplateFormatVersion: 2010-09-09
 Description: Kubernetes cluster
 Resources:
-{{if ne .Cluster.ConfigItems.delete_vpc_resources "true"}}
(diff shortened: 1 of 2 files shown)


cluster/cluster.yaml
```diff
@@ -1,6 +1,7 @@
 AWSTemplateFormatVersion: 2010-09-09
 Description: Kubernetes cluster
 Resources:
+{{if .Cluster.ConfigItems.delete_vpc_resources "false"}}
```
aermakov-zalando: `if ne`

muaazsaleem:
👍

aermakov-zalando:
👍

PR #3131, opened 2020-04-06
Enable OIDC-based Service Account tokens for Cluster Autoscaler

linki:
Enables Service Account credentials for CA.

Split of https://github.com/zalando-incubator/kubernetes-on-aws/pull/2918

- [x] This requires CA 1.18.


linki:
Current status: CA still cannot get credentials via WebIdentity. This was added in [aws-sdk-go v1.23.13](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) which this image of the autoscaler **should** [be using](https://github.com/zalando-incubator/autoscaler/blob/zalando-cluster-autoscaler-1-12/cluster-autoscaler/Godeps/Godeps.json#L159).

Double-checked all the configs and mounted files and they are right. It either uses an older version of the aws-sdk-go or the code for setting up a new AWS session is somewhat incompatible.

@aermakov-zalando Do you see any obvious mistake I'm making here?

mikkeloscar:
:+1:

aermakov-zalando:
👍

mikkeloscar:
:+1:

linki:
:+1:

PR #3143, opened 2020-04-07
CoreDNS: don't emit max_concurrent if it's 0

aermakov-zalando:
Follow-up to #3094. We run CoreDNS as `OnDelete`, so we'll have some nodes still running the old image after the update. If they restart for any reason, they'll never come up again because of the `max_concurrent` directive. Since we're not using it anyway at the moment, let's make it conditional.


aermakov-zalando:
👍

arjunrn:
:+1:

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3149, opened 2020-04-09
Easter 2020: configure default downtime

hjacobs:
Scale down test clusters over Easter 2020: Friday 20:00 CEST to Tuesday 07:00 CEST.

Kubernetes Downscaler supports recurring times (like Mon-Fri) and absolute ranges: https://github.com/hjacobs/kube-downscaler/#configuration

When uptime and downtime overlap, the downtime wins (relevant code: https://github.com/hjacobs/kube-downscaler/blob/master/kube_downscaler/scaler.py#L187).

Note that Friday is a public holiday only in Germany and Finland, not Ireland.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -127,8 +127,11 @@ kube_state_metrics_mem_max: "1Gi"
 # Kubernetes Downscaler (for non-production clusters)
 {{if eq .Environment "test"}}
 downscaler_default_uptime: "Mon-Fri 07:30-20:30 Europe/Berlin"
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
Let's use Cluster Registry for things like this. It's a one-off change that doesn't apply to clusters created in the future, and it's unclear whether it'll even make it to our clusters in time.

hjacobs:
@aermakov-zalando the PR is still needed to add the configuration option, right?

hjacobs:
@aermakov-zalando changed the PR so the actual value needs to come from the Cluster Registry.

aermakov-zalando:
👍

arjunrn:
:+1:

PR #3156, opened 2020-04-14
Add CSI Migration as a config item

arjunrn:
`CSIMigration` was made true by default in _1.17_ and is causing issues with existing volumes.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -111,7 +111,7 @@ prometheus_mem_min: "2Gi"
 prometheus_cpu_min: "0"
 prometheus_remote_write: "disabled"
 prometheus_tsdb_retention_size: "disabled"
-prometheus_csi_ebs: "true"
+prometheus_csi_ebs: "false"
 
 metrics_service_cpu: "100m"
 metrics_service_mem: "200Mi"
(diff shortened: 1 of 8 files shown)


cluster/config-defaults.yaml
```diff
@@ -111,7 +111,7 @@ prometheus_mem_min: "2Gi"
 prometheus_cpu_min: "0"
 prometheus_remote_write: "disabled"
 prometheus_tsdb_retention_size: "disabled"
-prometheus_csi_ebs: "true"
+prometheus_csi_ebs: "false"
```
mikkeloscar, line 114: This seem to not be in use anymore. The places where it was used was changed to `enable_csi_migration`

arjunrn: It's still used to create the prometheus stateful set.

mikkeloscar: Ah got it! :+1:

mikkeloscar:
:+1:

jrake-revelant:
👍

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -43,10 +43,9 @@ write_files:
 {{- end }}
       featureGates:
         BoundServiceAccountTokenVolume: {{ .Cluster.ConfigItems.rotate_service_account_tokens }}
-        CSINodeInfo: true
```
jrake-revelant, line 46: why is this not needed anymore?

arjunrn: It cannot be set to false anymore. It's GA in 1.17.

PR #3157, opened 2020-04-14
Allow audittrail adapter to read mounted token

linki:
Permissions of the filesystem didn't grant access to the service account token.


linki:
:+1:

aermakov-zalando:
👍

PR #3160, opened 2020-04-15
Remove obsolete aws-iam-controller roles and OIDC components

linki:
Removes AWS IAM roles when not needed and some left-overs from the OIDC implementation.


arjunrn:
These cannot be removed because some users are using them when they had issues with `kube2iam`

linki:
:+1:

linki:
:+1:

mikkeloscar:
:+1:

PR #3163, opened 2020-04-16
Add CSI Migration as a config item

arjunrn:
Cherry pick of #3156 to `beta`


aermakov-zalando:
Let's merge it normally via #3164.

PR #3179, opened 2020-04-21
Disable downscaling for statefulsets

arjunrn:
Many statefulsets are managed by operators and the downscaler is constantly scaling them down while the operators are bringing them up.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3181, opened 2020-04-21
Return early for IPv6 DNS queries in east-west domain

linki:
Might solve issues using alpine.


delivery.yaml
```diff
@@ -52,6 +52,8 @@ pipeline:
           env:
           - name: E2E_NODE_OS
             value: ubuntu
+          - name: E2E_SKIP_CLUSTER_UPDATE
```
aermakov-zalando: Why?

linki: Just to ensure e2e is running against the new config.

mikkeloscar:
:+1:

linki:
:+1:

PR #3196, opened 2020-04-24
Allow customizing etcd_endpoints per node pool

aermakov-zalando:
etcd apparently doesn't allow having http and https endpoints in the same client. 🤷‍♂️ This allows us to setup a temporary master node pool and then get rid of it after the migration.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3198, opened 2020-04-27
Migrate stackset-controller from the internal repo

aermakov-zalando:
This is way easier to manage than what we have right now.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3199, opened 2020-04-28
Admission controller: enable DaemonSet admitter

aermakov-zalando:
A basic sanity check for reserved resources. Defaults will be updated after testing it.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -168,6 +168,10 @@ teapot_admission_controller_deployment_default_max_surge: "5%"
 teapot_admission_controller_deployment_default_max_unavailable: "1"
 teapot_admission_controller_inject_aws_waiter: "true"
 
+## TODO update defaults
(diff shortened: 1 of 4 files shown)


cluster/config-defaults.yaml
```diff
@@ -168,6 +168,12 @@ teapot_admission_controller_deployment_default_max_surge: "5%"
 teapot_admission_controller_deployment_default_max_unavailable: "1"
 teapot_admission_controller_inject_aws_waiter: "true"
 
+teapot_admission_controller_check_daemonset_resources: "true"
```
linki: Do you want to immediately turn it on?

aermakov-zalando: We'd see how it work in `dev`, but yeah, we could also be on the safe side.

linki:
:+1:

aermakov-zalando:
👍

PR #3202, opened 2020-04-28
Add config item to separate OIDC IAM changes in userdata

linki:
This allows to separately control the userdata part of the OIDC IAM rollout.

The current setup doesn't allow for a quick rollback in case enabling the feature leads to problems. The rollback might get stuck because new deployments won't get credentials from the old masters, especially for components that are involved in the rollback itself, such as CLC.

This adds a new temporary config item that allows to separately control the userdata changes which allows a quick manifest-only rollback in case it's needed.


linki:
:+1:

aryszka:
:+1:

PR #3206, opened 2020-04-29
Support etcd secrets encryption by default

linki:
This enables the encryption provider sidecar by default for all environments. It also prevents people from accidentally removing support for encryption while still having encrypted secrets in etcd.

This will be a noop for dev, alpha, beta but will trigger a rolling update for stable. Changing the default is preferred because it allows us to combine the rolling of masters with other userdata changes.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -1761,7 +1761,6 @@ Resources:
             Resource: "*"
             Action:
             - "kms:Decrypt"
-{{- if eq .Cluster.ConfigItems.support_encryption "true" }}
   EtcdEncryptionKey:
     Type: "AWS::KMS::Key"
     Properties:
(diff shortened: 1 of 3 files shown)


mikkeloscar:
Can we roll back after this? Say we discover a problem in production that we didn't see in test, it would be good with a quick (as quick as possible) rollback option for the initial time it's running in production.

linki:
The encryption works two-fold: This config item configures the API server to go through the sidecar proxy, similar to e.g. the etcd-proxy. The sidecar is configured to not encrypt and simply does a 1 to 1 mapping. Another config item will actually instruct it do encrypt and decrypt.

At this point I trust the proxy enough to work for the first described case: Similar to how the image policy webhook in dev mode works, it cannot easily be turned off by a config item after this PR. However, it's still possible to enable/disable encryption but it requires the sidecar in both cases.

linki:
:+1:

mikkeloscar:
:+1:

PR #3183, opened 2020-04-21
upgrade ingress controller

aryszka:
- global and ingress specific WAF associations for LBs
- permissions to make WAF associations

Signed-off-by: Arpad Ryszka <arpad.ryszka@gmail.com>

PR #3211, opened 2020-04-30
Allow deploying WAFv2 resources

mikkeloscar:
Allows deploying `wafv2` resources which we now support in the ingress-controller #3183


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3216, opened 2020-05-04
Reduce priority for visibility-zmon

mikkeloscar:
Reduces the `visibility-zmon` priority value in order to ensure that the `visibility-logging` priority is higher as it's used for daemonset pods. Before they had the same priority and we could thus risk that zmon components land on a node making it impossible for `logging-agent` to be scheduled.
Since we're not using Priorities lower than this I just dropped a `0` to keep it simple.

Note that the priority will need to be deleted and re-created (as it can't be modified in place). The impact is that pods created with the previous priority will still keep the higher priority, but pods created with the new with get the lower priority value.

### DIFF
--- a/cluster/manifests/01-visibility/priority.yaml
+++ b/cluster/manifests/01-visibility/priority.yaml
@@ -2,6 +2,6 @@ apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
   name: visibility-zmon
-value: 1000000000
+value: 100000000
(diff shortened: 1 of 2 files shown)


mikkeloscar:
e2e fails because we can't update a `PriorityClass`. I will drop them in the other clusters on rollout.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3227, opened 2020-05-05
Update cronjob fixer

aermakov-zalando:
Fixes the issue with cronjobs that have a schedule that never matches.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3229, opened 2020-05-05
prometheus: Remove VPA max

mikkeloscar:
Since we introduced support for fallback node pools we no longer need to limit the max size that can be suggested by the VPA. Instead we can let the VPA suggest as high has it needs and then just add fallback pools (with lower priority) which can be used if Prometheus needs more memory than the default pool provides.

This enables us to not be limited by the size of the default node pool which is currently used for the calculation in the `autoscalingBufferSettings` function.


arjunrn:
:+1:

aermakov-zalando:
👍

PR #3236, opened 2020-05-06
enable compression by default on the ingress level

aryszka:
Enable compression by default on the ingress level.

This filter only compresses the responses when the client explicitly indicated in the request that it's ok, with the Accept-Encoding header. It doesn't compress already compressed responses, either. This means that it should be safe from functional perspective. Regarding the performance perspective: from earlier experience, we can assume that the positive impact on the traffic for the shared service landscape will be way bigger than the additional CPU usage in Skipper due to the compression.


aryszka:
we are testing this first with selected clusters using the config registry. Will reopen on success.

PR #3237, opened 2020-05-07
Enable encryption of secrets in etcd by default

linki:
Let's turn this on so new clusters will have it from the start.


mikkeloscar:
:+1:

linki:
:+1:

PR #3239, opened 2020-05-07
Allow dropping labels from kube-state-metrics scrapes

mikkeloscar:
* This is a quick fix to allow excluding labels per cluster for the kube-state-metrics scrape which otherwise could prevent Prometheus >2.16.0 from scraping anything from `kube-state-metrics`: https://github.com/kubernetes/kube-state-metrics/issues/1115 The idea is that this will not happen often but if it does we can easily exclude some labels. We have monitoring in place to alert us if Prometheus fails to scrape targets.
* Drops scraping stackset-controller from `default` namespace as it has been moved to `kube-system`.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -113,6 +113,11 @@ prometheus_remote_write: "disabled"
 prometheus_tsdb_retention_size: "disabled"
 prometheus_csi_ebs: "false"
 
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3241, opened 2020-05-07
kube-janitor v20.4.1

hjacobs:
Kubernetes Janitor [v20.4.1](https://github.com/hjacobs/kube-janitor/releases/tag/20.4.1) fixes an issue with deleting PVCs referenced by Deployments in test clusters.
That being said, most of the cases involve Redis which should be deployed as a StatefulSet instead.


aermakov-zalando:
👍

hjacobs:
:+1:

PR #3242, opened 2020-05-07
Treat scraper as component of dashboard

mikkeloscar:
To lower the number of applications we have we can define the `metrics-scraper` as simply a component of the `kubernetes-dashboard`.

~This change will cause a conflict on apply because of the `matchLabels` change, but I will just fix it manually.~ Renamed the deployments so this doesn't happen.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3245, opened 2020-05-11
etcd-backup: add TLS support

aermakov-zalando:
Use the certificates & keys configured for the API server when doing the backups (since we run on the masters anyway).


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3251, opened 2020-05-12
autoscaling_utilization_threshold: default to 1.0

aermakov-zalando:
We've been running with 1.0 instead of 0.75 for a couple of weeks already with no downsides, let's change the default.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3217, opened 2020-05-04
Change default replacement strategy for all pools to none

linki:
The current replacement strategy prevents spinning up spot replacement nodes for on-demand nodes during cluster updates. That means that pods are rarely moved away from on-demand nodes.

Now that we're using spot in production this setting allows replacement nodes to be spot during update which should increase the share of spot nodes in production. We use this setting in test clusters for the same reason as described here: https://github.bus.zalan.do/teapot/issues/issues/2198

PR #3252, opened 2020-05-13
Revert "Change default replacement strategy for all pools to none"

mikkeloscar:
Reverts zalando-incubator/kubernetes-on-aws#3217

Revert to avoid rolling the nodes right now as we had to do an "emergency" rollout of new certificates for the majority of production clusters.


mikkeloscar:
:+1:

linki:
:+1:

PR #3228, opened 2020-05-05
Update EC2 AMI to the latest versions

linki:
Update EC2 AMI to the latest versions

PR #3253, opened 2020-05-13
Revert "Update EC2 AMI to the latest versions"

mikkeloscar:
Reverts zalando-incubator/kubernetes-on-aws#3228

Revert to avoid rolling the nodes right now as we had to do an "emergency" rollout of new certificates for the majority of production clusters.


mikkeloscar:
:+1:

linki:
:+1:

PR #3257, opened 2020-05-15
Reduce CA log verbosity

aermakov-zalando:
It logs an insane amount of data, so let's try to reduce it. We can always edit the deployment for investigation.


mikkeloscar:
:+1:

jrake-revelant:
👍

PR #3258, opened 2020-05-15
Test clusters: pull images in parallel

aermakov-zalando:
This was only disabled by default due to a bug in an ancient Docker version, so let's enable it to improve the node startup time.

Some [documentation](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubelet/config/v1beta1/types.go#L540).


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3259, opened 2020-05-18
Update to Prometheus 2.18.1

mikkeloscar:
Just to stay on latest version.
The fix is in the snapshot API which we don't use right now.

https://github.com/prometheus/prometheus/releases/tag/v2.18.1


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #3262, opened 2020-05-18
Enable OIDC-based IAM credentials by default

linki:
This has been enabled in all clusters. In order to e2e test it we need to make it the default here. Note that you can still disable it like before.


linki:
:+1:

mikkeloscar:
:+1:

PR #3271, opened 2020-05-20
Allow customizing root volume size

aermakov-zalando:
Some of our clusters have lots of small pods with big images, so the AMI default doesn't work.


aermakov-zalando:
👍

szuecs:
:+1:

PR #3273, opened 2020-05-20
Fix the master node pool config for e2e

aermakov-zalando:
Some of our tests flake because we provision just one very small master node. Change the configuration to something more similar to what we have in actual clusters.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3277, opened 2020-05-25
Stackset controller: add a liveness probe

aermakov-zalando:
It gets stuck sometimes for an unclear reason. Let's add a liveness probe until the underlying issue is fixed.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3280, opened 2020-05-25
Set maximum number of bytes streamed to the client

linki:
This limits the number of bytes streamed by the API server for streaming connections, e.g. port-foward and most-likely others but I didn't check thoroughly.

The default is 0 which doesn't throttle the connection at all. Setting the flag to something higher triggers an entirely different branch of the code [here](https://github.com/kubernetes/kubernetes/blob/v1.17.4/staging/src/k8s.io/apimachinery/pkg/util/proxy/upgradeaware.go#L350-L354) which means that setting the value to something unreasonably high is not necessarily the same as not setting it at all.

This should fix https://github.bus.zalan.do/teapot/issues/issues/2575. More details there.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -352,3 +352,6 @@ stackset_controller_sync_interval: "10s"
 
 # EBS settings for the root volume
 ebs_root_volume_size: "50"
+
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

jrake-revelant:
👍

PR #3285, opened 2020-05-26
Inject parent-resource-hash labels in pods

mikkeloscar:
Updates the admission-controller to a version which supports injecting `parent-resource-hash` labels into pods owned by `StatefulSet` or `Deployment`.

These labels will be used to create non-overlapping PDBs in the pdb-controller.


jrake-revelant:
👍

mikkeloscar:
:+1:

PR #3291, opened 2020-05-29
pdb-controller: support parent-resource-hash

mikkeloscar:
Follow up to #3285 add support in the pdb-controller so it can create non-overlapping PDBs.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3295, opened 2020-06-02
Fix condition for enabling aws-iam controller for etcd-backup

mikkeloscar:
We probably won't need to roll back in the long run but the logic was wrong so `etcd-backup` didn't work when `teapot_admission_controller_service_account_iam=false` and `kube_aws_iam_controller_kube_system_enable=true`


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3301, opened 2020-06-04
Drop support for etcd_use_proxy

aermakov-zalando:
We want to get rid of it anyway, and it just complicates the configuration. Let's update the clusters instead.

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -111,13 +111,7 @@ write_files:
           - --apiserver-count={{ .Values.apiserver_count }}
           - --bind-address=0.0.0.0
           - --insecure-bind-address=0.0.0.0
-{{- if eq .Cluster.ConfigItems.etcd_use_proxy "true" }}
-          - --etcd-servers=http://127.0.0.1:2379
-{{- else if index .NodePool.ConfigItems.etcd_endpoints }}
           - --etcd-servers={{ .NodePool.ConfigItems.etcd_endpoints }}
-{{- else }}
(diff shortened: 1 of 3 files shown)


aermakov-zalando:
Making the tests pass is unfortunately going to be pretty annoying, so let's just merge this through when we don't have anything pending.

szuecs:
:+1:

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -111,9 +111,7 @@ write_files:
           - --apiserver-count={{ .Values.apiserver_count }}
           - --bind-address=0.0.0.0
           - --insecure-bind-address=0.0.0.0
-{{- if eq .Cluster.ConfigItems.etcd_use_proxy "true" }}
-          - --etcd-servers=http://127.0.0.1:2379
-{{- else if index .NodePool.ConfigItems.etcd_endpoints }}
+{{- if index .NodePool.ConfigItems.etcd_endpoints }}
```
linki: I think that's redundant since config items are propagated "down" to the node pools. So if `.Cluster.ConfigItems.etcd_endpoints` is set so will be `.NodePool.ConfigItems.etcd_endpoints`. that would mean we can just use `.NodePool.ConfigItems.etcd_endpoints` here without the if.

Here's an example of a global config item with default overwritable in node pool: https://github.com/zalando-incubator/kubernetes-on-aws/search?q=spot_allocation_strategy&unscoped_q=spot_allocation_strategy

aermakov-zalando: Good point, fixed.

linki:
:+1:

aermakov-zalando:
👍

gargravarr:
:+1:

PR #3308, opened 2020-06-05
Add support for spot.io Ocean pools

mikkeloscar:
Related change in CLM: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/316

This introduces:
* Deployment for the `spotinst/spot.io controller`. Looks for pending pods in the cluster and scales up those that should run on spot.io managed instances
* node pool profile: `worker-spotio-ocean`. Node pool which maps to an ocean. This means we can have `n` oceans per Kubernetes cluster. We deploy one spot.io controller per spot.io node pool. ~This is like a main spot.io pool that defines all the instance types to be available. THERE CAN ONLY BE ON OF THESE POOLS PER CLUSTER (or you need to run multiple controllers as well) Pods can be scheduled here with:~

* ~node pool profile: `worker-spotio`. This can act as secondary node pool with additional labels/taints. E.g. it can be used to isolate workloads on a group of instances. NOTE: You can't specify instance types at this level, the instance types of the `worker-spotio-ocean` pool will be used.~

~I don't know if modelling the node pools in this way really make sense, but was the simplest way to do it in CLM as a first PoC.~

This have a few quirks around the node pool handling which we need to address after the PoC:
1. CLM can't safely cleanup spot.io node pools 
2. If a spot.io node pool is removed the corresponding spotinst-controller deployment is not dropped and has to be done manually.

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -211,7 +211,4 @@ spec:
       - effect: NoSchedule
         key: dedicated
         value: skipper-ingress
-{{ else }}
-      nodeSelector:
-        node.kubernetes.io/role: worker
 {{ end }}
--- a/cluster/manifests/spotio-controller/deployment.yaml
+++ b/cluster/manifests/spotio-controller/deployment.yaml
@@ -1,0 +1,86 @@
+{{ if and (index .Cluster.ConfigItems "spotio_account_id") (index .Cluster.ConfigItems "spotio_access_token") }}
+{{ with $data := . }}
+{{ range $nodePool := .Cluster.NodePools }}
+{{ if eq $nodePool.Profile "worker-spotio-ocean" }}
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: spotinst-kubernetes-cluster-controller-{{ $nodePool.Name }}
+  namespace: kube-system
+  labels:
+    application: spotinst-kubernetes-cluster-controller
+    component: "{{ $nodePool.Name }}"
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      application: spotinst-kubernetes-cluster-controller # TODO: this label name is REQUIRED by the controller
+      component: "{{ $nodePool.Name }}"
+  template:
+    metadata:
+      labels:
+        application: spotinst-kubernetes-cluster-controller
+        component: "{{ $nodePool.Name }}"
+      annotations:
+        config/hash: {{"secret.yaml" | manifestHash}}
+    spec:
+      # TODO: run on master?
+      # nodeSelector:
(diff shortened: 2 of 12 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -211,7 +211,4 @@ spec:
       - effect: NoSchedule
         key: dedicated
         value: skipper-ingress
-{{ else }}
-      nodeSelector:
-        node.kubernetes.io/role: worker
```
mikkeloscar, line 216: This was dropped because it's not really needed in the general case and I just discovered we had it because it caused issues for the spot.io integration.

cluster/manifests/spotio-controller/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+        component: "{{ $nodePool.Name }}"
+      annotations:
+        config/hash: {{"secret.yaml" | manifestHash}}
+    spec:
+      # TODO: run on master?
+      # nodeSelector:
+      #   node.kubernetes.io/role: master
+      containers:
+      - name: controller
+        image: spotinst/kubernetes-cluster-controller:1.0.61 # TODO: setup pipeline
```
mikkeloscar, line 33: As long as we only enable spot.io node pools in test clusters this will work. I will follow up later with a proper image, but not required for the PoC.

cluster/manifests/spotio-controller/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+            cpu: "{{ $data.ConfigItems.spotio_ocean_controller_cpu }}"
+            memory: "{{ $data.ConfigItems.spotio_ocean_controller_memory }}"
+          requests:
+            cpu: "{{ $data.ConfigItems.spotio_ocean_controller_cpu }}"
+            memory: "{{ $data.ConfigItems.spotio_ocean_controller_memory }}"
+      serviceAccountName: spotinst-kubernetes-cluster-controller
+      tolerations:
+      - key: node.kubernetes.io/role
+        value: master
+        effect: NoSchedule
```
linki, line 82: Do you want to run it on masters or does it not matter at the moment? (You commented out the nodeSelector above)

linki:
:+1:

cluster/manifests/spotio-controller/deployment.yaml
mikkeloscar: I'm not sure what is best, for the test cluster tests it has to run on the master as I intend to remove the normal node pools so the two autoscalers don't fight. For the future it can run on workers as long as we have custom taints on the node pools. I will leave it like this for now and then we have to decide after the PoC.

mikkeloscar:
:+1:

PR #3323, opened 2020-06-11
Add possibility to configure NLB for Kubernetes API server

linki:
Prepares for replacing ELB with NLB. Doesn't create NLB yet.

Config Items:
```markdown
# defines the rollout status of the NLB for the API server. The options are:
#
#   disabled:    no NLB will be provisioned
#   provisioned: NLB will be provisioned but not hooked up to the ASG
#   hooked:      NLB will be provisioned and hooked up to the ASG but DNS still points to the ELB
#   active:      NLB will be fully functional and DNS points to the NLB
#   promoted:    NLB will be fully functional and ELB will be unhooked
#   exclusive:   NLB will be fully functional and ELB will be removed
#
apiserver_nlb: "disabled"
```

It's safe to go through the stages one-by-one in both directions.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -48,7 +48,7 @@ Resources:
           Value: owned
       ToPort: 10250
     Type: 'AWS::EC2::SecurityGroupIngress'
-{{- if or (eq .Cluster.ConfigItems.apiserver_active_load_balancer "nlb") (eq .Cluster.ConfigItems.apiserver_provision_nlb "true") }}
+{{- if ne .Cluster.ConfigItems.apiserver_nlb "disabled" }}
   MasterLoadBalancerNLB:
     Type: AWS::ElasticLoadBalancingV2::LoadBalancer
     Properties:
@@ -106,7 +106,7 @@ Resources:
       Port: 443
       Protocol: TLS
 {{- end }}
-{{- if or (eq .Cluster.ConfigItems.apiserver_active_load_balancer "elb") (eq .Cluster.ConfigItems.apiserver_provision_elb "true") }}
+{{- if ne .Cluster.ConfigItems.apiserver_nlb "exclusive" }}
   MasterLoadBalancer:
     Properties:
       ConnectionDrainingPolicy:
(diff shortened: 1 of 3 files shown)


cluster/cluster.yaml
```diff
@@ -48,6 +48,81 @@ Resources:
...
+    Properties:
+      Certificates:
+        - CertificateArn: "{{.Values.load_balancer_certificate}}"
+      DefaultActions:
+      - Type: forward
+        TargetGroupArn: !Ref MasterLoadBalancerNLBTargetGroup
+      LoadBalancerArn: !Ref MasterLoadBalancerNLB
+      Port: 443
+      Protocol: TLS
+  MasterLoadBalancerNLBDNSName:
```
aermakov-zalando: Is it for testing? I'd just switch the main record set for simplicity.

linki: Mostly just to keep it similar to the ELB case. Makes it easier to remove all the ELB related stuff later on I think.

linki: Fixed

aermakov-zalando:
Instead of modelling this as 5 separate booleans, most combinations of which don't make sense, you might want to consider using an enum-like value for this: `apiserver_nlb: disabled|provisioned|enabled|exclusive`. I'm not sure if `text/template` already has helper functions for this (e.g. return whether a string is in a static array), but even if it doesn't it's very easy to write something, or just slap a bunch of `or`s.

cluster/cluster.yaml
```diff
@@ -104,15 +163,27 @@ Resources:
           Value: owned
       VpcId: "{{.Cluster.ConfigItems.vpc_id}}"
     Type: 'AWS::EC2::SecurityGroup'
-  MasterLoadBalancerVersionDomain:
+{{- end }}
+  MasterLoadBalancerDNSRecord:
```
aermakov-zalando: What actually happens when you rename a CF resource? Does it leave the physical resource untouched, or is there a moment where it won't exist?

linki: Not sure, I'll try it out.

linki: In this case it doesn't work. It tries to create the new resource first (including the DNS name) and that fails because the DNS name already exists. I changed it back to use the original name.

muaazsaleem:
👍

linki:
Let me merge this.

aermakov-zalando:
`hooked` seems to only be present in the comments, is it a leftover?

aermakov-zalando:
I'm also not sure why we need `promoted`. Why not just attach all LBs to the ASG as long as they're present?

linki:
Nice catch but `hooked` is indeed used. It doesn't show up because I'm using both `eq` and `ne`.

The two states `promoted` and `provisioned` are needed to correctly terminate the respective load balancer (either when finally removing the ELB or in case of rollback when removing the NLB). In both cases the LB/Target Group needs to be detached from the ASG separately, otherwise CLM will be stuck (because load balancer and master node pool are different CF stacks).

I expect all that configuration to get removed once we fully commit to using NLBs. Therefore, I don't think it has to be perfect during that time. The benefit of the current approach is that's a very safe rollout and rollback by just going through the stages one-by-one in either direction from any point.

aermakov-zalando:
👍

PR #3332, opened 2020-06-12
Reduce activity of renovate to weekly

linki:
So that we don't get spammed too much while we're tweaking it.


linki:
Skipped e2e since it's not necessary.

linki:
:+1:

mikkeloscar:
:+1:

PR #3336, opened 2020-06-17
Revert "Set maximum number of bytes streamed to the client"

linki:
Since the original PR hasn't made it do `alpha` yet and NLB support is already in `dev` there's no need to have this quick fix in place.

It might still be a good idea but for now I can't estimate all the implications and it's safer to just leave it unchanged.


aermakov-zalando:
👍

linki:
:+1:

PR #3337, opened 2020-06-17
Get rid of priorities as much as possible

aermakov-zalando:
* Drop priority classes for things running on master nodes (they should be sized correctly from the beginning)
 * Drop `system-cluster-critical` from most of our components unconditionally
 * Keep `system-cluster-critical` if `system_pods_critical` is set to `true` (default in production clusters) for the following components:
   * emergency access service
   * external-dns
   * kube-ingress-aws-controller
   * skipper + skipper redis
   * stackset-controller
   * VPA admission controller

We'll then slowly disable priorities altogether if it doesn't cause any issues.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -357,3 +357,10 @@ stackset_controller_sync_interval: "10s"
 
 # EBS settings for the root volume
 ebs_root_volume_size: "50"
+
+# Migration off priority classes
+{{if eq .Environment "production"}}
+system_pods_critical: "true"
+{{else}}
+system_pods_critical: "false"
+{{end}}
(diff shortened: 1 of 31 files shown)


jrake-revelant:
👍

aermakov-zalando:
👍

szuecs:
I think external-dns, kube-ingress-aws-controller are also not that critical.

aermakov-zalando:
@szuecs They might be critical if someone is doing an urgent configuration fix or a deployment during an incident. They're also pretty small so it's unlikely they'll affect things. Let's see how it behaves with other deployments, we want to drop the priorities in general anyway.

PR #3343, opened 2020-06-18
Enable NonPreemptingPriority feature gate (scheduler/apiserver)

aermakov-zalando:
This will allow us to still keep the custom priority classes, but disable preemption.


aermakov-zalando:
I've checked the source code and it's only used in a couple of places so I don't think it's worth making it configurable.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3348, opened 2020-06-23
Test both NLB and ELB code paths

linki:
Test both NLB and ELB setups for the API server. That would allow us to roll it out more easily.


linki:
:+1:

szuecs:
:+1:

PR #3352, opened 2020-06-24
Test NLB setup alongside default setup

linki:
Makes sure to test the current state as well as enables testing for active NLB. Split in separate PRs to ensure that clusters correctly go through the upgrade path.


linki:
:+1:

mikkeloscar:
:+1:

PR #3353, opened 2020-06-24
Prevent kube-node-ready from running on spot.io nodes

mikkeloscar:
`kube-node-ready` can't run on spot.io nodes because it triggeres ASG events which are not relevant for spot.io as they are managed EC2 instances without an ASG.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3357, opened 2020-06-24
Add lifecycle-status=ready label to spot.io nodes

mikkeloscar:
Adds `lifecycle-status=ready` label to the spot.io configuration to indicate to spot.io autoscaler that new nodes will match pods with this label selector.


mikkeloscar:
:+1:

aryszka:
:+1:

PR #3358, opened 2020-06-24
Drop kube-dns-metrics

mikkeloscar:
We no longer rely on the metrics provided by this so let's drop it.

We have metrics from CoreDNS and dnsmasq which covers what we need.


mikkeloscar:
:+1:

szuecs:
@mikkeloscar do we need a change in e2e test and remove from zmon?

mikkeloscar:
@szuecs Not e2e, I think the failure we see is not related. For zmon and others I have created an internal ticket to track the cleanup.

szuecs:
:+1:

PR #3359, opened 2020-06-24
Update event-logger

aermakov-zalando:
Fixes too frequent snapshotting + event filtering in case of full resync.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3364, opened 2020-06-26
Use NLB by default for new clusters

mikkeloscar:
Enable NLB by default for new clusters and control the rollout via cluster-registry config-items per cluster.

This way we avoid running multiple Load balancers for a longer time in case the rollout gets stuck in some channel.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -266,7 +266,7 @@ apiserver_proxy: "true"
 #   promoted:    NLB will be fully functional and ELB will be unhooked
 #   exclusive:   NLB will be fully functional and ELB will be removed
 #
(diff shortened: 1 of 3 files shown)


aermakov-zalando:
Please update the second e2e in `delivery.yaml` as well, the test configurations are currently setup to test "default" + "active" instead of being explicit. This will result in just testing two of the same configuration.

mikkeloscar:
Updated: Now it will be `hooked` and `exclusive` for the two e2e as I understand. Is that what you meant?

aermakov-zalando:
@mikkeloscar Yep, although we could also test `disabled`/`exclusive`.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3366, opened 2020-06-26
Update Prometheus to v2.19.2

mikkeloscar:
https://github.com/prometheus/prometheus/releases/tag/v2.19.0

Interesting change for us:

```
[FEATURE] TSDB: Memory-map full chunks of Head (in-memory) block from disk. This reduces memory footprint and makes restarts faster. #6679
```


szuecs:
:+1:

mikkeloscar:
:+1:

PR #3375, opened 2020-07-01
Disable kube-downscaler in production clusters by default

aermakov-zalando:
There's no clear ownership of it so we've decided to disable it in production clusters. Some clusters will have it grandfathered in, though, so adding a config item for it.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3383, opened 2020-07-06
Make stackset-controller use networking API

mikkeloscar:
Updates the stackset-controller to a version that uses `networking.k8s.io/v1beta1` instead of `extensions/v1beta1` for interaction with the Ingress API.


mikkeloscar:
Broken because the docker image is not marked as trusted..

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3384, opened 2020-07-06
spotio: Configure on-demand ratio per pool

mikkeloscar:
Allow configuring on-demand ratio etc. per node pool instead of globally for the cluster.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #3386, opened 2020-07-06
spotinst controller image from Pierone

mikkeloscar:
Use spotinst controller image from Pierone so the controller can run in production clusters.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #3402, opened 2020-07-10
Autoscaling buffers: just use the static sizes

aermakov-zalando:
The automagical sizing function needs too many tweaks in our current setup, and we don't even use the buffer pods a lot. Let's just replace the sizes with static ones.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3407, opened 2020-07-15
Improve route group validation

aermakov-zalando:
* Add a regex for the hostname (copied from Ingress)
 * Make the hosts required (to avoid incidents breaking everything in the cluster)
 * Fix the array validation: `minLength` doesn't work for them, we need to use `minItems`

### DIFF
--- a/cluster/manifests/skipper/crd.yaml
+++ b/cluster/manifests/skipper/crd.yaml
@@ -35,15 +35,18 @@ spec:
           required:
           - backends
           - routes
+          - hosts
           type: object
           properties:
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
@szuecs @aryszka FYI

mikkeloscar:
:+1:

mikkeloscar:
@aermakov-zalando we should also update the CRD upstream similar to here: https://github.com/zalando/skipper/pull/1475

aermakov-zalando:
👍

PR #3409, opened 2020-07-16
HPA: make downscale stabilization configurable

aermakov-zalando:
This could be useful in a couple of clusters, and until 1.18 we can't customize it per-HPA.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3411, opened 2020-07-16
pre compute expensive queries for Prometheus with rules

szuecs:
pre compute expensive queries for Prometheus with rules to solve issue with zmon check 13278


szuecs:
:+1:

cluster/manifests/prometheus/configmap.yaml
```diff
@@ -6,12 +6,61 @@ metadata:
...
+        expr: sum(rate(skipper_serve_host_duration_seconds_sum{}[1m])) / sum(rate(skipper_serve_host_duration_seconds_count{}[1m]))
+
   prometheus.yml: |-
     global:
       scrape_interval: 15s
       external_labels:
         prometheus_replica: @@POD_NAME@@
         cluster: {{ .Cluster.Alias }}
+    rule_files:
+    - "prometheus.rules.yaml"
```
aermakov-zalando: This can be an absolute path apparently, so why not just have `/etc/prometheus/prometheus.rules.yaml` here and avoid the whole copy thing altogether?

szuecs: I jsut though there was a reason for doing it, so I did the same.
I am fine with changing it

aermakov-zalando: I think the only reason was replacing the pod name internally (don't remember why).

szuecs: changed as suggested

szuecs:
:+1:

aermakov-zalando:
👍

PR #3434, opened 2020-07-23
Drop nginx ReadOnly proxy from apiserver

mikkeloscar:
We no longer have any zmon checks depending on the nginx ReadOnly proxy. Now all access to the api server is handled via RBAC roles for ZMON.

Therefore we can get rid of the nginx container and the ports that were opened for it!


mikkeloscar:
:+1:

linki:
:+1:

PR #3439, opened 2020-07-24
Make it possible to disable stackset validation

mikkeloscar:
This is causing more problems than expected for some users, so make it possible to disable per cluster.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #3473, opened 2020-08-04
Fix permission and roles of aws-node-decommissioner

linki:
* Adds list nodes permission to aws-node-decommissioner
* Turns RoleBinding into ClusterRoleBinding in order to work for non-namespaced resource (nodes)


linki:
:+1:

gargravarr:
:+1:

PR #3474, opened 2020-08-04
Default stackset_validation_enabled to false

aermakov-zalando:
It cannot work in its current form, because we don't define a schema for the status field. Disable by default until we have that.


gargravarr:
:+1:

aermakov-zalando:
👍

PR #3485, opened 2020-08-10
Add a temporary config item for pinning Skipper to an old version

aermakov-zalando:
This allows us to at least unblock cluster updates until the corresponding issue is resolved.


aryszka:
:+1:

aermakov-zalando:
Turned out to not be necessary.

PR #3507, opened 2020-08-18
Update Skipper openid configuration

aermakov-zalando:
We're still using an old URL for PF IAM in production, update to use the new one.


mikkeloscar:
:+1:

aryszka:
:+1:

PR #3511, opened 2020-08-19
Add StackSet e2e (Split e2e into multiple CDP steps)

mikkeloscar:
Split e2e into multiple CDP steps

* create cluster
* Run e2e
* Run StackSet e2e tests
* Decommission cluster

The benefit of this is:
* Faster to retrigger just e2e if needed.
* Easier to force decommission if desired
* Easier to add other e2e tests in parallel e.g. stackset e2e tests.
* Measure time taking for each of the steps.

The Stackset e2e tests are just those from [stackset-controller repository](https://github.com/zalando-incubator/stackset-controller/tree/master/cmd/e2e) imported and run as a separate step.


mikkeloscar:
:+1:

cluster/config-defaults.yaml
```diff
@@ -386,4 +386,4 @@ spotio_ocean_controller_memory: "512Mi"
 kubernetes_event_logger_enabled: "true"
 
 # enable/disable stackset validation (preserveUnknownFields)
-stackset_validation_enabled: "false"
+stackset_validation_enabled: "true"
```
mikkeloscar, line 389: Note, this is disabled in cluster-registry for all clusters, but enabled and tested in e2e.

test/e2e/tools.go
```diff
@@ -4,4 +4,5 @@ package e2e
 
 import (
 	_ "github.com/jteeuwen/go-bindata/go-bindata"
+	_ "github.com/zalando-incubator/stackset-controller/cmd/e2e"
```
muaazsaleem, line 7: I'm curious why this is in a separate file. Is it because these aren't imported anywhere but the `run_e2e.sh` script?

muaazsaleem:
👍

test/e2e/tools.go
mikkeloscar: Just by convention. It's a "hack" to have the package managed with Go modules but we only use it to generate a separate binary in the `Makefile`.

PR #3514, opened 2020-08-19
Add FeatureGate EphemeralContainers (disabled by default)

mikkeloscar:
Makes it possible to enable `EphemeralContainers` alpha feature gate. This is disabled by default because it's an alpha feature AND it allows adding containers WITHOUT resource requirements in our setup, which is probably not desired.

It can be used for supporting `kubectl alpha debug` (attaching debug container to running pod) which could be an interesting feature to provide in the future.

https://kubernetes.io/docs/tasks/debug-application-cluster/debug-running-pod/

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -337,6 +337,10 @@ enable_hpa_scale_to_zero: "true"
 # Enabled by default since v1.18
 enable_even_pods_spread: "false"
 
+# Enable FeatureGate EphemeralContainers (Alpha)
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

aermakov-zalando:
Does this go through the image policy webhook?

mikkeloscar:
> Does this go through the image policy webhook?

I'm not sure. I'll create an internal ticket to track this whole topic, for now this PR is just adding the feature so we can test it for ourselves in a simpler way, not intended to be enabled right now.

mikkeloscar:
:+1:

muaazsaleem:
👍

PR #3530, opened 2020-08-26
Default PSP: allow NFS volumes

aermakov-zalando:
This was requested by some of the users, and there doesn't seem to be any security impact.

PR #3537, opened 2020-08-27
Default PSP: allow NFS volumes

aermakov-zalando:
Cherry-picking #3530 to stable so we could actually roll it out.


aermakov-zalando:
👍

aryszka:
:+1:

PR #3539, opened 2020-08-31
Add a VPA for ingress-template-controller

aermakov-zalando:
It's crashing in some of the clusters, and we apparently have no VPA defined.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3540, opened 2020-08-31
Use m5a.large for e2e master nodes

mikkeloscar:
We have observed negative impact when using `t3.large` instances. Switch to `m5a.large` for master nodes when running e2e.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3547, opened 2020-09-02
Decommission ingress-template-controller

mikkeloscar:
The `IngressTemplate` feature is no longer supported and no longer in use in any cluster. Thus we can remove all traces of it.

Note: We don't need the deletions because it has already been deleted from all clusters.


mikkeloscar:
:+1:

aermakov-zalando:
👍

gargravarr:
:+1:

PR #3557, opened 2020-09-07
Add InfrastructureComponent=true tags to CF stacks

mikkeloscar:
Depends on https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/357 to take effect.

This adds `InfrastructureComponent=true` tags on the CF stacks such that we can condition those resources in IAM role e.g. to not let normal users touch the stacks with such a tag.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3516, opened 2020-08-20
Update APIServer skipper

mikkeloscar:
Main motivation for updating is to run a version NOT build with Go 1.14 to avoid this issue: https://github.com/golang/go/issues/37436

Similar to #3512 and #3513

PR #3564, opened 2020-09-09
Update API Server skipper proxy

mikkeloscar:
In #3516 we updated to the latest working skipper version at the time. Later versions had a bug: https://github.com/zalando/skipper/issues/1508 which has now been fixed so we can upgrade to the latest version.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3569, opened 2020-09-11
Disable profiling on Kubernetes components by default

mikkeloscar:
Disable profiling by default and make it configurable via config item.

This will disable the `host:port/debug/pprof/` endpoint on `kube-apiserver`, `kube-controller-manager`, `kube-scheduler` ~and disable all debug endpoints on `kubelet`.~ Not disabled on `kubelet` because the [debug endpoint](https://github.com/kubernetes/kubernetes/blob/v1.18.8/staging/src/k8s.io/kubelet/config/v1beta1/types.go#L249-L256) is also serving log and exec access as well as profiling.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -388,3 +388,6 @@ kubernetes_event_logger_enabled: "true"
 
 # enable/disable stackset validation (preserveUnknownFields)
 stackset_validation_enabled: "true"
+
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
👍

mikkeloscar:
:+1:

mikkeloscar:
:+1:

jrake-revelant:
👍

PR #3575, opened 2020-09-14
Allow preserving the root volume after termination

aermakov-zalando:
This way we can at least look at the logs to see what went wrong. Note: spot.io pools won't support this for now, this will be added in a separate PR.


jrake-revelant:
👍

mikkeloscar:
:+1:

PR #3581, opened 2020-09-15
Make node updates faster

aermakov-zalando:
Having minimum sizes set for multiple node pools means that cluster rotations take a lot of time (almost 2 hours recently). We also don't use this in any of the clusters, so this PR changes the setup a bit.

 * Drop `min_size` to 0 for all node pools.
 * Change back to 1 master node since we use an instance type that should handle this now.
 * Create a marker deployment for every node pool we want to check and explicitly test that these are healthy after an update. This way we can be sure that a pool configuration works, but we can use the usual "drain and let the autoscaler deal with it" approach for the actual upgrade.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3587, opened 2020-09-17
Make it possible to configure AMI per Node Pool

mikkeloscar:
Allows specifying the AMI_ID per node pool if desired. Useful to test a new AMI just for a single node pool.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3604, opened 2020-09-28
[No rolling updates] Fold cadvisor/node-exporter into one daemonset

aermakov-zalando:
Changes:
 * Use a single DS for both cadvisor and node-exporter
 * Drop a weird security group rule
 * Restrict the access to the ports to worker SG only (instead of the whole VPC)

There would be a short (~5 minute) outage of the relevant metrics which could probably be avoided by splitting it into multiple PRs. Not sure if it's worth it, though.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3605, opened 2020-09-28
Add the journald log reader

aermakov-zalando:
Builds on #3604.

Adds a container to the `node-monitor` that tails the node's journal. Currently not enabled by default, we'll test it in a couple clusters first.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -119,6 +119,15 @@ cadvisor_memory: "150Mi"
 node_exporter_cpu: "20m"
 node_exporter_memory: "75Mi"
 
+# journald reader settings
+{{if eq .Environment "e2e"}}
+journald_reader_enabled: "true"
+{{else}}
+journald_reader_enabled: "false"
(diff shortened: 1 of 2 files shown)


jrake-revelant:
👍

aermakov-zalando:
👍

cluster/config-defaults.yaml
```diff
@@ -119,6 +119,15 @@ cadvisor_memory: "150Mi"
...
 node_exporter_memory: "75Mi"
 
+# journald reader settings
+{{if eq .Environment "e2e"}}
+journald_reader_enabled: "true"
+{{else}}
+journald_reader_enabled: "false"
+{{end}}
+journald_reader_cpu: "1m"
+journald_reader_memory: "75Mi"
```
anton-ryzhov, line 129: Isn't that too much?

aermakov-zalando: I'll reduce it later after figuring out how much it actually needs.

PR #3607, opened 2020-09-28
Don't drop the cadvisor/node-exporter PRs

aermakov-zalando:
This will be done in a separate PR afterwards, to avoid flaky E2E runs.


aermakov-zalando:
👍

jrake-revelant:
👍

PR #3609, opened 2020-09-28
Drop the textfile collector

aermakov-zalando:
The directory was dropped since we haven't used it, but I forgot to drop the configuration.


aermakov-zalando:
👍

jrake-revelant:
👍

PR #3613, opened 2020-09-29
Make it possible to disable preemption as much as possible

aermakov-zalando:
* Update admission controller to support enforcing preemptionPolicy Never
* Drop the `visibility-logging` priority class since it's not needed anymore
* Remove the manifest for the `visibility-zmon` priority class, it'll be moved into the internal repo
* Add a priority class with the highest value but preemptionPolicy set to Never, for use by system components
* Drop the `system_pods_critical` config item. Instead, add `system_priority_class` which is set to `system-cluster-critical` in production clusters and `cluster-critical-nonpreempting` in test ones for now

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -220,12 +220,15 @@ teapot_admission_controller_daemonset_reserved_memory: "64Gi"
 {{if eq .Environment "production"}}
 teapot_admission_controller_validate_application_label: "true"
 teapot_admission_controller_validate_pod_template_resources: "true"
+teapot_admission_controller_preemption_enabled: "true"
 {{else if eq .Environment "e2e"}}
 teapot_admission_controller_validate_application_label: "false"
 teapot_admission_controller_validate_pod_template_resources: "false"
+teapot_admission_controller_preemption_enabled: "true"
 {{else}}
 teapot_admission_controller_validate_application_label: "false"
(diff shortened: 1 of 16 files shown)


aermakov-zalando:
👍

cluster/config-defaults.yaml
```diff
@@ -220,12 +220,15 @@ teapot_admission_controller_daemonset_reserved_memory: "64Gi"
...
 teapot_admission_controller_validate_pod_template_resources: "true"
+teapot_admission_controller_preemption_enabled: "true"
 {{else if eq .Environment "e2e"}}
 teapot_admission_controller_validate_application_label: "false"
 teapot_admission_controller_validate_pod_template_resources: "false"
+teapot_admission_controller_preemption_enabled: "true"
 {{else}}
 teapot_admission_controller_validate_application_label: "false"
 teapot_admission_controller_validate_pod_template_resources: "true"
+teapot_admission_controller_preemption_enabled: "true"
```
mikkeloscar: Should this also be enabled in test? Considering we are switching to `cluster-critical-nonpreempting` this could also be `false` or?

aermakov-zalando: Oops. Will fix!

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3651, opened 2020-10-14
Use cluster-critical-nonpreempting in production clusters

aermakov-zalando:
This shouldn't cause any issues since we're using the highest priority class value anyway, and can help with rapid scale ups of system components.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3656, opened 2020-10-15
Don't allow unknown fields in RouteGroup validation

mikkeloscar:
With CRD v1beta1 the field `preserveUnknownFields` defaults to `true` making the validation not reject invalid fields.
Explicitly set it to `false` to avoid users having to debug why some fields with wrong indentation are not applied.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3657, opened 2020-10-15
Make kube-proxy resources configurable

aermakov-zalando:
We have some issues in big clusters already, so we might need to tune it.


szuecs:
:+1:

aermakov-zalando:
👍

PR #3658, opened 2020-10-16
Update to v1.18.10

aermakov-zalando:
Might as well do this, considering that we haven't rolled out the other update yet.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3660, opened 2020-10-19
Optimize/skipper pod setup for hpa

szuecs:
FIX hpa problems with skipper-ingress pod deployment

### DIFF
--- a/cluster/manifests/skipper/configmap-sandbox-tokeninfo-bridge.yaml
+++ b/cluster/manifests/skipper/configmap-sandbox-tokeninfo-bridge.yaml
@@ -1,0 +1,32 @@
+{{ if eq .ConfigItems.enable_tokeninfo_bridge "true"}}
+apiVersion: v1
+data:
+  tokeninfo-bridge.eskip: |
+    health: Path("/healthz") -> inlineContent("ok") -> <shunt>;
+
+    q2h: Path("/oauth2/tokeninfo") && QueryParam("access_token") ->
+         queryToHeader("access_token", "Authorization", "Bearer %s") ->
+         dropQuery("access_token") ->
+         <loopback>;
+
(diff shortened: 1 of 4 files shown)


cluster/manifests/skipper/configmap-sandbox-tokeninfo-bridge.yaml
```diff
@@ -0,0 +1,32 @@
+{{ if eq .ConfigItems.local_sandbox_bridge "true"}}
```
szuecs: @mikkeloscar is this ok?

mikkeloscar: Maybe we can prefix it with `skipper_` so it's easier to find, same for the other config items, WDYT?

mikkeloscar: It looks fine from the templating PoV

szuecs: I changed the cfg item keys

mikkeloscar:
In general it looks good. I'm just wondering, since we're changing fundamental parts about how we run skipper-ingress it may make sense to put it behind a config item like `skipper_single_container_enabled` or so, such that we have a quick way to switch back in case something doesn't perform as expected.
My main concern is that we don't have any experience running multiple processes in one container so we don't know how it performs under high load and if there are any gotchas in this way of running. We are also reducing the overall resources by default by consolidating the containers and not having separate resource requests like before so we may have to adjust this per cluster where local tokeninfo is in use. This we ofc. also needs for the HPA where it should now be easier to configure.

One way to have both configs available would be to simply have two deployment files, one for the old and one for the new guarded with an `if eq .Cluster.ConfigItems.skipper_single_container_enabled "true|false"` or so. It means updating the image in two places if needed, but it's probably simpler than a complicated branching in a single file.

szuecs:
@mikkeloscar good idea, fixed

szuecs:
:+1:

mikkeloscar:
:+1:

PR #3664, opened 2020-10-21
Add back tokeninfo cpu/memory config items

mikkeloscar:
Follow up to #3660 adding back the tokeninfo resource config items which were accidentally dropped, but are still required by the legacy setup.


mikkeloscar:
:+1:

muaazsaleem:
👍

PR #3668, opened 2020-10-22
Make flannel resources configurable

mikkeloscar:
Allows setting higher memory/cpu for flannel if needed. This is needed in clusters with a lot of pods because of the huge iptables table that is needed to have all the kube-proxy routing.


mikkeloscar:
:+1:

aryszka:
:+1:

PR #3677, opened 2020-10-26
multiple bf releavant skipper-ingress fixes 

szuecs:
fix: shunt fallback route if no endpoints
fix: smoother fade-in when all endpoints are fading in
feature: better tracing in proxy span write request, write request headers, time to first byte from response
fix: error response content-length for requests with Transfer-Encoding: chunked
fix: flush stream to client in error paths


szuecs:
:+1:

mikkeloscar:
:+1:

PR #3679, opened 2020-10-27
Use aws-credentials-waiter from OS Registry

jmcs:
Using aws-credentials-waiter from the internal Pierone prevents restoring it in case of an outage, because it makes it dependent on itself.


mikkeloscar:
Closed in favor of https://github.com/zalando-incubator/kubernetes-on-aws/pull/3680

PR #3680, opened 2020-10-27
Use aws-credentials-waiter from OS Registry

mikkeloscar:
Using aws-credentials-waiter from the internal Pierone prevents restoring it in case of an outage, because it makes it dependent on itself.


mikkeloscar:
:+1:

jmcs:
:+1:

gargravarr:
:+1:

PR #3697, opened 2020-10-29
Limit maximum scale-down to 10 instances at a time

linki:
AWS recommends to not scale down more than 10 instances at a time. Let's use the recommended number by default (this is also the default in CA itself). We do have a use case where we want to increase that number, therefore the config item is left in place.


szuecs:
As far as I see we get errors because delete timeout of e2e tests.
Maybe we should split the e2e tests into more than steps.

muaazsaleem:
👍

mikkeloscar:
:+1:

PR #3700, opened 2020-10-30
Support permanent scale-up for CA v1.18 [1/2]

linki:
This deploys CA 1.18 with the permanent scale up feature. Here: run e2e with 1.18

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -12,6 +12,11 @@ autoscaling_scale_down_unneeded_time: "10m"
 # the cluster autoscaler release to use, options are 1_12 and 1_18.
 cluster_autoscaler_release: "1_12"
 
+# When true, enables the legacy behaviour for 1.18 where pools that are backed off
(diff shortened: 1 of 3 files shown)


cluster/config-defaults.yaml
```diff
@@ -12,6 +12,11 @@ autoscaling_scale_down_unneeded_time: "10m"
 # the cluster autoscaler release to use, options are 1_12 and 1_18.
 cluster_autoscaler_release: "1_12"
 
+# When true, enables the legacy behaviour for 1.18 where pools that are backed off
+# don't get reset to their pre-scale-out values. This basically means the ASGs
+# keep trying to get a node while CA already moves on to the next ASG.
+cluster_autoscaler_keep_backed_off_pools_scaled_up: "true"
```
aermakov-zalando, line 18: Why? Do we really plan to go to the upstream behaviour at some point?

linki: It's just in case we quickly somehow need to change it during the next days. I expect to remove this soon-ish.

PR #3702, opened 2020-10-30
Enable cluster autoscaler release 1.18 by default

linki:
Purpose of this PR:
* Use CA 1.18 in all clusters
* Enable EvenPodSpread Feature Gate in all clusters (by using the default)
* ~~Automatic Topology Spread in Test clusters~~ postponed
* ~~Opt-In Topology Spread in Production clusters~~ postponed

- [x] ~~Don't merge yet because of incompatibility with spot.io pools.~~ Config is overwritten to `disable` for spot.io pools.

How to revert:
* Set config item
```
zregistry clusters config-items set -k cluster_autoscaler_release -v 1_12 --cluster-alias stups 
```

Alternatively: Revert this PR. The FeatureGate change in here should have no effect but in case it does, revert.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -413,10 +413,6 @@ enable_endpointsliceproxying: "false"
 # Enable FeatureGate HPAScaleToZero
 enable_hpa_scale_to_zero: "true"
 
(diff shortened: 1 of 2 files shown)


linki:
:+1:

aermakov-zalando:
👍

PR #3706, opened 2020-10-30
Disable PG admitter for now

aermakov-zalando:
We've found some issues with it, so let's disable the admission rule for now (and disable it by default).


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3707, opened 2020-10-30
Don't force restart for 1.18 CA

linki:
Forcing a restart isn't needed anymore for the more recent version as it discovers those changes by itself.


aermakov-zalando:
👍

muaazsaleem:
👍

PR #3373, opened 2020-07-01
Remove all traces of API server ELBv1 (part 1/2)

linki:
Removes the switch between NLB and ELBv1 and all related sections.

PR #3714, opened 2020-11-04
Add option to switch to ELBv1 for APIServer

mikkeloscar:
Based on revert of #3373

Add the option to switch API-server LB back to ELBv1 instead of NLB. This is helpful when migrating between subnets because the NLB does NOT allow removing subnets once created so it's impossible to migrate an NLB without deleting it. ELBv1 does not have this limitation.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #3715, opened 2020-11-04
spotio: use -XX:+ExitOnOutOfMemoryError

aermakov-zalando:
We want it to crash and restart on OOM errors, not get stuck doing nothing. The vendor didn't want to do it by default, so we have to specify it ourselves.


aermakov-zalando:
👍

aermakov-zalando:
👍

aryszka:
:+1:

PR #3716, opened 2020-11-04
Enable 1.18 CA by default in all test clusters

aermakov-zalando:
We haven't had any issues in the current set, let's extend it to more clusters.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -10,7 +10,11 @@ autoscaling_max_empty_bulk_delete: "10"
 autoscaling_scale_down_unneeded_time: "10m"
 
 # the cluster autoscaler release to use, options are 1_12 and 1_18.
+{{if eq .Environment "production"}}
 cluster_autoscaler_release: "1_12"
+{{else}}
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
👍

linki:
We also need to drop usage of the configitem if we remove it: https://github.com/zalando-incubator/kubernetes-on-aws/blob/12a97f2666708eeb823ea91e0e9bc85365f3ae6c/cluster/manifests/kube-cluster-autoscaler/daemonset.yaml#L71

I kept it in case the artificial scale-up doesn't work correctly.

aermakov-zalando:
@linki For some reason `ack` didn't show it for me so I thought we've already dropped this.

szuecs:
:+1:

mikkeloscar:
:+1:

linki:
@aermakov-zalando Shouldn't we enable `EvenPodsSpread` along with CA 1.18? https://github.com/zalando-incubator/kubernetes-on-aws/blob/ec8a08d3b7a31fcf53a614c059ce1115fd603be3/cluster/config-defaults.yaml#L400

PR #3725, opened 2020-11-05
Enable TSC for skipper and skipper redis

aermakov-zalando:
This should allow spreading them more evenly between AZs. Currently only enabled in e2e, we'll test in individual clusters first.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3736, opened 2020-11-09
Update cluster-autoscaler to include per-node-group metric

linki:
Exposes a metric to see whether a node group is in backoff state. Backoff state is permanent in our fork until a node group can get new instances. This helps to detect whether it is waiting for nodes unusually long.

```
cluster_autoscaler_node_group_in_backoff{node_group="AutoScalingGroup-1P6...DQ98"} 0
cluster_autoscaler_node_group_in_backoff{node_group="AutoScalingGroup-15U...68YB"} 1
```

Could also be used in combination with

```
cluster_autoscaler_node_groups_count{node_group_type="autoscaled"} 2
```

to get a fraction of affected node pools compared to the total. Might help identifying more/less affected clusters.


aermakov-zalando:
Let's plug it into Prometheus as well!

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3737, opened 2020-11-09
Add back VPA for spotio-controller

mikkeloscar:
With the OOMKills correctly handled in the spotio-controller https://github.com/zalando-incubator/kubernetes-on-aws/pull/3715 we can now add the VPA back.


mikkeloscar:
:+1:

aermakov-zalando:
This won't work correctly unfortunately. JVM will exit with a non-zero status code (`1` I think, it might be configurable but not in the version they run), and this won't be treated by the rest of Kubernetes as an OOM kill.

aermakov-zalando:
👍

aermakov-zalando:
👎

mikkeloscar:
:sob:

PR #3743, opened 2020-11-10
Automatic TSC: non-default node pool support + other improvements

aermakov-zalando:
* Generate per-node pool configuration for the automatic TSC injection
 * Update the admission controller (dedicated node pool support + explicit config for the default pool)
 * Move the ExtendedResourceToleration so it works before our webhook


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3745, opened 2020-11-10
Drop unused config items from e2e

mikkeloscar:
These config items are no longer in use and thus can be dropped from e2e config.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3750, opened 2020-11-11
fix: redis ratelimit used pipelines which are not safe to use

szuecs:
fix: redis ratelimit used pipelines which are not safe to use: TCP RST, because of re-use of socket which were closed by the client


aryszka:
:+1:

aryszka:
:+1:

szuecs:
:+1:

PR #3754, opened 2020-11-12
Autoscaling fixes

aermakov-zalando:
* Update CA to fix a bug which breaks autoscaling if upcoming nodes are present ([PR](https://github.com/zalando-incubator/autoscaler/pull/64)).
 * Reenable buffer pods by default (at 1m/10Mi) to ensure that we have at least one instance in each AZ. This is needed for TSC to work correctly.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3756, opened 2020-11-12
Simplify enabling sandbox bridge

mikkeloscar:
Currently we have 3 config items for enabling tokeninfo in test clusters `skipper_local_tokeninfo`, `skipper_local_sandbox_tokeninfo` and `skipper_local_sandbox_bridge`. It only makes sense to either run all of them or none of them because the bridge depends on the two tokeninfo instances.

This aims to simplify the configuration by reducing it to a single config item: `skipper_local_tokeninfo: "production|sandbox|disabled"` with the default being `production` for production clusters and `bridge` for test clusters.

* `production` - only run production tokeninfo
* `bridge` - run bridge + production and sandbox tokeninfo
* `disabled` - don't run local tokeninfo but rely on the remote one. This requires `tokeninfo_url` set to a remote endpoint.

Additionally when running in `bridge` mode the livenessProbe for `tokeninfo` is changed to check all sub processes to ensure the container is restarted in case one of them fails.

### TODO

* Tweak liveness probe, currently tokeninfo can be down for 50s before it's recovered.
* Consider if liveness probe is the correct approach for managing the sub processes. E.g. if tokeninfo fails the impact is not just for routes using the tokeninfo filter but all requests will suffer when the skipper container is restarted.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -127,14 +127,12 @@ skipper_ingress_tokeninfo_cpu: "1000m"
 skipper_ingress_tokeninfo_memory: "512Mi"
 {{if eq .Environment "production"}}
 tokeninfo_url: "http://127.0.0.1:9021/oauth2/tokeninfo"
-skipper_local_tokeninfo: "true"
-skipper_local_sandbox_tokeninfo: "false"          # TODO(sszuecs): delete CR entries
-skipper_local_sandbox_bridge: "false"             # TODO(sszuecs): delete CR entries
+# production|bridge|disabled
+skipper_local_tokeninfo: "production"
 {{else}}
-tokeninfo_url: "http://127.0.0.1:9000/oauth2/tokeninfo"
-skipper_local_tokeninfo: "true"
-skipper_local_sandbox_tokeninfo: "true"
-skipper_local_sandbox_bridge: "true"
+tokeninfo_url: "" # can be set when local tokeninfo is disabled
+# production|bridge|disabled
+skipper_local_tokeninfo: "bridge"
 {{end}}
 
 # Image Policy Webhook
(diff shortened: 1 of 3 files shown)


cluster/config-defaults.yaml
```diff
@@ -120,14 +120,12 @@ skipper_ingress_tokeninfo_cpu: "1000m"
 skipper_ingress_tokeninfo_memory: "512Mi"
 {{if eq .Environment "production"}}
 tokeninfo_url: "http://127.0.0.1:9021/oauth2/tokeninfo"
-skipper_local_tokeninfo: "true"
-skipper_local_sandbox_tokeninfo: "false"          # TODO(sszuecs): delete CR entries
-skipper_local_sandbox_bridge: "false"             # TODO(sszuecs): delete CR entries
```
mikkeloscar, line 132: I dropped these from all clusters where they were set in CR (they were all set to the default so not functional change)

szuecs: thanks for the cleanup

szuecs:
I think liveness probe makes sense in this case, or we need to add a govenor process that restarts the processes inside the container.
The good about the livenessProbe is that skipper process will start teardown (on SIGTERM) with ALB, which we might want to tune in kube-ingress-aws-controller.

mikkeloscar:
:+1:

mikkeloscar:
> The good about the livenessProbe is that skipper process will start teardown (on SIGTERM) with ALB, which we might want to tune in kube-ingress-aws-controller.

For reference the current worst case with this setup is about 70s unavailability for a skipper pod because of 5 times livenessProbe of interval 10s and 2 ELB healthcheck of interval 10s. (First 50s is only routes/reqs relying on tokeninfo affected)

szuecs:
:+1:

PR #3770, opened 2020-11-16
VPA: increase max memory

aermakov-zalando:
Increase VPA `maxAllowed.memory` to `4Gi` for most of the applications and drop `maxAllowed.cpu`. We're already bumping into these in some of the clusters, and `4Gi` should be schedulable anywhere.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3771, opened 2020-11-16
CA: disable node instances cache (test only)

aermakov-zalando:
It doesn't really do anything useful and causes issues when scaling down (see [the PR](https://github.com/zalando-incubator/autoscaler/pull/70) for more details). Let's try disabling it, starting with the test clusters.


linki:
:+1:

aermakov-zalando:
👍

PR #3772, opened 2020-11-16
allow host definitions with port to support for vintage migration

szuecs:

### DIFF
--- a/cluster/manifests/skipper/crd.yaml
+++ b/cluster/manifests/skipper/crd.yaml
@@ -45,7 +45,7 @@ spec:
               type: array
               items:
                 type: string
-                pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'
+                pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*(:[0-9]+)?$'
               minItems: 1
             backends:
               type: array


diegoroccia:
:+1:

cluster/manifests/skipper/crd.yaml
```diff
@@ -45,7 +45,7 @@ spec:
               type: array
               items:
                 type: string
-                pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'
+                pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*(:[0-9]+)?$'
```
mikkeloscar, line 48: I propose we put this behind a config item to have it disabled in most cases to avoid confusion for users of regular ingress stack where the port is not relevant.

szuecs: why do we need this?
Yet another feature toggle...

szuecs: I honestly think we should reduce the number of toggles we have. We have already a bloated config with template values...

mikkeloscar: > why do we need this?
> Yet another feature toggle...

IMO this is a special case where it's a feature only needed for a single use case and thus is makes sense to have it behind a toggle. Having it by default would allow invalid configuration by the users (it's not valid to have any other port than :443 in ingress) which I think is beneficial to avoid.

> I honestly think we should reduce the number of toggles we have. We have already a bloated config with template values...

I agree that there a things we could simplify and I'm all for it. I don't think this is one of those cases where we win anything by not introducing a feature toggle.

PR #3777, opened 2020-11-18
scale down policy for skipper-ingress 

szuecs:
use proper scale down policy for skipper-ingress to allow only 10 instance per 2 minutes to scale down at max, such that ALB target groups are not creating issues dropping out skipper members

### DIFF
--- a/cluster/manifests/skipper/hpa.yaml
+++ b/cluster/manifests/skipper/hpa.yaml
@@ -16,11 +16,15 @@ spec:
   - type: Resource
     resource:
       name: cpu
-      targetAverageUtilization: {{ .ConfigItems.skipper_ingress_target_average_utilization_cpu }}
+      target:
+        type: Utilization
+        averageUtilization: {{ .ConfigItems.skipper_ingress_target_average_utilization_cpu }}
   - type: Resource
     resource:
       name: memory
-      targetAverageUtilization: {{ .ConfigItems.skipper_ingress_target_average_utilization_memory }}
+      target:
+        type: Utilization
+        averageUtilization: {{ .ConfigItems.skipper_ingress_target_average_utilization_memory }}
   behavior:
     scaleDown:
       stabilizationWindowSeconds: 600
--- a/cluster/manifests/skipper/hpa.yaml
+++ b/cluster/manifests/skipper/hpa.yaml
@@ -1,4 +1,4 @@
-apiVersion: autoscaling/v2beta1
+apiVersion: autoscaling/v2beta2
 kind: HorizontalPodAutoscaler
 metadata:
   name: skipper-ingress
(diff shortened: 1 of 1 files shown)


cluster/manifests/skipper/hpa.yaml
```diff
@@ -21,3 +21,14 @@ spec:
     resource:
       name: memory
       targetAverageUtilization: {{ .ConfigItems.skipper_ingress_target_average_utilization_memory }}
+  behavior:
+    scaleDown:
+      stabilizationWindowSeconds: 600
+      policies:
+      - type: Pods
+        value: 10
+        periodSeconds: 60
```
aermakov-zalando: Shouldn't this be 120 seconds, according to the PR description?

cluster/manifests/skipper/hpa.yaml
```diff
@@ -21,3 +21,14 @@ spec:
...
+      - type: Percent
```
aermakov-zalando, line 35: Not sure why this is needed.

cluster/manifests/skipper/hpa.yaml
```diff
@@ -21,3 +21,14 @@ spec:
     resource:
       name: memory
       targetAverageUtilization: {{ .ConfigItems.skipper_ingress_target_average_utilization_memory }}
+  behavior:
+    scaleDown:
+      stabilizationWindowSeconds: 600
```
aermakov-zalando, line 30: Is the default of 5 minutes not good enough?

cluster/manifests/skipper/hpa.yaml
szuecs: did not commit this change :D

cluster/manifests/skipper/hpa.yaml
szuecs: to mimic the default behavior and keep it

cluster/manifests/skipper/hpa.yaml
szuecs: Better slower in this case for workloads that are spiking forth and back

cluster/manifests/skipper/hpa.yaml
aermakov-zalando: Shouldn't the first policy be enough, even for cases when we scale up by less than 10 pods? Or does it not apply at all, resulting in no scale down?

szuecs:
:+1:

cluster/manifests/skipper/hpa.yaml
szuecs: as far as I understood the docs it's either percentage or the value 10, so I want to be safe. Feel free to test to proof that we don't need it.

aermakov-zalando:
👍

szuecs:
:+1:

PR #3787, opened 2020-11-20
Support single Ocean spot.io setup

mikkeloscar:
This introduces a new node pool profile `worker-spotio` which is modeled different from `worker-spotio-ocean`. Instead of creating one Ocean per node pool, we can now create a single Ocean per cluster and create "Virtual Node Groups" as node pools.

Benefits of this are:
1. We no longer need to run one spotinst controller per spotio node pool, but only need one for the overarching Ocean cluster.
2. Better control of when spot.io should scale based on taints/labels. Allows us to run it along side the CA (when we correctly configure taints/labels)

The intention is to later drop support for the `worker-spotio-ocean` node profile, but it's left for now to ensure smooth migration path.

~Additionally this adds a proposed node label: `cluster-lifecycle-controller.zalan.do/instance-manager` to be read by CLC to decide how to terminate an instance (ec2 or via ASG), this is a bit more abstracted compared to relying on the node profile name as we currently do.~

Additionally this updates the Cluster Lifecycle Controller to fall back to terminate instances via EC2 API if they are not managed by an ASG.

## Dependency

* [x] Update to CLM: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/401
* [x] Update to CLC: ~Support `cluster-lifecycle-controller.zalan.do/instance-manager`~
* [x] Handle cleanup of spotio node pools?

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -1857,6 +1857,61 @@ Resources:
             Resource: "*"
             Action:
             - "kms:Decrypt"
+{{- if and (index .Cluster.ConfigItems "spotio_account_id") (index .Cluster.ConfigItems "spotio_access_token") }}
+  SpotIOOceanCluster:
+    Type: Custom::ocean
+    Properties:
+      ServiceToken: !Sub "arn:aws:lambda:${AWS::Region}:178579023202:function:spotinst-cloudformation"
+      accessToken: "{{ .Cluster.ConfigItems.spotio_access_token }}"
+      accountId: "{{ .Cluster.ConfigItems.spotio_account_id }}"
+      autoTag: true
+      updatePolicy:
+        shouldUpdateTargetCapacity: false
+        shouldRoll: "false" # disable rolling update, this is managed by CLM,CLC
+        rollConfig:
+          roll:
+            batchSizePercentage: "25"
+      # API Docs: https://help.spot.io/spotinst-api/ocean/ocean-cloud-api/ocean-for-aws/create-2/
+      ocean:
+        name: "{{ .Cluster.Alias }}-{{ .Cluster.ID }}"
+        controllerClusterId: "{{ .Cluster.LocalID }}"
+        region: !Ref AWS::Region
+        autoScaler:
+          isEnabled: true
+          cooldown: 180 # 3 min. This seems to be the minimum value
+          # maximum resources which can be allocated to the cluster
+          # Use very high values as we have the limit set via the
+          # capacity.maximum (NodePool.MaxSize)
+          resourceLimits:
+            maxMemoryGib: 1500000
+            maxVCpu: 75000
+          down:
+            # evaluationPeriods: 3
+            maxScaleDownPercentage: 100
+          headroom: # disable headroom/preprovisioned instances
+            cpuPerUnit: 0
+            memoryPerUnit: 0
+            numOfUnits: 0
+          isAutoConfig: false
+        capacity:
+          minimum: 0
+          maximum: {{ nodeCIDRMaxNodes (parseInt64 .Cluster.ConfigItems.node_cidr_mask_size) (parseInt64 .Cluster.ConfigItems.reserved_nodes) }}
+          target: 0 # default target (should just be zero)
+        strategy:
+          fallbackToOd: {{ .Cluster.ConfigItems.spotio_ocean_fallback_to_ondemand }}
+          utilizeReservedInstances: {{ .Cluster.ConfigItems.spotio_ocean_utilize_reserved_instances }}
+        compute:
+          subnetIds:
+          {{- with $values := .Values }}
+          {{- range $az := $values.availability_zones }}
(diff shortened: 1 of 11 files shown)


test/e2e/cluster_config.sh
```diff
@@ -82,6 +84,72 @@ clusters:
...
+    - m5dn.4xlarge
+    - r5dn.large
+    - r5dn.xlarge
+    - r5dn.2xlarge
+    - r5dn.4xlarge
+    discount_strategy: none
+    min_size: 0
+    max_size: 21
+    config_items:
+      labels: dedicated=spotio,node.kubernetes.io/node-pool-profile=worker-spotio-ocean
```
mikkeloscar: TODO: remove `node.kubernetes.io/node-pool-profile=worker-spotio-ocean` when CLC is updated.

mikkeloscar:
:+1:

cluster/cluster.yaml
```diff
@@ -1857,6 +1857,61 @@ Resources:
...
+            cpuPerUnit: 0
+            memoryPerUnit: 0
+            numOfUnits: 0
+          isAutoConfig: false
+        capacity:
+          minimum: 0
+          maximum: {{ nodeCIDRMaxNodes (parseInt64 .Cluster.ConfigItems.node_cidr_mask_size) (parseInt64 .Cluster.ConfigItems.reserved_nodes) }}
+          target: 0 # default target (should just be zero)
+        strategy:
+          fallbackToOd: {{ .Cluster.ConfigItems.spotio_ocean_fallback_to_ondemand }}
```
aermakov-zalando, line 1901: Shouldn't this be per-node pool?

cluster/cluster.yaml
```diff
@@ -1857,6 +1857,61 @@ Resources:
...
+          utilizeReservedInstances: {{ .Cluster.ConfigItems.spotio_ocean_utilize_reserved_instances }}
+        compute:
+          subnetIds:
```
aermakov-zalando, line 1904: Same here, this should be per-node pool I think.

cluster/node-pools/worker-default/userdata.yaml
```diff
@@ -8,7 +8,7 @@ write_files:
     path: /etc/kubernetes/secrets.env
     content: |
       NODEPOOL_TAINTS={{if index .NodePool.ConfigItems "taints"}}{{.NodePool.ConfigItems.taints}}{{end}}
-      NODE_LABELS={{ .Values.node_labels }}{{if eq .NodePool.Profile "worker-spotio-ocean"}},spot.io=true{{end}},node.kubernetes.io/distro=ubuntu{{if index .NodePool.ConfigItems "labels"}},{{.NodePool.ConfigItems.labels}}{{end}}
+      NODE_LABELS={{ .Values.node_labels }}{{if or (eq .NodePool.Profile "worker-spotio-ocean") (eq .NodePool.Profile "worker-spotio")}},spot.io=true{{end}},node.kubernetes.io/distro=ubuntu{{if index .NodePool.ConfigItems "labels"}},{{.NodePool.ConfigItems.labels}}{{end}}
```
aermakov-zalando: Add `.NodePool.IsSpotIO()` to CLM maybe?

test/e2e/cluster_config.sh
```diff
@@ -41,6 +41,8 @@ clusters:
     routegroups_validation: "enabled"
     stackset_routegroup_support_enabled: "true"
     stackset_ingress_source_switch_ttl: "1m"
+    spotio_account_id: "${SPOTIO_ACCOUNT_ID}"
```
aermakov-zalando, line 44: How would it work in E2E, where we have multiple clusters? Can they coexist together just fine?

mikkeloscar: Yeah, that's not a problem, in the former implementation we had one ocean per node pool, now we just have one per cluster, as they are named after the cluster and each have their own spotinst controller running it should work fine.

cluster/node-pools/worker-default/userdata.yaml
mikkeloscar: Added here: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/431

cluster/cluster.yaml
mikkeloscar: This can only be set on the ocean and not per "virtual node group"

aermakov-zalando: Oh well. :(

cluster/cluster.yaml
mikkeloscar: This is a required field for some reason. I have added it also to the node pool config where it was missing, but we must define the super set here..

aermakov-zalando: I see, thanks!

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3809, opened 2020-11-30
Revert the Skipper HPA changes _correctly_

aermakov-zalando:
Reverting the API version as well means that the behaviour changes will not be reverted.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3812, opened 2020-12-01
redis: default to 512Mi memory

aermakov-zalando:
We've increased the Skipper HPA max _a lot_, redis now OOMs sometimes just keeping the connections open. Let's overprovision it by default.


aermakov-zalando:
👍

szuecs:
:+1:

PR #3816, opened 2020-12-02
remove originmarkers and update skipper

szuecs:
remove originmarkers, which were a couple of times the reason for not needed route updates triggers which consume some CPU that lead to scale up in test clusters

update skipper also has
- apimonitoring filter fixes (remove spec reference and regexp are cached to speedup route creation)
- refactored the cluttered logic in makeBackendRequest
- refactored ingress tests
- fixed ingress named port mapping
- fixed routegroup supports backendprotocol annotation similar to ingress
- fixed GH code scanning findings (remove sensitive data from logs)
- fixed possible panic by concurrent access of non syncronized map in secrets.Registry#GetEncrypter function
- feature OIDC auth filters, Auth Code Options will be able to be dynamically set by a query parameter

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: skipper-ingress
-    version: v0.11.195
+    version: v0.11.204
     component: ingress
 spec:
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

aermakov-zalando:
👍

PR #3817, opened 2020-12-03
Use gp3 for the root volume

aermakov-zalando:
It's cheaper and provides much better baseline performance, no reason not to switch.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -444,6 +444,7 @@ stackset_controller_mem_min: "120Mi"
 stackset_controller_mem_max: "4Gi"
 
 # EBS settings for the root volume
+ebs_master_root_volume_size: "50"
 ebs_root_volume_size: "50"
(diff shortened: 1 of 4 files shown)


aermakov-zalando:
Doesn't seem to be supported in CF at the moment. 🤷‍♂️

mikkeloscar:
This is the CF error:

```
You must use a valid fully-formed launch template. Volume of size 30GB is smaller than snapshot 'snap-0194da2e1369f5937', expect size >= 50GB (Service: AmazonAutoScaling; Status Code: 400; Error Code: ValidationError; Request ID: 560d3ecc-bcd0-4130-8185-b4a1c739a8ff; Proxy: null)
```

aermakov-zalando:
@mikkeloscar They must've fixed it, let's see if it works with the correct size.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3818, opened 2020-12-03
Use special labels to allow metrics to be deduped

linki:
Dedup pushed metrics according to https://grafana.com/blog/2019/10/03/deduping-ha-prometheus-samples-in-cortex/


linki:
This doesn't work. Here's another approach: https://cortexmetrics.io/docs/guides/ha-pair-handling/#config

PR #3820, opened 2020-12-03
Customise tcp retries as well

aermakov-zalando:
Add `tcp_syn_retries`/`tcp_retries2` to `allowed_unsafe_sysctls`, with a plan to automatically inject better values by default in the future.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -364,6 +364,9 @@ pod_max_pids: "4096"
 # the cpu management policy which should be used by the kubelet
 cpu_manager_policy: "none"
 
+# sysctl names allowed to be used in security policies, comma-separated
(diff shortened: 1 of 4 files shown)


mikkeloscar:
:+1:

szuecs:
@aermakov-zalando why don't we put these better defaults into the image with a global setting, instead of this?
Does a workload require a changed retry strategy, that you are aware of?

aermakov-zalando:
@szuecs Because the only way to change the default is by recompiling the kernel.

szuecs:
@aermakov-zalando ? Can't we just add the different settings via sysctl.conf ?

aermakov-zalando:
@szuecs Nope. In newly created network namespaces (so all pods except those running with `hostNetwork`) these settings reset to hardcoded kernel defaults. 🤷‍♂️

aermakov-zalando:
👍

PR #3821, opened 2020-12-04
PG/namespace delete protection

aermakov-zalando:
* Update admission controller
 * Add back the hooks for postgresql resources
 * Implement namespace delete protection (both off by default)


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3828, opened 2020-12-08
Add RBAC for aws-tagging-service

mikkeloscar:
Adds RBAC for aws-tagging-service which is running in one cluster and accessing all other clusters to tag AWS resources for cost allocation.

It needs access to volumes and services to map information from Kubernetes to the corresponding AWS resources.


aermakov-zalando:
👍

mikkeloscar:
Added pods list permissions which is also needed in case pvc or service doesn't have enough metadata, then pods metadata is used.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3832, opened 2020-12-09
suppress route update logs

aryszka:
Currently, we are printing every route in the logs whenever there is an update or a full reset. This is not useful at scale, and we never used this detail when investigating past routing issues. For diagnostics of potential ongoing problems, we can use the /routes diagnostics endpoint.

This change suppresses the logs on updates, resulting in printing only the total number of reset/udpated/deleted routes.


aryszka:
:+1:

mikkeloscar:
:+1:

PR #3833, opened 2020-12-09
Revert gp3 for spot.io

aermakov-zalando:
I was looking at the wrong part of the API docs. The one that we're using doesn't support customising the block device mappings unfortunately.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3862, opened 2020-12-16
add psp for postgres-operator that wants to have nice(2)

szuecs:
This adds `SYS_NICE` as an allowed capability to the existing restricted PSP.

This means the capability is NOT enabled by default for pods running with the restricted PSP, but users can opt-in by adding:

```yaml
securityContext:
      capabilities:
        add: ["SYS_NICE"]
```

to their pods/containers. The only intended use case for now is for Postgres DBs.

### DIFF
--- a/cluster/manifests/psp/pod_security_policy.yaml
+++ b/cluster/manifests/psp/pod_security_policy.yaml
@@ -40,6 +40,7 @@ spec:
   - SETPCAP
   - SETUID
   - SYS_CHROOT
+  - SYS_NICE
 ---
 apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
@@ -83,3 +84,4 @@ spec:
   - SETPCAP
   - SETUID
   - SYS_CHROOT
+  - SYS_NICE


jonathanbeber (commented):
It does not add `SYS_NICE` to the `defaultAddCapabilities` as expected by ACID team. It's the right approach in my opinion, but it will require a change on the operator to create the pods with 

```
capabilities:
        add: ["SYS_NICE"]
```

Their request was:
> Add SYS_NICE to the allowedCapabilities + add it to the defaultAddCapabilities. That would automatically make it work without changing the operator.

mikkeloscar:
Updated it as we discussed:

-----

This adds `SYS_NICE` as an allowed capability to the existing restricted PSP.

This means the capability is NOT enabled by default for pods running with the restricted PSP, but users can opt-in by adding:

```yaml
securityContext:
      capabilities:
        add: ["SYS_NICE"]
```

to their pods/containers. The only intended use case for now is for Postgres DBs.

jrake-revelant:
👍

mikkeloscar:
:+1:

PR #3870, opened 2021-01-04
Drop admission controller proxy [1/2]

aermakov-zalando:
The admission controller can now use a secret directly, so we don't need the proxy hack anymore. This PR adds a credentials set for the admission controller explicitly so we could drop the proxy in a followup PR.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3871, opened 2021-01-04
Drop admission controller proxy [2/2]

aermakov-zalando:
Follow-up to https://github.com/zalando-incubator/kubernetes-on-aws/pull/3870, must be merged separately. The admission controller now uses the secret directly, so the proxy isn't needed.


gargravarr:
:+1:

aermakov-zalando:
👍

PR #3885, opened 2021-01-06
Improve the e2e steps [1/2]

aermakov-zalando:
This reduces the babysitting needed for the certificates in e2e clusters.

Changes:
 * Use a dedicated script to enable/disable autoscaling. This means we don't need to drag the whole cluster YAML around, plus it's faster than running CLM.
 * Generate the keys and certificates for every e2e run instead of hardcoding them statically. This means they'll never expire and we don't have to do anything when new certificates are added.
 * Simplify the script a bit by providing the configuration externally in envars. This also allows us to put secrets into actual secrets.
 * Move the configuration to this repository instead of splitting it around.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3907, opened 2021-01-08
Add kube-node-ready-controller [1/2]

mikkeloscar:
Adds a controller that "marks" nodes ready once all daemonset pods are running and ready.

The purpose is to only schedule user pods on nodes that are considered ready (has all daemonset pods running and ready). This ensures that no user pods are scheduled on nodes that are broken on startup and thereby we don't need to page 24x7 for new nodes that are broken on startup.

### How it works

* Nodes are started with the taint `zalando.org/node-not-ready=:NoSchedule` such that only daemonset pods will be scheduled at startup.
* The `kube-node-ready-controller` (running HA as a daemonset on the master nodes) will watch for new nodes with the taint and check if all expected daemonset pods are ready
* If the pods are ready, the taint will be removed from the node and user pods will be able to be scheduled.
* If the pods are not ready the controller will recheck their state until they are ready OR the CA removes the node because it didn't get ready in time.

**NOTE:** This only enables the controller in test clusters. Once this is fully rolled out we can enable setting the taint on new nodes.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -211,6 +211,9 @@ kube_state_metrics_mem_min: "120Mi"
 kubernetes_lifecycle_metrics_mem_max: "4Gi"
 kubernetes_lifecycle_metrics_mem_min: "120Mi"
 
+kube_node_ready_controller_cpu: "50m"
+kube_node_ready_controller_memory: "200Mi"
+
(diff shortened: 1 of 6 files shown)


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #3905, opened 2021-01-08
cdp-controller-rbac: create/delete batch/v1 Job

thilp:
Hello! To allow users to simulate HTTP traffic to their app during gradual traffic switching, CDP needs to create (and eventually delete) Job resources. More details at https://github.bus.zalan.do/automata/issues/issues/3455.

PR #3916, opened 2021-01-12
cdp: deletecollection batch/v1 Job

thilp:
Hello! Follow-up to #3905. cdp-controller actually deletes Jobs via label selector to avoid keeping track of Job names.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3917, opened 2021-01-12
Use simpler and familiar way to specify OIDC IAM role

linki:
https://github.bus.zalan.do/teapot/admission-controller/pull/103 enables a simpler and more familiar way to specify the OIDC-based IAM role:

Renames `eks.amazonaws.com/role-arn` to the name kube2iam introduced and is widely used: `iam.amazonaws.com/role`. It avoids confusion by clarifying that we don't run on EKS. Both annotations are supported during the transition.

Instead of `iam.amazonaws.com/role: <full-arn>` one can specify an account-local role and omit the ARN prefix, e.g. `iam.amazonaws.com/role: my-fancy-role`.

Cross-account roles aren't supported out of the box yet but will require to specify the full ARN.

Furthermore, there are checks during Pod and ServiceAccount admission that catch malformed IAM role definitions.


mikkeloscar:
:+1:

linki:
:+1:

PR #3921, opened 2021-01-14
Revert "Use simpler and familiar way to specify OIDC IAM role"

linki:
Let's add back support for the eks annotation as we have a bunch of cdp based deployments and I don't want this to go to alpha to soon.


mikkeloscar:
:+1:

aermakov-zalando:
Is this still needed?

PR #3922, opened 2021-01-14
add scale down policy for skipper-ingress

szuecs:
use proper scale down policy for skipper-ingress to allow only 10 instance per 2 minutes to scale down at max, such that ALB target groups are not creating issues dropping out skipper members

was fixed in v1.19.3 and v1.18.11


mikkeloscar:
:+1:

szuecs:
:+1:

PR #3929, opened 2021-01-15
Reduce stackset test parallelism

aermakov-zalando:
The stackset tests fail in half of the e2e runs, which is pretty annoying. Logs are full of throttling warnings, so let's see if reducing the parallelism would help. This shouldn't change the total time an e2e run takes because stackset tests run in parallel with the normal ones anyway.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3934, opened 2021-01-15
Disable privileged PSP by default

mikkeloscar:
Disable use of Privileged PSP by PowerUsers by default.

This will mean that PowerUsers (or CDP) can't run privileged pods. It will only be possible to run privileged pods in `kube-system` namespace where we have use cases for this.

I have set `privileged_psp_enabled=true` for the 13 clusters where privileged pods are still running.


aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #3939, opened 2021-01-18
journald-reader: enable by default

aermakov-zalando:
It's currently explicitly disabled in production clusters, we'll enable it one by one.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3948, opened 2021-01-19
Add OIDC provider and subject prefix to deployment config

linki:
This generates and adds OIDC provider and subject to the deployment config so it can be accessed by cdp-steps.


aermakov-zalando:
👍

linki:
:+1:

PR #3950, opened 2021-01-19
Rename 'restricted' PSP to 'default'

aermakov-zalando:
We want to prevent the users from relying on the `privileged` PSP, but there doesn't seem to be an easy way to do this. Even if the users update their pods to not use any features of the `privileged` PSP, the pods will still use it if the service account has access to it (because it comes before `restricted` lexicographically). Dropping the `use` permission later will then break pod updates completely.

The simplest way to fix this is to simply rename the `restricted` PSP so it comes before `privileged`. After the users' pods are recreated, they'll pick it instead of the `privileged` one even if they have access to the latter, and then they can safely revoke their access without any effect.

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -165,3 +165,9 @@ post_apply:
   kind: PriorityClass
 - name: autoscaling-buffer
   kind: PriorityClass
+- kind: ClusterRole
+  name: restricted-psp
+- kind: ClusterRoleBinding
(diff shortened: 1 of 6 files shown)


mikkeloscar:
:+1:

jonathanbeber (approved):

jonathanbeber:
:+1:

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3955, opened 2021-01-20
etcd: use TLS

aermakov-zalando:
Switch to the TLS-enabled etcd cluster. Note that the e2e will break until we redeploy it.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3956, opened 2021-01-21
Revert "Rename 'restricted' PSP to 'default'"

aermakov-zalando:
This creates issues for some of our users who have accidentally been relying on the bad defaults in the `privileged` PSP. Let's revert it for now and ensure that they're properly migrated first.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #3964, opened 2021-01-22
Revert to older image for admission-controller-proxy

mikkeloscar:
Newer image doesn't have the proxy inside anymore because of #3871 

Revert updating the image so we can still rely on the proxy for #3961


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #3984, opened 2021-01-26
Add OIDC provider creation and deletion to CF stack

linki:
Allows to create and delete OIDC Identity Providers as part of the CloudFormation Stack.

Allows to remove manual management in CLM: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/415

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -2006,6 +2006,101 @@ Resources:
             Resource: "*"
             Action:
             - "kms:Decrypt"
+  IdentityProvider:
+    Type: Custom::IdentityProvider
+    Properties:
+      ServiceToken: !GetAtt IdentityProviderCreator.Arn
+      Region: !Ref "AWS::Region"
+      ProviderName: "{{ .Cluster.LocalID }}.{{ .Values.hosted_zone }}"
+  IdentityProviderCreator:
+    Type: AWS::Lambda::Function
+    Properties:
+      Runtime: python2.7
+      Handler: index.lambda_handler
+      MemorySize: 128
+      Role: !GetAtt LambdaExecutionRole.Arn
+      Timeout: 30
+      Code:
+        ZipFile: !Sub |
+          import boto3
+          from botocore.exceptions import ClientError
+          import json
+          import cfnresponse
+
+          iam = boto3.client("iam")
+
+          def create_provider(provider_name):
+            try:
+              resp = iam.create_open_id_connect_provider(
+                  Url= "https://" + provider_name,
+                  ClientIDList=['sts.amazonaws.com'],
+                  ThumbprintList=['9e99a48a9960b14926bb7f3b02e22da2b0ab7280']
+              )
+              return(True, resp['OpenIDConnectProviderArn'])
+            except Exception as e:
+              return (False, "Cannot create OpenID Connect provider: " + str(e))
+
+          def delete_provider(arn):
+            try:
+              resp = iam.delete_open_id_connect_provider(OpenIDConnectProviderArn=arn)
+              return (True, "OpenID Connect provider with ARN " + arn + " deleted")
+            except ClientError as e:
+              if e.response['Error']['Code'] == "NoSuchEntity":
+                # no need to delete a thing that doesn't exist
+                return (True, "OpenID Connect provider with ARN " + arn + " does not exist, deletion succeeded")
+              else:
+                return (False, "Cannot delete OpenID Connect provider with ARN " + arn + ": " + str(e))
+            except Exception as e:
+              return (False, "Cannot delete OpenID Connect provider with ARN " + arn + ": " + str(e))
+
+          def lambda_handler(event, context):
+            provider_name = event['ResourceProperties']['ProviderName']
+            # create a default ARN from the name; will be overwritten if we are creating
+            provider_arn = "arn:aws:iam::${AWS::AccountId}:oidc-provider/" + provider_name
+
+            if event['RequestType'] == 'Create':
+              res, provider_arn = create_provider(provider_name)
+              reason = "Creation succeeded"
+            elif event['RequestType'] == 'Delete':
+              res, reason = delete_provider(provider_arn)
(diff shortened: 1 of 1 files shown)


cluster/cluster.yaml
```diff
@@ -2006,6 +2006,101 @@ Resources:
...
+              return (False, "Cannot create OpenID Connect provider: " + str(e))
+
+          def delete_provider(arn):
+            try:
+              resp = iam.delete_open_id_connect_provider(OpenIDConnectProviderArn=arn)
+              return (True, "OpenID Connect provider with ARN " + arn + " deleted")
+            except ClientError as e:
+              if e.response['Error']['Code'] == "NoSuchEntity":
+                # no need to delete a thing that doesn't exist
+                return (True, "OpenID Connect provider with ARN " + arn + " does not exist, deletion succeeded")
```
aermakov-zalando, line 2050: That's not very Pythonic, just throw an exception and catch it later.

cluster/cluster.yaml
```diff
@@ -2006,6 +2006,101 @@ Resources:
...
+              else:
+                return (False, "Cannot delete OpenID Connect provider with ARN " + arn + ": " + str(e))
```
aermakov-zalando, line 2052: That's not very Pythonic, just throw an exception and catch it later.

cluster/cluster.yaml
```diff
@@ -2006,6 +2006,101 @@ Resources:
...
+  IdentityProvider:
+    Type: Custom::IdentityProvider
+    Properties:
+      ServiceToken: !GetAtt IdentityProviderCreator.Arn
+      Region: !Ref "AWS::Region"
+      ProviderName: "{{ .Cluster.LocalID }}.{{ .Values.hosted_zone }}"
+  IdentityProviderCreator:
+    Type: AWS::Lambda::Function
+    Properties:
+      Runtime: python2.7
```
aermakov-zalando, line 2018: 2.7 has been EoL'd a year ago!

aermakov-zalando:
Is this really better than just doing it in CLM?

cluster/cluster.yaml
```diff
@@ -2006,6 +2006,101 @@ Resources:
...
+              res, provider_arn = create_provider(provider_name)
+              reason = "Creation succeeded"
+            elif event['RequestType'] == 'Delete':
+              res, reason = delete_provider(provider_arn)
+            else:
+              res = False
+              resp = "Unknown operation: " + event['RequestType']
+
+            responseData = {}
+            responseData['Reason'] = reason
```
aermakov-zalando, line 2071: `responseData = {"Reason": reason}`?

linki:
@aermakov-zalando In CLM it's difficult to handle the deletion part, e.g. when an OIDC provider is removed. With this we would just loop over the list and define a provider each. If a provider is not part of the Stack anymore CF takes care of deleting it.

If you have another idea I'm happy to discuss it in the PR: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/414

aermakov-zalando:
@linki No, I don't have any better ideas unfortunately (assuming we still want to do this with CLM).

cluster/cluster.yaml
```diff
@@ -2006,6 +2006,101 @@ Resources:
...
+                return (True, "OpenID Connect provider with ARN " + arn + " does not exist, deletion succeeded")
+              else:
+                return (False, "Cannot delete OpenID Connect provider with ARN " + arn + ": " + str(e))
+            except Exception as e:
+              return (False, "Cannot delete OpenID Connect provider with ARN " + arn + ": " + str(e))
+
+          def lambda_handler(event, context):
+            provider_name = event['ResourceProperties']['ProviderName']
+            # create a default ARN from the name; will be overwritten if we are creating
+            provider_arn = "arn:aws:iam::${AWS::AccountId}:oidc-provider/" + provider_name
```
aermakov-zalando, line 2059: Does this actually work? I'm not sure what would do the `${AWS::AccountId}` interpolation in this case.

cluster/cluster.yaml
```diff
@@ -2006,6 +2006,101 @@ Resources:
...
+  IdentityProviderCreator:
+    Type: AWS::Lambda::Function
+    Properties:
+      Runtime: python2.7
+      Handler: index.lambda_handler
+      MemorySize: 128
+      Role: !GetAtt LambdaExecutionRole.Arn
+      Timeout: 30
+      Code:
+        ZipFile: !Sub |
```
muaazsaleem, line 2024: nit: Maybe we could deploy using a container rather than a zip. WDYT?

cluster/cluster.yaml
muaazsaleem: How do you mean? Like with nested `try`s?

cluster/cluster.yaml
mikkeloscar: All the code is inside: `ZipFile: !Sub |`

cluster/cluster.yaml
aermakov-zalando: I've meant using multiple return of a bool + string instead of just letting the exception propagate upwards, this looks like something from Go. I don't see why the `try` blocks would even need to be nested, you can just have a `try/catch` around the `delete` part and parse the error reason, or just not return any string for the happy path.

linki:
Sorry, I should have mentioned that this is an adaption of https://github.com/cevoaustralia/cfn-identity-provider for OIDC providers. (Thanks @nonspecialist)

So yes, I'm happy to fix the python code if we want to use this approach. I didn't know you could extend CF this way.

cluster/cluster.yaml
linki: Do you mean we can use a docker image to hold the code instead on inlining it here?

linki:
Closed for now

aermakov-zalando:
This solves the problem of tracking OIDC providers in a nice way, so it makes sense to proceed with this approach. I'd definitely update the Python code a bit, though.

PR #3987, opened 2021-01-27
Move node-not-ready taint to admission-controller

mikkeloscar:
Instead of setting `zalando.org/node-not-ready` taints via kubelet flag, move it to admission-controller so it's simpler to enable/disable without rolling all the nodes.

Dropped `kube_node_ready_controller_enabled` as it was never actually used :facepalm: 

Enabled by default since we have run the controller for a while in test already

### DIFF
--- a/cluster/manifests/01-admission-control/config.yaml
+++ b/cluster/manifests/01-admission-control/config.yaml
@@ -55,3 +55,5 @@ data:
   pod.automatic-topology-spread.pools.dedicated.{{$group}}: "{{ $enabled }}"
 {{- end }}
 {{- end}}
+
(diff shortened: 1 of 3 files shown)


cluster/manifests/01-admission-control/config.yaml
```diff
@@ -55,3 +55,5 @@ data:
   pod.automatic-topology-spread.pools.dedicated.{{$group}}: "{{ $enabled }}"
 {{- end }}
 {{- end}}
+
+  node.node-not-ready-taint.enable: "{{ .Cluster.ConfigItems.teapot_addmission_controller_node_not_ready_taint }}"
```
aermakov-zalando, line 59: This should be set in the defaults?

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #3992, opened 2021-01-28
Update dnsmasq

aermakov-zalando:
* Update dnsmasq, getting rid of the `nanny` wrapper as well.
 * Get rid of the liveness probe (which went to a completely different container anyway), add a readiness probe instead.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4005, opened 2021-02-01
Always enable OIDC-based IAM roles and remove configuration option

linki:
Now that OIDC-based IAM roles are enabled on all clusters we can get rid of some configuration options.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4007, opened 2021-02-01
Serve OIDC discovery documents directly from API server

linki:
Avoids the need for extra and possibly wrong configuration in the skipper sidecar for OIDC Discovery Documents.

---

Changes:

FeatureGate `ServiceAccountIssuerDiscovery=true` will be default in Kubernetes 1.20 and can be removed then ([Feature Gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/))

```
--service-account-jwks-uri={{ .Cluster.APIServerURL }}/openid/v1/jwks
```
Changes the Host where keys can be looked up to the cluster's public endpoint (the default is an internal node ip)

Dropped:
> ```
> --anonymous-auth=true
> ```
> Is required to not reject all unauthenticated requests but rather let RBAC rules decide whether to grant anonymous access. This is used to open up the two paths needed to lookup the OIDC documents (Kubernetes provides a default cluster role for that).
> 
> Note: There's no binding to that role for unauthenticated users by default, hence this PR will probably fail. More on that in the [official docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery).

Instead skipper sidecar is still used but forwards directly to the API server, therefore removing the need to generate those files ourselves.


linki:
:+1:

mikkeloscar:
:+1:

PR #4011, opened 2021-02-03
Still support old OIDC keys path [1/2]

linki:
Instead of https://github.com/zalando-incubator/kubernetes-on-aws/pull/4010 to play it a bit safer.

Enable support for legacy OIDC keys path.

So this one is "no-op" for dev and alpha.


linki:
:+1:

mikkeloscar:
:+1:

PR #4018, opened 2021-02-08
Add oidc-subject-key helper variable

linki:
This adds another helper variable that only contains the key instead of the super-magic key-value combination.

Usage within CDP deployments can change from the current (and deprecated):

```yaml
          Condition:
            StringEquals:
              {{{CDP_OIDC_SUBJECT_PREFIX}}}:default:aws-limits
```

to

```yaml

          Condition:
            StringEquals:
              "{{{CDP_OIDC_SUBJECT_KEY}}}": "system:serviceaccount:default:aws-limits"
```


aermakov-zalando:
👍

linki:
:+1:

PR #4034, opened 2021-02-12
Add protocol to port definitions

aermakov-zalando:
There's [a bug](https://github.com/kubernetes-sigs/structured-merge-diff/issues/130) with server-side apply that fails for port definitions that lack a `protocol` field. Most of our manifests have it set properly, so let's fix the remaining ones just in case we actually switch to it.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4042, opened 2021-02-15
coredns/flannel: check API server availability first

aermakov-zalando:
Both flannel (or rather the apiserver-proxy container that we still have to run) and coredns crash during startup if the API server is not available. Because of a race with `kube-proxy` this happens very often and is usually mitigated by a restart or two. This is fairly annoying, because we still want to be alerted when these pods restart, but Kubernetes doesn't track enough information about the pods to exclude these restarts correctly. Instead of investing a lot into a custom monitoring tool, let's just add a small init container that would check the API server connectivity.


jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #4046, opened 2021-02-16
Configure logging/destination for all system pods

aermakov-zalando:
We can now send logs to a dedicated Scalyr account, which would help a lot with the querying and debugging. This PR configure the `logging/destination` annotation for all kube-system pods. Most of the pods are configured with `scalyr/stups`, so the logs would only be available to us, but there are exceptions:
 * `external-dns`, `kube-ingress-aws-controller` and `kube-static-egress-controller` log to both accounts. Unfortunately they don't report errors in a nice way (e.g. using events), so it'd be impossible for the users to debug otherwise.
 * `cluster-lifecycle-controller`, `emergency-access-service` and `kubernetes-event-logger` log to both accounts as well. These are used by the users when they investigate something, and they don't log that much.
 * `skipper-ingress` logs only to the cluster's account due to the amount of logs it creates. We don't want to duplicate those needlessly.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -172,6 +172,9 @@ journald_reader_memory: "30Mi"
 # Logging settings
 logging_s3_bucket: "zalando-logging-{{.InfrastructureAccount | getAWSAccountID}}-{{.Region}}"
 scalyr_team_token: ""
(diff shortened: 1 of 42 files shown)


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4048, opened 2021-02-16
Update ingress-controller and adjust EW range rule

jonathanbeber:
In #3952 we enabled the Skipper [EastWest Range][0] feature using the
predicates `ClientIP` and `SourceFromLast` as [discussed][1]. It
introduced a problem: other skipper deployments running inside clusters
(not the ingress one) could receive traffic and route it back to the
`skipper-ingress` deployment. This traffic would be denied, since the
`X-Forwarded-For` header won't be handled and its value is as of an
external IP.

This commit updates the kube-ingress-aws-controller to the version
`v0.11.19`. This new version brings among, a fix for NLB HealthCheck
timeouts and dependencies updates, the feature to [deny traffic for
internal domains][2] on the ALB level. It allow us to change the skipper
[east-west range][0] feature to block traffic based just on the
`ClientIP` predicate. This commit, therefore, removes the
`SourceFromLast` predicates previously configured.

[0]: https://skipper.docs.zalando.net/tutorials/operations/#east-west-range
[1]: https://github.com/zalando/skipper/issues/1526#issuecomment-756903746
[2]: https://github.com/zalando-incubator/kube-ingress-aws-controller#deny-traffic-for-internal-domains

### DIFF
--- a/cluster/manifests/ingress-controller/deployment.yaml
+++ b/cluster/manifests/ingress-controller/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: kube-ingress-aws-controller
-    version: v0.11.14
+    version: v0.11.20
 spec:
   replicas: 1
   selector:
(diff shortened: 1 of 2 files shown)


jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #4015, opened 2021-02-04
use kubernetes' default jwks path

linki:
Just to avoid unnecessary confusion by using different key paths.

This should be sprinkled in with other required master updates since it's not urgent in any way.

PR #4049, opened 2021-02-16
Remove legacy jwks path

linki:
With https://github.com/zalando-incubator/kubernetes-on-aws/pull/4015 merged to all channels we can get rid of supporting the old jwks path (and use Kubernetes' default).


linki:
:+1:

mikkeloscar:
:+1:

PR #4059, opened 2021-02-18
Add support for ZMON metrics in StackSet

mikkeloscar:
https://github.com/zalando-incubator/stackset-controller/releases/tag/v1.3.23

PR #4060, opened 2021-02-18
Update stackset CRDs to reflect latest version

mikkeloscar:
Follow up to #4059 also update the CRDs so the new metric can be defined!


mikkeloscar:
:+1:

jonathanbeber:
:+1:

PR #4062, opened 2021-02-19
PowerUser: allow get on pods/{exec, portforward, proxy}, services/proxy

aermakov-zalando:
This apparently exists as a separate verb, for whatever reason. It'd be nice if Kubernetes RBAC was based on semantics, not particularities of their HTTP API, but for now let's just allow it.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4069, opened 2021-02-19
Only run spotio controller when node pools are defined

mikkeloscar:
Only run spotio controller when spot.io node pools are defined!


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4072, opened 2021-02-22
Fix the AWSIAM test by using a static awscli image

aermakov-zalando:
The existing image has started crashing because `pip install awscli` was causing OOMs. Use a prebuilt image instead.


szuecs:
:+1:

aryszka:
:+1:

aermakov-zalando:
I was still working on this and tests were failing.

PR #4076, opened 2021-02-23
e2e-fix: Avoid zero padding of week number

mikkeloscar:
Fixes an issue where bash interprets the week numbers as an invalid octal:

```
printf "junit-reports/%04d-%02d/%s" "$(date +%Y)" "$(date +%V)" "local"
bash: printf: 08: invalid octal number
```

This issue only happens in CW08 and CW09 :sweat_smile: Fix is to avoid the zero padding in the `date` week number output.


muaazsaleem:
👍

mikkeloscar:
:+1:

PR #4077, opened 2021-02-23
Admission controller: add a token for the base image check

aermakov-zalando:
This will be needed for a follow-up PR. Additionally, deploy the PCS to all clusters, because it doesn't break anything.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4080, opened 2021-02-23
Admission controller: base image validation

aermakov-zalando:
If `teapot_admission_controller_validate_base_images` is set to `true` (currently defaults to `false` for all clusters), enable the base image check for all pod factories (deployments/statefulsets/jobs/etc). This requires the cluster's `docker_meta_url` to be set to be meaningful; it'll be configured in the private config repo.
Additionally, if `teapot_admission_controller_validate_base_images_namespaces` is set, the validation will only run in the namespaces that match this regex (`kube-system` is allowed by default, so this setting should only be used for testing).


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4086, opened 2021-02-25
Support scraping Prometheus metrics via pod entities [1/2]

mikkeloscar:
This introduces Prometheus scraping configuration for scraping pods with the annotations:

```yaml
annotations:
  prometheus.io/path: /metrics
  prometheus.io/port: "7980"
  prometheus.io/scrape: "true"
```

The intention is to replace it with the endpoints based scraping which we currently use. The reason for this is that it's undesirable to have to create a service just for scraping Prometheus metrics.
* It's inconvenient to configure
* It produces unneeded endpoints which is tracked by kube-proxy and configured in iptable on each node for no benefit. (During Cyberweek 2020 we had very big iptables causing flannel to fail, let's not contribute to the size via services just used for Prometheus metrics discovery).

This is the first of two PRs for migrating to pod based entities. This one just sets the relevant annotations on all components and adds the Prometheus configuration behind a config item flag.
Once all pods are annotated we can switch to pod entity based scraping and remove the services that are no longer used. We want to do it in two steps to not scrape double during the migration period.

### DIFF
--- a/cluster/manifests/coredns-local/daemonset-coredns.yaml
+++ b/cluster/manifests/coredns-local/daemonset-coredns.yaml
@@ -25,6 +25,10 @@ spec:
         component: cluster-dns
       annotations:
         logging/destination: "{{.Cluster.ConfigItems.log_destination_infra}}"
+        prometheus.io/path: /metrics
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "9153" # coredns
+        prometheus.io/sidecar-port: "9054" # dnsmasq/unbound
     spec:
       initContainers:
       - name: ensure-apiserver
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -201,6 +201,13 @@ prometheus_remote_max_backoff: "10s"
 # TODO: remove when kube-state-metrics is fixed
 prometheus_kube_state_metrics_drop_labels: ""
 
+# control migration from endpoint to pod based scraping entities
+# enable scraping metrics based on pod entities
+prometheus_pod_entity_scrape: "false"
+# enable scraping metrics based on endpoints entities (deprecated)
+# this is the default if prometheus_pod_entity_scrape=false
+prometheus_endpoints_entity_scrape: "false"
+
 metrics_service_cpu: "100m"
 metrics_service_mem: "200Mi"
 metrics_service_mem_max: "4Gi"
--- a/cluster/manifests/audittrail-adapter/daemonset.yaml
+++ b/cluster/manifests/audittrail-adapter/daemonset.yaml
@@ -20,6 +20,9 @@ spec:
         version: master-24
       annotations:
         logging/destination: "{{.Cluster.ConfigItems.log_destination_infra}}"
+        prometheus.io/path: /metrics
+        prometheus.io/port: "7980"
+        prometheus.io/scrape: "true"
     spec:
       serviceAccountName: audittrail-adapter
       priorityClassName: system-node-critical
--- a/cluster/manifests/external-dns/deployment.yaml
+++ b/cluster/manifests/external-dns/deployment.yaml
@@ -19,6 +19,9 @@ spec:
         version: v0.7.6
       annotations:
         logging/destination: "{{.Cluster.ConfigItems.log_destination_both}}"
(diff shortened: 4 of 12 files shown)


cluster/manifests/coredns-local/daemonset-coredns.yaml
```diff
@@ -25,6 +25,10 @@ spec:
         component: cluster-dns
       annotations:
         logging/destination: "{{.Cluster.ConfigItems.log_destination_infra}}"
+        prometheus.io/path: /metrics
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "9153" # coredns
+        prometheus.io/sidecar-port: "9054" # dnsmasq/unbound
```
mikkeloscar, line 31: This is sort of a hack because the pod has two containers exposing metrics. Ideas for how to do this better are welcome!

aermakov-zalando: How does this actually work? Wouldn't you get two labels remapped into the same one, so you'd still only get monitoring for one of the containers?

jonathanbeber: what do you think about adding to the config a `source_labels: [__meta_kubernetes_pod_container_port_name]` with `regex: metrics` (or even a more complex regex if needed, like `metrics(-.*)?`) and then define the port metrics on every pod? This way we don't need to specify the port of each container as an annotation.

mikkeloscar: > How does this actually work? Wouldn't you get two labels remapped into the same one, so you'd still only get monitoring for one of the containers?

From my understanding it explodes it into two entities this way. That is what I observed when I tested it.

@jonathanbeber 

Yeah, I also thought about this, maybe it's simpler. I was worried if there would be cases where we need a different port name for some other reason so it couldn't match, but this is only relevant to our pods, so I think it's actually fine.

I would do this if there are no objections?

jonathanbeber: @mikkeloscar +1 from my side

aermakov-zalando: Can you define the same port multiple times, with different names? If yes, then it's a very nice solution.

jonathanbeber: @aermakov-zalando I think in this case the little bit more complex regex will be needed.

mikkeloscar: @aermakov-zalando seems like you cannot. When you try to add another port with the same number but different name only the first entry in the list is persisted..

aermakov-zalando: Then I'm not sure if I like it, because in a lot of apps metrics would be served from the main port, and I wouldn't want to rename it.

jonathanbeber: Yeah, that would be bad, my suggestion would be renaming it announcing it's also metrics, something like `http-metrics` but it's not clean and weird. From this PR just two applications would actually conflict 

- `kube-static-egress-controller` uses the 8080 but does not define a name for the port, so it wouldn't be renaming;
- `kubernetes-lifecycle-metrics` uses 9090 and also does not define a name, but given the app name I don't think metrics would be a bad name for the port.

Are there other applications that this PR is not considering yet?

jonathanbeber: btw, I'm also fine with using the annotation, I don't think we need to make it more complicated just to avoid one extra annotation here. It's not like it will affect our users, it's just us and our own pods.

mikkeloscar: I don't have a strong opinion. I think it's a bit awkward to name the ports something like metrics, so it's maybe less awkward with the extra annotation for the special case?

jonathanbeber: @mikkeloscar :shipit: let's go for extra annotation

jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #4092, opened 2021-02-25
Support scraping Prometheus metrics via pod entities [2/2]

mikkeloscar:
Follow up to #4086 (must be fully rolled out)

This drops support for discovering scraping targets via services and uses the pod based discovery instead.
Drops all the services no longer needed as they were only used for Prometheus scraping configuration.


mikkeloscar:
:+1:

jonathanbeber:
:+1:

PR #4099, opened 2021-03-02
Run kube-proxy with a service account [node auth 1/X]

aermakov-zalando:
This is one of the preparation steps for rolling out proper node authnz.

Currently, `kube-proxy` runs with the kubelet's hardcoded credentials that have a lot of permissions. This will be a lot more annoying to implement once we switch to node auth, and we don't really need it. Implement the following changes instead to make it better:
 * Disable token automounting for the service account (we're not using it anyway in any of the configurations)
 * Provision a proper binding for the built-in `system:node-proxier` role
 * If `experimental_proxy_use_serviceaccount` is enabled (only enabled in test for now), run kube-proxy with a mode where it picks up the in-cluster configuration, but override the API server URL to the public hostname. Additionally, use a projected token in this case, because the non-rotated ones would be blocked by our security measures.

### DIFF
--- a/cluster/manifests/kube-proxy/daemonset.yaml
+++ b/cluster/manifests/kube-proxy/daemonset.yaml
@@ -36,6 +36,12 @@ spec:
         - --config=/config/kube-proxy.yaml
         - --v=2
         env:
+{{- if eq .Cluster.ConfigItems.experimental_proxy_use_serviceaccount "true" }}
+        - name: KUBERNETES_SERVICE_HOST
+          value: "{{.Cluster.LocalID}}.{{.Values.hosted_zone}}"
+        - name: KUBERNETES_SERVICE_PORT
+          value: "443"
+{{- end }}
         - name: HOSTNAME_OVERRIDE
           valueFrom:
             fieldRef:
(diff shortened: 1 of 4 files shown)


aryszka:
:+1:

cluster/manifests/kube-proxy/daemonset.yaml
```diff
@@ -33,9 +33,17 @@ spec:
         image: nonexistent.zalan.do/teapot/kube-proxy:fixed
         args:
         - --hostname-override=$(HOSTNAME_OVERRIDE)
+{{- if ne .Cluster.ConfigItems.experimental_proxy_use_serviceaccount "true" }}
         - --config=/config/kube-proxy.yaml
```
mikkeloscar, line 36: Don't we still need the configuration file if running in `serviceaccount` mode?

aermakov-zalando: Good point, I'm not sure where this came from!

aermakov-zalando: Moved to the correct place (the config).

mikkeloscar: This flag is still guarded by an if :)

aermakov-zalando: :(

aermakov-zalando: Now it's fixed!

aermakov-zalando:
Note: this changes how kube-proxy works on the masters (it'll now go via the NLB instead of directly), but I don't think this will be a problem.

mikkeloscar:
:+1:

jonathanbeber:
:-1:

jonathanbeber:
:+1:

PR #4102, opened 2021-03-03
Fix sidecar port scrape

mikkeloscar:
Turns out what I did and argued for in https://github.com/zalando-incubator/kubernetes-on-aws/pull/4086/files#r582626549 didn't actually work for two ports in a single pod :(

This fixes the problem by introducing two almost identical jobs just the port annotation being different. It's not super pretty, but based on our previous discussion and what I found in this long prometheus issue https://github.com/prometheus/prometheus/issues/3756, then I don't have a better idea right now.

WDYT? @aermakov-zalando @jonathanbeber

### DIFF
--- a/cluster/manifests/coredns-local/daemonset-coredns.yaml
+++ b/cluster/manifests/coredns-local/daemonset-coredns.yaml
@@ -27,8 +27,7 @@ spec:
         logging/destination: "{{.Cluster.ConfigItems.log_destination_infra}}"
         prometheus.io/path: /metrics
         prometheus.io/scrape: "true"
-        prometheus.io/port: "9153" # coredns
-        prometheus.io/sidecar-port: "9054" # dnsmasq/unbound
+        prometheus.io/port: "9153"
     spec:
       initContainers:
       - name: ensure-apiserver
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
Might be simpler to just have the annotation-based config for the 99% case of 1 port per pod, and then have additional rules for pods that have more than one container (apiserver has more than 8, for example, and we still want to scrape all of them).

jonathanbeber:
with additional rules, you mean creating new Prometheus' scraping jobs specifically for the apps with more than one container, @aermakov-zalando?

aermakov-zalando:
Yeah, just to avoid this kind of templating mess.

jonathanbeber:
I don't like all the logic inside the template as well and making it explicit sounds reasonable to me if it does not grow too much.

@mikkeloscar WDYT? So far we just have two applications with this "problem" and, please correct me if I'm wrong, `kube-apiserver` was not even on the previous PR, right? Is it still being scraped via a service?

mikkeloscar:
Fine by me, whatever is more understandable. Does it make sense how I added it now?

> @mikkeloscar WDYT? So far we just have two applications with this "problem" and, please correct me if I'm wrong, `kube-apiserver` was not even on the previous PR, right? Is it still being scraped via a service?

apiserver is special because it uses https and needs a token, that's why it's not included in the generic method.

jonathanbeber:
:+1:

aermakov-zalando:
👍

PR #4103, opened 2021-03-03
Update the AMI

aermakov-zalando:
Changes:
 * New spot termination handler (deletes instead of draining, posts events on the pods).


aermakov-zalando:
👍

jrake-revelant:
👍

PR #4104, opened 2021-03-03
Update the RBAC configuration for the spotinst controller

aermakov-zalando:
* Allow it to list CRDs (to stop the log spam)
 * Drop the Cleanup/Apply feature permissions, we don't need any of theses


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4108, opened 2021-03-04
E2E: use an OAuth token instead of a static secret [2/3]

aermakov-zalando:
Instead of using a static secret to connect to the E2E cluster, use an OAuth token.

**Important**: this must be merged separately from #4109, otherwise E2E will fail.


aermakov-zalando:
👍

mikkeloscar:
:+1:

jonathanbeber:
:+1:

PR #4109, opened 2021-03-04
E2E: support OAuth tokens as well [1/3]

aermakov-zalando:
This makes it possible to run E2E with an OAuth token in addition to the static credentials.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4112, opened 2021-03-04
E2E: generate a per-run WORKER_SHARED_SECRET [3/3]

aermakov-zalando:
Since we no longer need it for the actual E2E, just generate a new one when provisioning the cluster.

**Important**: has to be merged separately from #4108 (and the cleanup job has to be updated as well).


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4114, opened 2021-03-04
Add support for spot abstraction

mikkeloscar:
From the user perspective we introduce two pod annotations:

* **RuntimePolicy**: `zalando.org/runtime-policy=allow-spot|require-on-demand` - Pods with this annotation are either allowed to run on spot or must run on on-demand instances.
* **PreventScaleDown**: `zalando.org/prevent-scale-down=true` - Pods with this annotation doesn't get evicted from nodes where the autoscaler would like to clear a node for better packing of pods. (Less disruptions).

This is the first step of rolling out the feature where we only enable it in test clusters for now.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -328,6 +328,32 @@ teapot_admission_controller_topology_spread: optin
 
 teapot_admission_controller_validate_base_images: "false"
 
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4121, opened 2021-03-05
etcd stack: add certificate expiry tags

aermakov-zalando:
We use this for monitoring, so we should set them when we create a new stack.

Related CLM PR: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/444.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4133, opened 2021-03-10
Disable insecure address on kube-apiserver (+ Kubernetes v1.19.9) [4/x]

mikkeloscar:
This disables the insecure address on the kube-apiserver (`--insecure-bind-address`) which will no longer be available in Kubernetes v1.20

In order to do this we migrate everything that was previously talking to the insecure address to the secure address. This means that all components must have "some" identity and corresponding permissions. The table below shows the changes

| component | old: identity/permissions | new: identity/permissions | comment |
| ------------- | ------------- | ------ | -- |
| kube-controller-manager | `system:kube-controller-manager` / per controller service account | `system:kube-controller-manager` / per controller service account  | x |
| kube-scheduler | `<none>` / `system:masters` | `system:kube-scheduler` / `system:kube-scheduler` ClusterRole  | x |
| admission-controller | `<none>` / `system:masters` | `teapot-admission-controller` / `system:masters` | x |
| kubelet (on control plane nodes) | `kubelet` / `system:masters` | `control-plane-kubelet` / `system:masters` | x |

Further more we set `--anonymous-auth=true` which means all unauthenticated calls gets the identity `system:anonymous`. We use this to create an RBAC role that allows access to `/openid/v1/jwks` without credentials which is needed for our AWS IAM OIDC setup.

This also depends on a new AMI where we generate per component kubeconfig files

### TODO

* [x] Migrate NLB to TCP healthcheck (HTTP is no longer possible when port `8080` is disabled) #4138
* [x] Update AMI
* [x] Update admission-controller image
* [x] Update k8s-authnz-webhook image
* [x] Validate that `--anonymous-auth=true` is ok (or find alternative solution)
  * [x] Find existing bindings
  * [x] Prevent creation by users in admission-controller
  * [x] Handle skipper livenessprobe?
* [x] ~Add dedicated role for admission-controller~ This is complicated because it needs to create cluster roles for CRD resources, so it essentially needs full access anyway. Sticking with `system:masters` permissions for now.

**Important**: Don't merge before #4138,#4143,#4164 has been fully rolled out and switch to new NLB has been done!


cluster/node-pools/master-default/userdata.yaml
```diff
@@ -455,13 +441,13 @@ write_files:
             h: Path("/kube-system/healthz")
               -> setPath("/healthz")
               -> disableAccessLog()
-              -> "http://127.0.0.1:8080";
+              -> "https://127.0.0.1:443";
```
linki, line 441: We should be able to remove these then because they were only to bypass authentication with [anonymous auth still disabled](https://github.com/zalando-incubator/kubernetes-on-aws/pull/4007/commits/b15ada51e28b2a634b9a1b41d79b9035880303c6). (See the equal `all` rule below.)

linki: Also you can replace the custom RBAC rules with this: https://github.com/zalando-incubator/kubernetes-on-aws/pull/4007/commits/b15ada51e28b2a634b9a1b41d79b9035880303c6#diff-829c7c7be2a35683dae916778f24edfe9e91ff89fd17d07bc18f25953a0c49b8

There's a predefined rule to allow OIDC discovery access.

linki: Also you can replace the custom RBAC rules with this: https://github.com/zalando-incubator/kubernetes-on-aws/pull/4007/commits/b15ada51e28b2a634b9a1b41d79b9035880303c6#diff-829c7c7be2a35683dae916778f24edfe9e91ff89fd17d07bc18f25953a0c49b8

There's a predefined role to allow OIDC discovery access.

mikkeloscar: Thanks for pointing it out! https://github.com/zalando-incubator/kubernetes-on-aws/pull/4145

aermakov-zalando:
We could create a proper role for the admission controller, just to reduce its permissions a bit. It doesn't read secrets, so it should be ok with readonly + access to its PCS.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4134, opened 2021-03-11
Don't check application label on pod template of stackset

mikkeloscar:
Many users rely on labels from the top level stackset being injected into the pod template so they don't have to define it multiple times within a manifest.

In the previous version of the admission-controller we changed the behavior so it's no longer possible. This drops the application label check on pod template for stacksets again so we don't block users which have valid defined manifests.


jonathanbeber:
:+1:

jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #4138, opened 2021-03-11
Migrate to NLB with TCP healthcheck [1/x]

mikkeloscar:
This sets up migration of the control plane NLB from HTTP healthcheck to port 8080 (insecure port) to TCP 443. This is needed in preparation of #4133 

The way it works:

1. By default deploy a new NLB with the suffix `-tcp`
2. Via the `apiserver_nlb_tcp` config item control the rollout and switch to the new NLB and decommissioning of the old one.

```yaml
# defines temporary rollout of NLB with TCP healthcheck. This assumes
# apiserver_nlb=exclusive. The options are:
#
#   hooked:    New NLB is hooked up to ASG but DNS still points to old NLB
#   active:    New NLB will be fully functional and DNS points to the new NLB
#   promoted:  New NLB will be fully functional and old NLB will be unhooked
#   exclusive: New NLB will be fully functional and old NLB will be removed
#
apiserver_nlb_tcp: "hooked"
```

We have to setup a new NLB because AWS doesn't support changing the healthcheck protocol of a running NLB :facepalm:

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -52,6 +52,7 @@ Resources:
       ToPort: 10250
     Type: 'AWS::EC2::SecurityGroupIngress'
 {{- if ne .Cluster.ConfigItems.apiserver_nlb "disabled" }}
+{{- if ne .Cluster.ConfigItems.apiserver_nlb_tcp "exclusive" }}
   MasterLoadBalancerNLB:
     Type: AWS::ElasticLoadBalancingV2::LoadBalancer
     Properties:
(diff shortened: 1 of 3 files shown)


jonathanbeber:
:+1:

jonathanbeber:
IMHO, that's a lot of templating logic :( Took me some minutes to fully comprehend it, but I don't have any better idea.

szuecs:
I agree with @jonathanbeber . 
Seems fine, but 2 times in row and `if not` statement is not easy to grasp, everything else looks good.
We should run a proper test before rolling out to understand if postgres-operator will see a connection bleep or not.
From the roll out description looks good to me, too.

szuecs:
:+1:

mikkeloscar:
I agree it's a bit convoluted, but it will only be for a short migration phase and then I will drop it all to only have a single NLB.

Postgres operator uses the internal service (kube-proxy iptables) so it's not affected by this change.

PR #4143, opened 2021-03-12
Allow anonymous access to openid endpoint [2/x]

mikkeloscar:
As a prerequisite for #4133 we need to provision an RBAC role that allows access to the openID endpoint which is requested by AWS OIDC in order to validate the AWS IAM credentials.

We can't do it together with #4133 because the role would be applied after the APIserver has been updated.

Note this requires setting `--anonymous-auth=true` on the APIserver (in #4133) which is discouraged in the [CIS Kubernetes Benchmark](https://www.cisecurity.org/benchmark/kubernetes/) unless RBAC is enabled as is the case for us.

(snippet from the benchmark)
image

### DIFF
--- a/cluster/manifests/roles/public-openid-discovery.yaml
+++ b/cluster/manifests/roles/public-openid-discovery.yaml
@@ -1,0 +1,20 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: public-openid-discovery
+rules:
+- nonResourceURLs: ["/openid/v1/jwks", "/.well-known/openid-configuration"]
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

aermakov-zalando:
👍

linki:
IIRC, there's a pre-defined role for this: https://github.com/zalando-incubator/kubernetes-on-aws/pull/4007/commits/b15ada51e28b2a634b9a1b41d79b9035880303c6#diff-829c7c7be2a35683dae916778f24edfe9e91ff89fd17d07bc18f25953a0c49b8 (`system:service-account-issuer-discovery`)

PR #4145, opened 2021-03-15
Use existing role for anonymous OIDC discovery

mikkeloscar:
Update to #4143 

This uses an already existing RBAC role which does the same as what we created manually. Additionally it allows access for both `authenticated` and `unauthenticated` (Thanks for pointing it out @linki).

I will manually drop the one called `public-openid-discovery` as it's only in dev so far.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4146, opened 2021-03-15
Base image check: enable by default

aermakov-zalando:
It's enabled in all clusters now, let's switch to using the default instead.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4147, opened 2021-03-15
Add the missing InfrastructureComponent tag to spotinst nodes

aermakov-zalando:
It's not propagated from the CF stack, so we have to set it manually.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4153, opened 2021-03-16
Add support for STS-based node authentication

aermakov-zalando:
This PR implements (but not enables) proper node auth instead of the hacky shared secret we use currently. The rollout is controlled by a single config item (`node_auth`) that can be set to one of the following:
 * `disabled` (default): only the auth webhook is updated, no other changes
 * `supported`: rolls the masters. Node auth support is enabled in the auth webhook and in the API server configuration, but no other changes are done.
 * `enabled`: rolls the workers. They now use AWS IAM for authentication instead of the shared secret, and the admission controller is reconfigured to disable the node relabeling in preparation for the next step. Shared secret auth still works.
 * `exclusive`: rolls the masters to remove support for the shared secrets and enable `NodeRestriction` plus the additional checks in the admission controller.

We still have a bunch of potential issues with the security side of this, but it's much better than what we have now.

**Important**: `enabled` and above also expect `experimental_proxy_use_serviceaccount` to be enabled. However I'd just enable it by default before going forward with the node auth changes because we haven't had any issues with it.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -574,3 +574,14 @@ enable_control_plane_profiling: "false"
 clm_new_userdata_path: "true"
 
 experimental_proxy_use_serviceaccount: "true"
+
+# Defines the rollout status of the node auth feature. The possible values are:
+#
+#   disabled:   node auth is disabled on the API server side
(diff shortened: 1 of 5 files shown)


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4154, opened 2021-03-16
Enable node auth support by default

aermakov-zalando:
Other than sporadic DescribeInstances requests, this shouldn't have any effects. Let's enable it by default.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4157, opened 2021-03-16
Set the max_pid config item as node label

jonathanbeber:
This commit adds the nodepool config item `pod_max_pids` as a label to
allow users to get the value for each pod in the monitoring.

### DIFF
--- a/cluster/node-pools/master-default/stack.yaml
+++ b/cluster/node-pools/master-default/stack.yaml
@@ -37,6 +37,9 @@ Resources:
       - Key: node.kubernetes.io/role
         PropagateAtLaunch: true
         Value: master
+      - Key: zalando.org/pod-max-pids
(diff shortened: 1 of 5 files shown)


cluster/node-pools/master-default/stack.yaml
```diff
@@ -37,6 +37,9 @@ Resources:
       - Key: node.kubernetes.io/role
         PropagateAtLaunch: true
         Value: master
+      - Key: zalando.org/pod_max_pids
```
aermakov-zalando: `pod-max-pids` maybe? We don't use underscores in other labels.

jonathanbeber: true, good one, will change!

jonathanbeber:
:+1:

szuecs:
:+1:

PR #4158, opened 2021-03-16
E2E: wait for the initial cluster to stabilise

aermakov-zalando:
Because we don't pre-create the worker nodes anymore, it's likely that CLM will apply the second update before CA even starts requesting nodes. This means that updates that only change the workers will not test the rolling part because there would be no nodes to roll yet. Fix by waiting for the cluster to become ready before we proceed with the update (and then wait again after the update).
Also fixes a minor issue in wait-for-update.py where it didn't print in realtime because of Python output buffering.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #4161, opened 2021-03-17
kube-proxy: use SA by default

aermakov-zalando:
We didn't have any issues in the test clusters, let's enable it by default. Config item is still there to allow for a quick rollback, I'd like to wait another couple of weeks before dropping it fully.


linki:
:+1:

aermakov-zalando:
👍

PR #4163, opened 2021-03-17
Drop privileged PSP support for user workloads

mikkeloscar:
We migrated all clusters away from using privileged PSP.

This PR:

 1. Removes the config_item that is no longer needed.
 2. Updates the authnz webhook to not support privileged PSP for legacy `operator` service account.


jonathanbeber:
:+1:

szuecs:
:+1:

PR #4164, opened 2021-03-18
Promote TCP NLB for e2e [3/X]

mikkeloscar:
This has already been done for all clusters via config items. However in order to drop the old NLB also in e2e, we need to promote it first.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4167, opened 2021-03-19
Kubernetes v1.19.9

aermakov-zalando:
* Update the AMI:
  * Kubernetes v1.19.9
  * Update spot termination handler to support EKS-style auth
* Increase the instance sizes so the rolling update is faster


aermakov-zalando:
This needs to be rebased on top of #4133.

aermakov-zalando:
This AMI update is included in #4133, closing.

PR #4169, opened 2021-03-19
Add an E2E test for the spot termination handler

aermakov-zalando:
This is now fairly easy to test, and we'll know when it's broken again (see #4167 for example).

### DIFF
--- a/test/e2e/cluster_config.sh
+++ b/test/e2e/cluster_config.sh
@@ -85,6 +85,15 @@ clusters:
     profile: worker-default
     min_size: 0
     max_size: 21
+  - discount_strategy: spot
+    instance_types: ["m4.large", "m5.large", "m5.xlarge", "m4.xlarge"]
+    min_size: 0
+    max_size: 3
+    profile: worker-default
+    name: worker-spot-termination-handler
+    config_items:
+      labels: dedicated=spot-termination-handler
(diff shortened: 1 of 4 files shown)


aermakov-zalando:
👍

test/e2e/cluster_config.sh
```diff
@@ -85,6 +85,15 @@ clusters:
     profile: worker-default
     min_size: 0
     max_size: 21
+  - discount_strategy: spot
+    instance_types: ["m4.large", "m5.large", "m5.xlarge", "m4.xlarge"]
```
jonathanbeber, line 89: does it respect order? If so why we go with m4 first instead of m5?

aermakov-zalando: IIRC it's whatever has the least potential for disruption, I'm not sure if order matters. It also doesn't matter anyway, the node would run a pause container and a small shell script and then get cordoned.

jonathanbeber: Got it. Ofc it does not matter for the process, just thinking that `m5`s are cheaper.

jonathanbeber:
:+1:

test/e2e/cluster_config.sh
jonathanbeber: but we can't be sure for spots anyway, just ignore.

PR #4176, opened 2021-03-22
Increase the default instance sizes to avoid creating too many instances

aermakov-zalando:
We run a lot of things by default, and in practice never get the smaller instances anyway. Avoid scale downs and extra node rotations by creating bigger instances from the start.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4182, opened 2021-03-22
Use HTTPS for Control Plane LB health checks

mikkeloscar:
In #4138 we introduced a way to migrate to TCP based health check (instead of `HTTP`). Turns out I read the docs wrong :facepalm: and we can use `HTTPS` instead which will avoid error logs on the skipper-proxy.

```
2021/03/22 13:15:22 http: TLS handshake error from 172.31.7.147:2250: EOF
2021/03/22 13:15:22 http: TLS handshake error from 172.31.3.130:36167: EOF
2021/03/22 13:15:22 http: TLS handshake error from 172.31.16.106:27014: EOF
 ```

This implements the same (convoluted) migration path as in #4138 but from TCP to HTTPS instead of HTTP to TCP. As always we need to create a new LB and switch DNS as you can't modify the health check protocol of a running NLB :man_facepalming: 

```yaml
# defines temporary rollout of NLB with HTTPS healthcheck. This assumes
# apiserver_nlb=exclusive. The options are:
#
#   hooked:    New NLB is hooked up to ASG but DNS still points to old NLB
#   active:    New NLB will be fully functional and DNS points to the new NLB
#   promoted:  New NLB will be fully functional and old NLB will be unhooked
#   exclusive: New NLB will be fully functional and old NLB will be removed
#
apiserver_nlb_https: "hooked"
```

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -52,6 +52,7 @@ Resources:
       ToPort: 10250
     Type: 'AWS::EC2::SecurityGroupIngress'
 {{- if ne .Cluster.ConfigItems.apiserver_nlb "disabled" }}
+{{- if ne .Cluster.ConfigItems.apiserver_nlb_https "exclusive" }}
   MasterLoadBalancerNLBTCP:
     Type: AWS::ElasticLoadBalancingV2::LoadBalancer
     Properties:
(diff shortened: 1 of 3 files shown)


aermakov-zalando:
It seems like it can be done simpler, or am I missing something? Instead of creating a whole new NLB, just rename the target group resource and update the reference. This would cause CF to create a new TG (which we'd do anyway), update the NLB to use it, and then delete the old one.

mikkeloscar:
We depend on the target group in the master node pool stack via CF import. So we can't just drop the target group in the main stack unfortunately

aermakov-zalando:
Argh, I see.

aermakov-zalando:
👍

mikkeloscar:
This depends on #4133 being rolled out otherwise e2e will fail because I reused the resource names which was dropped in #4133. Alternatively I can rename the resources so they won't overlap which the previous ones.

aermakov-zalando:
Probably easier to wait for #4133.

mikkeloscar:
:+1:

PR #4183, opened 2021-03-22
Restart ExternalDNS after update just in case it's stuck

linki:
Sometimes ExternalDNS gets stuck during master-rolling updates.

As a short-term solution restart ExternalDNS as part of the final apply phase to unblock it just in case it got stuck.


linki:
:+1:

aermakov-zalando:
👍

PR #4185, opened 2021-03-23
Add basic support for Okta auth

aermakov-zalando:
This PR implements very basic support for Okta auth. If `okta_auth_enabled` is set to `true`, the API server is configured with an OIDC provider at `okta_auth_issuer_url`, using `okta_auth_client_id` for the client ID (defaults to `kubernetes.cluster.<alias>`, but can be customised for things like e2e/pet clusters). Users and groups will be prefixed with `okta:`, and the bindings for ReadOnly/PowerUser/Collaborator would include the corresponding Okta groups (`common/read-only`, `common/engineer` and `common/collaborator`).

The auth webhook and the admission controller are updated to add support for an extra administrator role in addition to `system:masters` (`okta:common/administrator`).

### DIFF
--- a/cluster/manifests/roles/poweruser-binding.yaml
+++ b/cluster/manifests/roles/poweruser-binding.yaml
@@ -21,3 +21,11 @@ subjects:
 - kind: Group
   name: Emergency
   apiGroup: rbac.authorization.k8s.io
+{{- if eq .Cluster.ConfigItems.okta_auth_enabled "true" }}
+- kind: Group
+  name: "okta:common/engineer"
+  apiGroup: rbac.authorization.k8s.io
+- kind: Group
+  name: "okta:common/collaborator"
+  apiGroup: rbac.authorization.k8s.io
+{{- end }}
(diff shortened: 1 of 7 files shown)


cluster/manifests/roles/poweruser-binding.yaml
```diff
@@ -21,3 +21,11 @@ subjects:
 - kind: Group
   name: Emergency
   apiGroup: rbac.authorization.k8s.io
+{{- if eq .Cluster.ConfigItems.okta_auth_enabled "true" }}
+- kind: Group
+  name: "okta:common/engineer"
```
aermakov-zalando, line 26: Is this clear enough? Or should we rename it?

aermakov-zalando:
Please don't merge until 2FA is configured properly on our Okta instance.

cluster/manifests/roles/poweruser-binding.yaml
jonathanbeber: I believe we already discussed it in the doc and we can live with that.

aermakov-zalando: I don't have any better proposals, so let's keep it like this.

jrake-revelant:
👍

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4187, opened 2021-03-23
Add liveness probe to unbound DNS cache

mikkeloscar:
This adds a `livenessProbe` to the unbound DNS cache container which uses `dig` to check if the internal kubernetes apiserver service endpoint is resolvable.

The purpose of this is to detect if `unbound` gets stuck for some reason such that it can be killed and re-created.

Depends on a new `unbound` image with `dig` (internal PR).

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -458,6 +458,8 @@ dns_cache: "dnsmasq"
 dns_cache: "unbound"
 {{ end }}
 
+expirimental_dns_unbound_liveness_probe: "true"
+
 # DNS container resources
 dns_dnsmasq_cpu: "100m"
 dns_dnsmasq_mem: "50Mi"
--- a/cluster/manifests/coredns-local/daemonset-coredns.yaml
+++ b/cluster/manifests/coredns-local/daemonset-coredns.yaml
@@ -43,7 +43,11 @@ spec:
       containers:
 {{ if eq .Cluster.ConfigItems.dns_cache "unbound" }}
       - name: unbound
-        image: registry.opensource.zalan.do/teapot/unbound:1.13.1
+        image: registry.opensource.zalan.do/teapot/unbound:1.13.1-1
+        args:
+        - -d
+        - -c
+        - /etc/unbound/unbound.conf
         ports:
         - containerPort: 53
           name: dns-udp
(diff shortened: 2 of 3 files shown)


cluster/manifests/coredns-local/daemonset-coredns.yaml
```diff
@@ -51,6 +51,18 @@ spec:
...
           name: dns-tcp
           protocol: TCP
+{{- if eq .Cluster.ConfigItems.dns_unbound_liveness_probe "true" }}
+        livenessProbe:
+          exec:
+            command:
+            - dig
+            - "+short"
+            - "@127.0.0.1"
+            - "kubernetes.default.svc.cluster.local"
```
aermakov-zalando, line 65: I'd use an external hostname here.

cluster/manifests/coredns-local/daemonset-coredns.yaml
```diff
@@ -51,6 +51,18 @@ spec:
...
+          initialDelaySeconds: 5
```
aermakov-zalando: Let's increase this a bit? This depends on CoreDNS being up, which can take more than 5 seconds if it needs to be pulled. It's also very unlikely that it'll get stuck in the first 5 seconds.

cluster/config-defaults.yaml
```diff
@@ -458,6 +458,8 @@ dns_cache: "dnsmasq"
 dns_cache: "unbound"
 {{ end }}
 
+dns_unbound_liveness_probe: "true"
```
aermakov-zalando: Mark it as `experimental_`? Not sure why we'd want to disable it, if it works.

aermakov-zalando: Or just make it unconditional.

cluster/manifests/coredns-local/daemonset-coredns.yaml
aermakov-zalando: On the other hand, this will cause all containers to restart if there's a hiccup with AWS DNS, so maybe an internal one would work better.

mikkeloscar: Yeah, this is also what I worry about. We just want to test that the unbound cache is working, not the whole DNS infra.

cluster/manifests/coredns-local/daemonset-coredns.yaml
mikkeloscar: Increased to 60, similar to coredns livenessProbe

cluster/config-defaults.yaml
mikkeloscar: Added `experimental_`. I just want to be able to disable it in case something didn't work as expected after we roll it out to different channels. At least for a while.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4199, opened 2021-03-29
Cleanup the metrics for the apiserver components [1/2]

aermakov-zalando:
The current setup for the API server components leaves a lot to be desired:
 * metrics for some of them aren't scraped at all
 * we have to open the component's port to scrape it, even though they have other endpoints that should not be accessible
 * these ports are opened to the whole VPC instead of only the workers (or even something with auth).

This PR improves that. Instead of continuing with the current approach, we instead add a new skipper container that proxies the metrics of other API server components, and then configure them as individual Prometheus targets.

In addition to that, it also updates all the Skipper images in the API server to something that's not ancient, and fixes the TLS configuration for the API server target.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -403,6 +403,17 @@ Resources:
         - Key: 'kubernetes.io/cluster/{{.Cluster.ID}}'
           Value: owned
     Type: 'AWS::EC2::SecurityGroupIngress'
+  MasterSecurityGroupIngressFromWorkerToSkipperMetrics:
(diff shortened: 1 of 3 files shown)


aermakov-zalando:
Note: this doesn't cover the two Skipper sidecars because of the potential impact on the metrics. We can add them a bit later once we fix the various ZMON checks.

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4201, opened 2021-03-29
Master SG: drop the ingress rules that were only used for metrics [2/2]

aermakov-zalando:
These are now exposed via a dedicated Skipper (see #4199), so we can drop the ingress rules.

**Warning**: #4199 should be fully applied first, otherwise some of the metrics will be unavailable until the update is finished.


aermakov-zalando:
👍

gargravarr:
:+1:

PR #4204, opened 2021-03-31
Revert "Worker SG: allow UDP as well"

aermakov-zalando:
We don't actually have any users for this, and we've agreed that we could do better even with the TCP setup. Let's revert this for now.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4216, opened 2021-04-06
Make it possible to disable the egress controller

aermakov-zalando:
We have a cluster where it won't be needed, so let's make it configurable.


aermakov-zalando:
👍

gargravarr:
:+1:

PR #4217, opened 2021-04-06
Drop Muaaz and Henning from the approvers

aermakov-zalando:
Truncate this to only the active team members.


aermakov-zalando:
👍

gargravarr:
:+1:

PR #4221, opened 2021-04-09
Update the admission controller

aermakov-zalando:
Changes:
 * Treat the Okta administrators as cluster administrators for the rolebinding admitter
 * Drop the node relabeling feature, since it was moved to the node-ready-controller
 * Improve the base image error messages a bit


aermakov-zalando:
👍

jrake-revelant:
👍

PR #4225, opened 2021-04-13
Update kube-state-metrics to v2.0.0

mikkeloscar:
https://kubernetes.io/blog/2021/04/13/kube-state-metrics-v-2-0/

https://github.com/kubernetes/kube-state-metrics/releases/tag/v2.0.0

In this version it doesn't include all resource labels by default as Prometheus metric labels. I have limited it to `application` and `component` for pods and `topology.kubernetes.io/zone` for nodes as those are the only labels we use in our ZMON checks. This will reduce the cardinality in prometheus!


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4237, opened 2021-04-19
Promote https health check for APIserver NLB

mikkeloscar:
The migration from `hooked` -> `exclusive` is already done for the active clusters. This change is mostly to allow smooth transition for e2e. After this we can drop the whole config item.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4257, opened 2021-04-28
Enable node_auth=exclusive by default

aermakov-zalando:
We're close to enabling it fully, so update the defaults to avoid having to update new clusters. I've manually set it to `supported` in the clusters that haven't been updated yet.


mikkeloscar:
:+1:

aermakov-zalando:
Let's see if e2e works, although I might have to do it as 2 separate PRs. :/

aermakov-zalando:
This unfortunately won't work.

PR #4259, opened 2021-04-28
node_auth: supported->enabled [1/2]

aermakov-zalando:
This is a replacement for #4257. Unfortunately, going directly from supported to exclusive doesn't work because the workers lose connection to the masters and the upgrade breaks down. We have to roll it out in two steps if we want e2e to work.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4260, opened 2021-04-28
node_auth: enabled->exclusive [2/2]

aermakov-zalando:
This is a replacement for #4257. Unfortunately, going directly from supported to exclusive doesn't work because the workers lose connection to the masters and the upgrade breaks down. We have to roll it out in two steps if we want e2e to work.

Do not merge at the same time as #4259!


jonathanbeber:
:+1:

aermakov-zalando:
👍

PR #4267, opened 2021-04-30
cdp-controller: drop the old role binding

aermakov-zalando:
This is a follow-up to https://github.com/zalando-incubator/kubernetes-on-aws/pull/2891. We somehow forgot to drop the old binding after CDP has migrated, let's drop it now.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4272, opened 2021-05-03
Reduce the deregistration delay

aermakov-zalando:
The deregistration delay setting for the LBs managed by the ingress controller doesn't really make a lot of sense. By the time we're terminating the node, there are no pods running on it, so the only thing we achieve with the delay is an extra 5 minute wait interval on every node termination. This slows down rolling upgrades and downscaling a lot, especially in large clusters.

Unfortunately I don't think it can be disabled (because the default is 5 minutes), but 10 seconds is low enough to not matter.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4273, opened 2021-05-03
Allow to scale skipper based on time

jonathanbeber:
This commit add an external metric to the skipper hpa when the config
item `skipper_time_based_scaling_enabled` is set to true. It allow us to
define zmon checks for skipper scaling based on time.

### DIFF
--- a/cluster/manifests/skipper/hpa.yaml
+++ b/cluster/manifests/skipper/hpa.yaml
@@ -25,6 +25,22 @@ spec:
       target:
         type: Utilization
         averageUtilization: {{ .ConfigItems.skipper_ingress_target_average_utilization_memory }}
+{{ if ne .ConfigItems.skipper_time_based_scaling_check_id "" }}
+  - type: External
+    external:
+      target:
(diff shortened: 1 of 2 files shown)


cluster/manifests/skipper/hpa.yaml
```diff
@@ -25,6 +25,22 @@ spec:
...
+    external:
+      target:
+        averageValue: {{ .ConfigItems.skipper_time_based_scaling_target }}
+        type: AverageValue
+      metric:
+        name: zmon-check
+        selector:
+          matchLabels:
+            type: "zmon"
+            check-id: {{ .ConfigItems.skipper_time_based_scaling_check_id }}
```
aermakov-zalando: This should be quoted I think, or does it work even if it's an int?

jonathanbeber: good point. I pushed a change also removing the `skipper_time_based_scaling_enabled` as we talked.

aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #4305, opened 2021-05-17
Add cronjob-monitor

aermakov-zalando:
This is a workaround for https://github.com/kubernetes/kubernetes/issues/75674, a small controller that tracks the last successful completion time in an annotation. Will be dropped once (if) it's actually implemented in Kubernetes.


jonathanbeber:
:+1:

aermakov-zalando:
👍

PR #4307, opened 2021-05-17
Enable prevent scale down annotation in all clusters

aermakov-zalando:
We forgot to follow up on this, let's just enable it everywhere.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4312, opened 2021-05-20
Upgrade VPA crd to v1

mikkeloscar:
Updates the VPA CRD to v1, available since v1.16 (v1beta1 will be removed in v1.22). This adds proper validation and descriptions which are helpful for `kubectl explain`

I have tested all existing VPAs and VPA checkpointers against this validation and they all validate with it.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4243, opened 2021-04-19
[2/3] Reduce skipper metrics dimension

jonathanbeber:
This commit removes the response http status code and the http method
from the skipper_serve_host_duration_seconds_bucket metric. It reduces
the number of metrics generated by skipper ingress. More details in
https://github.com/zalando/skipper/pull/1755.

Also, it  configure the prometheus jobs to use the new metric enabled
during #4235.

PR #4315, opened 2021-05-25
Remove skipper prometheus job

jonathanbeber:
The job skipper_serve_host_duration_2xx was not collecting any data
since #4243, once the requested metric does not have the dimension
required.

This commit is a cleanup removing the job completely. The job is no
longer used in our monitoring systems.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #4316, opened 2021-05-25
kube-metrics-adapter CRDs for time-based scaling

jonathanbeber:
This commit updates the kube-metrics-adapter version, to include the new
feature of time-based scaling. It installs the new CRDs and starts the
server with a defined config flag. It also updates the RBAC permissions
to allow the server to get, list and watch the new CRDs.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -79,6 +79,10 @@ skipper_redis_write_timeout: "25ms"
 # skipper api GW features
 enable_apimonitoring: "true"                       # TODO(sszuecs): cleanup candidate to reduce amount of branches in deployment
 
+
+# Enable the kube-metrics-adapter time-based metrics.
+enable_scaling_schedule_metrics: "true"
+
 # skipper east-west feature
 # enable_skipper_eastwest is the legacy feature gate for the automatic
 # ingress.cluster.local addresses created by skipper.
--- a/cluster/manifests/kube-metrics-adapter/01-rbac.yaml
+++ b/cluster/manifests/kube-metrics-adapter/01-rbac.yaml
@@ -81,6 +81,15 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - zalando.org
+  resources:
+  - clusterscalingschedules
+  - scalingschedules
+  verbs:
+  - get
+  - list
+  - watch
 ---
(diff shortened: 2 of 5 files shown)


jonathanbeber:
@mikkeloscar could you take a look, please? 

This PR is not what I intended, but I got into a situation a bit hard. I want to include e2e tests for that. I also want to make sure everything is working so we can make a proper release in kube-metrics-adapter. 

The kube-metrics-adapter generated clients are in v1.20.5 :confused:. It makes it impossible (at least for me) to make the e2e compile with the 1.19. It's also not an option to upgrade the e2e k8s version now (I think). 

I thought about downgrading the version in kube-metrics-adapter, but not sure that's what we want. At the same time, I'm not sure if what we want is to have yet another dependency with the k8s version strictly attached to this repository, as it's the case for the stackset controller. 

I also thought about creating the `ScalingSchedule` resources without the clients, with a `yaml` file or something like that, since we don't get or watch the resources in the tests, we just interact with the HPA. The `ScalingSchedule` resources just have to be there. Being in the `apply` folder is not an option because I want to interact with the schedules inside it before creation, just to set the time to a few minutes from "now".

This PR is to start rolling the feature, at least in a silent manner, it's not like I'm gonna make an announcement or start using it to all the applications, but at least it's rolled out for tests. If you think it's too risky and do not agree with it I'm ok on waiting and finding another solution.

A later PR will come e2e tests based on our discussion here. 

Let me know what you think and thanks in advance!

mikkeloscar:
First of all, I think it's fine to roll this forward without e2e tests straight away. We can still test it in different ways.

For the e2e tests we can IMO do two things:
1. We roll back to 1.19 in kube-metrics-adapter (I don't believe we really need 1.20, apart from: https://github.com/zalando-incubator/kube-metrics-adapter/pull/326, but that we can do separately with the 1.20 rollout)
2. Considering we are soon to move forward with 1.20 we can also just implement the e2e tests there. They are IMO not strictly needed for the feature to be ready but helpful for us making changes to the metrics-adapter and/or config going forward.

jonathanbeber:
Thanks, for reviewing. I think we should go with the second option. Let's merge this one and I'll work on tests and making a release for kube-metrics-adapter with 1.20.

jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #4318, opened 2021-05-27
set write timeout to 0, Go http server default

aryszka:
needs to be explicitly set to 0 because skipper has a 60s default


szuecs:
:+1:

aryszka:
:+1:

PR #4326, opened 2021-05-31
Drop node_auth and worker_shared_secret from the manifests [1/2]

aermakov-zalando:
We don't plan on reverting this, so let's drop the feature flag.


aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #4328, opened 2021-06-01
Downgrade back to v1.19.10

aermakov-zalando:
v1.19.11 has a bug in kubelet that causes panics on port-forwards, let's downgrade until it's fixed. To avoid rolling the nodes again, we downgrade the whole AMI.


jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #4349, opened 2021-06-07
Downgrade to Kubernetes v1.20.6

mikkeloscar:
Downgrade to Kubernetes v1.20.6 to avoid kubelet panics bug introduced in v1.20.7 (https://github.com/kubernetes/kubernetes/issues/102480)


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4360, opened 2021-06-09
Update CLC

aermakov-zalando:
The existing version uses an ancient client-go that can't post events if it talks to a 1.20 API server.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4374, opened 2021-06-16
Make it possible to disable exec probe timeout

mikkeloscar:
The `ExecProbeTimeout` feature gate was introduced in Kubernetes v1.20 and is enabled by default. It fixes timeouts set for `exec` probes. This change could have a side-effect where existing exec probes start to fail if they were never working as expected in the first place.

This let's us disable the feature per clusters (or node pool) by settings `enable_exec_probe_timeout=false`. Note that this will roll the nodes as it's a setting on the `kubelet` config.

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md#no-really-you-must-read-this-before-you-upgrade

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -607,3 +607,7 @@ deploy_allow_lakeformation: "false"
 deploy_allow_ram: "false"
 
 experimental_nlb_alpn_h2_enabled: "true"
+
+# Enable/Disable ExecProbeTimeout (default in v1.20)
+# Can be disabled in case some workloads depends on the old behavior.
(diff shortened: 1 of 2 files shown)


jonathanbeber:
:+1:

aermakov-zalando:
`exec_probe_timeout_enabled`? Just to match the style we use in other boolean config items.

mikkeloscar:
@aermakov-zalando I think we have a mix of different conventions.. I have changed this one for now.

aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #4375, opened 2021-06-16
Disable unbound until further tests

jonathanbeber:
Most probably due to 1.20 [fix on the kubelet exec probe timeouts][0],
unbound is constantly restarting on different clusters. Until further
investigation and decision on the dns cache topic, let's disable unbound
by default. It can still be enabled by setting `dns_cache` to `unbound`.

[0]: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md#no-really-you-must-read-this-before-you-upgrade


mikkeloscar:
:+1:

jonathanbeber:
:+1:

PR #4385, opened 2021-06-18
egress controller: support non-default VPCs

aermakov-zalando:
Automatic detection doesn't work, but we already have the VPC ID anyway.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4390, opened 2021-06-22
Fix the bucket policy when DS is disabled

aermakov-zalando:
The roles don't exist if the deployment is not enabled, so the bucket policy needs to adjusted accordingly.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4392, opened 2021-06-23
Remove spotio support [2/2]

mikkeloscar:
Drop all spot.io support as we will no longer use it.

Should be merged after #4393 is fully rolled out.


aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #4393, opened 2021-06-23
Remove spotio node pool from e2e [1/2]

mikkeloscar:
Prepare for dropping all spot.io support (#4392)

First we need to drop the node pool created as part of e2e.


cluster/manifests/e2e-resources/pool-reserve.yaml
```diff
@@ -1,5 +1,5 @@
 {{ if eq .Environment "e2e" }}
-{{ range $pool := split "default-worker-splitaz,default-worker,worker-limit-az,worker-instance-storage,default-worker-spotio,worker-node-tests" "," }}
+{{ range $pool := split "default-worker-splitaz,default-worker,worker-limit-az,worker-instance-storage,worker-node-tests" "," }}
```
aermakov-zalando, line 2: I think you'll need a deletion as well, otherwise the update would fail.

mikkeloscar:
:+1:

cluster/manifests/e2e-resources/pool-reserve.yaml
mikkeloscar: Thanks! I added this and it's working now!

aermakov-zalando:
👍

PR #4396, opened 2021-06-23
Remove affinity configs from skipper

jonathanbeber:
Skipper is already running as host network and host port, so we don't
need to rely on affinity to spread the pods. At the same time we
distrust cluster autoscaler is taking bad decisions when launching new
nodes based on that.

This commit removes the affinity configs currently configured for
skipper.


jonathanbeber:
:+1:

aermakov-zalando:
Let's test it first? I'm not 100% sure that CA would work correctly.

aermakov-zalando:
Seems to be working fine.

aermakov-zalando:
👍

PR #4410, opened 2021-06-30
Update deployment service roles

aermakov-zalando:
This is needed to support the new probe result cache in the status service.


aermakov-zalando:
👍

gargravarr:
:+1:

PR #4423, opened 2021-07-06
remove extra char in kairosDB URL

jonathanbeber:
This extra char makes the query return ` unexpected response code: 404"`


gargravarr:
:+1:

jonathanbeber:
:+1:

PR #4452, opened 2021-07-09
Node auth: tweak the rate limits

aermakov-zalando:
We have some clusters that scale up in chunks of 80+ nodes at a time. The node auth handler in our auth webhook defaults to a rate limit of 0.5 'DescribeInstances' which is extremely conservative and something we've set to avoid overloading AWS. Since these calls should be cached 99.9% of the time anyway, let's just increase the defaults to a much higher value and let the AWS throttling to rate limit us.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4462, opened 2021-07-15
Allow customizing the kube-scheduler image

aermakov-zalando:
This is a temporary setup that will be migrated to a better one slightly later, right now I just want to test it in a couple of clusters. This allows customising the kube-scheduler image with an experimental config item.


aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #4474, opened 2021-07-20
use the library version and update patch version to latest available

szuecs:

### DIFF
--- a/cluster/manifests/skipper/skipper-redis.yaml
+++ b/cluster/manifests/skipper/skipper-redis.yaml
@@ -44,6 +44,10 @@ spec:
       containers:
       - image: registry.opensource.zalan.do/library/redis-6-alpine:6-alpine-20210712
         name: skipper-ingress-redis
+        args:
+        - /usr/local/bin/docker-entrypoint.sh
+        - --save
+        - ""
         ports:
         - containerPort: 6379
           protocol: TCP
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

cluster/manifests/skipper/skipper-redis.yaml
```diff
@@ -42,7 +42,7 @@ spec:
                 - skipper-ingress-redis
       priorityClassName: "{{ .Cluster.ConfigItems.system_priority_class }}"
       containers:
-      - image: registry.opensource.zalan.do/zmon/redis:6-master-12
+      - image: registry.opensource.zalan.do/library/redis-6-alpine:6-alpine-20210712
         name: skipper-ingress-redis
```
mikkeloscar, line 46: The zmon image runs with the flag `--save ""` which to my understanding means to disable saving to disk: https://raw.githubusercontent.com/redis/redis/6.0/redis.conf

For this reason we probably want to explicitly set this flag via `args` to run redis in the same way with the new image.

szuecs: where do you see that they use this flag?
Is it in the script they have in the docker container? 
I could not easily cat it at runtime so I thought this is fine. Honestly quite unreasonable to have "library" with completely 0 information about the defaults.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #4477, opened 2021-07-21
Upgrade metrics-server to v0.5.0

mikkeloscar:
https://github.com/kubernetes-sigs/metrics-server/releases/tag/v0.5.0

### DIFF
--- a/cluster/manifests/metrics-server/deployment.yaml
+++ b/cluster/manifests/metrics-server/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: metrics-server
-    version: v0.4.2
+    version: v0.5.0
 spec:
   replicas: 1
   selector:
@@ -16,7 +16,7 @@ spec:
       name: metrics-server
       labels:
         application: metrics-server
-        version: v0.4.2
+        version: v0.5.0
       annotations:
         logging/destination: "{{.Cluster.ConfigItems.log_destination_infra}}"
     spec:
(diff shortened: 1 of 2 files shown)


jonathanbeber:
:+1:

cluster/manifests/metrics-server/deployment.yaml
```diff
@@ -27,11 +27,12 @@ spec:
       serviceAccountName: metrics-server
       containers:
       - name: metrics-server
-        image: registry.opensource.zalan.do/teapot/metrics-server:v0.4.2
+        image: registry.opensource.zalan.do/teapot/metrics-server:v0.5.0
         args:
         - --cert-dir=/tmp
         - --secure-port=4443
         - --kubelet-use-node-status-port
+        - --metric-resolution=15s
```
jonathanbeber: IIUIC the default in 0.4.x [is 60s](https://github.com/kubernetes-sigs/metrics-server/blob/release-0.4/FAQ.md#how-often-metrics-are-scraped), do we have an idea of the impact (or want to assess) of changing it?

I also don't understand why in the [release notes](https://github.com/kubernetes-sigs/metrics-server/releases/tag/v0.5.0) the default is mentioned as 15s if the FAQ [is still saying 60s](https://github.com/kubernetes-sigs/metrics-server/blob/release-0.5/FAQ.md#how-often-metrics-are-scraped) and so does [the code](https://github.com/kubernetes-sigs/metrics-server/blob/release-0.5/cmd/metrics-server/app/options/options.go#L83).

mikkeloscar: They probably refer to this as the "default": https://github.com/kubernetes-sigs/metrics-server/blob/04e6101884fd9633c545369bb0d7d1778b2a1fbc/manifests/base/deployment.yaml#L28 (assuming you use their manifests directly)

The release notes suggests it's possible to increase the resolution because of performance improvements, but we could make it configurable via a config-item so we have the option to increase if we see issues?

jonathanbeber: Yes, that seems a good solution. We should be able to recognize its performance in alpha too.

jonathanbeber:
:+1:

cluster/manifests/metrics-server/deployment.yaml
mikkeloscar: Added

jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #4482, opened 2021-07-22
added telemetry_infrastructure_logs_bucket variable

kostz:
https://github.bus.zalan.do/logging/logging-platform/issues/545

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -1576,16 +1576,16 @@ Resources:
             Statement:
             - Action: ["s3:ListBucket"]
               Effect: "Allow"
-              Resource: 'arn:aws:s3:::{{.Cluster.ConfigItems.telemetry_infrastructure_logs_bucket}}'
+              Resource: 'arn:aws:s3:::{{.Cluster.ConfigItems.logging_infrastructure_s3_bucket}}'
             - Action: ["s3:GetObject"]
               Effect: "Allow"
-              Resource: 'arn:aws:s3:::{{.Cluster.ConfigItems.telemetry_infrastructure_logs_bucket}}/*'
+              Resource: 'arn:aws:s3:::{{.Cluster.ConfigItems.logging_infrastructure_s3_bucket}}/*'
             - Action: ["s3:PutObject"]
               Effect: "Allow"
-              Resource: 'arn:aws:s3:::{{.Cluster.ConfigItems.telemetry_infrastructure_logs_bucket}}/*'
+              Resource: 'arn:aws:s3:::{{.Cluster.ConfigItems.logging_infrastructure_s3_bucket}}/*'
             - Action: [ "s3:PutObjectAcl" ]
               Effect: "Allow"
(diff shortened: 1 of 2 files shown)


kostz:
👍

aermakov-zalando:
Normally we use `logging_` for the config items manage by your team, which would then allow you to change them in Cluster Registry. Do you want to use a different prefix from now on?

kostz:
> Normally we use `logging_` for the config items manage by your team, which would then allow you to change them in Cluster Registry. Do you want to use a different prefix from now on?

that's a tricky thing since from one hand `logging` team was recently renamed to `telemetry-infrastructure`, but from other hand I agree that in this case maybe it's better to support legacy conventions with `logging_` prefix, we'll discuss this internally and come back soon

anyway am I correct that it's only naming and nothing really relies on that names, like write policy, any filters, etc, or there's something that I'm not aware of?

aermakov-zalando:
As I've said in the comment, `logging_*` keys are available for your team to modify in Cluster Registry (along with `zmon_`, `tracing_` and `scalyr_`). If you want a new prefix, we will need to add it to the list as well.

mikkeloscar:
This needs to be rebased on `dev`

cluster/cluster.yaml
```diff
@@ -1569,6 +1569,24 @@ Resources:
...
+            - Action: ["s3:ListBucket"]
+              Effect: "Allow"
+              Resource: 'arn:aws:s3:::{{.Cluster.ConfigItems.logging_infrastructure_s3_bucket}}'
+            - Action: ["s3:GetObject"]
+              Effect: "Allow"
+              Resource: 'arn:aws:s3:::{{.Cluster.ConfigItems.logging_infrastructure_s3_bucket}}/*'
+            - Action: ["s3:PutObject"]
+              Effect: "Allow"
+              Resource: 'arn:aws:s3:::{{.Cluster.ConfigItems.logging_infrastructure_s3_bucket}}/*'
+            - Action: [ "s3:PutObjectAcl" ]
```
aermakov-zalando: Can we merge these into the same statement?

cluster/cluster.yaml
```diff
@@ -1569,6 +1569,24 @@ Resources:
             - Action: ["s3:PutObject"]
               Effect: "Allow"
               Resource: 'arn:aws:s3:::{{.Cluster.ConfigItems.logging_s3_bucket}}/*'
+{{- if ne .Cluster.ConfigItems.telemetry_infrastructure_logs_bucket "" }}
```
aermakov-zalando: Should be renamed here as well.

kostz: yes, that was corrected

cluster/cluster.yaml
kostz: merged

kostz:
👍

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4483, opened 2021-07-22
Remove all artificates related to old image policy webhook

linki:
Removes image policy webhook and uses admission-controller instead. Depends on https://github.com/zalando-incubator/kubernetes-on-aws/pull/4503 so that we keep the image policy related e2e tests.


linki:
I think we have that huge problem that `image-policy-webhook` currently validates images for compliance in our e2e tests. For instance, if someone configures a system image that is not trusted, then it will be visible in e2e tests clusters.

If we completely rely on admission-controller+docker-meta we cannot make those checks in e2e nor in dev. It will only be visible in production when it's already too late.

@aermakov-zalando Any thoughts on this? I know you are aware but I don't know how to best tackle that once we want to get rid of image-policy-webhook completely.

aermakov-zalando:
Definitely, but I'm not sure if we want do anything here, or if this is something that should be handled by team Automata. I would obviously prefer the latter, but please bring it up with them in the chat.

linki:
👍

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -496,10 +483,6 @@ write_files:
               -> disableAccessLog()
               -> setPath("/metrics")
               -> "https://127.0.0.1:9085";
-            image_policy_webhook: Path("/image-policy-webhook")
```
aermakov-zalando, line 500: This is referenced in the Prometheus configuration, could you update that as well?

linki: Thanks!

aermakov-zalando:
👍

linki:
Let's hold until a final test in production: https://github.bus.zalan.do/teapot/admission-controller/pull/143#issuecomment-3125699

linki:
👍

aermakov-zalando:
👍

PR #4486, opened 2021-07-22
EAS: add Opsgenie support

aermakov-zalando:
* Update EAS to the new version
 * Add a new temporary config item, `emergency_access_service_incident_backend` (with `jira` or `opsgenie` as valid values) to control which service to use


aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #4490, opened 2021-07-26
preconfigure admission-controller with upcoming pod-image-check fields

linki:
First of several PRs to replace `image-policy-webhook` with functionality in the `admission-controller`.

Rollout plan:
1. Preconfigure the config fields so we can enable admission-controller imaging checking in one go in the next PR
2. Rollout new version of admission-controller that supports validating images on pod admission (disabled by default in e2e und test, enabled in production)
3. Remove image-policy-webhook from the manifests. Admission controller takes over fully.

PR #4496, opened 2021-07-27
do not enable admission-controller's pod image in production by default

linki:
Follow up on https://github.com/zalando-incubator/kubernetes-on-aws/pull/4490

Let's not enable image checks in pod-admission by default in production. This way we can safely roll out new versions of admission-controller to all channels and enable image-checks per cluster afterwards.


jonathanbeber:
:+1:

linki:
👍

PR #4503, opened 2021-07-28
Add back e2e tests for image policy

linki:
This disables image-policy-webhook and enables admission-controller in e2e as well so that it matches what's currently in production already.

To do that it limits image-checks to a particular namespace prefix and then runs the old e2e image policy tests on it to validate that it allows and rejects the correct images.

In addition it also enforces image-policy checks in `kube-system` and `visibility` such that we can detect if we accidentally introduce invalid images in the system part. The reason to not check each and every namespace in e2e is that the e2e containers make use of non-compliant images.

Furthermore, we allow the pause container from k8s.io in e2e because it's deployed to `kube-system` as part of the e2e tests (https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/psp.go#L108). This we should try to tackle and then remove the config again.

It also changes the cluster alias for e2e clusters from the generic `teapot-e2e` to the stack name. This is needed to make the "scoped" platform tokens working.

Another change that was necessary is the `--cluster-domain` field for stackset-controller. In e2e tests we don't have a separate hosted zone for each cluster. Because each e2e cluster now has its own alias, the previous `alias+zalan.do` calculation doesn't work anymore. Instead I switched it to `Values.hosted_zone` which results in the same values for all existing clusters and to `teapot-e2e.zalan.do` for e2e clusters.


cluster/config-defaults.yaml
```diff
@@ -343,14 +343,17 @@ teapot_admission_controller_daemonset_reserved_memory: "64Gi"
 teapot_admission_controller_validate_application_label: "true"
 teapot_admission_controller_validate_base_images: "true"
 teapot_admission_controller_validate_pod_images: "true"
+teapot_admission_controller_validate_pod_images_soft_fail_namespaces: "^pierone$"
```
mikkeloscar: Should this be set for all clusters or only for the one where pierone runs?

linki: Only for one cluster. Thanks! 🤦

mikkeloscar:
:+1:

linki:
👍

linki:
👍

cluster/config-defaults.yaml
```diff
@@ -350,7 +350,9 @@ teapot_admission_controller_namespace_delete_protection_enabled: "true"
 {{else if eq .Cluster.Environment "e2e"}}
 teapot_admission_controller_validate_application_label: "false"
 teapot_admission_controller_validate_base_images: "false"
-teapot_admission_controller_validate_pod_images: "false"
+teapot_admission_controller_validate_pod_images: "true"
+teapot_admission_controller_validate_pod_images_namespaces: "^(image-policy-test-enabled-.*|kube-system|visibility)$"
```
aermakov-zalando, line 354: This will disable validation in most namespaces, no? This shouldn't be configured by default at all, only for E2E and in specific test clusters.

aermakov-zalando: Might be a good idea to rename the field as well, to prevent people from configuring it accidentally. Alternatively, we could also just ignore it completely in the template for production clusters.

cluster/config-defaults.yaml
```diff
@@ -350,7 +350,9 @@ teapot_admission_controller_namespace_delete_protection_enabled: "true"
...
+teapot_admission_controller_validate_pod_images_ignored_images: "^k8s.gcr.io/pause:.+$"
```
aermakov-zalando, line 355: Again, why? This is only needed in E2E, production clusters don't depend on it.

cluster/config-defaults.yaml
linki: Discussed internally: Those settings are limited to e2e clusters. Usability and safety improvements are addressed in https://github.com/zalando-incubator/kubernetes-on-aws/pull/4582.

cluster/config-defaults.yaml
linki: This also applies to e2e only.

PR #4511, opened 2021-07-29
Disable image-policy-webhook in all clusters

linki:
Disables image-policy-webhook in production clusters in favour of admission-controller's checks.

~~It also drops an e2e test that wouldn't pass with that setting.~~ Let's keep it in e2e to catch non-compliant system images.

Once we can access _some_ docker-meta from e2e clusters we switch to proper testing, see https://github.com/zalando-incubator/kubernetes-on-aws/pull/4503


linki:
👍

aermakov-zalando:
👍

PR #4514, opened 2021-07-30
Use the internal scheduler build in all clusters

aermakov-zalando:
We haven't observed any issues so far, and the metrics in the clusters where it was enabled seemed better.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4536, opened 2021-08-19
DS: add an optional ML Experiment deployment role

aermakov-zalando:
Right now it's disabled in all clusters, once the PoC phase is done we'll enable it in production ones.


jonathanbeber:
:+1:

aermakov-zalando:
👍

PR #4539, opened 2021-08-19
EAS: drop JIRA intergration

aermakov-zalando:
No longer in use, drop all related config items and secrets.


jonathanbeber:
:+1:

aermakov-zalando:
👍

PR #4543, opened 2021-08-20
etcd stack: support vpc-less security groups

aermakov-zalando:
Very old stacks don't specify the VPC ID for the security group. Unfortunately, CF doesn't allow to change this without recreating the resource (even if it'd be set to the same effective value), which is impossible due to cross-stack references. Let's add a config item for now and then figure out how to migrate it properly later.


aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #4546, opened 2021-08-20
etcd stack: fix the group references for vpc-less SGs

aermakov-zalando:
Apparently `!Ref <SecurityGroup>` has [different semantics](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html#aws-properties-ec2-security-group-return-values) depending on whether the SG specifies the VPC or not. Switch to `GetAtt` instead.


aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #4548, opened 2021-08-20
Add a name to the etcd launch template

aermakov-zalando:
It's not really necessary, but it's a bit less ugly and makes it easier to find in the UI.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4564, opened 2021-08-25
CA: use scaling events for faster error detection

aermakov-zalando:
Add an experimental config item (`experimental_cluster_autoscaler_check_scaling_events`) to roll out the new Cluster Autoscaler version that can detect scaling issues and fallback a lot faster (see https://github.com/zalando-incubator/autoscaler/pull/78 for details). Enable it by default for all test clusters.

PR #4569, opened 2021-08-26
Use scaling events in all clusters

aermakov-zalando:
Enable #4564 in all clusters by default. We still keep the config item just in case, let's drop it a week later.


aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #4571, opened 2021-08-26
Allow skipper to scale based on ScalingSchedules

jonathanbeber:
Currently we have skipper scaling based on ZMON. Since we are migrating
away from ZMON for time based scaling, we need to move skipper too for
`ClusterScalingSchedules`. This is the first one, for hyped-article
releases. We will have more once SRE department decide on the
distribution of `ScalingSchedules`. We will have to decide which ones
will scale skipper and in which average.

This commit adds a new object metric for skipper's HPA targeting on the
hyped-article-release `ClusterScalingSchedule` object deployed to all
clusters. The clusters' skippers will scale iff the config item
`skipper_scaling_schedule_hyped_articles_target` is defined.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -200,8 +200,12 @@ skipper_oauth2_ui_login: "true"
 #
 # This section contains the config items related to time based scaling
 # for skipper
+# kube-zmon-scaling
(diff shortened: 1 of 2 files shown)


jonathanbeber:
:+1:

aryszka:
:+1:

PR #4572, opened 2021-08-27
CA: don't override terminationGracePeriodSeconds

aermakov-zalando:
The autoscaler apparently caps the pods' `terminationGracePeriodSeconds` to just 10 minutes by default, which causes issues for applications that need a long time to shutdown (unfortunately we have some that take hours). Let's set it to a much higher value for now and see what happens. This will most likely increase the costs because the existing scale-down logic can't handle more than one node at a time, but I don't see any other solution that won't require rewriting a lot of the autoscaler code.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -17,6 +17,9 @@ cluster_autoscaler_max_pod_eviction_time: "1h"
 cluster_autoscaler_max_pod_eviction_time: "3h"
 {{end}}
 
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
We should ideally not have deployment/statefulset pods terminating for hours, but I don't think that's realistic in the short term.

jonathanbeber:
:+1:

aermakov-zalando:
👍

PR #4573, opened 2021-08-27
E2E: cleanup the cluster asynchronously

aermakov-zalando:
Instead of cleaning up the cluster as part of the pipeline, just mark it for decommissioning and let the cleaner do its job. This saves us an extra 8–10 minutes.


aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #4577, opened 2021-08-30
Update node-exporter configuration

aermakov-zalando:
* Use the new argument name for `collector.filesystem.ignored-mount-points`
 * Ignore veth devices in netdev/netclass collectors, these are not really needed and there can be quite a lot of them on busy nodes


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4582, opened 2021-08-31
Do not make image ignoring settings available as config items

linki:
This doesn't expose the image-ignoring settings in admission-controller as config items since they are only needed in e2e clusters and we want to make it hard for people to change them.

The only exposed configuration is a general knob for turning it on and off with a big warning and the soft-fail setting which is unused as long as docker-meta is available.

Also added more comments to explain these settings.


aermakov-zalando:
👍

linki:
👍

PR #4583, opened 2021-08-31
CA: handle 'Cancelled' ASG events as well

aermakov-zalando:
Treat 'Cancelled' events as scale-up failures, ASG seems to only report those for weird issues on their side (see https://github.com/zalando-incubator/autoscaler/pull/81 for more details).


linki:
👍

mikkeloscar:
:+1:

PR #4584, opened 2021-09-01
Restrict access to nodes/proxy, services/proxy

aermakov-zalando:
We shouldn't allow access for `services/proxy` for the ReadOnly role because of the potential security impact. It was most likely only added for the dashboard to work, so let's explicitly provide access to the dashboard only and restrict the rest. `nodes/proxy` shouldn't be allowed at all.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4587, opened 2021-09-01
Kubernetes 1.21

jonathanbeber:
- New AMI
- batch/v2alpha1 deprecated, switching to batch/v1. Also: removed from
  kube-api-server runtime config
- Update stackset controller version
- Update e2e tests
  - Create new describe function since kubernetes/e2e deprecated
    KubeDescribe


test/e2e/go.mod
```diff
@@ -3,32 +3,32 @@ module github.com/zalando-incubator/kubernetes-on-aws/tests/e2e
...
+	k8s.io/client-go v0.18.9
+	k8s.io/kubernetes v0.0.0-00010101000000-000000000000
 )
 
 replace (
 	github.com/go-openapi/spec => github.com/go-openapi/spec v0.19.3
 	github.com/googleapis/gnostic => github.com/googleapis/gnostic v0.4.1
 	github.com/onsi/ginkgo => github.com/onsi/ginkgo v1.11.0
 	github.com/onsi/gomega => github.com/onsi/gomega v1.7.0
+	github.com/szuecs/routegroup-client => github.com/jonathanbeber/routegroup-client v0.18.9
```
jonathanbeber: temporary until https://github.com/szuecs/routegroup-client/pull/9 is merged and new tag created

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -118,7 +118,7 @@ write_files:
           - --enable-admission-plugins=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,DefaultTolerationSeconds,ExtendedResourceToleration,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota,StorageObjectInUseProtection,PodSecurityPolicy,ImagePolicyWebhook,Priority,NodeRestriction
           - --tls-cert-file=/etc/kubernetes/ssl/apiserver.pem
           - --tls-private-key-file=/etc/kubernetes/ssl/apiserver-key.pem
-          - --runtime-config=extensions/v1beta1/networkpolicies=true,batch/v2alpha1=true,policy/v1beta1=true,imagepolicy.k8s.io/v1alpha1=true,authorization.k8s.io/v1beta1=true,scheduling.k8s.io/v1alpha1=true,admissionregistration.k8s.io/v1beta1=true
```
mikkeloscar: Is the new API already available in `1.20` otherwise we will break at least deployments of jobs for all users still using the old APIVersion. We need some transition period I believe.

jonathanbeber: we have it tracked already (see GHE issue)

linki:
We need to keep an eye out for the `BoundServiceAccountTokens` feature (aka Service Account Tokens that have a limited validity time).

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md#api-change-1 (under "The BoundServiceAccountTokenVolume feature has been promoted to beta, and enabled by default.")

jonathanbeber:
Early reviews are more than welcome, concerns regarding new features/versions would be more easily tracked in GHE

test/e2e/run_e2e.sh
```diff
@@ -174,12 +174,19 @@ if [ "$e2e" = true ]; then
     #   https://github.com/kubernetes/kubernetes/blob/v1.19.2/test/e2e/network/service.go#L2522
     # * "[Fail] [sig-network] Services [It] should be able to switch session affinity for NodePort service [LinuxOnly] [Conformance]"
     #   https://github.com/kubernetes/kubernetes/blob/v1.19.2/test/e2e/network/service.go#L2538
+    #
+    # These are disabled because the nodePorts are not reachable via
+    # public IP, as the ones above and because it tries to schedule a
+    # pod directly in the node, what the admission controller only
+    # allows to system pods.
```
jonathanbeber: I was actually wrong here. It uses the [internal IP](https://github.com/kubernetes/kubernetes/blob/v1.21.4/test/e2e/network/hostport.go#L78) and tries to connect from another pod. The only problem is that our admissions controller doesn't allow the scaling into a node:

```
err:admission webhook "pod-admitter.teapot.zalan.do" denied the request: only system pods are allowed to schedule directly on nodes
```
we might want to revisit this one and disable this check during the e2e. I will investigate further, comments are welcome.

jonathanbeber: reactivated.

jonathanbeber: Ok, now the problem is a bit bigger, will detail here and also in the internal issue. 

For using hostPort we will need to enable it in our [CNI configmap](https://github.com/zalando-incubator/kubernetes-on-aws/blob/51a5fc424f2457231558ad0982673d2a86fe5a18/cluster/manifests/flannel/configmap.yaml#L9-L22) like the [doc](https://v1-21.docs.kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#support-hostport) suggests. The doc also suggests we shouldn't use hostPort at all. Currently, we use it in the skipper pods but with a host network. It's a very specific use case and I don't see why to enable it for builders. Therefore, let's not enable it for now and discuss later if we want that and reactivate this test.

aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #4589, opened 2021-09-02
Consolidate apiserver/controller-manager/scheduler [1/X]

aermakov-zalando:
Change the mirror pods on the masters (apiserver, controller-manager and scheduler) to components of the `kubernetes` application instead of using their own application IDs. We will lose Prometheus metrics for the duration of the rotation, but it doesn't seem like a big deal.

### DIFF
--- a/cluster/manifests/admission-control-credentials/credentials.yaml
+++ b/cluster/manifests/admission-control-credentials/credentials.yaml
@@ -4,9 +4,10 @@ metadata:
   name: admission-controller-credentials
   namespace: kube-system
   labels:
-    application: kube-apiserver
+    application: kubernetes
+    component: kube-apiserver
 spec:
-  application: kube-apiserver
+  application: kubernetes
   tokens:
     kio:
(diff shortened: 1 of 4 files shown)


cluster/manifests/admission-control-credentials/credentials.yaml
```diff
@@ -4,9 +4,10 @@ metadata:
   name: admission-controller-credentials
   namespace: kube-system
   labels:
-    application: kube-apiserver
+    application: kubernetes
+    component: apiserver
```
mikkeloscar: nit: I think this should be `kube-apiserver` because that is the name of the component in Kubernetes. Similar for `kube-controller-manager` and `kube-scheduler`.

aermakov-zalando: Makes sense!

mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4594, opened 2021-09-06
Add e2e tests for image policy during pod creation

linki:
Adds additional e2e tests for image policy.

### DIFF
--- a/test/e2e/apiserver.go
+++ b/test/e2e/apiserver.go
@@ -27,14 +27,19 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
 )
 
-var _ = framework.KubeDescribe("API Server webhook tests (enabled namespace)", func() {
+const (
+	compliantImage    = "registry.opensource.zalan.do/teapot/skipper:v0.13.98"       // this image tag is compliant
+	compliantImage2   = "registry.opensource.zalan.do/teapot/skipper:v0.13.97"       // this image tag is compliant as well
+	nonCompliantImage = "registry.opensource.zalan.do/teapot/skipper-test:pr-1845-1" // this image tag is not compliant
+)
+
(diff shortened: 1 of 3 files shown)


test/e2e/apiserver.go
```diff
@@ -131,3 +131,87 @@ var _ = framework.KubeDescribe("API Server webhook tests (ignored namespace)", f
...
+		}()
+		label := map[string]string{
+			"app": podname,
+		}
+		labelSelector := labels.SelectorFromSet(labels.Set(label))
+		_, err = e2epod.WaitForPodsWithLabelRunningReady(cs, ns, labelSelector, 1, 1*time.Minute)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("Should not create pod with un-compliant image [Image-Webhook] [Non-Compliant] [Zalando]", func() {
```
aermakov-zalando: `non-compiant`?

test/e2e/apiserver.go
```diff
@@ -131,3 +131,87 @@ var _ = framework.KubeDescribe("API Server webhook tests (ignored namespace)", f
 		Expect(err).NotTo(HaveOccurred())
 	})
 })
+
+var _ = framework.KubeDescribe("API Server webhook tests for pods (enabled namespace)", func() {
```
aermakov-zalando: This should have a test for changing the image later (which we've missed).

test/e2e/apiserver.go
```diff
@@ -262,7 +262,7 @@ var _ = framework.KubeDescribe("API Server webhook tests for pods (update path)
 		image = "skipper"
 		tag = "v0.13.97" // this image tag is compliant as well
 
-		pod = createImagePolicyWebhookTestPod(pod.Name, ns, image, tag, podname)
+		pod = updateImagePolicyWebhookTestPod(pod.Name, ns, image, tag, podname)
 		_, err = cs.CoreV1().Pods(ns).Update(context.TODO(), pod, metav1.UpdateOptions{})
```
aermakov-zalando: This won't work because of the versioning. You need to update the previous pod object you get from the API, or send a patch instead.

linki: 🤦   That makes sense

test/e2e/apiserver.go
linki: This was added: https://github.com/zalando-incubator/kubernetes-on-aws/blob/bbeda1afea5de87d117ecaa9bfe09082c028c53d/test/e2e/apiserver.go#L270-L281

mikkeloscar:
:+1:

linki:
👍

PR #4600, opened 2021-09-08
Admission controller: AZ block feature

aermakov-zalando:
Add a new config item, `blocked_availability_zone`, that allows us to prevent the use of a particular availability zone as much as possible. Using a more generic name here because while it's only used with the admission controller right now, it could affect more things in the future.

### DIFF
--- a/cluster/manifests/01-admission-control/config.yaml
+++ b/cluster/manifests/01-admission-control/config.yaml
@@ -84,6 +84,15 @@ data:
 {{- end }}
 {{- end}}
 
+  pod.az-block.zone-name: "{{ .Cluster.ConfigItems.blocked_availability_zone }}"
+{{- range $group, $enabled := zoneDistributedNodePoolGroups .Cluster.NodePools }}
+{{- if eq $group "" }}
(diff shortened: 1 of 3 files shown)


cluster/manifests/01-admission-control/config.yaml
```diff
@@ -82,6 +82,15 @@ data:
 {{- else }}
   pod.automatic-topology-spread.pools.dedicated.{{$group}}: "{{ $enabled }}"
 {{- end }}
+{{- end}}
+
+  pod.az-block.zone-name: "{{ .Cluster.ConfigItems.blocked_availability_zone }}"
```
aermakov-zalando, line 87: This could technically work a bit better, but I'd rather drop `worker-default` instead and just get rid of this thing altogether.

aermakov-zalando:
👍

mikkeloscar:
:+1:

aermakov-zalando:
Seems to work fine. The current setup doesn't cover node pool groups where some pools use `worker-splitaz` and others use `worker-default`, but I'd rather try to get rid of `worker-default` altogether instead of creating yet another template function.

PR #4605, opened 2021-09-08
kube-state-metrics: add more node labels to the allowlist

aermakov-zalando:
With these in place, we could easily run aggregated queries to e.g. calculate the total amount of wasted space on all the nodes in a node pool or a node pool group. Doing it without those is still possible, obviously, but would result in hundreds of queries to Prometheus and would require us to do the aggregation in ad-hoc code instead.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4610, opened 2021-09-09
Drop the experimental_cluster_autoscaler_check_scaling_events CI

aermakov-zalando:
It's unlikely we'll have to roll back, and we should just fix the autoscaler instead.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4625, opened 2021-09-14
Start with the maximum memory for vertically-scaled deployments

aermakov-zalando:
VPA OOM handling logic doesn't do anything unless we already have a recommendation value for this VPA. This means that new deployments can completely fail to start without manual intervention if they don't have enough memory out of the box. This should ideally be fixed in the VPA controller, but I'm not sure if anyone would have time to do this any time soon.

Instead, let's initialise the vertically scaled deployments with the maximum memory values instead of some other ones. This will lead to slightly higher resource consumption for new deployments, but it'll notice and adjust almost immediately. It also gives us a nice failsafe for the cases where VPA's admitter is down (since we didn't invest a lot into its availability).

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -261,16 +261,13 @@ prometheus_remote_min_backoff: "3s"
 prometheus_remote_max_backoff: "10s"
 
 metrics_service_cpu: "100m"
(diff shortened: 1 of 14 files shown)


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4628, opened 2021-09-14
Disable TLS1.0/1.1 on APIServer Load Balancer

mikkeloscar:
For Ingress we set the default TLS policy on ALBs/NLBs to `ELBSecurityPolicy-TLS-1-2-2017-01`. Set the same policy for the load balancer used for the APIServers in order to disable TLS1.0/1.1 and improve security in our setup.

If not set specifically it will default to: `ELBSecurityPolicy-2016-08`

Ref: https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-security-policy-table.html


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4634, opened 2021-09-15
Cleanup the deletions

aermakov-zalando:
Most of these have been rolled out to all clusters years ago, we don't need to keep them.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4636, opened 2021-09-15
Rename worker-default to worker-combined

aermakov-zalando:
It's unclear whether we need it at all, but for now let's rename it to something more explicit. This is safe to do because we don't have any pools using this profile, so it doesn't seem worth it to roll it out in multiple PRs.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4651, opened 2021-09-21
Enables NLB HTTP redirect

AlexanderYastrebov:
Updates `skipper` and `kube-ingress-aws-controller` and enables NLB HTTP redirect

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -361,7 +361,7 @@ Resources:
 {{- end }}
 {{- end }}
         - CidrIp: {{ if eq .Cluster.ConfigItems.kube_aws_ingress_controller_nlb_enabled "true" }}"0.0.0.0/0"{{else}}"{{.Values.vpc_ipv4_cidr}}"{{end}}
-          FromPort: 9999
+          FromPort: 9998
           IpProtocol: tcp
           ToPort: 9999
 {{- if ne .Cluster.ConfigItems.skipper_redis_replicas "0"}}
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: skipper-ingress
-    version: v0.13.100
+    version: v0.13.108
     component: ingress
 spec:
   strategy:
@@ -19,7 +19,7 @@ spec:
     metadata:
       labels:
         application: skipper-ingress
-        version: v0.13.100
+        version: v0.13.108
         component: ingress
       annotations:
         kubernetes-log-watcher/scalyr-parser: |
@@ -46,12 +46,18 @@ spec:
       hostNetwork: true
       containers:
       - name: skipper-ingress
-        image: registry.opensource.zalan.do/teapot/skipper-internal:v0.13.100-172
+        image: registry.opensource.zalan.do/teapot/skipper-internal:v0.13.108-177
         ports:
         - name: ingress-port
           containerPort: 9999
           hostPort: 9999
           protocol: TCP
+{{ if or (eq .ConfigItems.nlb_switch "pre") (eq .ConfigItems.nlb_switch "exec") }}
+        - name: http-redirect
+          containerPort: 9998
+          hostPort: 9998
+          protocol: TCP
+{{ end }}
         env:
         - name: LIGHTSTEP_TOKEN
           valueFrom:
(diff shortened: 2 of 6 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -208,11 +218,18 @@ spec:
...
+            command:
+            - /bin/sh
+            - -ce
+            # Check all sub processes in the container and fail if one of them fails.
+            # wget will exit with non-zero status code if the HTTP code is no 2xx.
+            # production tokeninfo, http-redirect
+            - >
+              wget -T 1 -q -O /dev/null http://127.0.0.1:9021/health;
+{{- if or (eq .ConfigItems.nlb_switch "pre") (eq .ConfigItems.nlb_switch "exec") }}
+              wget -T 1 -q -O /dev/null http://127.0.0.1:9998/healthz;
```
AlexanderYastrebov: TODO: add to `-status-checks`?

szuecs: For the redirect listener?
I would say we can omit it, because it's not waiting for an external dependency and we start it before main skipper, so no need for extra safety.

cluster/cluster.yaml
```diff
@@ -359,6 +359,12 @@ Resources:
           IpProtocol: {{ $element.Protocol }}
           ToPort: {{ $element.ToPort }}
 {{- end }}
+{{- end }}
+{{- if or (eq .ConfigItems.nlb_switch "pre") (eq .ConfigItems.nlb_switch "exec") }}
```
szuecs: We should not guard this here, because it might introduce a race condition. Just put it into the section with kube_aws_ingress_controller_nlb_enabled "true" below and change just FromPort to 9998

cluster/manifests/skipper/deployment.yaml
```diff
@@ -226,8 +243,14 @@ spec:
...
             # wget will exit with non-zero status code if the HTTP code is no 2xx.
-            # production tokeinfo, sandbox tokeninfo, bridge
-            - "wget -T 1 -q -O /dev/null http://127.0.0.1:9021/health; wget -T 1 -q -O /dev/null http://127.0.0.1:9121/health; wget -T 1 -q -O /dev/null http://127.0.0.1:9000/healthz"
+            # production tokeninfo, sandbox tokeninfo, bridge, http-redirect
+            - >
+              wget -T 1 -q -O /dev/null http://127.0.0.1:9021/health;
+              wget -T 1 -q -O /dev/null http://127.0.0.1:9121/health;
+              wget -T 1 -q -O /dev/null http://127.0.0.1:9000/healthz;
+{{- if or (eq .ConfigItems.nlb_switch "pre") (eq .ConfigItems.nlb_switch "exec") }}
+              wget -T 1 -q -O /dev/null http://127.0.0.1:9998/healthz;
```
szuecs: Similar to above:
I think we should omit having this livenessProbe change, because it's too much interruption for the redirect case, that should be rarely happen anyways

cluster/cluster.yaml
```diff
@@ -359,6 +359,12 @@ Resources:
           IpProtocol: {{ $element.Protocol }}
           ToPort: {{ $element.ToPort }}
 {{- end }}
+{{- end }}
+{{- if or (eq .Cluster.ConfigItems.nlb_switch "pre") (eq .Cluster.ConfigItems.nlb_switch "exec") }}
```
szuecs: Don't guard this and add it into the section below only changing FromPort, otherwise we can have a race condition here for no reason.

cluster/cluster.yaml
AlexanderYastrebov: Do you mean if `kube_aws_ingress_controller_nlb_enabled` is `false`? In that case we need to change all other occurrences  of `nlb_switch` to respect `kube_aws_ingress_controller_nlb_enabled`

cluster/manifests/skipper/deployment.yaml
szuecs: please no livenessProbe for this, because we don't want to loose production traffic only because the redirect does not work for a single instance. tokeninfo is different because we serve ~10% of overall traffic with it.

szuecs (changes requested):
Some small but important changes

cluster/cluster.yaml
szuecs: Just open the port without guard

AlexanderYastrebov: Fixed

cluster/manifests/skipper/deployment.yaml
AlexanderYastrebov: Removed HTTP redirect liveness probe

szuecs (approved):
thanks, great PR

szuecs:
:+1:

AlexanderYastrebov:
:+1:

aryszka:
:+1:

PR #4655, opened 2021-09-23
added option to route traces traffic to local availability zone

kostz:
https://github.bus.zalan.do/teapot/issues/issues/3102

### DIFF
--- a/cluster/manifests/coredns-local/configmap-local.yaml
+++ b/cluster/manifests/coredns-local/configmap-local.yaml
@@ -78,28 +78,14 @@ data:
     }
 {{ end }}
 
-{{ if eq .ConfigItems.coredns_route_traces_to_local_zone "true"}}
-    {{ .ConfigItems.coredns_global_traces_endpoint }}:9254 {
-        errors
-
-        rewrite name exact {{ .ConfigItems.coredns_global_traces_endpoint }} {{ .ConfigItems.coredns_local_zone_traces_endpoint }}
-
-        kubernetes cluster.local {
-            pods insecure
-        }
-        cache 30
-{{ if eq .ConfigItems.coredns_log_svc_names "true"}}
-        log svc.svc.cluster.local.
-{{ end }}
-        prometheus :9153
-        ready :9155
-    }
-{{ end }}
 
     # 10.2.0.0/16, 10.3.0.0/16 defines that this server is authority for revese
     # lookups for these ranges.
-    cluster.local:9254 10.2.0.0/16:9254 10.3.0.0/16:9254 {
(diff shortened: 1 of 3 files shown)


cluster/manifests/coredns-local/configmap-local.yaml
```diff
@@ -78,6 +78,24 @@ data:
     }
 {{ end }}
 
+{{ if eq .ConfigItems.coredns_route_traces_to_local_zone "true"}}
+    {{ .ConfigItems.coredns_global_traces_endpoint }}:9254 {
+        errors
+
+        rewrite name exact {{ .ConfigItems.coredns_global_traces_endpoint }} {{ .ConfigItems.coredns_local_zone_traces_endpoint }}
+
+        kubernetes cluster.local {
```
mikkeloscar: We are a bit worried that having `kubernetes` defined both in this server block and below will double the memory for tracking kubernetes resources which would be a problem in big clusters.

Would you be able to verify if this is the case or not? Or alternatively merge it into a single server block to avoid defining the kubernetes block twice?

For context we have separate server blocks because of the guidance here: https://github.com/coredns/coredns/issues/2593#issuecomment-466888239 it may need to be re-evaluated considering this new use-case.

kostz: That makes sense, moved rewrite rule for traces endpoint into existed server, removed extra server

aermakov-zalando: Could you include the guidance from [this comment](https://github.com/coredns/coredns/issues/2593#issuecomment-815162945)?

kostz: Not sure if I understand you correctly, 

- Second server with a `kubernetes` plugin was removed
- I'm not touching `kubernetes` plugin definition at all in the existed server, it [stays the same](https://github.com/zalando-incubator/kubernetes-on-aws/pull/4655/files#diff-37253e16272476123f699244af6ce1ee6a8e7baf7d6515a52bcd083099067f65R88)  
- That enhancement you're referring to is probably a good case for individual ticket, testing (performance?) and individual pull request

kostz: I mean I can replace
```
 kubernetes  {
            pods insecure
        }
```

with 

```
 kubernetes cluster.local 2.10.in-addr.arpa 3.10.in-addr.arpa {
            pods insecure
        }
```

but I'm not sure about the potential impact and it looks like not much related to the feature I'm implementing

mikkeloscar:
:+1:

kostz:
👍

aermakov-zalando:
👍

PR #4658, opened 2021-09-23
Exclude master node from being targets for services of type LoadBalancer

linki:
> To prevent control plane nodes from being added to load balancers automatically, upgrade users need to add "node.kubernetes.io/exclude-from-external-load-balancers" label to control plane nodes. (#97543, @pacoxu)

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md#changelog-since-v1200

Interestingly, this was guarded in 1.20 with a FeatureGate (which was turned off in 1.21). However, it uses the `node-role.kubernetes.io/master` label which we removed. Hence, currently masters are added to Service ELBs.

This is the 1.21 way of excluding master nodes from Services of Type LoadBalancer.

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -4,7 +4,7 @@ write_files:
     path: /etc/kubernetes/secrets.env
     content: |
       NODEPOOL_TAINTS=node.kubernetes.io/role=master:NoSchedule{{if index .NodePool.ConfigItems "taints"}},{{.NodePool.ConfigItems.taints}}{{end}}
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
Shouldn't this go onto the 1.21 branch, if it doesn't work with 1.20?

linki:
Yes, I thought I needed to follow https://teapot.docs.zalando.net/howtos/kubernetes-on-aws/ to always make changes against dev. But here, the label is really only recognised by 1.21 so it makes more sense there.

aermakov-zalando:
👍

linki:
👍

PR #4659, opened 2021-09-23
Don't use an FQDN for the status service URL

aermakov-zalando:
There's no real reason to do this, so let's drop the trailing dot.


aermakov-zalando:
👍

linki:
👍

PR #4662, opened 2021-09-24
Drop apiserver_nlb config item

aermakov-zalando:
We don't want to support two complex configurations at the same time, especially when one causes issues with client-go.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4669, opened 2021-09-27
Ensure that extended tokens have to actively be turned off

linki:
Ensure that extended tokens have to actively be turned off

===

~~Enables expiring service account tokens.~~ Some previous findings can be found here: https://github.com/zalando-incubator/kubernetes-on-aws/pull/2875#issuecomment-578115791

KEP: https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/1205-bound-service-account-tokens

| Token behaviour                   | Timeframe |
|----------------------------------|-----------|
| Recommended refresh interval     | 1 minute  |
| Warnings and metrics occur after | 1 hour    |
| Tokens are valid for             | 1 year    |


cluster/config-defaults.yaml
```diff
@@ -505,7 +505,7 @@ kubernetes_scheduler_image: "zalando"
 # when set to true, service account tokens can be used from outside the cluster
 allow_external_service_accounts: "false"
 # issue service account tokens with expiration time.
-rotate_service_account_tokens: "false"
+rotate_service_account_tokens: "true"
```
aermakov-zalando: Should we maybe enable it just in a couple of clusters first?

linki: Yes, I'll do that. The PR was to trigger the latest e2e tests with the other changes in here. I didn't realise how long it was ago since they last run. If it's green I'll enable it in one cluster and let it be for a while.

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -140,6 +140,7 @@ write_files:
           - --service-account-signing-key-file=/etc/kubernetes/ssl/service-account-private-key.pem
           - --service-account-issuer={{ .Cluster.APIServerURL }}
           - --service-account-jwks-uri={{ .Cluster.APIServerURL }}/openid/v1/jwks
+          - --service-account-extend-token-expiration={{ .Cluster.ConfigItems.rotate_service_account_tokens_extended_expiration }}
```
linki, line 143: The default is `true` already but this way we have to explicitly disable it or remove the flag.

linki:
👍

jonathanbeber:
:+1:

PR #4676, opened 2021-09-28
Import DS resources from the internal repo

aermakov-zalando:
We don't change it that often, so let's migrate it to the public repo. Dropping the config item to disable it as well, for now (although it's still kept in place until the manifests have been deleted from the internal one).


aermakov-zalando:
👍

jonathanbeber:
:+1:

PR #4677, opened 2021-09-28
Align dns names so they work in e2e

linki:
We changed the cluster alias in e2e to be unique so that platform credential tokens could be provisioned into them.

However, the domain name for all e2e clusters didn't change which in combination broke several Ingresses that expect to be under the teapot-e2e.zalan.do domain, especially the ones that need an ACM cert.

This fixes it by using `Values.hosted_zone` similar to other places in our manifests.


jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #4681, opened 2021-09-28
e2e: increase the memory requests for the e2e run

aermakov-zalando:
The main e2e run has started crashing due to OOM kills recently, let's bump it a bit. Reduce the others since we don't really need that much memory there.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #4686, opened 2021-09-29
Set pdb max unavailable to 50% in test clusters

jonathanbeber:
PDBs are setting a max unavailable of 1%. It allows disruption to
deployments with more than 10 replicas, which works great for production
workloads. The test clusters do not have large deployments and nodes
could contain pods of deployments with 0-9 replica, what makes it update
slow.

This commit sets the max unavailable used by the pdb controller to 50%
in tests clusters. This value would handle deployments with too few
replicas.


aermakov-zalando:
The percentage is rounded up, so you still get 1 pod even with a small deployment.

jonathanbeber:
You are right, I found a deployment failing to be evicted with 2 replicas but that was due to a different issue :confused:  
I tested again with an isolated deployment and the behaviour is the one your described.

PR #4694, opened 2021-10-04
Fix the DS configuration

aermakov-zalando:
Changes:
 * Use the same policy for the status service as for the rest of the VPA-scaled deployments (see https://github.com/zalando-incubator/kubernetes-on-aws/pull/4625 for more details),
 * Deploy the configmap _before_ the deployment service, and rotate the pods once we change it.


jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #4702, opened 2021-10-05
Skipper/routesrv [1/2]

szuecs:
This is the first step to deploy skipper kubernetes dataclient service to reduce load on apiserver

We need to deploy this in two steps such that we can validate that this component is serving the same routes as the current running skipper-ingress deployment.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -105,6 +105,11 @@ skipper_redis_pool_timeout: "250ms"
 skipper_redis_read_timeout: "25ms"
 skipper_redis_write_timeout: "25ms"
 
+# skipper routesrv settings
+skipper_routesrv_replicas: 3
+skipper_routesrv_cpu: "100m"
+skipper_routesrv_memory: "1Gi"
+
 # skipper api GW features
 enable_apimonitoring: "true"                       # TODO(sszuecs): cleanup candidate to reduce amount of branches in deployment
 
--- a/cluster/manifests/skipper/routesrv-deployment.yaml
+++ b/cluster/manifests/skipper/routesrv-deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     application: skipper-ingress
-    version: v0.13.120
+    version: v0.13.121
     component: routesrv
 spec:
   replicas: "{{ .ConfigItems.skipper_routesrv_replicas }}"
@@ -21,7 +21,7 @@ spec:
     metadata:
       labels:
         application: skipper-ingress
-        version: v0.13.120
+        version: v0.13.121
         component: routesrv
       annotations:
         kubernetes-log-watcher/scalyr-parser: |
@@ -46,7 +46,7 @@ spec:
       terminationGracePeriodSeconds: {{ .Cluster.ConfigItems.skipper_termination_grace_period }}
       containers:
       - name: routesrv
-        image: registry.opensource.zalan.do/teapot/skipper:v0.13.120
+        image: registry.opensource.zalan.do/teapot/skipper:v0.13.121
         ports:
         - name: ingress-port
           containerPort: 9990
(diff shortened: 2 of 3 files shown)


cluster/manifests/skipper/routesrv-deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+          - "-enable-api-usage-monitoring" ## TODO: required in routesrv no?
+          - "-api-usage-monitoring-realm-keys=https://identity.zalando.com/realm"
+          - "-api-usage-monitoring-client-keys=https://identity.zalando.com/managed-id,sub"
+          - "-api-usage-monitoring-default-client-tracking-pattern=services[.].*"
+          - "-default-filters-dir=/etc/config/default-filters"
+{{ end }}
+#{{ if eq .ConfigItems.skipper_oauth2_ui_login "true" }}
+#          - "-enable-oauth2-grant-flow" ## TODO: required in routesrv no?
+#          - "-oauth2-callback-path={{ .ConfigItems.skipper_oauth2_redirect_url }}"
+#{{ end }}
```
jonathanbeber, line 80: can't we drop it and get back if needed? we have the config in skipper deployment, right?

jonathanbeber:
:+1:

cluster/manifests/skipper/routesrv-deployment.yaml
szuecs: I want to check first the e2e cluster :)

jonathanbeber: got it, will reapply my +1 later so. Thanks for confirming.

szuecs: ah already gone, will merge and cleanup in [2/2] :)

szuecs:
:+1:

jonathanbeber:
:+1:

cluster/manifests/skipper/routesrv-deployment.yaml
jonathanbeber: sure!

cluster/config-defaults.yaml
```diff
@@ -105,6 +105,11 @@ skipper_redis_pool_timeout: "250ms"
 skipper_redis_read_timeout: "25ms"
 skipper_redis_write_timeout: "25ms"
 
+# skipper routesrv settings
+skipper_routesrv_replicas: 3
+skipper_routesrv_cpu: "100m"
```
aryszka, line 110: does 100m make any sense in this case? I think in production clusters cpu requests non-divisible by 1000 can cause hard to detect problems

szuecs: what kind of problem?

szuecs: changed it to 1 CPU for production in https://github.com/zalando-incubator/kubernetes-on-aws/pull/4704

cluster/config-defaults.yaml
```diff
@@ -105,6 +105,11 @@ skipper_redis_pool_timeout: "250ms"
 skipper_redis_read_timeout: "25ms"
 skipper_redis_write_timeout: "25ms"
 
+# skipper routesrv settings
```
marcinzaremba, line 108: Curious: how did we arrive at `replicas`, `cpu` and `memory` values? Was it a guess, some sort of a load test, ball-park calculation, experience?

cluster/config-defaults.yaml
marcinzaremba: What sort of problem does it impose @aryszka ?

cluster/config-defaults.yaml
szuecs: It was an experienced guess.
The thing is that it won't have much load, but we want to have availability so 3 replicas was required. 
CPU and memory are IMO much more than we actually need. When we merged to dev we can test it in teapot and scale out skipper-ingress to 150 replicas and see if 1 CPU and 1Gi is enough. (We need to change the setting in teapot before scale up otherwise we would have 100m only.

PR #4715, opened 2021-10-11
Remove unused port 8080 from control plane SG

mikkeloscar:
Since https://github.com/zalando-incubator/kubernetes-on-aws/pull/4133 we disabled the insecure port (8080) on the api-server. We can therefore also remove the port from the security group as it's no longer relevant.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #4730, opened 2021-10-13
maxPods: make the extra capacity customizable

aermakov-zalando:
This gives us an extra tunable to potentially reduce the capacity issues in some of the big clusters until we can come up with other measures.


mikkeloscar:
:+1:

jonathanbeber:
:+1:

PR #4734, opened 2021-10-14
Update the etcd instances to use an external volume

aermakov-zalando:
This updates both stacks to use an EBS volume with a hardcoded size of 32Gi. We could, of course, make the whole thing configurable and allow switching between the ramdisk and EBS setups, or to allow customising the storage size, but at the moment we hope that it won't actually be needed.


linki:
👍

gargravarr:
:+1:

PR #4742, opened 2021-10-19
fix: apimonitoring was not enabled so metrics were not exposed

szuecs:


tkrop:
:+1:

tkrop:
But I'm not sure, that this will fix the issue for *out-of-the-box* API monitoring, since if `-default-filters-dir=/etc/config/default-filters` are not set, I assume the configuration of the API monitoring is not read - isn't it?

szuecs:
@tkrop default filters are applied at routesrv, because skipper-ingress will not use kubernetes dataclient anymore

szuecs:
:+1:

aryszka:
:+1:

PR #4745, opened 2021-10-20
Change Deployment Role policies by existing ones managed centrally

tcondeixa:
Replace Deployment Role locally defined policy by Centrally managed policies. This allow us to centrally enable new services for Teams and their Deployments.

Since we don't know the order that CF templates will be applied I'm first changing the role to use the new polices
Later I will create another PR to clean the role and config-defaults.yaml

The Policy names are not the best one, but should be enough for now. We will review Managed Policies content and names after Cyberweek, so we will create another PR later when needed.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -680,10 +680,10 @@ Resources:
         Version: 2012-10-17
       Path: /
       ManagedPolicyArns:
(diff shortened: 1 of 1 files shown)


mikkeloscar:
👍

gargravarr:
:+1:

gargravarr:
:+1:

PR #4757, opened 2021-10-27
Add support for automatic injection of topology spread timeout

linki:
Update admission-controller to enable support for automatic injection of topology spread timeout.

Topology spread timeout is the time after which automatically injected topology spread constraints are removed so that pods can run although they are zone-skewed.


linki:
Closing in favour of https://github.com/zalando-incubator/kubernetes-on-aws/pull/4758

PR #4767, opened 2021-11-02
Update admission-controller to master-132

linki:
This version fixes a bug where an AZ block is considered active although the configured blocked `zoneName` is an empty string. This turned off automatic pod spread injection even though no zone is actually supposed to be blocked.


jonathanbeber:
:+1:

gargravarr:
:+1:

PR #4772, opened 2021-11-05
fix to set grace period, change was rolled out already via config items

szuecs:

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -44,17 +44,15 @@ kube_aws_ingress_default_lb_type: "application"
 #
 # "exec": same as "pre" and kube-ingress-aws-controller will default to NLB
 # after removing it you need to set kube_aws_ingress_default_lb_type: "network" above and cleanup skipper and ingress-ctl deployment.yaml
+#
+# new clusters should use NLB as default, everything else has already
+# set config items via cluster registry
+nlb_switch: "exec"
 # TODO(sszuecs): grace period need also to be set for NLB and in skipper deployment we should clean i up
+# skipper termination settings
+skipper_wait_for_healthcheck_interval: "65s"
 # skipper_termination_grace_period: 302
-{{if eq .Cluster.Channel "dev"}}
-nlb_switch: "pre"
-{{else if eq .Cluster.Channel "alpha"}}
-nlb_switch: "false"
-{{else if eq .Cluster.Channel "beta"}}
-nlb_switch: "false"
-{{else}}
-nlb_switch: "false"
-{{end}}
+skipper_termination_grace_period: "70"
 
 # skipper ingress settings
 skipper_ingress_target_average_utilization_cpu: "60"
@@ -96,9 +94,6 @@ skipper_liveness_init_delay_seconds: 30
 skipper_readiness_init_delay_seconds: 1
 skipper_liveness_init_delay_seconds: 30
 {{end}}
-# skipper termination settings
-skipper_termination_grace_period: "70"
-skipper_wait_for_healthcheck_interval: "65s"
 
 # skipper redis settings
 enable_dedicate_nodepool_skipper_redis: "false"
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

cluster/config-defaults.yaml
```diff
@@ -44,6 +44,8 @@ kube_aws_ingress_default_lb_type: "application"
 #
 # "exec": same as "pre" and kube-ingress-aws-controller will default to NLB
 # after removing it you need to set kube_aws_ingress_default_lb_type: "network" above and cleanup skipper and ingress-ctl deployment.yaml
+# TODO(sszuecs): grace period need also to be set for NLB and in skipper deployment we should clean i up
+# skipper_termination_grace_period: 302
```
AlexanderYastrebov: Not sure I fully understand the comment. It sounds like that behind NLB termination grace period should be 302 seconds instead of default 70 we have now:
https://github.com/zalando-incubator/kubernetes-on-aws/blob/33a818d170340955fa30027af29b277a3e899a06/cluster/config-defaults.yaml#L102-L104

Should we rather increase default value to 302 unconditionally?

szuecs: we change it to 302s if the majority of the high traffic clusters are migrated

AlexanderYastrebov:
:+1:

cluster/manifests/skipper/deployment.yaml
```diff
@@ -43,7 +43,11 @@ spec:
 {{- end }}
       priorityClassName: "{{ .Cluster.ConfigItems.system_priority_class }}"
       serviceAccountName: skipper-ingress
+{{ if eq .Cluster.ConfigItems.nlb_switch "exec" }}
+      terminationGracePeriodSeconds: {{ 302 }}
```
mikkeloscar: This looks wrong with the `{{` `}}` around the value. I guess it's not caught in e2e because we don't have `nlb_switch "exec"` in e2e.

szuecs: that's interesting how I got to this state :D

szuecs:
I also cleaned up the defaults for the nlb switch as we are having this done via cluster registry, now.
We don't need this environment based switch anymore

cluster/manifests/skipper/deployment.yaml
szuecs: I guess it's also not wrong to have a text template that returns 302, so why should it fail in e2e tests?

cluster/config-defaults.yaml
```diff
@@ -44,15 +44,15 @@ kube_aws_ingress_default_lb_type: "application"
...
-{{end}}
+#
+# new clusters should use NLB as default, everything else has already
+# set config items via cluster registry
+nlb_switch: "exec"
+# TODO(sszuecs): grace period need also to be set for NLB and in skipper deployment we should clean i up
+# skipper termination settings
+skipper_wait_for_healthcheck_interval: "65s"
+# skipper_termination_grace_period: 302
+skipper_termination_grace_period: "70"
```
AlexanderYastrebov, line 55: Lets maybe not move these properties around but just add a comment to the existing defaults (since they are not changed)
```yaml
# skipper termination settings
# When nlb_switch is "exec" skipper_termination_grace_period value is ignored and termination grace period is explicitly set to 302 seconds
skipper_termination_grace_period: "70"
skipper_wait_for_healthcheck_interval: "65s"
```

cluster/manifests/skipper/deployment.yaml
mikkeloscar: > I guess it's also not wrong to have a text template that returns 302, so why should it fail in e2e tests?

You're right, I didn't think about that you just define the int in the template, so it would work fine I guess. Still it's easier to read without `{{` around :)

cluster/config-defaults.yaml
szuecs: why does it matter?

AlexanderYastrebov: * Moving without change makes it harder to track when/why current value was introduced/changed
* commented out `# skipper_termination_grace_period: 302` creates confusion IMO
* I am struggling to understand `TODO(sszuecs): ` comment clearly
* `skipper_` configs are somewhat grouped together

AlexanderYastrebov:
These configs were updated by #5024 and #5094

PR #4773, opened 2021-11-05
Update cluster autoscaler

jonathanbeber:
This commit updates autoscaler with the changes in
https://github.com/zalando-incubator/autoscaler/pull/87. It just updates
the available instance types.


jonathanbeber:
This PR is required as a kind of `bugfix` since when unknown instances are used by some node pools, the cluster-autoscaler can't react and scale even other node-pools with know instance types.

mikkeloscar:
:+1:

jonathanbeber:
:+1:

PR #4777, opened 2021-11-08
Enable okta authentication by default

mikkeloscar:
Enable okta authentication by default in all clusters.

(Already enabled in all clusters via config-item)


mikkeloscar:
:+1:

jonathanbeber:
:+1:

szuecs:
@mikkeloscar why can't we clean up the config item?

mikkeloscar:
> @mikkeloscar why can't we clean up the config item?

We can, but I would propose to wait until okta is the default as it's still in a testing phase. I have added a step to remove the config item in the internal ticket tracking the okta rollout.

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #4780, opened 2021-11-09
Allow logging-agent to use OIDC AWS credentials

jonathanbeber:
To allow the logging agent pods to access its own role via OIDC we have
to adjust the IAM role with a new assume role policy. This commit
creates it and removes the older policies attached to the worker and
master role. It is under a config `logging_agent_enable_oidc_iam` which
defaults to false.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -1473,6 +1473,14 @@ Resources:
                 - ''
                 - - 'arn:aws:iam::{{.Cluster.InfrastructureAccount | getAWSAccountID}}:role/'
                   - !Ref MasterIAMRole
+          - Action:
+              - 'sts:AssumeRoleWithWebIdentity'
+            Effect: Allow
+            Principal:
+              Federated: "arn:aws:iam::{{ .Cluster.InfrastructureAccount | getAWSAccountID }}:oidc-provider/{{.Cluster.LocalID}}.{{.Values.hosted_zone}}"
+            Condition:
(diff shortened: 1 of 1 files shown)


cluster/cluster.yaml
```diff
@@ -1473,6 +1474,16 @@ Resources:
...
+          {{ else }}
+          - Action:
+              - 'sts:AssumeRoleWithWebIdentity'
+            Effect: Allow
+            Principal:
+              Federated: "arn:aws:iam::{{ .Cluster.InfrastructureAccount | getAWSAccountID }}:oidc-provider/{{.Cluster.LocalID}}.{{.Values.hosted_zone}}"
+            Condition:
+              StringLike:
+                "{{ .Cluster.LocalID }}.{{ .Values.hosted_zone }}:sub": "system:serviceaccount:visibility:logging-agent"
+          {{ end }}
```
linki: There's a race condition here because the role is updated before the logging-agent Deployment is updated to use the OIDC approach. All logging-agent Pods are rolled during the update. Old Pods are still using kube2iam's approach but this role is already updated and the permissions are dropped such that kube2iam's approach doesn't work anymore. So it might be safer to just leave both `assumeRole` permissions in there during the migration phase.

jonathanbeber: thanks, updated.

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #4790, opened 2021-11-18
Upgrade nvidia-gpu-device-plugin v0.10.0

elgalu:
The GPU plugin is now officially provided by NVIDIA.

The default image `command` has changed:

```diff
- /usr/bin/nvidia-gpu-device-plugin
+ /usr/bin/nvidia-device-plugin
```

The command line arguments also changed, only these are supported and we don't need to customize them AFAIK:

```
NAME:
   nvidia-device-plugin - A new cli application

USAGE:
   nvidia-device-plugin [global options] command [command options] [arguments...]

COMMANDS:
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --mig-strategy value          the desired strategy for exposing MIG devices on GPUs that support it:
                                   [none | single | mixed] (default: "none") [$MIG_STRATEGY]
   --fail-on-init-error          fail the plugin if an error is encountered during initialization, otherwise block indefinitely (default: true) [$FAIL_ON_INIT_ERROR]
   --pass-device-specs           pass the list of DeviceSpecs to the kubelet on Allocate() (default: false) [$PASS_DEVICE_SPECS]
   --device-list-strategy value  the desired strategy for passing the device list to the underlying runtime:
                                   [envvar | volume-mounts] (default: "envvar") [$DEVICE_LIST_STRATEGY]
   --device-id-strategy value    the desired strategy for passing device IDs to the underlying runtime:
                                   [uuid | index] (default: "uuid") [$DEVICE_ID_STRATEGY]
   --nvidia-driver-root value    the root path for the NVIDIA driver installation (typical values are '/' or '/run/nvidia/driver') (default: "/") [$NVIDIA_DRIVER_ROOT]
   --help, -h                    show help (default: false)
```


linki:
@elgalu Why was this closed?

linki:
Note that this doesn't support `g2.*` instances. From the error message it looks like the GPUs are too old for the driver (NVIDIA GRID K520).

elgalu:
Thanks for catching that! let's get rid of g2 support, I only added that pool for testing but is unnecessary, we can use the other pools.

mikkeloscar:
~@linki Do we use `g2.*` anywhere? In any node pools? If so I guess we need to upgrade those with this rollout?~

I see you already checked this :+1:

cluster/manifests/nvidia/nvidia-gpu-device-plugin.yaml
```diff
@@ -42,27 +42,23 @@ spec:
...
+        - --pass-device-specs
         resources:
           requests:
             cpu: 50m
-            memory: 10Mi
+            memory: 25Mi
           limits:
             cpu: 50m
-            memory: 10Mi
+            memory: 25Mi
```
linki, line 57: The plugin got OOMKilled on some instance types. That's why I bumped it here.

Do we need to adjust the reserved resources for DaemonSets for this?

linki: We might have to bump `teapot_admission_controller_daemonset_reserved_memory` by `15Mi` in all clusters to be absolutely on the safe side I think.

mikkeloscar:
:+1:

linki:
👍

PR #4797, opened 2021-11-26
Allow to configure prometheus retention time and size

linki:
It's handy to tweak this setting to keep Prometheus disk usage under control.


linki:
Or we just go with the retention-size flag.

szuecs:
:+1:

AlexanderYastrebov:
IMO if the goal is to control the data size we should steer it explicitly. The docs https://prometheus.io/docs/prometheus/latest/storage/#operational-aspects do not explicitly prohibit having two options at the same time and docker seems to start just fine with both:
```
docker run --rm -it prom/prometheus --storage.tsdb.retention.time=1d --storage.tsdb.retention.size=49GB --config.file=/etc/prometheus/prometheus.yml
```

From the code it looks like they remove blocks that are both beyond retention time and size (i.e. use OR logic) https://github.com/prometheus/prometheus/blob/b63d7c786e045385b1abdff6fe7e5ac3b6f13a7a/tsdb/db.go#L1191-L1197

test/e2e/cluster_config.sh
```diff
@@ -35,7 +35,7 @@ clusters:
     nlb_switch: "pre"
     vm_dirty_bytes: 134217728
     vm_dirty_background_bytes: 67108864
-    prometheus_tsdb_retention_size: enabled
+    prometheus_retention_size: 1258292KB
```
linki: This is the value that would have been used for test clusters, the only user of this config item.

linki: Changed to the more readable 1280MB.

test/e2e/cluster_config.sh
```diff
@@ -35,7 +35,7 @@ clusters:
     nlb_switch: "pre"
     vm_dirty_bytes: 134217728
     vm_dirty_background_bytes: 67108864
-    prometheus_tsdb_retention_size: enabled
+    prometheus_retention_size: 1280MB
```
linki: This is the value that would have been used for test clusters, the only user of this config item (to be exact: 1258292KB)

AlexanderYastrebov: I think this is used for e2e test clusters, the actual value probably does not make any difference as these clusters are short-lived so maybe it should be just dropped and rely on defaults

mikkeloscar:
:+1:

linki:
We decided to drop the custom TSDB size limit for e2e clusters because we don't fully understand why it's needed.

@mikkeloscar If you happen to remember, feel free to comment (see related PR https://github.com/zalando-incubator/kubernetes-on-aws/pull/2999). Otherwise we'll go ahead with this approach.

mikkeloscar:
@linki @AlexanderYastrebov I agree and also don't see a need for this special config in e2e.

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #4798, opened 2021-11-30
enable: ingress v1 in test clusters

szuecs:

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -61,6 +61,11 @@ skipper_ingress_max_replicas: "50"
 skipper_ingress_min_replicas: "3"
 skipper_ingress_cpu: "1000m"
 skipper_ingress_memory: "1Gi"
+{{if eq .Cluster.Environment "test"}}
+skipper_ingress_kubernetes_ingress_v1: "true"
+{{else}}
+skipper_ingress_kubernetes_ingress_v1: "false"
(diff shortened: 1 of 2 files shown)


cluster/config-defaults.yaml
```diff
@@ -61,6 +61,11 @@ skipper_ingress_max_replicas: "50"
 skipper_ingress_min_replicas: "3"
 skipper_ingress_cpu: "1000m"
 skipper_ingress_memory: "1Gi"
+{{if eq .Cluster.Environment "test"}}
```
AlexanderYastrebov: I have noticed that we mostly check `production` elsewhere but I guess there is no difference if we revert the condition to disable for `production`?..

szuecs: we would switch e2e tests to v1 if we change to disable production, maybe it's better to do it as you pointed out

szuecs: updated the PR

mikkeloscar:
:+1:

AlexanderYastrebov:
Do we want to bump https://github.com/zalando-incubator/kubernetes-on-aws/blob/4cac0485e475ed02a2ce031cf597301a55a9201d/cluster/manifests/skipper/routesrv-deployment.yaml#L51 or we do it separately?

szuecs:
:+1:

PR #4801, opened 2021-11-30
allow reading ingress v1

szuecs:

### DIFF
--- a/cluster/manifests/skipper/rbac.yaml
+++ b/cluster/manifests/skipper/rbac.yaml
@@ -9,6 +9,13 @@ kind: ClusterRole
 metadata:
   name: skipper-ingress
 rules:
+- apiGroups:
+  - networking.k8s.io
+  resources:
+  - ingresses
+  verbs:
+  - get
+  - list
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

cluster/manifests/skipper/rbac.yaml
```diff
@@ -9,6 +9,13 @@ kind: ClusterRole
 metadata:
   name: skipper-ingress
 rules:
+- apiGroups:
+  - networking.k8s.io
+  resources:
+  - ingresses
+  verbs:
+  - get
+  - list
```
AlexanderYastrebov, line 18: IMO would be nice to use either yaml or json array notation consistently

AlexanderYastrebov:
:+1:

cluster/manifests/skipper/rbac.yaml
szuecs: yes I also thought the same and then I thought the part below we will drop after v1 migration...

mikkeloscar:
:+1:

PR #4804, opened 2021-12-01
kube-ingress-aws-controller: ingress v1 in test

szuecs:
kube-ingress-aws-controller switch to ingress v1 in all non production environments

### DIFF
--- a/cluster/manifests/ingress-controller/deployment.yaml
+++ b/cluster/manifests/ingress-controller/deployment.yaml
@@ -51,6 +51,7 @@ spec:
         - --load-balancer-type={{ .Cluster.ConfigItems.kube_aws_ingress_default_lb_type }}
 {{end}}
         - --cert-polling-interval={{ .Cluster.ConfigItems.kube_aws_ingress_controller_cert_polling_interval }}
+        - --ingress-api-version=networking.k8s.io/{{ .Cluster.ConfigItems.kube_aws_ingress_controller_ingress_version }}
         env:
         - name: CUSTOM_FILTERS
           value: "tag:kubernetes.io/cluster/{{ .Cluster.ID }}=owned tag:node.kubernetes.io/role=worker tag:zalando.org/ingress-enabled=true"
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

cluster/manifests/ingress-controller/deployment.yaml
```diff
@@ -51,6 +51,7 @@ spec:
         - --load-balancer-type={{ .Cluster.ConfigItems.kube_aws_ingress_default_lb_type }}
 {{end}}
         - --cert-polling-interval={{ .Cluster.ConfigItems.kube_aws_ingress_controller_cert_polling_interval }}
+        - --ingress-api-version=networking.k8s.io/{{ .Cluster.ConfigItems.kube_aws_ingress_controller_ingress_version }}
```
AlexanderYastrebov, line 54: There are three values possible
```go
	IngressAPIVersionExtensions       = "extensions/v1beta1"
	IngressAPIVersionNetworking       = "networking.k8s.io/v1beta1"
	IngressAPIVersionNetworkingV1     = "networking.k8s.io/v1"
```
so maybe it would be better to use full name verion as a config value, i.e.
```yaml
kube_aws_ingress_controller_ingress_version: "networking.k8s.io/v1beta1"
{{else}}
kube_aws_ingress_controller_ingress_version: "networking.k8s.io/v1"
{{end}}

--ingress-api-version={{ .Cluster.ConfigItems.kube_aws_ingress_controller_ingress_version }}
```

szuecs: The default in the controller is kubernetes.IngressAPIVersionNetworking ("networking.k8s.io/v1beta1") and we do not set it in the flags of the deployment `args:`, so "extensions/v1beta1" is not in use.
So actually we don't need to set the 3rd one which is the oldest API version.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

szuecs:
:+1:

szuecs:
:+1:

AlexanderYastrebov:
:+1:

M4NTR4:
:+1:

PR #4805, opened 2021-12-01
add ingress v1 e2e test

szuecs:

### DIFF
--- a/test/e2e/ingress.go
+++ b/test/e2e/ingress.go
@@ -22,6 +22,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
+	netv1 "k8s.io/api/networking/v1"
 	"k8s.io/api/networking/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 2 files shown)


test/e2e/ingress.go
```diff
@@ -596,7 +597,316 @@ var ____ = describe("Ingress tests custom routes", func() {
...
+
+		_, err = cs.NetworkingV1().Ingresses(ns).Get(context.TODO(), ing.Name, metav1.GetOptions{ResourceVersion: "0"})
+		Expect(err).NotTo(HaveOccurred())
+
+		// skipper http -> https redirect
+		By("Waiting for skipper route to default redirect from http to https, to see that our ingress-controller and skipper works")
+		err = waitForResponse(addr, "http", waitTime, isRedirect, true)
+		Expect(err).NotTo(HaveOccurred())
+
+		// ALB ready
```
AlexanderYastrebov, line 666: If we do not set LB type on ingress the type would be default LB type which could be NLB

AlexanderYastrebov:
:+1:

test/e2e/ingress.go
szuecs: yes true and we might want to convert the e2e tests in the future to default to NLB and use some specific tests for ALB.

szuecs:
:+1:

PR #4611, opened 2021-09-09
Consolidate more applications [3/X]

aermakov-zalando:
Change more stuff to use the `kubernetes` application:
 * `kube-aws-iam-controller`
 * `kube-metrics-adapter`
 * `metrics-server`
 * `csi-node-driver`/`ebs-csi-controller`
 * `kube-static-egress-controller`
* `kube-state-metrics`
* `kubernetes-lifecycle-metrics`
* `pdb-controller`
* `event-logger`
* `cronjob-monitor`
* `cronjob-fixer`
* `aws-node-decommissioner`
* `kube-downscaler`
* `kube-janitor`
* `cluster-lifecycle-controller`
* `efs-provisioner`
* `kube-node-ready-controller`
* `kube-cluster-autoscaler`/`autoscaling-buffer`

PR #4816, opened 2021-12-08
Uses `podManagementPolicy: Parallel` for skipper-ingress-redis

AlexanderYastrebov:
Converts `skipper-ingress-redis` into component to consolidate under the
same `skipper-ingress` application and to update `podManagementPolicy`.

`Parallel` pod management tells the StatefulSet controller to launch or
terminate all Pods in parallel, and to not wait for Pods to become
Running and Ready or completely terminated prior to launching or
terminating another Pod.

https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#parallel-pod-management

This should speedup Redis re-scaling

### DIFF
--- a/cluster/manifests/skipper/skipper-redis.yaml
+++ b/cluster/manifests/skipper/skipper-redis.yaml
@@ -2,7 +2,8 @@ apiVersion: apps/v1
 kind: StatefulSet
 metadata:
   labels:
-    application: skipper-ingress-redis
+    application: skipper-ingress
+    component: redis
     version: v6.2.4
   name: skipper-ingress-redis
   namespace: kube-system
@@ -11,12 +12,14 @@ spec:
   podManagementPolicy: Parallel
   selector:
     matchLabels:
-      application: skipper-ingress-redis
+      statefulset: skipper-ingress-redis
   serviceName: skipper-ingress-redis
   template:
     metadata:
       labels:
-        application: skipper-ingress-redis
+        statefulset: skipper-ingress-redis
+        application: skipper-ingress
+        component: redis
         version: v6.2.4
       annotations:
         cluster-autoscaler.kubernetes.io/safe-to-evict: "false"
@@ -42,7 +45,11 @@ spec:
               - key: application
                 operator: In
                 values:
-                - skipper-ingress-redis
+                - skipper-ingress
+              - key: component
+                operator: In
+                values:
+                - redis
       priorityClassName: "{{ .Cluster.ConfigItems.system_priority_class }}"
       containers:
       - image: registry.opensource.zalan.do/library/redis-6-alpine:6-alpine-20210712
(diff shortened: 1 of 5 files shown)


AlexanderYastrebov:
https://sunrise.zalando.net/cdp/gh/zalando-incubator/kubernetes-on-aws/l-jdxogqzbpm3o397u9s7qu3twu/steps/1/logs
```
time="2021-12-08T14:53:59Z" level=error msg="The StatefulSet \"skipper-ingress-redis\" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden" module=skipper
```

linki:
This probably happens during the e2e test's cluster update.

If it's fine for redis to be down and its state being dropped you can add it to `deletions.yaml`, the  but keep in mind that it runs for any kind of cluster update (e.g. config items changes). An alternative is to use an intermediate  second StatefulSet.

You can also do a `kubectl delete sts redis --cascade=false` to keep the pods alive. But this isn't supported in CLM. (Would probably need to add it to the DeleteOptions [here](https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/3a891a57dfaaf9ff3f4601ffd87dc3eb05261f00/provisioner/clusterpy.go#L1118))

AlexanderYastrebov:
@linki
> This probably happens during the e2e test's cluster update.

Yes

We have another goal to consolidate skipper-ingress-redis into skipper-ingress application. I am experimenting in this PR if that would work.

> If it's fine for redis to be down and its state being dropped you can add it to deletions.yaml, the but keep in mind that it runs for any kind of cluster update (e.g. config items changes). An alternative is to use an intermediate second StatefulSet.

I have added old non-component statefulset to deletions so it should be deleted only once on rollout.

cluster/manifests/skipper/skipper-redis.yaml
```diff
@@ -2,20 +2,24 @@ apiVersion: apps/v1
...
   name: skipper-ingress-redis
   namespace: kube-system
 spec:
   replicas: {{ .ConfigItems.skipper_redis_replicas }}
+  podManagementPolicy: Parallel
   selector:
     matchLabels:
-      application: skipper-ingress-redis
+      application: skipper-ingress
+      component: redis
```
mikkeloscar: If you change this anyway, you may want to go with what we did for a bunch of other components where we use the matchLabel: `<resource-type>=<name>` e.g. `statefulset=skipper-ingress-redis`. That let's you decouple `application` and `component` from the selector which is only really about ownership between statefulset and pods and you can freely adjust `application` and `component` should it be needed in the future.

Ref: https://github.com/zalando-incubator/kubernetes-on-aws/pull/4611

szuecs: We also need to check zmon alerts/checks, such that we do not catch by application label all the things

AlexanderYastrebov:
Closing for now as component change requires preparation of the monitoring changes.

AlexanderYastrebov:
Created https://github.com/zalando-incubator/kubernetes-on-aws/pull/4819 that does not modify application/component labels

PR #4818, opened 2021-12-09
Update cluster-autoscaler to internal.37

linki:
* Fixes an issue that the autoscaler stops doing anything when it encounters an unknown instance type
* Makes sure to fetch the instance template from the most up-to-date instance type list (e.g. dynamically fetched)

See https://github.com/zalando-incubator/autoscaler/pull/88


mikkeloscar:
:+1:

linki:
👍

PR #4819, opened 2021-12-09
Uses `podManagementPolicy: Parallel` for skipper-ingress-redis [take 2]

AlexanderYastrebov:
Similar to https://github.com/zalando-incubator/kubernetes-on-aws/pull/4816 but without application/component label changes.

`Parallel` pod management tells the StatefulSet controller to launch or
terminate all Pods in parallel, and to not wait for Pods to become
Running and Ready or completely terminated prior to launching or
terminating another Pod.

https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#parallel-pod-management

This should speedup Redis re-scaling.

To avoid
```
updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden
```
due to change of `podManagementPolicy` the statefulset is recreated via
pre-apply deletion that runs once.

Also changes selector `matchLabels` to allow template `application` label change in the future.

### DIFF
--- a/cluster/manifests/skipper/skipper-redis.yaml
+++ b/cluster/manifests/skipper/skipper-redis.yaml
@@ -3,18 +3,25 @@ kind: StatefulSet
 metadata:
   labels:
     application: skipper-ingress-redis
-    version: v6.2.4
+    #
+    # skipper-ingress-redis podManagementPolicy update
+    # - version is updated with `-parallel` suffix to not match the version configured in deletions.yaml
+    # - `-parallel` suffix can be dropped after completion along with deletions.yaml cleanup
+    #
+    version: v6.2.4-parallel
   name: skipper-ingress-redis
   namespace: kube-system
 spec:
   replicas: {{ .ConfigItems.skipper_redis_replicas }}
+  podManagementPolicy: Parallel
   selector:
     matchLabels:
-      application: skipper-ingress-redis
+      statefulset: skipper-ingress-redis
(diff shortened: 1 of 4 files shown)


cluster/manifests/skipper/skipper-redis.yaml
```diff
@@ -3,18 +3,25 @@ kind: StatefulSet
 metadata:
   labels:
     application: skipper-ingress-redis
-    version: v6.2.4
+    #
+    # skipper-ingress-redis podManagementPolicy update
+    # - version is updated with `-parallel` suffix to not match the version configured in deletions.yaml
+    # - `-parallel` suffix can be dropped after completion along with deletions.yaml cleanup
+    #
+    version: v6.2.4-parallel
```
szuecs, line 11: As far as I know the deletions would be applied before applying manifests so it should not be a problem ti have the same version.

AlexanderYastrebov: If this has the same version it will delete sts on every config change while we need to delete the old one (`version: v6.2.4`) only once

AlexanderYastrebov:
The https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/540 introduces support for `propagationPolicy` for the deletions which should enable statefulset re-creation without terminating pods (i.e. without downtime). I'll create an alternative PR when propagation policy is supported.

PR #4821, opened 2021-12-10
Update KubeMetricsAdapter

katyanna:
https://github.com/zalando-incubator/kube-metrics-adapter/pull/376
- Use labels hash in the custom metrics store
- Define stronger types for metrics store 💅

### DIFF
--- a/cluster/manifests/kube-metrics-adapter/deployment.yaml
+++ b/cluster/manifests/kube-metrics-adapter/deployment.yaml
@@ -27,7 +27,7 @@ spec:
       serviceAccountName: custom-metrics-apiserver
       containers:
       - name: kube-metrics-adapter
-        image: registry.opensource.zalan.do/teapot/kube-metrics-adapter:v0.1.17
+        image: registry.opensource.zalan.do/teapot/kube-metrics-adapter:v0.1.17-6-g9d8359b
         env:
         - name: AWS_REGION
           value: {{ .Region }}


cluster/manifests/01-admission-control/teapot.yaml
```diff
@@ -132,7 +132,7 @@ webhooks:
     rules:
       - operations: ["CREATE", "UPDATE"]
         apiGroups: ["extensions", "networking.k8s.io"]
-        apiVersions: ["v1beta1"]
+        apiVersions: ["v1"]
```
mikkeloscar: This change also needs a change to the admission-controller code because we currently hardcode `v1beta1` only. We should also remove `extentions` from the `apiGroups` above as there is not v1 `extensions`.

gargravarr:
:+1:

katyanna:
:+1:

cluster/manifests/kube-metrics-adapter/deployment.yaml
```diff
@@ -27,7 +27,7 @@ spec:
       serviceAccountName: custom-metrics-apiserver
       containers:
       - name: kube-metrics-adapter
-        image: registry.opensource.zalan.do/teapot/kube-metrics-adapter:v0.1.17
+        image: registry.opensource.zalan.do/teapot/kube-metrics-adapter:v0.1.17-6-g9d8359b
```
mikkeloscar, line 30: Would be nice to tag a new release `v0.1.18` now that the bug is fixed. Not only for us, but also for our open source users. :)

PR #4846, opened 2021-12-19
fix: rules should be updated more often

szuecs:
fix: evaluation_interval is one minute but we scrape every 15s and we loose granularity


szuecs:
:+1:

mikkeloscar:
:+1:

PR #4858, opened 2022-01-03
Handle CRDs with overlapping plural names

mikkeloscar:
When creating/updating/deleting CRDs our admission-controller will update corresponding RBAC roles to give users, edit/read access to the CRD based resources.

Before the roles were generated with a name based on the plural name defined in the CRD. Since this is not unique across CRDs in a cluster we could get overlapping roles meaning the role would only work for one of the CRD types. This update to the admission-controller fixes this by using the CRD names which are cluster unique for the role name generation.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #4865, opened 2022-01-05
default to nlb

szuecs:
all clusters in cluster-registry have their current settings set, only applies to new created clusters


mikkeloscar:
:+1:

szuecs:
:+1:

PR #4874, opened 2022-01-10
Fix DENY policy so oversight is easier

msooszalando:
Currently DENY policy makes it difficult to properly govern this bucket. In particular, oversight systems cannot see the configuration. This makes it difficult to spot issues and have confidence all is good.

NOTE: This does not grant access. One still needs access to be able to be able to log in to the bucket. However, this explicitly denies listing, getting, deleting, and putting objects.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #4880, opened 2022-01-11
Enable OIDC provider via CF by default

mikkeloscar:
Follow up to #4861 to enable OIDC Provider management via cloudformation by default everywhere.

This will not work for e2e because we rely on CLM retrying the applying of the stack as it would fail the first time as explained in #4861 So we need to merge this through all channels to avoid breaking e2e for other PRs. 

Depends on #4861 being fully rolled out.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -389,9 +389,7 @@ Resources:
 
 # end of vpc dependent resources
 {{ end }}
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #4881, opened 2022-01-11
Add support for pulling images directly from ECR

linki:
Adds support for pulling images from ECR by granting the correct IAM permissions on our side.


mikkeloscar:
:+1:

linki:
👍

PR #4847, opened 2021-12-20
Adds `statefulset: skipper-ingress-redis` label

AlexanderYastrebov:
This is a preparation step for `skipper-ingress` application
consolidation, see similar #4611

This allows `matchLabels` to use unique `statefulset: skipper-ingress-redis`
label instead of the `application` label.

The change of `matchLabels` requires statefulset re-creation which would
be performed separately along with `podManagementPolicy` change, see
https://github.com/zalando-incubator/kubernetes-on-aws/pull/4819#issuecomment-997974635

Signed-off-by: Alexander Yastrebov <alexander.yastrebov@zalando.de>

PR #4885, opened 2022-01-12
skipper-ingress: update `selector.matchLabels`

AlexanderYastrebov:
`skipper-ingress` HPA, although targets deployment by name, uses its
selector labels to get metrics
https://github.com/kubernetes/kubernetes/issues/78761#issuecomment-670815813
which also match `routesrv` pods.

Updates `selector.matchLabels` to use `kind: name` label that matches
only `ingress` component pods (similar to #4611 and #4847).

Deployment selector is immutable therefore update requires deployment re-creation.
To re-create deployment without deleting pods `propagation_policy: Orphan` is used.

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -22,6 +22,20 @@ pre_apply:
   namespace: kube-system
   kind: StatefulSet
   propagation_policy: Orphan
+#
+# skipper-ingress selector.matchLabels update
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
Rolled back by #4887

PR #4893, opened 2022-01-13
Build on `dev` merges to recreate pipeline cache

AlexanderYastrebov:

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -4,7 +4,8 @@ pipeline:
 - id: build
   depends_on: []
   when:
-    event: pull_request
+  - branch: dev
+  - event: pull_request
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
The build condition was previously changed by #1973.
The `cache` configuration was already there so apparently after that change cache stopped to be updated (cache is saved/persisted only for pipeline runs that resulted from a push to a protected branch) and then at some point old cache became incompatible with the build step and started to produce
```
Downloading cache
Downloaded cache archive (1.7G)
zstd: /tmp/cache.tar.zst: unsupported format 
```
within `Setting up environment` prestep.

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #4897, opened 2022-01-14
Enable mlbuilder support in deployment service by default

mikkeloscar:
This is now at a stage where it can be enabled everywhere. Aligned with the responsible team.

Avoids us having to enable it on a case-by-case basis.


mikkeloscar:
:+1:

elgalu:
👍🏽

gargravarr:
:+1:

PR #4898, opened 2022-01-14
Consolidate DaemonSet applications/components [1/3]

mikkeloscar:
This is the first of a 3-step change to migrate 6 daemonsets to be components of `kubernetes` instead of individual applications.

1. Label the pod spec with `daemonset: <name>`, roll out to all clusters and wait for pods/nodes to be completely rotated.
2. Change the selector to use the label `daemonset: <name>` and use the deletions.yaml with `propagation_policy: Orphan` to delete the old daemonset (without deleting the pods) and create a new daemonset with the new selector.
3. Change the pod template labels for `application` and `component` and wait for all nodes/pods to be rotated.

### DIFF
--- a/cluster/manifests/coredns-local/daemonset-coredns.yaml
+++ b/cluster/manifests/coredns-local/daemonset-coredns.yaml
@@ -5,10 +5,11 @@ metadata:
   namespace: kube-system
   labels:
     application: coredns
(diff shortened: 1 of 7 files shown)


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #4899, opened 2022-01-14
Allow using session manager for connecting to instances

mikkeloscar:
This enables the use of AWS session manager as an alternative to connect to kubernetes nodes compared to ssh.

Ref: https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager.html

Essentially you would be able to do the following:

```
aws ssm start-session --target <instance-id-of-node>
```

And you will get a shell on the node. Eventually this could enable us to close port 22 on the nodes. For now it's just to test how well it works.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -450,6 +450,15 @@ Resources:
                   - 'ecr:GetAuthorizationToken'
                 Effect: Allow
                 Resource: '*'
+              # allow connecting via Session Manager
+              - Action:
+                - "ssm:UpdateInstanceInformation"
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

gargravarr:
:+1:

mikkeloscar:
I had to double check that this won't allow non admin users access to connect to Kubernetes nodes via session manager. This access was possible but is now blocked based on the instance tag `InfrastructureComponent=true`. Therefore it's now fine to proceed with this :)

mikkeloscar:
:+1:

szuecs:
:+1:

PR #4900, opened 2022-01-14
Consolidate prometheus into kubernetes application [1/3]

mikkeloscar:
This is the first of a 3-step change to migrate prometheus statefulset to be a component of `kubernetes` instead of individual an application.

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -41,6 +41,12 @@ pre_apply:
 #   namespace: kube-system
 #   kind: DaemonSet
 #   propagation_policy: Orphan
+# step 2 (prometheus consolidation)
+# - labels:
+#     application: prometheus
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

gargravarr:
:+1:

cluster/manifests/deletions.yaml
```diff
@@ -10,6 +10,13 @@ pre_apply:
   kind: Deployment
 {{ end }}
 
+# step 2
```
gargravarr: Will step 2 be a single PR?

mikkeloscar: Yes, adding the deletions here and making the label selector changes in the statefulset will be done in a single PR.

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #4903, opened 2022-01-17
Double check that e2e is failing on missing permissions

linki:
To ensure the e2e test in https://github.com/zalando-incubator/kubernetes-on-aws/pull/4881 works correctly.


linki:
It works.

PR #4908, opened 2022-01-19
Fix DENY policy so oversight is easier

linki:
https://github.com/zalando-incubator/kubernetes-on-aws/pull/4874 with e2e tests

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -807,15 +807,14 @@ Resources:
               AWS: "{{.Cluster.ConfigItems.deployment_service_api_role_arn}}"
 {{- end }}
           - Action:
-              - s3:Put*
-              - s3:Create*
-              - s3:Update*
-              - s3:List*
-              - s3:GetObject*
-              - s3:Delete*
-              - s3:Restore*
-              - s3:Update*
-              - s3:Replicate*
+              - 's3:Put*'
+              # - 's3:Create*'
+              # - 's3:Update*'
+              - 's3:List*'
+              - 's3:GetObject*'
+              - 's3:Delete*'
+              - 's3:Restore*'
+              - 's3:Replicate*'
             Effect: Deny
             Resource:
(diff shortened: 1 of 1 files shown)


msooszalando:
Ah, the problem is the single quotes maybe?

linki:
No, I thought that initially, too, but that's not it.

I tracked it down to the two actions that I commented out. If you pick one of them it will fail. (The e2e tests are currently failing for another reason.)

If there's no fundamental reason why you cannot Deny those actions (e.g. List on a Bucket item might be invalid?) then I suspect it might just contradict some other statement in the role?

linki:
@msooszalando Do you still need this. We had to rollback your initial PR https://github.com/zalando-incubator/kubernetes-on-aws/pull/4874

linki:
@idotan Do you still need this? We had to revert the original PR https://github.com/zalando-incubator/kubernetes-on-aws/pull/4874

mikkeloscar:
@linki I believe it fails because `Create` and `Update` actions are not relevant for a bucket resource. E.g. the create/update actions are the following:

```
CreateAccessPoint
CreateAccessPointForObjectLambda
CreateBucket
CreateJob
CreateMultiRegionAccessPoint
UpdateJobPriority
UpdateJobStatus
```

Which doesn't make sense to do on a bucket I guess.

The better solution here would be to use `NotAction` and only list the action we want to allow with this change, instead of trying to list all the ones to be denied.

https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html

linki:
Can we close this?

This was an attempt to bring back the changes of https://github.com/zalando-incubator/kubernetes-on-aws/pull/4874 (by @msooszalando) which got reverted.

I don't know what the purpose actually was.

mikkeloscar:
The goal was to read bucket policy information e.g. by the security audit role, but since this hasn't been brought up in 6 months, let's close it for now.

PR #4911, opened 2022-01-19
Add more gpu instance types to increase chances for spot

linki:
e2e tests are currently failing very often due to spot shortage on GPU nodes.

We [recently reduced](https://github.com/zalando-incubator/kubernetes-on-aws/pull/4790/files#diff-9a99d741a98cf62976440d87f64f503aab0266cd1e406365c9cd05c7ccdf012e) the possible GPU instance types in e2e clusters which might be a contributing factor.

Let's add a couple more instance types with GPUs to increase our chances.

### DIFF
--- a/test/e2e/cluster_config.sh
+++ b/test/e2e/cluster_config.sh
@@ -97,7 +97,7 @@ clusters:
       labels: dedicated=node-tests
       taints: dedicated=node-tests:NoSchedule
   - discount_strategy: spot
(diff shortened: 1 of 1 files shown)


mikkeloscar:
@linki Maybe we want a fallback node pool on on-demand? Risk is ofc. that it will always be used and it costs more, but right now it's not very efficient when all the tests fail :/

linki:
@mikkeloscar Good idea, added. Let's keep an eye on the cost to make sure it doesn't explode for some reason.

linki:
👍

katyanna:
👍

PR #4917, opened 2022-01-21
Config items to migrate Etcd security group.

gargravarr:
Adds two config items:
* `etcd_stack_security_group_temp`: when set to `true`, creates a temporary Etcd security group.
*  `cluster_reference_etcd_sg_temp`: controls which etcd security group to reference in the main cluster CloudFormation template.

This will allow for clusters _without_ the `VpcId` set in the ETCD security grroup to use the pure CloudFormation etcd-`stack.yaml`, when we set `experimental_new_etcd_stack` to `true`. 

### DIFF
--- a/cluster/etcd-stack.yaml
+++ b/cluster/etcd-stack.yaml
@@ -10,6 +10,13 @@ Outputs:
     Value: !GetAtt EtcdSecurityGroup.GroupId
     Export:
       Name: "etcd-cluster-etcd:security-group-id"
+{{- if eq .Cluster.ConfigItems.etcd_stack_security_group_temp "true"}}
+  EtcdSecurityGroupIdTemp:
+    Description: "Security Group ID of the etcd cluster"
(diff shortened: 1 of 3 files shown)


cluster/etcd-stack.yaml
```diff
@@ -144,9 +151,29 @@ Resources:
...
+        ToPort: 2381
+        CidrIp: "{{.Values.vpc_ipv4_cidr}}"
+      - IpProtocol: tcp
+        FromPort: 9100
+        ToPort: 9100
+        CidrIp: "{{.Values.vpc_ipv4_cidr}}"
+      VpcId: "{{.Cluster.ConfigItems.vpc_id}}"
       Tags:
         - Key: InfrastructureComponent
           Value: true
```
mikkeloscar, line 178: nit: seems like the `Tags` is missing from the old security group with this diff. It should be on both

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #4926, opened 2022-01-24
Include EtcdIngress in the template.

gargravarr:
The SecurityGroupIngress also depends on the etcd security group. This PR allows controlling which security group is part of the ingress.


gargravarr:
:+1:

mikkeloscar:
:+1:

PR #4929, opened 2022-01-25
Fiy yet another etcd template

gargravarr:
Fix templating missing from https://github.com/zalando-incubator/kubernetes-on-aws/pull/4926

### DIFF
--- a/cluster/etcd-stack.yaml
+++ b/cluster/etcd-stack.yaml
@@ -190,10 +190,10 @@ Resources:
       FromPort: 2379
       ToPort: 2479
 {{- if eq .Cluster.ConfigItems.etcd_stack_security_group_temp "false"}}
-      GroupId: !GetAtt EtcdSecurityGroup.GroupId
+      SourceSecurityGroupId: !GetAtt EtcdSecurityGroup.GroupId
 {{- end }}
 {{- if eq .Cluster.ConfigItems.etcd_stack_security_group_temp "true"}}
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

cluster/etcd-stack.yaml
```diff
@@ -189,7 +189,12 @@ Resources:
       IpProtocol: tcp
       FromPort: 2379
       ToPort: 2479
+{{- if eq .Cluster.ConfigItems.etcd_stack_security_group_temp "false"}}
       SourceSecurityGroupId: !GetAtt EtcdSecurityGroup.GroupId
+{{- end }}
+{{- if eq .Cluster.ConfigItems.etcd_stack_security_group_temp "true"}}
+      SourceSecurityGroupId: !GetAtt EtcdSecurityGroupTemp.GroupId
+{{- end }}
```
mikkeloscar, line 195: Maybe it's better to copy the whole `EtcdIngressMembers` resource into `EtcdIngressMembersTemp` so you can have both Security Groups with all settings at the same time, otherwise I expect you break one security group during the switch.

gargravarr: But can we have two `AWS::EC2::SecurityGroupIngress` resources?

gargravarr: I simplified it.

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #4934, opened 2022-01-25
Consolidate prometheus into kubernetes application [2/3]

mikkeloscar:
Follow up to #4900

This is the second of a 3-step change to migrate prometheus statefulset to be a component of `kubernetes` instead of individual an application.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #4937, opened 2022-01-26
skipper/routesrv: Delete routesrv deployment when not enabled

AlexanderYastrebov:
Signed-off-by: Alexander Yastrebov <alexander.yastrebov@zalando.de>

PR #4940, opened 2022-01-26
ingress/routesrv: Makes skipper_routesrv_enabled a three state switch

AlexanderYastrebov:

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -107,7 +107,13 @@ skipper_redis_write_timeout: "25ms"
 
 skipper_cluster_ratelimit_max_group_shards: 1
 
+#
 # skipper routesrv settings
+#
+# skipper_routesrv_enabled is a three state switch:
+# - "false" - routesrv deployment is removed, skipper uses own k8s dataclient
+# - "pre" - routesrv is deployed, skipper uses own k8s dataclient
+# - "exec" - routesrv is deployed, skipper uses routesrv
 {{if eq .Cluster.Environment "production"}}
 skipper_routesrv_enabled: "false"
 skipper_routesrv_replicas: 3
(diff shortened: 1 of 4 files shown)


cluster/manifests/skipper/routesrv-deployment.yaml
```diff
@@ -1,4 +1,4 @@
-{{ if ne .ConfigItems.skipper_routesrv_enabled "true" }}
+{{ if ne .ConfigItems.skipper_routesrv_enabled "false" }}
```
AlexanderYastrebov, line 1: Fixes a wrong condition introduced by #4937

cluster/config-defaults.yaml
```diff
@@ -107,7 +107,13 @@ skipper_redis_write_timeout: "25ms"
 
 skipper_cluster_ratelimit_max_group_shards: 1
 
+#
 # skipper routesrv settings
+#
+# skipper_routesrv_enabled is a three state switch:
+# - "false" - routesrv deployment is removed, skipper uses own k8s dataclient
+# - "pre" - routesrv is deployed, skipper uses own k8s dataclient
+# - "exec" - routesrv is deployed, skipper uses routesrv
```
AlexanderYastrebov, line 116: Used `exec` instead of `true` similar to `nlb_switch` and to hint that its a three state config

cluster/manifests/deletions.yaml
```diff
@@ -130,7 +130,7 @@ post_apply:
   kind: ClusterRoleBinding
 {{- end }}
 
-{{ if ne .Cluster.ConfigItems.skipper_routesrv_enabled "true" }}
+{{ if .Cluster.ConfigItems.skipper_routesrv_enabled "false" }}
```
mikkeloscar: missing `eq`

cluster/config-defaults.yaml
szuecs: no that's not needed here. skipper-ingress waits for first successful route load and routesrv has the time to startup before it hurts skipper-ingress.

aryszka:
:+1:

mikkeloscar:
:+1:

PR #4941, opened 2022-01-27
Update CoreDNS to v1.8.7

mikkeloscar:
https://coredns.io/2021/12/09/coredns-1.8.7-release/
https://coredns.io/2021/10/07/coredns-1.8.6-release/
https://coredns.io/2021/09/10/coredns-1.8.5-release/

PR #4945, opened 2022-01-28
Consolidate prometheus into kubernetes application [3/3]

mikkeloscar:
Follow up to #4900, #4934

This is the third of a 3-step change to migrate prometheus statefulset to be a component of `kubernetes` instead of individual an application.

Depends on #4934 being fully rolled out!

### DIFF
--- a/cluster/manifests/prometheus/service.yaml
+++ b/cluster/manifests/prometheus/service.yaml
@@ -4,7 +4,8 @@ metadata:
   name: prometheus
   namespace: kube-system
   labels:
-    application: prometheus
+    application: kubernetes
+    component: prometheus
 spec:
   type: ClusterIP
   ports:
@@ -12,4 +13,4 @@ spec:
       targetPort: 9090
       protocol: TCP
   selector:
-    application: prometheus
+    component: prometheus
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -57,12 +57,6 @@ pre_apply:
 #   namespace: kube-system
 #   kind: DaemonSet
 #   propagation_policy: Orphan
-# step 2 (prometheus consolidation)
-- labels:
-    application: prometheus
-  namespace: kube-system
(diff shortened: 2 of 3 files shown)


mikkeloscar:
:+1:

cluster/manifests/coredns-local/daemonset-coredns.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
     application: coredns
     # application: kubernetes # step 2
     component: coredns
-    version: v1.8.4
+    version: v1.8.7
```
szuecs: this is not part of the message. Do you want to change the message and description to make this through?
Happy to have this too within this PR, because of e2e runtime

mikkeloscar: Weird, I somehow included this in a rebase, that was a mistake, thanks for noticing it!

It was already rolled out here: https://github.com/zalando-incubator/kubernetes-on-aws/pull/4941 :shrug: 

I have fixed the commits to not include this.

mikkeloscar:
:+1:

cluster/manifests/prometheus/service.yaml
```diff
@@ -4,12 +4,13 @@ metadata:
...
+    component: prometheus
 spec:
   type: ClusterIP
   ports:
     - port: 80
       targetPort: 9090
       protocol: TCP
   selector:
-    application: prometheus
+    component: prometheus
```
AlexanderYastrebov, line 16: Why not match both application+component?

mikkeloscar: I expect a short downtime during rollout because it takes a while to roll the prometheus pods

I would add the `application` label in a follow up. (so this should be [3/4] really).

AlexanderYastrebov: Ah, I guess service should match pods before and after labels update

katyanna:
:+1:

PR #4949, opened 2022-01-28
default NLB with internal annotation should fallback to ALB

szuecs:
- default NLB with internal annotation should fallback to ALB https://github.com/zalando-incubator/kube-ingress-aws-controller/releases/tag/v0.12.16
- can be overridden by users with lb annotation set to nlb


szuecs:
:+1:

gargravarr:
:+1:

PR #4955, opened 2022-01-31
skipper-ingress: cleanup after `selector.matchLabels` update

AlexanderYastrebov:
- [ ] requires #4948 rollout


AlexanderYastrebov:
Will create a new one

PR #4956, opened 2022-01-31
skipper add profilers

szuecs:
- add block and mutex profilers which are disabled by default
- memory profiler is now configurable

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -69,6 +69,11 @@ skipper_default_filters: 'enableAccessLog(4,5) -> lifo(2000,20000,"3s")'
 skipper_ingress_inline_routes: ""
 skipper_ingress_refuse_payload: ""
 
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

szuecs:
Because stackset tests work it seems reasonable to say this change was tested successfully. e2e shows no logs but since 14h in progress, weird

AlexanderYastrebov:
I am not sure its a good idea to change skipper deployment until we rollout selector update #4948
Changing template will trigger rolling update but I would prefer we change one thing at a time, I've also tested the rollout without changing the template.

mikkeloscar:
:+1:

AlexanderYastrebov:
> I am not sure its a good idea to change skipper deployment until we rollout selector update #4948

#4948 was rolled back

szuecs:
:+1:

PR #4912, opened 2022-01-20
Update Ubuntu Image

katyanna:
- Terminate restarted nodes
  A restarted node must not be trusted, it should be deleted.

PR #4958, opened 2022-02-01
E2e update to v1.21.9

mikkeloscar:
* Update e2e build to Kubernetes v1.21.9 to be aligned with the AMI update from #4912
* Cleanup some old build settings e.g. `GO111MODULE=on` which is no longer relevant
* ~Switch to CDP runtime instead of overlay.~


mikkeloscar:
:+1:

katyanna:
👍

PR #4959, opened 2022-02-01
Revert "skipper-ingress: update `selector.matchLabels`"

AlexanderYastrebov:
There is an edge case when HPA does not have status which results in
deployment replicas set to zero by the `zalando.org/create-using-hpa-replicas` which triggers implicit maintenance mode
and prevents HPA from scaling up the deployment.
The edge case is observed during e2e tests or on the new clusters.
See https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#implicit-maintenance-mode-deactivation

This reverts commit 2c54a541509b5eb09cd2b2ab36fe070a304587b4.

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -63,17 +63,15 @@ pre_apply:
   namespace: kube-system
   kind: StatefulSet
   propagation_policy: Orphan
+
 #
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #4963, opened 2022-02-02
admission-controller: update version to ignore zero HPA replicas on deployment creation

AlexanderYastrebov:
admission-controller supports `zalando.org/create-using-hpa-replicas: <HPA name>` annotation that sets deployment replicas from HPA desired replicas. This change ignores zero desired replicas value to avoid moving deployment into [implicit maintenance mode](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#implicit-maintenance-mode-deactivation).


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #4969, opened 2022-02-07
Drop permission to gerry mint bucket credentials for ZMON

mikkeloscar:
This was used in the past by ZMON, but since September 2021 this is no longer in use, so drop the unneeded permissions.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #4970, opened 2022-02-07
Enable gerry (zmon user) to monitor coredns endpoints from the outside

mikkeloscar:
This role is used to check coredns endpoints of a cluster from the outside.
The intention is to monitor a cluster from the outside in case DNS is broken and ZMON inside the cluster would not work.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #4975, opened 2022-02-08
Consolidate DaemonSet applications/components [2/3]

mikkeloscar:
This is the second of a 3-step change to migrate 6 daemonsets to be components of `kubernetes` instead of individual applications.

1. Label the pod spec with `daemonset: <name>`, roll out to all clusters and wait for pods/nodes to be completely rotated.
2. Change the selector to use the label `daemonset: <name>` and use the deletions.yaml with `propagation_policy: Orphan` to delete the old daemonset (without deleting the pods) and create a new daemonset with the new selector.
3. Change the pod template labels for `application` and `component` and wait for all nodes/pods to be rotated.

Ref: https://github.com/zalando-incubator/kubernetes-on-aws/pull/4898

This also removes the `node-monitor` service as it was never really used for anything.


mikkeloscar:
:+1:

cluster/manifests/node-monitor/service.yaml
```diff
@@ -1,20 +0,0 @@
-kind: Service
```
gargravarr, line 1: Is this service to disappear? Shouldn't you apply similar selector / labels as in the other services?

mikkeloscar: Forgot to mention why this is to be removed. It was never really used for anything so it can just be dropped. I added this to the description on top now.

cluster/manifests/node-monitor/daemonset.yaml
```diff
@@ -4,16 +4,14 @@ metadata:
   name: node-monitor
```
gargravarr, line 4: Then should this daemonset disappear as well?

mikkeloscar: It's just the service that is not used. The daemonset itself is.

The main motivation for removing an unused service is that kube-proxy have to generate iptable rules for service -> pod routing on EVERY node. And when we don't use the service this can be avoided.

gargravarr:
:+1:

cluster/manifests/node-monitor/service.yaml
szuecs: this broke monitoring to gather data prometheus -> node-exporter and cadvisor

PR #4976, opened 2022-02-08
Add application label selector to Prometheus service

mikkeloscar:
Follow up to #4945

Use both `application` and `component` label for service selector.


gargravarr:
:+1:

mikkeloscar:
:+1:

PR #4982, opened 2022-02-10
ingress controller handles fabricgateway resources 

szuecs:
ingress controller handles fabricgateway resources to create cloud load balancers and adds status to fabricgateway resources, but this will be overwritten by fabric controller. If we stop fabric controller the status field would stay, such that we can create DNS records with that data


szuecs:
:+1:

mikkeloscar:
:+1:

PR #4989, opened 2022-02-14
Tag control plane EC2 instances with application/component tags

mikkeloscar:
In order to make it simpler to see control plane cost in AWS cost explorer tag the control plane instances with `application=kubernetes,component=control-plane`.

Use the generic component name `control-plane` as we run multiple components on the node, but just need a way to identify the whole EC2 instance in AWS Cost explorer costs.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #5020, opened 2022-02-18
add step: start loadtest and query/check the result

szuecs:

### DIFF
--- a/test/e2e/loadtest/client/loadtest-config.yaml
+++ b/test/e2e/loadtest/client/loadtest-config.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    application: e2e-vegeta
+  name: e2e-vegeta
+  namespace: loadtest-e2e
+data:
+  routes.eskip: |
+    h: PathSubtree("/healthz") -> status(200) -> inlineContent("OK") -> <shunt>;
+    r: * -> "https://%TARGET%.%ZONE%";
+  get.txt: |
+    GET http://127.0.0.1:9090
--- a/test/e2e/loadtest/client/prom-ingress.yaml
+++ b/test/e2e/loadtest/client/prom-ingress.yaml
@@ -1,0 +1,18 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: prometheus
+  namespace: loadtest-e2e
+  labels:
+    application: prometheus
+spec:
+  rules:
+  - host: prometheus-%TARGET%.%ZONE%
+    http:
+      paths:
+      - backend:
+          service:
+            name: prometheus
+            port:
+              number: 80
+        pathType: ImplementationSpecific
--- a/test/e2e/loadtest/client/prom-statefulset.yaml
+++ b/test/e2e/loadtest/client/prom-statefulset.yaml
@@ -1,0 +1,105 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  labels:
+    application: loadtest-prometheus
+    version: v2.26.0
+  namespace: loadtest-e2e
+  name: loadtest-prometheus
+spec:
+  replicas: 1
+  podManagementPolicy: Parallel
+  selector:
+    matchLabels:
+      application: loadtest-prometheus
+  serviceName: loadtest-prometheus
+  template:
+    metadata:
+      labels:
+        application: loadtest-prometheus
+        version: v2.26.0
+      annotations:
+        prometheus.io/path: /metrics
+        prometheus.io/port: "9090"
+        prometheus.io/scrape: "true"
+    spec:
+      serviceAccountName: loadtest-prometheus
+      dnsConfig:
+        options:
+        - name: ndots
+          value: "1"
+      initContainers:
+      - name: generate-config
+        image: registry.opensource.zalan.do/library/alpine-3.13:3.13-20210216
+        command:
+        - /bin/sh
+        args:
+        - -c
+        - sed s/'@@POD_NAME@@'/${POD_NAME}/g /etc/prometheus/prometheus.yml > /prometheus/prometheus.yaml ; cp /etc/prometheus/prometheus.rules.yml /prometheus/prometheus.rules.yaml
+        env:
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        resources:
+          limits:
+            cpu: 1m
+            memory: 128Mi
+        volumeMounts:
+        - name: prometheus-loadtest-config-volume
+          mountPath: /etc/prometheus
+        - name: prometheus-loadtest-storage-volume
+          mountPath: /prometheus
+      containers:
+      - name: prometheus
+        image: registry.opensource.zalan.do/teapot/prometheus:v2.26.0
+        args:
+        - --config.file=/prometheus/prometheus.yaml
+        - --storage.tsdb.path=/prometheus/
+        - --storage.tsdb.retention.time=7d
+        - --storage.tsdb.wal-compression
+        - --storage.tsdb.min-block-duration=30m
+        ports:
+        - name: ingress-port
+          containerPort: 9090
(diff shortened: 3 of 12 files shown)


szuecs:
:+1:

test/e2e/loadtest/client/prom-statefulset.yaml
```diff
@@ -0,0 +1,105 @@
+apiVersion: apps/v1
+kind: StatefulSet
```
AlexanderYastrebov, line 2: Do we need separate prometheus? Why not use the existing one?

test/e2e/run_e2e.sh
```diff
@@ -114,6 +119,11 @@ if [ "$create_cluster" = true ]; then
 
         # Wait for the resources to be ready
         ./wait-for-update.py --timeout 1200
+
+        # provision and start load test
+        echo "provision and start load test"
+        ZONE=$(echo "${API_SERVER_URL}" | awk -F . '{print substr($0, index($0,$2))}')
```
AlexanderYastrebov:
```sh
export API_SERVER_URL="https://${LOCAL_ID}.${HOSTED_ZONE}"
```
`ZONE` and `HOSTED_ZONE` are the same, aren't they?

test/e2e/loadtest/client/prom-ingress.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: prometheus
+  namespace: loadtest-e2e
+  labels:
+    application: prometheus
+spec:
+  rules:
+  - host: prometheus-%TARGET%.%ZONE%
```
AlexanderYastrebov, line 10: Do we need public host maybe we could use cluster.local?
Then we could also hardcode it in the verification step as it would be the same for any cluster

test/e2e/run_e2e.sh
```diff
@@ -231,6 +241,37 @@ if [ "$stackset_e2e" = true ]; then
     E2E_NAMESPACE="${namespace}" ./stackset-e2e -test.parallel 20
 fi
 
+if [ "$loadtest_e2e" = true ]; then
+  >&2 echo "collect loadtest e2e data"
+  prometheus=$(kubectl -n loadtest-e2e get ing prometheus -o json | jq -r '.spec.rules[0].host')
```
AlexanderYastrebov, line 245: Maybe use hardcoded cluster.local hostname?

test/e2e/run_e2e.sh
```diff
@@ -231,6 +241,37 @@ if [ "$stackset_e2e" = true ]; then
...
+
+  >&2 echo "target prometheus: ${prometheus}"
+
+  # get data for the last 30m
+  curl -s -H"Accept: application/json" \
```
AlexanderYastrebov: curl supports `--data-urlencode "step=60"` that works together with `--get`, I think it could make the command cleaner

test/e2e/loadtest/client/loadtest-config.yaml
```diff
@@ -0,0 +1,3994 @@
...
+          `) -> "https://%TARGET%.%ZONE%";
+  ew-get.txt: |
+    GET http://127.0.0.1:9090/ew
+  get-long.txt: |
+          GET http://127.0.0.1:9090/ei7choo2eexaph8mahboo5shi6eeyo9phaishae3Naibei6Siiwu9Fu8Rabie5Zo0exoo3aetheeNaiPohvoo8jein9eNgo6neemaef4nohcheag0Pheitheif4fe6Me7xuyo2ENg1sa2ahgheikaeWookoe1dae0nee5aifaeyiis3jahvieru7ooDaetoh6Oi9aeNg0uo6ahsa5Lahzae6chapifeedeiPh0kiagh4iethukainu1cahsahh7fu7ahJ7ahT1thee3hohCeeMei4Un2aiyaijiangularo9cae2Higheit1voPhae7EejuPuYe7ohPie9Shophiphu8oopheP2IeCh9eu7oKiengeit0Tha8Ieshahc8pae6ahnaih0caisieL0tugh5och2hica8ahQui9yeis0ZaiBish2aghooneirohxaeghoshoamiegeiZolee5oPh9Ahmeecaep5Ohv1yeepo4osh8iwiequeiloh1shubaij2Fi1sha6jaijiquaeghiem0koochiach1eiRacha4aijeigou3iemah0eghic6Pahfeepeilee3je3iev6ieL8ALaehao0goocaemier7biaheeMoh9AGedaequ2chieThohshoh7Ungee2ahH6oWaiph7eiluodohTee5uthaegaiCi8aequ6xa8zahheisaDo9eakeLoo5than6opeiMeih7aetoox4aub8uShaeB2oveiloadtestfucaigegh5ohnaeteo1sine8EthohTauj3xierah8aiseikoDoofe6ahW5Coh3aisaeZ7wohda3xai6ohnee3uw9auroh8yo2Ip4eiku3TauW4Oong0ungeimoi0loochi2eiche3eeShech2uiz0aehee5WugheilaeSiV0zohdeK3Ahv7Yahn4ooch6Mah7uu0tu4oog8Dei1eeJie1aekaef9ze3Phaich4yie4uo3iyaeS3huangeeNgei0weesooyie8Oon
+  get.txt: |
```
AlexanderYastrebov, line 12: I think load test uses only `get.txt`. Lets remove everything else (and blobs especially) - we can add it anytime when needed

test/e2e/run_e2e.sh
szuecs: no because we can't query from the cluster were the e2e orchestrator is executed to internal domains of the e2e cluster

test/e2e/loadtest/client/prom-ingress.yaml
szuecs: yes we need to query it from the outside (pipeline is executed in another cluster not in e2e)

test/e2e/loadtest/client/prom-statefulset.yaml
szuecs: The ones in the cluster are rotated in case they are updated/config changed/..., so they are part of the cluster update.
This one is used as a self contained load test infrastructure

test/e2e/run_e2e.sh
szuecs: no it would send the query in the body and IMO would not be cleaner this way, because you need extra handling for the query anyways, because the PromQL query consists of characters you need to extra handle anyways.

test/e2e/loadtest/client/loadtest-config.yaml
szuecs: done

test/e2e/run_e2e.sh
szuecs: fixed

szuecs:
:+1:

test/e2e/run_e2e.sh
AlexanderYastrebov: With `--get`, `--data-urlencode` will do proper urlencoding of parameters into the url, not the body:
```sh
curl -v --get \
  --data-urlencode 'query=sum by(code) (rate(skipper_serve_host_count{application="e2e-vegeta"}[1m]))' \
  --data-urlencode "start=$(( $(date +%s) - (30*60) ))" \
  --data-urlencode "end=$(( $(date +%s) ))" \
  --data-urlencode "step=60" \
  "https://${prometheus}/api/v1/query_range"
```

szuecs: Interesting I tested `-X GET` and that is using the body, but `--get` is putting it into the query.

szuecs: fixed

szuecs:
https://github.com/zalando-incubator/kubernetes-on-aws/pull/5047 was successful with rolling nodes
You can check the e2e-load-test-result step result:
```
DEBUG: e2e loadtest not OK: 0
DEBUG: e2e loadtest OK: 83587.3
```

szuecs:
:+1:

mikkeloscar:
:+1:

PR #5029, opened 2022-02-24
Include component label in prometheus metrics

mikkeloscar:
This includes `component` label in metrics collected from pods. Prometheus collect metrics from itself (on port :9090) and since we changed to `application=kubernetes,component=prometheus` we are not able to filter on metrics from `prometheus` because the component label is missing. This fixes that issue.

Ref: #4945


mikkeloscar:
:+1:

katyanna:
👍

PR #5038, opened 2022-02-25
Add gp3 support for Persistent Volumes

mikkeloscar:
WIP: Attempt to add `gp3` volume support via patched controller manager

Using this patch not accepted upstream: https://github.com/kubernetes/kubernetes/pull/97237

Depends on #5045

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -580,7 +580,7 @@ write_files:
           effect: NoSchedule
         containers:
         - name: kube-controller-manager
-          image: nonexistent.zalan.do/teapot/kube-controller-manager:fixed
+          image: {{if eq .Cluster.ConfigItems.kubernetes_controller_manager_image "zalando" }}registry.opensource.zalan.do/teapot/kube-controller-manager-internal:v1.21.9-2{{else}}nonexistent.zalan.do/teapot/kube-controller-manager:fixed{{end}}
           args:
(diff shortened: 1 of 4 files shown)


mikkeloscar:
:+1:

gargravarr:
:+1:

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -580,7 +580,7 @@ write_files:
           effect: NoSchedule
         containers:
         - name: kube-controller-manager
-          image: nonexistent.zalan.do/teapot/kube-controller-manager:fixed
+          image: {{if eq .Cluster.ConfigItems.kubernetes_controller_manager_image "zalando" }}registry.opensource.zalan.do/teapot/kube-controller-manager-internal:v1.21.9-2{{else}}nonexistent.zalan.do/teapot/kube-controller-manager:fixed{{end}}
```
linki, line 583: This should follow the same approach as for `kube-scheduler`, shouldn't it? (using `:fixed` tag for both images)

linki: * https://github.bus.zalan.do/teapot/kubernetes-on-ubuntu/pull/221
* https://github.com/zalando-incubator/kubernetes-on-aws/pull/5055

mikkeloscar: Yes, this was only to enable quicker roll out of just the controller-manager changes. The idea was to bring it back once we also update the AMI.

PR #5060, opened 2022-03-10
allow zmon to access given buckets

vetinari:
so we do not override what users might have set


cluster/config-defaults.yaml
```diff
@@ -703,3 +703,6 @@ deployment_service_enabled: "true"
 
 # Standard storageclass gp3 volume type
 storageclass_standard_gp3: "false"
+
+# default for ZMON is not to be able to get any objects from S3
+zmon_s3_allow_get_object: ""
```
linki: How about we make the default "false" similar to the `storageclass_standard_gp3` setting above.

linki: Nevermind, I see now that it's a list of buckets and not a bool.

linki: Then let's rename the field to something like: `zmon_accessible_s3_buckets` and add add a comment that states that it's a list of buckets separated by comma.

linki:
Where can users set this?

mikkeloscar:
> Where can users set this?

I guess they would have to request this from the ZMON team. Right @vetinari?

mikkeloscar:
:+1:

vetinari:
> I guess they would have to request this from the ZMON team. Right @vetinari?

yes

vetinari:
:+1:

linki:
👍

mikkeloscar:
:+1:

mikkeloscar:
:+1:

PR #5077, opened 2022-03-17
Fix e2e build for go 1.18

mikkeloscar:
Fixes the e2e build for Go 1.18 which broke after the `golang` docker image was updated which we use in the `test/e2e/Dockerfile`.

Also uses the minor version tag on the `golang` image to ensure breaking changes doesn't happen in the future in an uncontrolled way.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #5098, opened 2022-03-30
Run e2e tests against a separate cluster with arm64 nodes

linki:
This runs the entire test suite again against a cluster using arm64 master and worker nodes.

It's quite overkill as it 2x the time to run e2e tests and increases the potential to fail due to flakes. I didn't find a more clever way to do that but if we had more arm64 related stuff to test, e.g. skipper-ingress on arm64, we could test that here as well.

One idea to improve would be to run this "full" test suite only on channel propagation, e.g. dev->alpha, alpha->beta and so on. This way we have everything tested when it matters but don't impact every feature branch.

Eventually we can get rid of it but currently it's easy to make mistakes mixing up container images for x86 and arm64.

Let me know what you think.


demonCoder95:
Looks good. Maybe the part where we say "we will get rid of this later" we can add as a comment in the ``delivery.yaml``? 🤔

linki:
Solved via https://github.com/zalando-incubator/kubernetes-on-aws/pull/6251

PR #5101, opened 2022-04-04
Update pause container to current version

linki:
This is the default pause container defined by kubelet for our version. Since we overwrite it or define it add various places we need to update it from time to time.

```console
$ kubelet --version
Kubernetes v1.21.11
$ kubelet --help | grep pause
      --pod-infra-container-image string                         Specified image will not be pruned by the image garbage collector. When container-runtime is set to 'docker', all containers in each pod will use the network/ipc namespaces from this image. Other CRI implementations have their own configuration to set this image. (default "k8s.gcr.io/pause:3.4.1")
```


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #5111, opened 2022-04-29
skipper-ingress: add configurable `ClusterScalingSchedule`

AlexanderYastrebov:

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -230,13 +230,8 @@ skipper_time_based_scaling_target: "1"
 # ClusterScalingSchedules
 # hyped-article-releases
 skipper_scaling_schedule_hyped_articles_target: ""
-# custom schedules
-skipper_scaling_schedule_name_1: ""
-skipper_scaling_schedule_target_1: ""
-skipper_scaling_schedule_name_2: ""
-skipper_scaling_schedule_target_2: ""
-skipper_scaling_schedule_name_3: ""
-skipper_scaling_schedule_target_3: ""
+# custom schedules in form of "<name1>=<target1>,<name2>=<target2>"
+skipper_scaling_schedule_custom: ""
 
 # cadvisor settings
 cadvisor_cpu: "150m"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -230,6 +230,9 @@ skipper_time_based_scaling_target: "1"
 # ClusterScalingSchedules
 # hyped-article-releases
 skipper_scaling_schedule_hyped_articles_target: ""
+# custom
+skipper_scaling_schedule_custom_name: ""
+skipper_scaling_schedule_custom_target: ""
 
 # cadvisor settings
 cadvisor_cpu: "150m"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -230,7 +230,11 @@ skipper_time_based_scaling_target: "1"
 # ClusterScalingSchedules
 # hyped-article-releases
 skipper_scaling_schedule_hyped_articles_target: ""
-# custom schedules in form of "<name1>=<target1>,<name2>=<target2>"
+# One or multiple custom schedules can be configured as a
+# comma-separated list of <schedule name>=<target value> pairs.
+# E.g. to configure "weekdays" schedule with a target value of 3 and
+# "weekends" with a target value of 5 set
+# skipper_scaling_schedule_custom: "weekdays=3,weekends=5"
 skipper_scaling_schedule_custom: ""
 
 # cadvisor settings
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -230,9 +230,13 @@ skipper_time_based_scaling_target: "1"
 # ClusterScalingSchedules
 # hyped-article-releases
 skipper_scaling_schedule_hyped_articles_target: ""
-# custom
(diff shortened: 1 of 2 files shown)


szuecs:
from my side lgtm

cluster/manifests/skipper/hpa.yaml
```diff
@@ -53,6 +53,19 @@ spec:
       target:
         averageValue: {{ .ConfigItems.skipper_scaling_schedule_hyped_articles_target }}
         type: AverageValue
+{{ end }}
+{{ if and (ne .ConfigItems.skipper_scaling_schedule_custom_name "") (ne .ConfigItems.skipper_scaling_schedule_custom_target "") }}
```
mikkeloscar: Maybe we would want to have a list of schedules such that we could subscribe to more schedules if needed? E.g. some clusters may want to subscribe to the lounge AND the hyped article schedules?

mikkeloscar:
:+1:

cluster/manifests/skipper/hpa.yaml
AlexanderYastrebov: >  E.g. some clusters may want to subscribe to the lounge AND the hyped article schedules?

This particular example is possible since we keep hyped schedule. But if you mean a generic scheme then maybe we should go for something extensible like
`skipper_scaling_schedule_1_name and skipper_scaling_schedule_1_target`, `skipper_scaling_schedule_2_name and skipper_scaling_schedule_2_target`.
WDYT?

mikkeloscar: Yeah, either that or encoded in one string list: `<name>=<target>,...` and then decode the string and loop through it. Maybe it's too complicated to configure though? :) We do it for some other use cases in this way.

cluster/manifests/skipper/hpa.yaml
```diff
@@ -53,6 +53,25 @@ spec:
       target:
         averageValue: {{ .ConfigItems.skipper_scaling_schedule_hyped_articles_target }}
         type: AverageValue
+{{ end }}
+{{ with $config_items := .ConfigItems }}
+	{{ range $i := split "1 2 3" " " }}
+		{{ $name := index $config_items (printf "skipper_scaling_schedule_name_%s" $i) }}
+		{{ $target := index $config_items (printf "skipper_scaling_schedule_target_%s" $i) }}
+		{{ if and (ne $name "") (ne $target "") }}
```
szuecs: I wonder if a little copy is not simpler than this. :)
We also do not win anything, because if we need "4" than we need to add a "4" here....

AlexanderYastrebov: yeah, I would do a copy for two but for three+ I thought it would be to much copy-paste.

AlexanderYastrebov: > because if we need "4" than we need to add a "4" here....

this we can do right now up to arbitrary number

AlexanderYastrebov: I think I'll do like @mikkeloscar suggests because we use such encoding elsewhere, e.g.
https://github.com/zalando-incubator/kubernetes-on-aws/blob/dc0d37ed3e2c86c7ca462d5fe2e6e5e2cd3a35f4/cluster/node-pools/worker-combined/stack.yaml#L88-L95

cluster/manifests/skipper/hpa.yaml
AlexanderYastrebov: I've updated as suggested

mikkeloscar:
:+1:

cluster/config-defaults.yaml
```diff
@@ -230,6 +230,8 @@ skipper_time_based_scaling_target: "1"
 # ClusterScalingSchedules
 # hyped-article-releases
 skipper_scaling_schedule_hyped_articles_target: ""
+# custom schedules in form of "<name1>=<target1>,<name2>=<target2>"
+skipper_scaling_schedule_custom: ""
```
szuecs, line 238: `name` should be distinct, right?
Otherwise we will create a likely failing metrics service and maybe impair all HPA of the cluster

AlexanderYastrebov: Name is the schedule name

AlexanderYastrebov: and schedule name is unique per cluster

mikkeloscar: @szuecs is right that the names need to be unique. I guess you mean we just need to clarify this in the comment or?

I assume using the same name twice would result in the HPA not applying OR the HPA applying but those scheduled scaling metrics not working. Would need to test this. I don't assume other HPAs would break, because this is isolated in the metrics-adapter.

AlexanderYastrebov: I have updated the comment a bit. It would be interesting to test how HPA behaves when configured with duplicate object metrics; from the description
> If you provide multiple such metric blocks, the HorizontalPodAutoscaler will consider each metric in turn. The HorizontalPodAutoscaler will calculate proposed replica counts for each metric, and then choose the one with the highest replica count.

I would say it would not be different as using a single metric.

mikkeloscar: @AlexanderYastrebov Yeah, I think you're right it would be the same as a single metric, the details are in our implementation that's why I'm not 100% sure without testing, but I think you're right.

mikkeloscar:
:+1:

szuecs:
👍

PR #5116, opened 2022-05-03
external-dns: enable DNSEndpoints CRD

AlexanderYastrebov:

### DIFF
--- a/cluster/manifests/external-dns/01-rbac.yaml
+++ b/cluster/manifests/external-dns/01-rbac.yaml
@@ -26,6 +26,12 @@ rules:
 - apiGroups: ["zalando.org"]
   resources: ["routegroups"]
   verbs: ["list"]
+- apiGroups: ["externaldns.k8s.io"]
+  resources: ["dnsendpoints"]
+  verbs: ["get", "watch", "list"]
+- apiGroups: ["externaldns.k8s.io"]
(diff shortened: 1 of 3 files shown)


cluster/manifests/external-dns/01-rbac.yaml
```diff
@@ -26,6 +26,12 @@ rules:
 - apiGroups: ["zalando.org"]
   resources: ["routegroups"]
   verbs: ["list"]
+- apiGroups: ["externaldns.k8s.io"]
+  resources: ["dnsendpoints"]
+  verbs: ["get", "watch", "list"]
+- apiGroups: ["externaldns.k8s.io"]
+  resources: ["dnsendpoints/status"]
+  verbs: ["*"]
```
AlexanderYastrebov, line 34: How do we prevent cluster users from creating `DNSEndpoint` resources?

mikkeloscar: Currently we don't have a way. What enables users to create CRD based resources is that our admission-controller automatically adds the related permissions on CRD creation. We could have a list of CRDs for which the admission-controller shouldn't do this for instance?

AlexanderYastrebov:
Not needed, new controller will translate FabricGateways into RouteGroups that are already supported.

PR #5120, opened 2022-05-05
Set protectKernelDefaults=true for kubelet

mikkeloscar:
Suggested configuration from kube-bench. Test that it works in our setup.

This just makes sure that kubelet does NOT make sysctl changes but instead verify the settings and fails to start if the settings are not as expected.


mikkeloscar:
depends on change to the AMI.

mikkeloscar:
:+1:

linki:
👍

PR #5124, opened 2022-05-09
update admission-controller

zeitlinger:
update admission-controller

* Adds support for an additional environment variable used by the observability SDKs.
* Removed Ingress admitter because it was only used for migrating traffic ownership between ingress resources and stackset resources. This is no longer needed as all stacksets have the traffic ownership as desired.


mikkeloscar:
:+1:

zeitlinger:
:+1:

szuecs:
:+1:

PR #5136, opened 2022-05-19
Introduce Karpenter

mikkeloscar:
Changes needed for introducing Karpenter as an optional autoscaler

**This PR is just showing of the Proof of Concept as presented in the demo**


szuecs:
@mikkeloscar as far as I saw in notification e-mails, this was not properly load tested. We need:
1. overlap test of without karpenter, to karpenter
2. ingress controller create/delete/update ALB and NLB
3.  svc type load balancer test. 

One open question would be if it makes sense to switch the ingress controller to AWS CNI mode, because it will detect the nodes faster like this.

mikkeloscar:
@szuecs you're right, so far there is no decision when to try this out. I demoed it for the team yesterday and I made these PRs to show what was done and to work as a starting point for going forward with this.

This is done in a way where it's only enabled if there are node pools of _profile_ `worker-karpenter`, so we could even run it next to the skipper-ingress node-pools as a start and skipper-ingress/ingress-controller would have the old behavior. If we see a benefit of karpenter then we could explore the skipper-ingress/ingress-controller use case how you suggest. At least that's how I think about it right now.

cluster/node-pools/worker-karpenter/stack.yaml
```diff
@@ -0,0 +1,50 @@
...
+    eu-central-1:
+      # Use the node pool's architecture to construct the config item name that we're using to get the AMI name.
+      MachineImage: '{{ index .NodePool.ConfigItems (print "kuberuntu_image_v1_21_" .Values.InstanceInfo.Architecture) }}'
+
+Resources:
+# create InstanceProfile wrapper on NodeRole
+  KarpenterNodeInstanceProfile:
+    Type: "AWS::IAM::InstanceProfile"
+    Properties:
+      InstanceProfileName: "KarpenterNodeInstanceProfile"
```
myaser:
```suggestion
      InstanceProfileName: "{{ .Cluster.ID | awsValidID }}-{{ .NodePool.Name }}-InstanceProfile"
```

myaser: creating more than one karpenter node pool per AWS account fails

cluster/manifests/z-karpenter/deployment.yaml
```diff
@@ -38,7 +38,7 @@ spec:
       #   node.kubernetes.io/role: master
       containers:
         - name: controller
-          image: "public.ecr.aws/karpenter/controller:v0.16.0@sha256:15cb6fe00279442b752beaf642dc9d6a3a2a9570d6da77b9bd7681e4c61b0958"
+          image: "public.ecr.aws/karpenter/controller:v0.16.1@sha256:ed1f97bc2cd67e63b45546955cf763890182f1f97586fca5e6cbd25f175b9c56"
```
mikkeloscar: Note, we also need/should update the webhook container image which is the second container in this deployment pod :)

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -0,0 +1,125 @@
...
+metadata:
+  name: {{ .NodePool.Name }}-template
+spec:
+  instanceProfile: {{ .Cluster.ID | awsValidID }}-{{ .NodePool.Name }}-InstanceProfile
+  amiFamily: Custom
+  amiSelector:
+    environment: production
+    purpose: zalando-ubuntu-kubernetes
+    version: v1.21.14
+    revision: master-236
```
mikkeloscar: Is this selection based on tags? If so it may make sense to just use a single tag like `Name` such that you don't need to update several different values here. Would make it simpler to have in a config-item and overwrite it for testing etc.

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -0,0 +1,125 @@
...
+    application: kubernetes
+    component: worker
+    environment: "{{ .Cluster.Environment }}"
+    Name: "{{ .NodePool.Name }} ({{ .Cluster.ID }})"
+    # TODO: are these tags still needed?
+    # Do we want a node pool tag for cost tracking reasons?
+    node.kubernetes.io/role: worker
+    node.kubernetes.io/node-pool: "{{ .NodePool.Name }}"
+# only node pools without taints should be attached to Ingress Load balancer
+{{- if or (not (index .NodePool.ConfigItems "taints")) (eq (index .NodePool.ConfigItems "taints") "dedicated=skipper-ingress:NoSchedule") }}
```
mikkeloscar: nit: This logic should be updated to be compatible with the changes made here: https://github.com/zalando-incubator/kubernetes-on-aws/pull/5158

We can argue that it's less relevant with skipper-ingress on a dedicated node pool, but we may wan't to optimize this in the future for cost savings so IMO better to keep it flexible for now according to the referenced PR.

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -0,0 +1,125 @@
...
+    - key: "node.kubernetes.io/instance-type"
+      operator: In
+      values:
+{{- range $type := .NodePool.InstanceTypes }}
+      - "{{ $type }}"
+{{- end }}
+    - key: karpenter.sh/capacity-type
+      operator: In
+      values:
+{{if .NodePool.IsSpot}}
```
mikkeloscar: I commented this internally, but to have it reflected here as well. I think we really should consider if defining pools as spot or not really makes sense with Karpenter. My proposal would be that ALL pools have both spot and on-demand and then we control via admission (by injecting node selectors on pods) if spot is the default or if on-demand is. That way we will automatically have a setup with spot falling back to on-demand or just only considering on-demand from the pool if the pods should not be allowed to run on spot.

cluster/cluster.yaml
```diff
@@ -301,6 +301,8 @@ Resources:
       Tags:
         - Key: 'kubernetes.io/cluster/{{.Cluster.ID}}'
           Value: owned
+        - Key: 'karpener.sh/discovery'
```
linki: Typo: -> `karpenter.sh/discovery`

cluster/cluster.yaml
```diff
@@ -301,6 +301,8 @@ Resources:
...
+          Value: 'WorkerNodeSecurityGroup'
```
linki: We could use the `{{.Cluster.ID}}` or alias here. Would be more in line with the docs.

cluster/manifests/deletions.yaml
```diff
@@ -122,3 +122,37 @@ post_apply:
...
+  namespace: kube-system
+  kind: Role
+- name: karpenter
+  namespace: kube-system
+  kind: RoleBinding
+- name: karpenter
+  namespace: kube-system
+  kind: Service
+- name: provisioners.karpenter.sh
+  namespace: kube-system
```
linki: We can drop namespace for global resources, I believe.

cluster/manifests/kube-node-ready/daemonset.yaml
```diff
@@ -25,6 +25,15 @@ spec:
...
         logging/destination: "{{.Cluster.ConfigItems.log_destination_infra}}"
     spec:
+      # consider tagging ASG based instance with a label such that a simple
+      # nodeSelector can be used which is more performant.
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: karpenter.sh/capacity-type
```
linki: Let's think about if it makes sense to have our own, e.g. `zalando.org/node-provisioner=[asg|karpenter]`. This is more descriptive and solves the point mentioned in the comment.

linki: If not then `karpenter.sh/initialized` or `karpenter.sh/provisioner-name` might be better alternatives.

cluster/manifests/z-karpenter/deployment.yaml
```diff
@@ -0,0 +1,128 @@
...
+      securityContext:
+        fsGroup: 1000
+      priorityClassName: "{{ .Cluster.ConfigItems.system_priority_class }}"
+      tolerations:
+      - key: node.kubernetes.io/role
+        value: master
+        effect: NoSchedule
+      # TODO: should or can run on master?
+      # nodeSelector:
+      #   node.kubernetes.io/role: master
```
linki: I think it makes sense to run it on master nodes (It allows for backing all worker nodes by karpenter). Note, that we run cluster-autoscaler as a DaemonSet instead of a Deployment (also on the masters).

cluster/manifests/z-karpenter/secret-webhook.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
```
linki, line 3: This is not wrapped by `{{ if karpenterNodePools .Cluster.NodePools }}` like the other files. Is that intended?

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -0,0 +1,125 @@
+{{- if eq .NodePool.Profile "worker-karpenter" }}
```
linki: Do we really need this here? When we take a template from this folder it should always have this Profile.

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -0,0 +1,125 @@
...
+kind: Provisioner
+metadata:
+  name: "{{.NodePool.Name}}"
+spec:
+  {{- if index .NodePool.ConfigItems "scaling_priority"}}
+  weight: {{.NodePool.ConfigItems.scaling_priority}}
+  {{- end}}
+  consolidation:
+    {{- if index .NodePool.ConfigItems "consolidation" }}
+    enabled: {{ or .NodePool.ConfigItems.consolidation true }}
```
linki: Do we need the `or` or can we use the plain value of `.NodePool.ConfigItems.consolidation` here?

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -0,0 +1,125 @@
+{{- if eq .NodePool.Profile "worker-karpenter" }}
+---
+apiVersion: karpenter.sh/v1alpha5
+kind: Provisioner
```
linki, line 4: For easier review and rollout we could separate the actual node pool creation into a separate PR. (It would be similar to deploying an IngressController but not deploying any Ingress yet). Of course the final version must probably support deploying both at the same time for cluster creation.

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -201,7 +201,8 @@ write_files:
             requests:
               cpu: 100m
               memory: 200Mi
-        - image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/admission-controller:master-151
+        # - image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/admission-controller:master-151
+        - image: mikkeloscar/admission-controller:latest
```
myaser: TODO: merge the admission controller first

cluster/node-pools/worker-karpenter/provisioners.yaml
myaser: The name has the arch part in it that's why it is not usable here. But I can introduce a single config item and a single tag

cluster/manifests/z-karpenter/deployment.yaml
```diff
@@ -0,0 +1,125 @@
...
+      securityContext:
+        fsGroup: 1000
+      priorityClassName: "{{ .Cluster.ConfigItems.system_priority_class }}"
+      tolerations:
+      - key: node.kubernetes.io/role
+        value: master
+        effect: NoSchedule
+      containers:
+        - name: controller
+          image: "public.ecr.aws/karpenter/controller:v0.16.1@sha256:ed1f97bc2cd67e63b45546955cf763890182f1f97586fca5e6cbd25f175b9c56"
```
mikkeloscar: Now there is v0.18.0 :sweat_smile: https://github.com/aws/karpenter/releases/tag/v0.18.0

mikkeloscar: nit: we also need to have a copy of this in our registry

myaser: yes, I noticed. but then decided to merge this one and update later. it is getting bigger and bigger 😆

cluster/node-pools/worker-karpenter/provisioners.yaml
myaser: Wow, what an If statement. It would be better to have it as a template helper function.
will improve it in the next PR

cluster/manifests/z-karpenter/deployment.yaml
linki: Fixed in https://github.com/zalando-incubator/kubernetes-on-aws/pull/5446

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -0,0 +1,121 @@
...
+---
+apiVersion: karpenter.k8s.aws/v1alpha1
+kind: AWSNodeTemplate
+metadata:
+  name: {{ .NodePool.Name }}
+spec:
+  instanceProfile: {{ .Cluster.ID | awsValidID }}-WorkerKarpenter-InstanceProfile
+  amiFamily: Custom
+  amiSelector:
+    Name: {{ .Cluster.ConfigItems.kuberuntu_image_karpenter_label  }}
```
linki: Should this be a lower case `name`? See https://github.com/aws/karpenter/blob/main/pkg/cloudproviders/aws/cloudprovider/amifamily/ami.go#L172

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -84,7 +84,7 @@ spec:
   instanceProfile: {{ .Cluster.ID | awsValidID }}-WorkerKarpenter-InstanceProfile
   amiFamily: Custom
   amiSelector:
-    Name: {{ .Cluster.ConfigItems.kuberuntu_image_karpenter_label  }}
```
myaser: @linki 
Can you also remove the kuberuntu_image_karpenter_label from the config_defaults.yaml please

cluster/node-pools/worker-karpenter/provisioners.yaml
linki: Solved by using `aws-ids` for now.

myaser:
We are still waiting for this fix to be released. then we will upgrade karpenter version
https://github.com/aws/karpenter/issues/3697

cluster/config-defaults.yaml
```diff
@@ -27,6 +27,12 @@ cluster_autoscaler_max_graceful_termination_sec: "1209600" # 2 weeks
 # Prevent CA lock-ups caused by large amounts of pending pods. Can be disabled completely by setting to 0.
 cluster_autoscaler_max_usnchedulable_pods_considered: "1000"
 
+# karpenter settings
+karpenter_controller_cpu: "1000m"
+karpenter_controller_memory: "1Gi"
+#karpenter_webhook_cpu: "200m"
+#karpenter_webhook_memory: "100Mi"
```
mikkeloscar: let's drop these commented out fields if not used anymore.

cluster/manifests/z-karpenter/deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+      tolerations:
+      - key: node.kubernetes.io/role
+        value: master
+        effect: NoSchedule
+      containers:
+        - name: controller
+          image: "container-registry.zalando.net/teapot/karpenter-controller:v0.27.3-main-9"
+          env:
+            - name: KUBERNETES_MIN_VERSION
+              value: "1.19.0-0"
```
mikkeloscar: What is the significance of this? Does it make sense to change it to 1.22/23?

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -0,0 +1,121 @@
...
+    # TODO: maybe use application label on dedicated node pools?
+    application: kubernetes
+    component: worker
+    environment: "{{ .Cluster.Environment }}"
+    Name: "{{ .NodePool.Name }} ({{ .Cluster.ID }})"
+    # TODO: are these tags still needed?
+    # Do we want a node pool tag for cost tracking reasons?
+#    node.kubernetes.io/role: worker
+#    node.kubernetes.io/node-pool: "{{ .NodePool.Name }}"
+#    kubernetes.io/cluster/{{ .Cluster.ID }}: "owned"
```
mikkeloscar: Are these tags set from labels on the node or not set at all now?

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -0,0 +1,121 @@
...
+metadata:
+  name: {{ .NodePool.Name }}
+spec:
+  amiFamily: Custom
+  amiSelector:
+    aws-ids: "{{ .Cluster.ConfigItems.kuberuntu_image_v1_22_amd64  }},{{ .Cluster.ConfigItems.kuberuntu_image_v1_22_arm64 }}"
+  metadataOptions:
+    httpTokens: optional
+  subnetSelector:
+    Name: "dmz-eu-central-1*" # TODO: use a better tag + it has to have auto assign public IP enabled
```
mikkeloscar: I assume fixing the public IP settings (either on subnet or in karpenter) is sort of important for merging this?

cluster/config-defaults.yaml
```diff
@@ -27,6 +27,12 @@ cluster_autoscaler_max_graceful_termination_sec: "1209600" # 2 weeks
 # Prevent CA lock-ups caused by large amounts of pending pods. Can be disabled completely by setting to 0.
 cluster_autoscaler_max_usnchedulable_pods_considered: "1000"
 
+# karpenter settings
+karpenter_controller_cpu: "1000m"
+karpenter_controller_memory: "1Gi"
```
mikkeloscar: Maybe we can run with a lower default and use a VPA to scale in bigger clusters? Or we run it on the master nodes 🤔

cluster/manifests/z-karpenter/deployment.yaml
myaser: checking...

cluster/node-pools/worker-karpenter/provisioners.yaml
myaser: only the `kubernetes.io/cluster/` is set automatically, will add the other 2 tags

myaser: karpenter also tag them with:
`karpenter.sh/provisioner-name`

but I guess we have some tooling around `node.kubernetes.io/node-pool`

cluster/node-pools/worker-karpenter/provisioners.yaml
myaser: yes it is, for teapot cluster I switched it on manually after consulting with linus. for wider rollout we need them to enable it or we do it in karpenter and update the version here

cluster/config-defaults.yaml
myaser: I will double-check the recommendations on karpenter side. VPA sounds good. and yes it is also schedulable on master nodes

cluster/node-pools/worker-karpenter/provisioners.yaml
mikkeloscar: > but I guess we have some tooling around `node.kubernetes.io/node-pool`

I'm not 100% sure but I imagine we have some ZMON check/alerts that may depend on it and potentially some cost reporting could also. To avoid that we have to clean this up as part of karpenter it can be safer to just keep them and then address it later if we want to clean it up.

There are also things like the `dedicated` tag which can be useful e.g. in the case of skipper-ingress node pools to get their cost in AWS Cost Explorer. That would currently not work with Karpenter, but we also will not run skipper-ingress here right away so we can solve that later.

cluster/node-pools/worker-karpenter/provisioners.yaml
mikkeloscar: Ok, it doesn't have to be a blocker for merging this IMO, it just means we can only realistically test in dev clusters until it's resolved. But it's fine to fix it later in a follow up.

cluster/manifests/z-karpenter/deployment.yaml
myaser: Screenshot 2023-04-27 at 09 51 23
then I will change it to 1.22

cluster/node-pools/worker-karpenter/provisioners.yaml
myaser: I added the tags back, and also added the dedicated tag.

cluster/config-defaults.yaml
myaser: the recommendations have changed in the recent versions
```
  # We usually recommend not to specify default resources and to leave this as a conscious
  # choice for the user. This also increases chances charts run on environments with little
  # resources, such as Minikube.
  ```
  I will then lower the resource reservation and add VPA

File: cluster/node-pools/worker-karpenter/provisioners.yaml [hunk 24]
```diff
@@ -99,20 +99,25 @@ spec:
   userData: |
 {{.UserData | indent 4}}
   tags:
-    InfrastructureComponent: "true"
-    # TODO: maybe use application label on dedicated node pools?
-    application: kubernetes
-    component: worker
-    environment: "{{ .Cluster.Environment }}"
```
mikkeloscar: are application/component tags gonna be set in another way? The TODO was an idea to improve the data in the future, but we should ensure we have application and component labels on instances.

cluster/manifests/z-karpenter/configmap.yaml
```diff
@@ -0,0 +1,64 @@
...
+  "aws.clusterName": "{{ .Cluster.ID  }}"
+  "aws.defaultInstanceProfile": "{{ .Cluster.ID | awsValidID }}-WorkerKarpenter-InstanceProfile"
+  "aws.enableENILimitedPodDensity": "true" # DEPRECATED: Use `.spec.kubeletConfiguration.maxPods` to set pod density on a per-provisioner basis
+  "aws.enablePodENI": "false"
+  "aws.interruptionQueueName": "{{.Cluster.LocalID}}-karpenter-interruption-queue"
+  "aws.isolatedVPC": "false"
+  "aws.nodeNameConvention": "ip-name"
+  "aws.vmMemoryOverheadPercent": "0.075"
+  #The global tags to use on all AWS infrastructure resources (launch templates, instances, etc.) across node templates
+  "aws.tags": '{"InfrastructureComponent":"true","application":"kubernetes","component":"worker","environment":"{{ .Cluster.Environment }}"}'
```
myaser, line 22: @mikkeloscar 
the alternative way to set the tags on all AWS infrastructure resources.
I gathered all the static/generic tags here and left the node-pool specific ones in the AWSNodeTemplate

mikkeloscar: Great!

cluster/config-defaults.yaml
myaser: added

cluster/node-pools/worker-karpenter/provisioners.yaml
myaser: this was resolved for dev clusters
https://github.bus.zalan.do/Base-Infrastructure/aws-network-infrastructure/pull/448/files

linki:
👍

myaser:
👍🏻

PR #5142, opened 2022-05-23
Consolidate nvidia-gpu-device-plugin into kubernetes

mikkeloscar:
Consolidate the `nvidia-gpu-device-plugin` daemonset into a component of `kubernetes` instead of being a dedicated application.

Since the Daemonset pods only runs on node startup and only runs on GPU nodes it's save to just delete the daemonset and then re-create it. Therefore we simply migrate it by:

1. Deleting the daemonset if it has `application=nvidia-gpu-device-plugin` as label (via `deletions.yaml`)
2. Applying the new daemonset with `application=kubernetes,component=nvidia-gpu-device-plugin`

In a follow up PR after the rollout, the deletions.yaml entry can be dropped again.

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -19,6 +19,10 @@ pre_apply:
     volume-type: gp3
   kind: StorageClass
 {{ end }}
+- labels:
+    application: nvidia-gpu-device-plugin
+  namespace: kube-system
+  kind: DaemonSet
 
 # everything defined under here will be deleted after applying the manifests
 post_apply:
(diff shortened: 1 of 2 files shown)


myaser:
👍🏻

rodrigodelmonte:
:+1:

mikkeloscar:
:+1:

mikkeloscar:
@myaser @rodrigodelmonte I was missing the namespace defined in the deletions.yaml. Now this has been fixed and the tests are passing

rodrigodelmonte:
:+1:

PR #5146, opened 2022-05-27
Switch to AMI that to pulls core docker images from ECR

linki:
Switch to an AMI that pulls from ECR (for testing).

Also updates cgroupDriver because this AMI is already on 22.04.


linki:
Doesn't work yet because fixes images are wrongly tagged: `nonexistent.zalan.do/production_namespace/teapot/kube-scheduler`

linki:
Replaced by https://github.com/zalando-incubator/kubernetes-on-aws/pull/5161

PR #5152, opened 2022-05-31
Disable tracking of zmon-appliance worker pods in Prometheus

mikkeloscar:
Our current monitoring works like this:

1. Prometheus scrapes information about pods
2. For each pod, ZMON worker executes a check which is one or more calls to prometheus querying data about the pod.
3. If the check execution takes longer, ZMON worker scales up to compensate.
4. Since we track all pods in Prometheus, more zmon worker pods leads to more load on prometheus which lead to response times, which leads to longer check execution time, which again leads to zmon worker scaling out.

This circular effect can sometimes result in hundreds or thousands of zmon worker pods running which is not very cost effecient.

As an attempt to easy the problem a bit, this PR introduces a config item (true by default) to disable tracking of ZMON worker pods in Prometheus. The theory is that if we don't track the pods, there is less load and it's less likely that this crazy scale-up happen. It needs to be validated in practice.

### DIFF
--- a/cluster/manifests/node-monitor/daemonset.yaml
+++ b/cluster/manifests/node-monitor/daemonset.yaml
@@ -45,7 +45,7 @@ spec:
             - --docker_only
             - --raw_cgroup_prefix_whitelist=/system.slice/kubelet.service
             - --store_container_labels=false
-            - --whitelisted_container_labels=io.kubernetes.container.name,io.kubernetes.pod.name,io.kubernetes.pod.namespace,io.kubernetes.pod.uid,application
+            - --whitelisted_container_labels=io.kubernetes.container.name,io.kubernetes.pod.name,io.kubernetes.pod.namespace,io.kubernetes.pod.uid,application,component
           resources:
             requests:
               cpu: "{{ .ConfigItems.cadvisor_cpu }}"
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

cluster/manifests/node-monitor/daemonset.yaml
```diff
@@ -45,7 +45,7 @@ spec:
             - --docker_only
             - --raw_cgroup_prefix_whitelist=/system.slice/kubelet.service
             - --store_container_labels=false
-            - --whitelisted_container_labels=io.kubernetes.container.name,io.kubernetes.pod.name,io.kubernetes.pod.namespace,io.kubernetes.pod.uid,application
+            - --whitelisted_container_labels=io.kubernetes.container.name,io.kubernetes.pod.name,io.kubernetes.pod.namespace,io.kubernetes.pod.uid,application,component
```
rodrigodelmonte, line 48: Why do we need this change? I'm just curious

rodrigodelmonte:
:+1:

cluster/manifests/node-monitor/daemonset.yaml
mikkeloscar: This is such that we can match exactly on `application=zmon-appliance,component=worker` when we exclude it in the scraping. Without this we would exclude all zmon-appliance pods, also those that are not workers which we don't need/want to do.

It's also anyway helpful to have the metrics tagged with component because it's a standardized label (in Zalando) that we just forgot to include.

PR #5158, opened 2022-06-07
fix: only add skipper-ingress node pools to the TG 

szuecs:
fix: only add skipper-ingress node pools to the TG members to increase node scalability

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -60,10 +60,11 @@ skipper_ingress_cpu: "1000m"
 skipper_ingress_memory: "1Gi"
 # skipper node-pool
 enable_dedicate_nodepool_skipper: "true"
-skipper_attach_only_to_skipper_node_pool: "true"
 {{if eq .Cluster.Environment "e2e"}}
+skipper_attach_only_to_skipper_node_pool: "false"
 skipper_topology_spread_enabled: "true"
 {{else}}
+skipper_attach_only_to_skipper_node_pool: "true"
 skipper_topology_spread_enabled: "false"
 {{end}}
 skipper_suppress_route_update_logs: "true"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -58,17 +58,15 @@ skipper_ingress_min_replicas: "2"
 {{end}}
 skipper_ingress_cpu: "1000m"
 skipper_ingress_memory: "1Gi"
-{{if eq .Cluster.Environment "e2e"}}
-enable_dedicate_nodepool_skipper: "false"
-{{else}}
+# skipper node-pool
 enable_dedicate_nodepool_skipper: "true"
-{{end}}
-skipper_suppress_route_update_logs: "true"
+skipper_attach_only_to_skipper_node_pool: "true"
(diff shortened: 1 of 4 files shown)


mikkeloscar:
Maybe we want a config-item for enabling/disabling this? E.g. `skipper_attach_only_to_skipper_node_pool` that would allow a smoother migration for e2e and if it's ever needed/desired in the future to switch back.

mikkeloscar:
Maybe we want a config-item for enabling/disabling this? E.g. `skipper_attach_only_to_skipper_node_pool` that would allow a smoother migration for e2e and if it's ever needed/desired in the future to switch back.

cluster/node-pools/worker-splitaz/stack.yaml
```diff
@@ -55,8 +55,8 @@ Resources:
       - Key: node.kubernetes.io/role
         PropagateAtLaunch: true
         Value: worker
-# only node pools without taints should be attached to Ingress Load balancer
-{{- if or (not (index $data.NodePool.ConfigItems "taints")) (eq (index $data.NodePool.ConfigItems "taints") "dedicated=skipper-ingress:NoSchedule") }}
+# only skipper-ingress node pools should be attached to Ingress Load balancer
+{{- if and (eq .Cluster.ConfigItems.skipper_attach_only_to_skipper_node_pool "true") (eq (index $data.NodePool.ConfigItems "taints") "dedicated=skipper-ingress:NoSchedule") }}
```
linki: I believe this must be `$data.Cluster` (or similar) because you're in a loop here.

linki: Also matching with the error message:
```
Fail to provision: failed to provision node pool worker-gpu: template: main/cluster/node-pools/worker-splitaz/stack.yaml:59:23: executing \"main/cluster/node-pools/worker-splitaz/stack.yaml\" at <.Cluster.ConfigItems.skipper_attach_only_to_skipper_node_pool>: can't evaluate field Cluster in type string, failed to provision node pool worker-node-tests: template: main/cluster/node-pools/worker-splitaz/stack.yaml:59:23:
```

cluster/config-defaults.yaml
```diff
@@ -58,17 +58,15 @@ skipper_ingress_min_replicas: "2"
...
 skipper_ingress_cpu: "1000m"
 skipper_ingress_memory: "1Gi"
-{{if eq .Cluster.Environment "e2e"}}
-enable_dedicate_nodepool_skipper: "false"
-{{else}}
+# skipper node-pool
 enable_dedicate_nodepool_skipper: "true"
-{{end}}
-skipper_suppress_route_update_logs: "true"
+skipper_attach_only_to_skipper_node_pool: "true"
```
mikkeloscar: This should be false for e2e now, and then we can enable it in a follow up PR, then the migration to skipper pool should be successful.

mikkeloscar: This should be false for e2e now, and then we can enable it in a follow up PR, then the migration to skipper pool should be successful.

szuecs: thanks fixed

szuecs:
:+1:

mikkeloscar:
:+1:

PR #5161, opened 2022-06-08
Update Kubernetes to v1.21.12

linki:
This is an update of Kubernetes from `v1.21.11` to `v1.21.12`.

It's a little phony: The latest AMI was based on Ubuntu 22.04 but not rolled out. We reverted the AMI back to Ubuntu 20.04 for the meantime but kept all the changes that happened in-between. This AMI also pre-pulls its (multi-arch) images from our new registry.

This leaves us with this update, which doesn't change the Kernel version but increases the Kubernetes version, amongst a couple other small changes.

This is mostly here to run e2e tests on the latest AMI. It replaces https://github.com/zalando-incubator/kubernetes-on-aws/pull/5146.


demonCoder95:
👍

myaser:
👍🏻

mikkeloscar:
:+1:

PR #5162, opened 2022-06-08
Pull admission-controller from the new registry

linki:
Would be nice if it worked like that but I think we cannot pull the vanity URL from a static pod. Let's run e2e to confirm.


linki:
👍

linki:
👍

demonCoder95:
👍

myaser:
👍🏻

PR #5166, opened 2022-06-10
Add Default Config-items to Enable Kubernetes ServiceCIDR Migration.

rodrigodelmonte:
- The admission-controller denies externalIPs by default, but we need to change the externalIP to keep the old serviceIP available during the network service migration. the `teapot_admission_controller_check_service_resources=false` allows the usage of  externalIPs
- For the Kubernetes service migration, we need to use patched docker images for kube-proxy and kube-apiserver. The patch allows us to change the service clusterIP for the new IP in the new ServiceCIDR range, this behavior is not allowed by default.
Once the Kubernetes Service CIDR migration has completed this change should be undone by setting again the default docker images for kube-apiserver and kube-proxy
Patch: https://github.bus.zalan.do/teapot/kubernetes-pipeline/pull/71/files
- To avoid disruptions during the migration the configuration from coredns, flannel, and skipper-internal need to be updated with the new service CIDR range

For: https://github.bus.zalan.do/teapot/issues/issues/3243

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -373,6 +373,9 @@ teapot_admission_controller_deployment_default_max_unavailable: "1"
 teapot_admission_controller_inject_aws_waiter: "true"
 teapot_admission_controller_parent_resource_hash: "true"
 
+# Define ExternalIPs is not allowed
+teapot_admission_controller_check_service_resources: "true"
+
 ## Defaults are set per-cluster
 teapot_admission_controller_check_daemonset_resources: "true"
 teapot_admission_controller_daemonset_reserved_cpu: "8"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -167,6 +167,10 @@ zmon_kairosdb_url: "https://data-service.zmon.zalan.do/kairosdb-proxy"
 enable_skipper_eastwest_dns: "true"
 enable_skipper_eastwest: "false"
 
+# Configure skipper-internal ClusterIP to Coredns query
+skipper_internal_cluster_ip: "10.3.99.99"
+skipper_internal_cluster_ip_additional: "10.3.99.99"
+
 # enable temporay logging of ingress.cluster.local names
 # used to find services for which it's being used.
 skipper_eastwest_dns_log_enabled: "false"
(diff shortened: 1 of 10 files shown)


cluster/config-defaults.yaml
```diff
@@ -511,6 +525,18 @@ audittrail_url: ""
 {{end}}
 audittrail_root_account_role: ""
 
+# kube-apiserver / kube-proxy container image name.
+# Temporarly added to help during the Kubernetes service network range migration.
+kube_apiserver_container_image_name: nonexistent.zalan.do/teapot/kube-apiserver:fixed
+kube_proxy_container_image: nonexistent.zalan.do/teapot/kube-proxy:fixed
```
mikkeloscar, line 531: I think it's more clear if we make these a true/false flag and template if/else for which image to chose. Otherwise you need to define the correct image in the config-item.

mikkeloscar: This could already be configured to the patched image so we roll this out as the first step.

cluster/config-defaults.yaml
```diff
@@ -511,6 +525,18 @@ audittrail_url: ""
 {{end}}
 audittrail_root_account_role: ""
 
+# kube-apiserver / kube-proxy container image name.
+# Temporarily added to help during the Kubernetes service network range migration.
+kube_apiserver_container_image_name: registry.opensource.zalan.do/teapot/kube-apiserver-internal:v1.21.12-pr-71-5
+kube_proxy_container_image: registry.opensource.zalan.do/teapot/kube-proxy-internal:v1.21.12-pr-71-2
```
mikkeloscar: These must be the images built from master, maybe align with @linki to ensure it's the right multi-arch versions we use.

rodrigodelmonte: Currently, the e2e tests are failing due to the error:

```sh
Error creating: admission webhook "pod-admitter.teapot.zalan.do" denied the request: container 
kube-proxy uses a non-compliant image (registry.opensource.zalan.do/teapot/kube-proxy-internal:v1.21.12-pr-71-2): 
The pull request was not merged.
```

cluster/manifests/01-admission-control/teapot.yaml
```diff
@@ -226,6 +226,7 @@ webhooks:
         apiGroups: [""]
         apiVersions: ["v1"]
         resources: ["namespaces"]
+{{- if eq .Cluster.ConfigItems.teapot_admission_controller_check_service_resources "true" }}
```
rodrigodelmonte, line 229: The service-admitter only has one rule to block the usage of `ExternalIP`, that is the reason why I opted to disable it at this level.
https://github.bus.zalan.do/teapot/admission-controller/blob/master/pkg/admitters/service/admit_service.go#L22

cluster/config-defaults.yaml
rodrigodelmonte: e2e tests fixed after git rebase dev

rodrigodelmonte:
Cluster `teapot` tested and serviceCIDR migrated using this branch.
The kube-apiserver and kube-proxy did not add side effects to the cluster.

rodrigodelmonte:
:+1:

mikkeloscar:
:+1:

PR #5183, opened 2022-06-20
Reenable extensions/v1beta1 for ingresses for ReadOnly role

mikkeloscar:
This permission is used by CDP/deployment-service as long as ingress resources of apiversion `extensions/v1beta1` are deployed. Therefore we can't disable this before the `extensions/v1beta1` api has been disabled.


mikkeloscar:
:+1:

myaser:
👍🏻

PR #5196, opened 2022-06-22
add v1beta1 crd enable config item for testing

demonCoder95:
In the effort to upgrade the CRD versions, there's a need to disable entry of ``v1beta1`` CRDs into clusters while dealing with the migration of existing ones.

This ``config-item`` named ``allow_v1beta1_crds`` is being added to be used as follows:
1. The default value will remain ``true`` to match with the existing ``runtime-config`` of the ``api-server`` of a cluster.
2. Using ``zregistry`` the value will be switched to ``false`` on a test cluster to ensure it works as expected.
3. Upon there being no side-effects per testing, it will be slowly rolled out to other clusters, starting with the ones that have no ``v1beta1`` CRD object in them. Both production and test clusters.
4. The default value will then be changed to ``false``, which will rollout the change to the all the clusters.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -237,6 +237,11 @@ skipper_time_based_scaling_target: "1"
 # skipper_cluster_scaling_schedules: "schedule1=3,schedule2=5"
 skipper_cluster_scaling_schedules: ""
 
+# kube-api-server settings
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #5197, opened 2022-06-22
Update to kube-metrics-adapter v0.1.19

katyanna:
Update to latest version of kube-metrics-adapter which limits ingress
support to only `networking.k8s.io/v1`.

Signed-off-by: Katyanna Moura <amelie.kn@gmail.com>

### DIFF
--- a/test/e2e/kube_metrics_adapter_test.go
+++ b/test/e2e/kube_metrics_adapter_test.go
@@ -71,40 +71,6 @@ var _ = describe("[HPA] Horizontal pod autoscaling (scale resource: Custom Metri
 
 	})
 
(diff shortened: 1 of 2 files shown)


test/e2e/kube_metrics_adapter_test.go
```diff
@@ -71,40 +71,6 @@ var _ = describe("[HPA] Horizontal pod autoscaling (scale resource: Custom Metri
 
 	})
 
-	It("should scale down with Custom Metric of type Object from Skipper [Ingress] [CustomMetricsAutoscaling] [Zalando]", func() {
```
mikkeloscar, line 74: Did this test fail? If so we need to verify that no HPAs has this defined in our clusters. (I thought it would still work with this update)

gargravarr:
:+1:

demonCoder95:
👍

PR #5201, opened 2022-06-22
Make sure next CLM version works well

linki:
Makes sure that when using the upcoming CLM version in https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/590 will work with the current setup.


linki:
Works fine

PR #5210, opened 2022-06-24
Use multiarch image for skipper routesrv

mikkeloscar:
The component is not yet active, but since we have multi-arch image, let's already switch.


mikkeloscar:
:+1:

linki:
👍

PR #5211, opened 2022-06-27
Enable GenericEphemeralVolume

mikkeloscar:
[Generic Ephemeral Volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes) are beta since Kubernetes v1.21 and thereby enabled by default. We have them disabled, but they are generally very useful and work fine in our environment. Let's enable them and allow the usage for normal user pods (restricted PSP).


mikkeloscar:
:+1:

myaser:
👍🏻

demonCoder95:
👍

PR #5219, opened 2022-06-28
Update unbound to multi-arch

mikkeloscar:
Updates the unbound containers to be multi-arch and latest version.

We had trouble with the previous version sometimes getting stuck in the past, hopefully this is fixed in the new version as `unbound` performs better than `dnsmasq` as a cache (when it's not stuck).

This is all still behind a config-item and disabled by default, but enables us to test unbound again also on ARM instances :)


mikkeloscar:
:+1:

linki:
👍

PR #5224, opened 2022-06-30
Close apiserver port from the internet

mikkeloscar:
Follow up to #5223 disabling public access to the control plane nodes and only allow access via the NLB on port 8443.

Depends on #5223 being fully rolled out!


mikkeloscar:
:+1:

demonCoder95:
👍

PR #5234, opened 2022-07-05
Add external-dns config-item policy and annotation

rodrigodelmonte:
This commit adds the config-tiem `external_dns_policy` that allows
change the syncronazition policy between Kuberntes and AWS Route53,
the main idea is to use it together with the annotation
`external-dns.alpha.kubernetes.io/exclude: true`.

To the serviceCIDR migration  we need to add an temporary ExternalIP to the
Kubernetes services, but we want to keep the Loadbalancer hostname and avoid
external-dns updates the Route53 entry with the ExternalIP.

To achieve that behavior, we need to temporarily set the external_dns_policy to
upsert-only, and add the annotatation `external-dns.alpha.kubernetes.io/exclude: true`
to the Kubernetes service that has the ExternalIP. Then, external-dns should create
and update the Route53, but will not delete AWS Route53 entries.

For https://github.bus.zalan.do/teapot/issues/issues/3243

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -616,6 +616,8 @@ custom_dns_zone_nameservers: "" # space seperated list of nameserver IP addresse
 external_dns_ownership_prefix: ""
 # domains that should be ignored by ExternalDNS
 external_dns_excluded_domains: cluster.local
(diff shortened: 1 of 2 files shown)


demonCoder95:
👍

rodrigodelmonte:
:+1:

PR #5240, opened 2022-07-06
Fix deployment-controller rbac for crd

demonCoder95:
Final fix to the RBAC that is known to work with the pipeline :)


demonCoder95:
👍

gargravarr:
:+1:

PR #5248, opened 2022-07-12
Update teapot/event-logger container image

rodrigodelmonte:
This PR updates the teapot/event-logger container image which also adds muti-arch capabliiy.
Ref: https://github.bus.zalan.do/teapot/issues/issues/3283


rodrigodelmonte:
:+1:

demonCoder95:
👍

PR #5249, opened 2022-07-12
Multi arch support for k8s-lifecycle-metrics

myaser:
teapot/issues#3283

### DIFF
--- a/cluster/manifests/kubernetes-lifecycle-metrics/deployment.yaml
+++ b/cluster/manifests/kubernetes-lifecycle-metrics/deployment.yaml
@@ -33,7 +33,7 @@ spec:
       serviceAccountName: kubernetes-lifecycle-metrics
       containers:
         - name: kubernetes-lifecycle-metrics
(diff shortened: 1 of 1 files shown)


cluster/manifests/kubernetes-lifecycle-metrics/deployment.yaml
```diff
@@ -33,7 +33,7 @@ spec:
       serviceAccountName: kubernetes-lifecycle-metrics
       containers:
         - name: kubernetes-lifecycle-metrics
-          image: "pierone.stups.zalan.do/teapot/kubernetes-lifecycle-metrics:master-9"
+          image: "926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/kubernetes-lifecycle-metrics:master-12"
```
mikkeloscar: We should be able to use `container-registry.zalando.net` URL here as it goes through the admission-controller.

myaser: 👍🏻

myaser:
👍🏻

mikkeloscar:
:+1:

PR #5258, opened 2022-07-15
Enable SizeMemoryBackedVolumes by default

katyanna:
Set `enable_size_memory_backed_volumes` config to true so every cluster
has it enabled by default.


mikkeloscar:
:+1:

katyanna:
👍

PR #5272, opened 2022-07-28
Feature/autoscaling skipper redis

szuecs:
By setting `skipper_ingress_redis_autoscaling="true"` we can enable autoscaling.
By setting `skipper_ingress_redis_autoscaling="false"` we can revert autoscaling.

@mikkeloscar do you know if setting `replicas: 2` in statefulset will interfere with hpa?
I remember there was an edge case which scaled down deployments in some kubernetes versions (don't know if this was fixed in the meantime.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -132,6 +132,7 @@ skipper_redis_pool_timeout: "250ms"
 skipper_redis_read_timeout: "25ms"
 skipper_redis_write_timeout: "25ms"
 
+skipper_ingress_redis_swarm_enabled: "true"
 skipper_ingress_redis_target_average_utilization_cpu: "30"
 skipper_ingress_redis_target_average_utilization_memory: "60"
 skipper_ingress_redis_min_replicas: "1"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -132,6 +132,10 @@ skipper_redis_pool_timeout: "250ms"
 skipper_redis_read_timeout: "25ms"
 skipper_redis_write_timeout: "25ms"
 
+# "true" enable autoscaling
+# "false" remove autoscaling
+# "" do not change
+skipper_ingress_redis_autoscaling: ""
 skipper_ingress_redis_swarm_enabled: "true"
 skipper_ingress_redis_target_average_utilization_cpu: "30"
 skipper_ingress_redis_target_average_utilization_memory: "60"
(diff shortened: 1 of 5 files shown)


cluster/manifests/skipper/hpa-redis.yaml
```diff
@@ -0,0 +1,56 @@
...
+      metric:
+        name: {{ index $name_target 0 }}
+      target:
+        averageValue: {{ index $name_target 1 }}
+        type: AverageValue
+  {{ end }}
+{{ end }}
+  behavior:
+    scaleDown:
+      stabilizationWindowSeconds: 600
```
mikkeloscar, line 47: What's the motivation for delayed scale-down here? Seems reasonable to just let it scale down as fast as HPA would normally do it.

cluster/config-defaults.yaml
```diff
@@ -132,6 +132,16 @@ skipper_redis_pool_timeout: "250ms"
 skipper_redis_read_timeout: "25ms"
 skipper_redis_write_timeout: "25ms"
 
+# "true" enable autoscaling
+# "false" remove autoscaling
+# "" do not change
+skipper_ingress_redis_autoscaling: ""
+skipper_ingress_redis_swarm_enabled: "true"
+skipper_ingress_redis_target_average_utilization_cpu: "30"
```
mikkeloscar, line 140: We currently have a paging alert triggering on 85% usage. So we can probably comfortably set this higher now that it will also be automatic and not depend on a human waking up and scaling :)

cluster/manifests/skipper/hpa-redis.yaml
szuecs: It's copied from skipper-ingress, but honestly I never saw that it stopped downscaling.
In general for regular load spikes it's better to downscale slower.
I would let it like this.

cluster/config-defaults.yaml
szuecs: I know but this is average and I want to try to scale up as fast as we need and not only when it's too late. 
Let's keep it like this and learn by inspecting metrics and set better target if we see that it's a waste of money.

szuecs:
:+1:

cluster/manifests/skipper/hpa-redis.yaml
mikkeloscar: For skipper-ingress the motivation was to slow down target group detaching.

here the theory is regular load spikes, but as we don't now the impact without testing we might as well start less conservative and adjust if it's needed (we already have the conservative and low HPA CPU target so we don't need this in both directions up front IMO)

mikkeloscar:
:+1:

PR #5276, opened 2022-07-29
Patch image with termination logic patch

demonCoder95:
Updates the CLC image to the version which has the pod forced termination logic improvement implemented [here](https://github.bus.zalan.do/teapot/cluster-lifecycle-controller/pull/32).


demonCoder95:
👍

mikkeloscar:
:+1:

PR #5286, opened 2022-08-03
Fix/skipper redis autoscaling

szuecs:
- fix: hpa manifest default was nil
- fix: set min replicas to current value to not scale in to 1 because of hpa
- hpa use 60% target to run more efficiently


mikkeloscar:
:+1:

szuecs:
:+1:

PR #5290, opened 2022-08-05
Disable validate query log by default

mikkeloscar:
Follow up to #5222 where validate query logging was introduced.

This disable the log by default and allows enabling per cluster. The reason we want to disable by default is to avoid excess logging and thereby excess logging costs.


rodrigodelmonte:
:+1:

mikkeloscar:
:+1:

PR #5295, opened 2022-08-09
Readd test using extensions/v1beta1

katyanna:
Keep test while we still have HPAs with old Ingress versions.


mikkeloscar:
This being green is enough to validate that we can move forward with #5197 which is part of #5297

PR #5296, opened 2022-08-09
Change Configs That Supported Kubernetes Servicecidr Migration 

rodrigodelmonte:
This PR changes service CIDR network to `10.5.0.0/16` since all the current clusters were migrated as part of the task https://github.bus.zalan.do/teapot/issues/issues/3243.

The changes include:
 - Remove patched kube-apiserver and kube-proxy images
 - Remove feature-flag to allow externalIPs
 - Configure coredns and skipper to use IPs from the service CIDR  range 10.5.0.0/16
 - Set cluster service CIDR to 10.5.0.0/16 

Relate to #5166

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -393,9 +393,6 @@ teapot_admission_controller_deployment_default_max_unavailable: "1"
 teapot_admission_controller_inject_aws_waiter: "true"
 teapot_admission_controller_parent_resource_hash: "true"
 
-# Define ExternalIPs is not allowed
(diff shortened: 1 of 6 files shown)


rodrigodelmonte:
:warning: 
The e2e test will not pass because it will use first the range `10.3.0.0/16` and then will try to apply the changes to `10.5.0.0/16`, the default kube-apiserver version doesn't allow this change.
So, the idea is to merge the PR even with the e2e tests failing, and monitor how the next PRs will behave.

rodrigodelmonte:
:+1:

gargravarr:
:+1:

PR #5313, opened 2022-08-19
admission-controller: STS replicas from HPA

mikkeloscar:
Add support for annotations:

* `zalando.org/create-using-hpa-replicas: <HPA name>`
* `zalando.org/update-using-hpa-replicas: <HPA name>`

Such that it's possible to add an HPA for an existing statefulset without replicas being reset to 1.

A similar feature was implemented for `Deployment`s in #4963 and now `StatefulSet`s are also supported.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #5314, opened 2022-08-19
Set skipper-ingress-redis replicas based on HPA

mikkeloscar:
This sets the replicas based on the HPA to ensure they are not reset to 1 when dropping the replicas field on the statefulset.

Depends on #5313 being fully rolled out!


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #5321, opened 2022-08-25
Prepare etcd CloudFormation for storing certificates in S3 

gargravarr:
This Pull Request reorganizes the structure of etcd CloudFormation, in preparation for using the new etcd Taupage. The structure is somewhat similar to master and worker node pools. The `files.yaml` specifies the certificates needed for etcd. Subsequent Pull Requests on the Cluster Lifecycle Manager will handle the shipping of the certificates to S3 to be consumed by the new AMI, as well as the new location of the etcd CloudFormation file.

The previous path of the etcd CloudFormation is now a symbolic link, for compatibility with the current setup. 
So the result of this change is a no-op in the point-of-view of CLM (already tested with `teapot` cluster). This prevents using awkward Config Items to switch between old and new file locations.

For now the etcd CloudFormation remains unaltered. After the updates in CLM the next step is to update the etcd CloudFormation to use the new AMI.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #5324, opened 2022-08-26
Increase timeout for image compliance e2e tests

mikkeloscar:
The tests checking for pods with compliant images only wait 1 minute for the pods to be running and ready. It could be that it takes longer because of a node needing to be scaled out so increase the timeout to 5 min. to reduce the likely-hood of this being an issue.


szuecs:
:+1:

myaser:
👍🏻

PR #5347, opened 2022-09-08
Use new etcd AMI

gargravarr:
This **Pull Request (PR)** updates the etcd CloudFormation template to support the new etcd AMI. The new AMI is also set as the default for etcd.

Merging this PR _won't rotate_ the etcd instances. The effective rotation will still be done manually using [AWS Instance Refresh](https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-instance-refresh.html).

### DIFF
--- a/cluster/etcd/stack.yaml
+++ b/cluster/etcd/stack.yaml
@@ -44,32 +44,29 @@ Resources:
         InstanceInitiatedShutdownBehavior: terminate
         ImageId: {{.Cluster.ConfigItems.etcd_ami}}
         InstanceType: {{.Cluster.ConfigItems.etcd_instance_type}}
-        UserData: !Base64 |
-          #taupage-ami-config
-          application_id: etcd-cluster
-          application_version: etcd
-          environment:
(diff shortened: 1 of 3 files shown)


gargravarr:
:+1:

cluster/etcd/stack.yaml
```diff
@@ -78,7 +75,7 @@ Resources:
         Timeout: PT15M
     Properties:
       DesiredCapacity: {{.Cluster.ConfigItems.etcd_instance_count}}
-      HealthCheckGracePeriod: 300
+      HealthCheckGracePeriod: 0
```
demonCoder95, line 78: Any specific reason to set this to 0? Are we relying on the `HeartbeatTimeout` in `AutoscalingLifecycleHook` instead? 🤔

gargravarr: Yes, exactly that :slightly_smiling_face:

demonCoder95:
👍

PR #5355, opened 2022-09-13
stackset-e2e: Don't run in parallel

mikkeloscar:
Don't run stackset e2e in parallel to see if it makes a difference on the flakyness.

If this works it's just a band-aid on the problem.


mikkeloscar:
This increases the time it takes for running the stackset e2e from around 15 min. to 45 min...

rodrigodelmonte:
:+1:

mikkeloscar:
:+1:

mikkeloscar:
Merging this is a simple mitigation which is better than nothing.

PR #5357, opened 2022-09-14
Add application tag to CF resources

demonCoder95:
Adding the `application` tag to CF resources for teapot. 

Reference issue: https://github.bus.zalan.do/teapot/issues/issues/3332

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -37,6 +37,8 @@ Resources:
       Tags:
         - Key: 'kubernetes.io/cluster/{{.Cluster.ID}}'
           Value: owned
+        - Key: 'kubernetes:application'
+          Value: kube-ingress-aws-controller
       VpcId: "{{.Cluster.ConfigItems.vpc_id}}"
     Type: 'AWS::EC2::SecurityGroup'
   MasterKubeletToMasterKubeletSecurityGroup:
@@ -770,6 +772,9 @@ Resources:
         BlockPublicPolicy: true
         IgnorePublicAcls: true
         RestrictPublicBuckets: true
+      Tags:
+        - Key: component
+          Value: deployment-service
       OwnershipControls:
         Rules:
           - ObjectOwnership: BucketOwnerPreferred
@@ -1733,7 +1738,7 @@ Resources:
         Status: Suspended
       Tags:
       - Key: component
-        Value: resend
+        Value: audittrail-adapter
   AuditTrailBucketPolicy:
     Type: AWS::S3::BucketPolicy
     Properties:
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -37,8 +37,6 @@ Resources:
       Tags:
         - Key: 'kubernetes.io/cluster/{{.Cluster.ID}}'
           Value: owned
-        - Key: 'kubernetes:application'
-          Value: kube-ingress-aws-controller
       VpcId: "{{.Cluster.ConfigItems.vpc_id}}"
     Type: 'AWS::EC2::SecurityGroup'
   MasterKubeletToMasterKubeletSecurityGroup:
@@ -69,8 +67,6 @@ Resources:
       Tags:
         - Key: 'kubernetes.io/cluster/{{.Cluster.ID}}'
           Value: owned
-        - Key: "application"
-          Value: kube-apiserver
         - Key: "component"
           Value: "kube-apiserver"
       Type: network
@@ -89,8 +85,6 @@ Resources:
       Tags:
         - Key: 'kubernetes.io/cluster/{{.Cluster.ID}}'
           Value: owned
-        - Key: "application"
-          Value: kube-apiserver
         - Key: "component"
           Value: "kube-apiserver"
       VpcId: "{{.Cluster.ConfigItems.vpc_id}}"
@@ -776,9 +770,6 @@ Resources:
         BlockPublicPolicy: true
         IgnorePublicAcls: true
         RestrictPublicBuckets: true
-      Tags:
-        - Key: application
-          Value: deployment-service
       OwnershipControls:
         Rules:
           - ObjectOwnership: BucketOwnerPreferred
@@ -1741,8 +1732,6 @@ Resources:
       VersioningConfiguration:
         Status: Suspended
       Tags:
-      - Key: application
-        Value: audittrail-adapter
       - Key: component
         Value: resend
   AuditTrailBucketPolicy:
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -3,6 +3,7 @@ Description: Kubernetes cluster
 Metadata:
   Tags:
     InfrastructureComponent: "true"
+    application: "kubernetes"
 Resources:
 {{ if ne .Cluster.ConfigItems.delete_vpc_resources "true" }}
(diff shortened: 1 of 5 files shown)


cluster/cluster.yaml
```diff
@@ -3,6 +3,7 @@ Description: Kubernetes cluster
 Metadata:
   Tags:
     InfrastructureComponent: "true"
+    application: "kubernetes"
```
mikkeloscar, line 6: We have this defined individually below for some specific resources. Would make sense to remove those per resource tags where it would be overwritten by this. For instance we have some `application: kube-apiserver` tags defined which are no longer correct and should be dropped.

cluster/etcd/stack.yaml
```diff
@@ -4,6 +4,7 @@ Metadata:
   Tags:
     InfrastructureComponent: "true"
     "kubernetes:component": "etcd-cluster"
+    application: "kubernetes"
```
mikkeloscar, line 7:
```suggestion
    application: "kubernetes"
    component: "etcd-cluster"
```

cluster/node-pools/master-default/stack.yaml
```diff
@@ -3,6 +3,7 @@ Description: Kubernetes default master node pool
 Metadata:
   Tags:
     InfrastructureComponent: "true"
+    application: "kubernetes"
```
mikkeloscar, line 6:
```suggestion
    application: "kubernetes"
    component: "control-plane"
```

cluster/cluster.yaml
demonCoder95: Agreed. Will remove per-resource tags for this template and will keep only the "top level" tag already added.

cluster/cluster.yaml
```diff
@@ -37,8 +37,6 @@ Resources:
       Tags:
         - Key: 'kubernetes.io/cluster/{{.Cluster.ID}}'
           Value: owned
-        - Key: 'kubernetes:application'
-          Value: kube-ingress-aws-controller
```
mikkeloscar: We may depend on this in the ingress-controller

cluster/cluster.yaml
```diff
@@ -774,9 +768,6 @@ Resources:
         BlockPublicPolicy: true
         IgnorePublicAcls: true
         RestrictPublicBuckets: true
-      Tags:
-        - Key: application
-          Value: deployment-service
```
mikkeloscar, line 780: Maybe we can have `component: deployment-service` at this level instead.

cluster/cluster.yaml
```diff
@@ -1739,8 +1730,6 @@ Resources:
       VersioningConfiguration:
         Status: Suspended
       Tags:
-      - Key: application
-        Value: audittrail-adapter
```
mikkeloscar, line 1744: Same here, `component` tag instead.

cluster/etcd/stack.yaml
```diff
@@ -94,9 +94,6 @@ Resources:
         - Key: StackVersion
           PropagateAtLaunch: true
           Value: etcd
-        - Key: kubernetes:component
-          PropagateAtLaunch: true
-          Value: etcd-cluster
```
mikkeloscar: I'm not 100% sure if we depend on this somewhere, maybe in zmon checks 🤔

cluster/node-pools/master-default/stack.yaml
```diff
@@ -39,12 +36,6 @@ Resources:
       - Key: zalando.org/pod-max-pids
         PropagateAtLaunch: true
         Value: "{{ .NodePool.ConfigItems.pod_max_pids }}"
-      - Key: application
-        Value: kubernetes
-        PropagateAtLaunch: true
```
mikkeloscar, line 42: Is this true for tags set at the CF stack level?

demonCoder95: I'm not sure. I was wondering if we can add this at the stack level. Explicitly setting it `true` here tells me that it is `false` by default. 🤔

demonCoder95: Here, it says the tagging implementation can vary by resource, and that this `PropagateAtLaunch` is available only for `AutoScalingGroup` resources. 🤦 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html

cluster/cluster.yaml
demonCoder95: I see, I'll restore this.

cluster/cluster.yaml
demonCoder95: Done.

cluster/cluster.yaml
demonCoder95: Done. It already had a tag `component` with value `resend`, I just changed it to `audittrail-adapter`.

cluster/etcd/stack.yaml
demonCoder95: Alright. I'll restore the tag then.

cluster/node-pools/master-default/stack.yaml
demonCoder95: For resources that support tagging, tags propagate from stack level to those resources created by the stack. So, I believe this will be `true` and EC2 instances should have the tag added to the stack.

demonCoder95:
Permission denied when attempting to tag resources during rollout:
```bash


2022-09-15 10:40:27 UTC+0200 | e2e-pr-5357-8 | UPDATE_ROLLBACK_IN_PROGRESS | The following resource(s) failed to update: [WorkerFilesEncryptionKey, DeploymentSecretKey, RemoteFilesEncryptionKey, MasterFilesEncryptionKey, EtcdEncryptionKey].
-- | -- | -- | --
2022-09-15 10:40:26 UTC+0200 | MasterFilesEncryptionKey | UPDATE_FAILED | Resource update cancelled
2022-09-15 10:40:26 UTC+0200 | WorkerFilesEncryptionKey | UPDATE_FAILED | Resource update cancelled
2022-09-15 10:40:26 UTC+0200 | RemoteFilesEncryptionKey | UPDATE_FAILED | Resource update cancelled
2022-09-15 10:40:26 UTC+0200 | EtcdEncryptionKey | UPDATE_FAILED | Resource update cancelled
2022-09-15 10:40:25 UTC+0200 | DeploymentSecretKey | UPDATE_FAILED | Resource handler returned message: "Access denied for operation 'TagResource'." (RequestToken: 368ceb5f-b19a-c75a-22bf-b10736fe24f2, HandlerErrorCode: AccessDenied)


```

demonCoder95:
Added permission for the operation `tag:TagResources` to the key policies for all keys that had tagging fail on them.

Update: Had to add key policy update in a separate PR as CF wasn't applying the policy update before attempting to tag.

demonCoder95:
Rebased with the key policy upgrade for `DeploymentSecretKey`.

demonCoder95:
Rebase after fixing the plural in key-policy.

mikkeloscar:
Marked as do-not-merge because it depends on #5369 being fully rolled out.

mikkeloscar:
@demonCoder95 Since we established that the stack level tags are propagated to instances launched via the ASG I have dropped the tags on the ASG level to avoid duplication.

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #5269, opened 2022-07-25
Update to Kubernetes 1.22

katyanna:
Kubernetes 1.22: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md

- Remove support to Ingress versions `extensions/v1beta1` and `kuberentes.k8s.io/v1beta1`

PR #5358, opened 2022-09-14
Remove BoundSATokenVolume flag from kube-proxy-config

demonCoder95:
Removing the `BoundServiceAccountTokenVolume` feature flag from the `kube-proxy` configmap in order to move forward with the v1.22 upgrade. 

This flag causes a race condition where it never gets updated during an upgrade rollout and the old value causes the control-plane provisioning to fail. 

See [here](https://github.com/zalando-incubator/kubernetes-on-aws/pull/5269#issuecomment-1245329930) for reference.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #5337, opened 2022-09-02
stactset-e2e: control wait timeout with flags

MustafaSaber:
Add environment variables to the script with old default values and pass them from CI.

PR #5359, opened 2022-09-14
Increase stackset e2e timeout

mikkeloscar:
Re-try #5337 to increase stackset wait timeout.


mikkeloscar:
Not relevant after #5364

PR #5361, opened 2022-09-15
Add VPA version config-item

demonCoder95:
In order to safely roll-out the latest version of the `vertical-pod-autoscaler` (rebase on latest upstream), this config-item is introduced as a flag to be able to restore to the legacy version in case of running into any issues.

More information in this issue: https://github.bus.zalan.do/teapot/issues/issues/3328

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -379,6 +379,9 @@ horizontal_pod_autoscaler_sync_period: "30s"
 horizontal_pod_autoscaler_tolerance: "0.1"
 horizontal_pod_downscale_stabilization: "5m0s"
 
+# Vertical pod autoscaler version config for controlling roll-out
+vertical_pod_autoscaler_version: "v0.11.0-internal.17"
+
 # Cluster update settings
 {{if eq .Cluster.Environment "production"}}
 drain_grace_period: "6h"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -379,8 +379,10 @@ horizontal_pod_autoscaler_sync_period: "30s"
 horizontal_pod_autoscaler_tolerance: "0.1"
 horizontal_pod_downscale_stabilization: "5m0s"
 
-# Vertical pod autoscaler version config for controlling roll-out
-vertical_pod_autoscaler_version: "v0.11.0-internal.17"
+# Vertical pod autoscaler version for controlling roll-out, can be "current" or "legacy"
+# current => v0.11.0-internal.17
+# legacy => v0.6.1-internal.16
+vertical_pod_autoscaler_version: "current"
 
 # Cluster update settings
 {{if eq .Cluster.Environment "production"}}
--- a/cluster/manifests/01-vertical-pod-autoscaler/updater-deployment.yaml
+++ b/cluster/manifests/01-vertical-pod-autoscaler/updater-deployment.yaml
@@ -25,7 +25,10 @@ spec:
       serviceAccountName: vpa-updater
       containers:
       - name: updater
-        image: container-registry.zalando.net/teapot/vpa-updater:{{.Cluster.ConfigItems.vertical_pod_autoscaler_version}}
+        {{if eq .Cluster.ConfigItems.vertical_pod_autoscaler_version "current"}}
+        image: container-registry.zalando.net/teapot/vpa-updater:v0.11.0-internal.17
+        {{else if eq .Cluster.ConfigItems.vertical_pod_autoscaler_version "legacy"}}
+        image: container-registry.zalando.net/teapot/vpa-updater:v0.6.1-internal.16
         command:
           - ./updater
         args:
--- a/cluster/manifests/01-vertical-pod-autoscaler/updater-deployment.yaml
+++ b/cluster/manifests/01-vertical-pod-autoscaler/updater-deployment.yaml
@@ -25,7 +25,7 @@ spec:
       serviceAccountName: vpa-updater
       containers:
       - name: updater
-        image: container-registry.zalando.net/teapot/vpa-updater:v0.6.1-internal.16
+        image: container-registry.zalando.net/teapot/vpa-updater:{{.Cluster.ConfigItems.vertical_pod_autoscaler_version}}
         command:
           - ./updater
         args:
--- a/cluster/manifests/01-vertical-pod-autoscaler/updater-deployment.yaml
+++ b/cluster/manifests/01-vertical-pod-autoscaler/updater-deployment.yaml
@@ -36,7 +36,6 @@ spec:
           - --v=1
           - --logtostderr
           - --min-replicas=1
-          - --pod-lifetime-update-threshold=12h
           - --evict-after-oom-threshold=12h
         resources:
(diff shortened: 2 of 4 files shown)


cluster/config-defaults.yaml
```diff
@@ -379,6 +379,9 @@ horizontal_pod_autoscaler_sync_period: "30s"
 horizontal_pod_autoscaler_tolerance: "0.1"
 horizontal_pod_downscale_stabilization: "5m0s"
 
+# Vertical pod autoscaler version config for controlling roll-out
+vertical_pod_autoscaler_version: "v0.11.0-internal.17"
```
mikkeloscar: Maybe we can define this in a way so you don't need to know the old version to switch back. So instead of defining the version here we could have the value `current|legacy` or so and then have an if/else when choosing which image version to use.

cluster/manifests/01-vertical-pod-autoscaler/updater-deployment.yaml
```diff
@@ -25,7 +25,7 @@ spec:
       serviceAccountName: vpa-updater
       containers:
       - name: updater
-        image: container-registry.zalando.net/teapot/vpa-updater:v0.6.1-internal.16
+        image: container-registry.zalando.net/teapot/vpa-updater:{{.Cluster.ConfigItems.vertical_pod_autoscaler_version}}
```
mikkeloscar: Note that there are 3 components in total, those also need to be update, but see my suggestion above :)

cluster/config-defaults.yaml
demonCoder95: Hmm good idea. I'll add that!

cluster/manifests/01-vertical-pod-autoscaler/updater-deployment.yaml
demonCoder95: Ah I see. the `recommender` and `admission-controller` as well. I'll update those with the new value types as well :)

demonCoder95:
Permission denied when attempting to tag resources during rollout:
```bash


2022-09-15 10:40:27 UTC+0200 | e2e-pr-5357-8 | UPDATE_ROLLBACK_IN_PROGRESS | The following resource(s) failed to update: [WorkerFilesEncryptionKey, DeploymentSecretKey, RemoteFilesEncryptionKey, MasterFilesEncryptionKey, EtcdEncryptionKey].
-- | -- | -- | --
2022-09-15 10:40:26 UTC+0200 | MasterFilesEncryptionKey | UPDATE_FAILED | Resource update cancelled
2022-09-15 10:40:26 UTC+0200 | WorkerFilesEncryptionKey | UPDATE_FAILED | Resource update cancelled
2022-09-15 10:40:26 UTC+0200 | RemoteFilesEncryptionKey | UPDATE_FAILED | Resource update cancelled
2022-09-15 10:40:26 UTC+0200 | EtcdEncryptionKey | UPDATE_FAILED | Resource update cancelled
2022-09-15 10:40:25 UTC+0200 | DeploymentSecretKey | UPDATE_FAILED | Resource handler returned message: "Access denied for operation 'TagResource'." (RequestToken: 368ceb5f-b19a-c75a-22bf-b10736fe24f2, HandlerErrorCode: AccessDenied)


```

demonCoder95:
The `vpa-updater` has an outdated-flag in the container arguments:
```bash
unknown flag: --pod-lifetime-update-threshold
Usage of ./updater:
      --add-dir-header                                                  If true, adds the file directory to the header of the log messages
      --address string                                                  The address to expose Prometheus metrics. (default ":8943")
      --alsologtostderr                                                 log to standard error as well as files
      --evict-after-oom-threshold duration                              Evict pod that has only one container and it OOMed in less than
                                                                        		evict-after-oom-threshold since start. (default 10m0s)
      --eviction-rate-burst int                                         Burst of pods that can be evicted. (default 1)
unknown flag: --pod-lifetime-update-threshold
      --eviction-rate-limit float                                       Number of pods that can be evicted per seconds. A rate limit set to 0 or -1 will disable
                                                                        		the rate limiter. (default -1)
      --eviction-tolerance float                                        Fraction of replica count that can be evicted for update, if more than one pod can be evicted. (default 0.5)
      --in-recommendation-bounds-eviction-lifetime-threshold duration   Pods that live for at least that long can be evicted even if their request is within the [MinRecommended...MaxRecommended] range (default 12h0m0s)
      --kube-api-burst float                                            QPS burst limit when making requests to Kubernetes apiserver (default 10)
      --kube-api-qps float                                              QPS limit when making requests to Kubernetes apiserver (default 5)
      --kubeconfig string                                               Path to a kubeconfig. Only required if out-of-cluster.
      --log-backtrace-at traceLocation                                  when logging hits line file:N, emit a stack trace (default :0)
      --log-dir string                                                  If non-empty, write log files in this directory
      --log-file string                                                 If non-empty, use this log file
      --log-file-max-size uint                                          Defines the maximum size a log file can grow to. Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
      --logtostderr                                                     log to standard error instead of files (default true)
      --min-replicas int                                                Minimum number of replicas to perform update (default 2)
      --one-output                                                      If true, only write logs to their native severity level (vs also writing to each lower severity level)
      --pod-update-threshold float                                      Ignore updates that have priority lower than the value of this flag (default 0.1)
      --skip-headers                                                    If true, avoid header prefixes in the log messages
      --skip-log-headers                                                If true, avoid headers when opening log files
      --stderrthreshold severity                                        logs at or above this threshold go to stderr (default 2)
      --updater-interval duration                                       How often updater should run (default 1m0s)
      --use-admission-controller-status                                 If true, updater will only evict pods when admission controller status is valid. (default true)
  -v, --v Level                                                         number for the log level verbosity (default 0)
      --vmodule moduleSpec                                              comma-separated list of pattern=N settings for file-filtered logging
      --vpa-object-namespace string                                     Namespace to search for VPA objects. Empty means all namespaces will be used.
```

demonCoder95:
The `--pod-lifetime-update-threshold` parameter was removed in favor of `pod-update-threshold` and `in-recommendation-bounds-eviction-lifetime-threshold` in release `v0.10.0` of the `vertical-pod-autoscaler`. See the PR here https://github.com/kubernetes/autoscaler/pull/3962

Simply removing the flag from the `vpa-recommender` container should fix this problem.

demonCoder95:
The flag is safe to remove since the previous default in the code [0] was 12h, same as the value we had set in our manifest [1]

[0] https://github.com/kubernetes/autoscaler/pull/3962/files#diff-2d1951e8948b2e86bd416513edadea4f736079ed4648203400b7e200f422e354L37

[1] https://github.com/zalando-incubator/kubernetes-on-aws/pull/5361/commits/b8a6cb7e8f3d81e8d70a6604b991639c563da278#diff-40f82f16b6e3b890123b895ba55211ab3c4c471f4ac060dda689908930677ae0L39

So, the change is backwards compatible.

demonCoder95:
👍

demonCoder95:
I think the PR is ready. Let's roll out! :)

mikkeloscar:
:+1:

PR #5362, opened 2022-09-15
Update key policies to allow tagging

demonCoder95:
Updating the key policies for the following keys to add `tag:TagResources` permission as part of administration actions:
1. DecryptionSecretKey
2. WorkerFilesEncryptionKey
3. RemoteFilesEncryptionKey
4. MasterFilesEncryptionKey
5. EtcdEncryptionKey

This is done separately in the PR in order to succeed with the CF resource tagging work here: https://github.com/zalando-incubator/kubernetes-on-aws/pull/5357

This is because of the chicken and egg problem of updating the policies in the tagging PR itself, causing a race condition where the policy is never updated _before_ applying the tagging update to the CF stacks.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -738,6 +738,7 @@ Resources:
               - 'kms:Delete*'
               - 'kms:ScheduleKeyDeletion'
               - 'kms:CancelKeyDeletion'
+              - 'tag:TagResources'
             Effect: Allow
             Principal:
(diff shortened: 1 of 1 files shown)


demonCoder95:
I see on the console that CF stacks upgraded successfully to update the key policies! Let's merge this!

demonCoder95:
👍

gargravarr:
:+1:

PR #5382, opened 2022-09-21
fabric: make resources configurable

AlexanderYastrebov:
* adds missing cpu limit config


mikkeloscar:
Considering this runs as a singleton it probably also makes sense to add vertical pod autoscaling such that you don't have to care about these settings but can scale it based on e.g. amount of fabric gateway resources or whatever is mostly driving the resources.

You can see example of this for all singleton components we run in kube-system e.g. prometheus.

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #5393, opened 2022-09-26
[Proposal] Scale down etcd in e2e outside of working hours

mikkeloscar:
Proposal on how we can scale down the etcd cluster used in e2e outside of working hours.

It uses the scheduled action feature of AWS Autoscaling Groups: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-as-scheduledaction.html

This has two benefits:
* We clean etcd storage every day so we don't have a build-up of data which leads to slower etcd
* We save a bit of cost outside of office hours.


gargravarr:
:+1:

mikkeloscar:
:+1:

PR #5400, opened 2022-09-28
Scale down empty nodes faster

mikkeloscar:
Instead of waiting 10m before scaling down an empty node, scale down faster to save a bit of cost. It's unclear exactly how much we can safe but there is no need to run empty nodes for 10m.

We have been running with this reduced setting in a couple of clusters. Enable it everywhere as we don't see any negative effects of reducing this.


rodrigodelmonte:
:+1:

mikkeloscar:
:+1:

PR #5406, opened 2022-09-29
Disable Prometheus PCS when not used

mikkeloscar:
Disables the Prometheus PCS when remote write is not enabled. Also changed the application from `prometheus` to `kubernetes` which is the active application ID.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #5436, opened 2022-10-10
Enable spot node rescheduling via CronJob

katyanna:
Currently once a spot node dies, the pods fall to on-demand and can only go back to spot when the nodes are eventually rolled out. This PR adds a CronJob that every 5 minutes checks for spot availability and reintroduces spot nodes as soon as possible. To enable the behaviour the `spot_node_rescheduler` flag must be set to `true`.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -506,6 +506,10 @@ teapot_admission_controller_runtime_policy_default: "allow-spot"
 teapot_admission_controller_runtime_policy_enforced: ""
 # Enable hard spot assignment. Only relevant when node_lifecycle_provider=zalando
 teapot_admission_controller_runtime_policy_spot_hard_assignment: "false"
+# Enable experimental rescheduling of spot nodes after spot decommission
+spot_node_rescheduler: "false"
+spot_node_rescheduler_memory: "348Mi"
+spot_node_rescheduler_cpu: "50m"
 
 # Enable and configure prevent scale down annotation
 teapot_admission_controller_prevent_scale_down_enabled: "true"
--- a/cluster/manifests/spot-node-rescheduler/cronjob.yaml
+++ b/cluster/manifests/spot-node-rescheduler/cronjob.yaml
@@ -1,0 +1,35 @@
+{{ if eq .ConfigItems.spot_node_rescheduler "true" }}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: node-spot-rescheduler
+  namespace: kube-system
+  labels:
+    application: kubernetes
+    component: node-spot-rescheduler
+spec:
+  schedule: "*/5 * * * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      activeDeadlineSeconds: 1800
+      backoffLimit: 1
+      template:
+        metadata:
+          labels:
+            application: kubernetes
+            component: node-spot-rescheduler
+        spec:
+          serviceAccountName: node-spot-rescheduler
+          restartPolicy: Never
+          containers:
+          - name: node-spot-rescheduler
+            image: container-registry.zalando.net/teapot/spot-node-rescheduler:main-1
+            resources:
+            limits:
+              cpu: "{{ .ConfigItems.spot_node_rescheduler_cpu }}"
+              memory: "{{ .ConfigItems.spot_node_rescheduler_memory }}"
+            requests:
+              cpu: "{{ .ConfigItems.spot_node_rescheduler_cpu }}"
+              memory: "{{ .ConfigItems.spot_node_rescheduler_memory }}"
+{{ end }}
--- a/cluster/manifests/spot-node-rescheduler/rbac.yaml
+++ b/cluster/manifests/spot-node-rescheduler/rbac.yaml
@@ -1,0 +1,33 @@
+{{ if eq .ConfigItems.spot_node_rescheduler "true" }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: spot-node-rescheduler
+  namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: spot-node-rescheduler
+rules:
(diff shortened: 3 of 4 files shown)


cluster/config-defaults.yaml
```diff
@@ -506,6 +506,8 @@ teapot_admission_controller_runtime_policy_default: "allow-spot"
 teapot_admission_controller_runtime_policy_enforced: ""
 # Enable hard spot assignment. Only relevant when node_lifecycle_provider=zalando
 teapot_admission_controller_runtime_policy_spot_hard_assignment: "false"
+# Enable experimental rescheduling of spot nodes after spot decommission
+node-spot-rescheduler: "false"
```
mikkeloscar: nit: This should be `snake_case`

cluster/manifests/spot-node-rescheduler/cronjob.yaml
```diff
@@ -0,0 +1,36 @@
...
+            component: node-spot-rescheduler
+        spec:
+          serviceAccountName: node-spot-rescheduler
+          restartPolicy: Never
+          containers:
+          - name: node-spot-rescheduler
+            image: container-registry-test.zalando.net/teapot/spot-node-rescheduler-test:pr-1-4
+            env:
+            - name: EXPERIMENTAL_SPOT_RESCHEDULER
+              value: "{{ .ConfigItems.experimental_spot_rescheduler }}"
```
mikkeloscar: I think better than having this as a flag we can just have and if around this whole manifest. Like we do here: https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/kube-downscaler/deployment.yaml#L1


And the corresponding config in the `deletions.yaml` to be able to drop it when the config-item is disabled: https://github.com/zalando-incubator/kubernetes-on-aws/blob/c5823ca2d00d73c4bb2cc5b365cf80abf4704d42/cluster/manifests/deletions.yaml#L89-L103

cluster/manifests/spot-node-rescheduler/cronjob.yaml
```diff
@@ -0,0 +1,35 @@
+{{ if ne .ConfigItems.spot_node_rescheduler "true" }}
```
mikkeloscar:
```suggestion
{{ if eq .ConfigItems.spot_node_rescheduler "true" }}
```

mikkeloscar: Should also have the corresponding change in the deletions.yaml:

Example:

https://github.com/zalando-incubator/kubernetes-on-aws/blob/c5823ca2d00d73c4bb2cc5b365cf80abf4704d42/cluster/manifests/deletions.yaml#L89-L103

cluster/manifests/spot-node-rescheduler/rbac.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - get
+  - watch
+  - list
+  - update
+  - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: spot-node-rescheduler-privileged-psp
```
mikkeloscar: I don't think we need privileged psp. This should be a binding to the `spot-node-rescheduler` `ClusterRole` as defined above.

cluster/manifests/spot-node-rescheduler/rbac.yaml
```diff
@@ -0,0 +1,40 @@
+{{ if eq .ConfigItems.spot_node_rescheduler "true" }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: spot-node-rescheduler
+  namespace: kube-system
+  annotations:
+    iam.amazonaws.com/role: "{{ .Cluster.LocalID }}-spot-node-rescheduler"
```
mikkeloscar: There is no AWS role needed for the spot-node-rescheduler so we can drop this annotation.

cluster/manifests/spot-node-rescheduler/cronjob.yaml
```diff
@@ -0,0 +1,35 @@
...
+          containers:
+          - name: node-spot-rescheduler
+            image: container-registry.zalando.net/teapot/spot-node-rescheduler:main-1
+            resources:
+              limits:
+                cpu: 50m
+                memory: 384Mi
+              requests:
+                cpu: 50m
+                memory: 384Mi
```
mikkeloscar: It's good practice to define these with a config-item such that we can increase per cluster if needed. We don't know ahead of time how much it will consume so better to make it easy to change. See other manifests for examples.

cluster/config-defaults.yaml
```diff
@@ -506,6 +506,10 @@ teapot_admission_controller_runtime_policy_default: "allow-spot"
 teapot_admission_controller_runtime_policy_enforced: ""
 # Enable hard spot assignment. Only relevant when node_lifecycle_provider=zalando
 teapot_admission_controller_runtime_policy_spot_hard_assignment: "false"
+# Enable experimental rescheduling of spot nodes after spot decommission
+spot_node_rescheduler: "false"
+spot_node_rescheduler_memory: "1Gi"
+spot_node_rescheduler_cpu: "1000m"
```
katyanna: Are these reasonable values?

mikkeloscar: I think they are too high.

What was there e.g. `50m` and `384Mi` doesn't sound like a bad starting point.

cluster/manifests/spot-node-rescheduler/rbac.yaml
```diff
@@ -0,0 +1,30 @@
...
+metadata:
+  name: spot-node-rescheduler
+  namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: spot-node-rescheduler
+rules:
+- apiGroups:[""]
```
mikkeloscar: nit: looks from the syntax that a space is needed here after `apiGroups:`

mikkeloscar:
:+1:

katyanna:
👍

PR #5438, opened 2022-10-12
Shorten expiration of etcd backup objects.

gargravarr:
The etcd S3 bucket contains backups taken from etcd clusters every minute. We use these backups when we need to recover an etcd cluster after an event that compromises the current data.

When recovering an etcd cluster, we normally use the most recent backup, reflecting the last known state of etcd. Therefore having the current bucket object expiration policy set to 2 days does not makes sense and incur in unnecessary S3 storage costs.

This pull request reduces the object expiration policy to 1 day, hopefully reducing by more or less half the size of the etcd bucket. Ideally we could have an expiration policy of 1 or 2 hours, but unfortunately the S3 lifecycle configuration only allows to specify the number of days for the expiration time.
 

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #5447, opened 2022-10-14
skipper-ingress: allow disabling filters

AlexanderYastrebov:

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -82,9 +82,8 @@ skipper_suppress_route_update_logs: "true"
 skipper_validate_query: "true"
 skipper_validate_query_log: "false"
 
-# skipper default filters
 skipper_default_filters: 'enableAccessLog(4,5) -> fifo(2000,20000,"3s")'
-# inline routes
+skipper_disabled_filters: ""
 skipper_ingress_inline_routes: ""
 skipper_ingress_refuse_payload: ""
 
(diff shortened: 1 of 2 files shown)


cluster/config-defaults.yaml
```diff
@@ -82,9 +82,8 @@ skipper_suppress_route_update_logs: "true"
 skipper_validate_query: "true"
 skipper_validate_query_log: "false"
 
-# skipper default filters
 skipper_default_filters: 'enableAccessLog(4,5) -> fifo(2000,20000,"3s")'
-# inline routes
+skipper_disabled_filters: ""
```
AlexanderYastrebov, line 86: Named similar to `skipper_default_filters`, i.e. did not use `skipper_ingress_` prefix

cluster/manifests/skipper/deployment.yaml
```diff
@@ -201,6 +201,7 @@ spec:
           - "-read-timeout-server={{ .ConfigItems.skipper_read_timeout_server }}"
           - "-write-timeout-server={{ .ConfigItems.skipper_write_timeout_server }}"
           - '-default-filters-prepend={{ .ConfigItems.skipper_default_filters }}'
+          - "-disabled-filters={{ .ConfigItems.skipper_disabled_filters }}"
```
AlexanderYastrebov, line 204: Empty is the default flag value so there is no need for if-empty check

szuecs:
:+1:

MustafaSaber:
👍  - I think my vote still doesn't count here

AlexanderYastrebov:
:+1:

PR #5463, opened 2022-10-21
Allow to configure registry pull qps and burst

linki:
During the load test we saw that kubelet massively rate-limited itself when pulling from the registries. According to the team managing those there's enough capacity to increase these values. Let's make them configurable so we can adjust them before the next load test.

For reference: https://github.bus.zalan.do/teapot/issues/issues/3353


szuecs:
:+1:

mikkeloscar:
This will rotate all the nodes :/

@linki Maybe we can have an if condition for setting this in the manifest so we only set it if it's different from the default?

linki:
This hasn't become necessary in subsequent load tests. Closing for now to avoid further introduction of configuration options.

PR #5466, opened 2022-10-21
Update Admission Controller

katyanna:
Validate hostname length on Ingresses, RouteGroups, and Services
with the external-dns annotation, by ensuring that the hostname
length is still in accordance with _Route53_ limit after _external-dns_
`cname-` prefix addition on _v0.12.2_.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -811,4 +811,4 @@ hyped_article_lifecycle_management: "false"
 enable_size_memory_backed_volumes: "true"
 
 # specify hostname max allowed length
-hostname-max-length: 57
+hostname_max_length: 57
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -809,3 +809,6 @@ hyped_article_lifecycle_management: "false"
 
 # enable SizeMemoryBackedVolumes feature flag
 enable_size_memory_backed_volumes: "true"
+
+# specify hostname max allowed length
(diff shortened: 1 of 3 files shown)


cluster/config-defaults.yaml
```diff
@@ -809,3 +809,6 @@ hyped_article_lifecycle_management: "false"
 
 # enable SizeMemoryBackedVolumes feature flag
 enable_size_memory_backed_volumes: "true"
+
+# specify hostname max allowed length
+hostname-max-length: 57
```
linki: In this file we use snake case 😆

demonCoder95:
👍

katyanna:
👍

cluster/config-defaults.yaml
```diff
@@ -809,3 +809,6 @@ hyped_article_lifecycle_management: "false"
 
 # enable SizeMemoryBackedVolumes feature flag
 enable_size_memory_backed_volumes: "true"
+
+# specify hostname max allowed length
+hostname_max_length: 57
```
mikkeloscar, line 814: I'm a bit surprised CLM doesn't complain. Config Items are always strings so it should ideally be `"57"` defined here.

katyanna: interesting, will fix it.

katyanna: https://github.com/zalando-incubator/kubernetes-on-aws/pull/5468

PR #5469, opened 2022-10-21
fabric: drop unused header

AlexanderYastrebov:
The `X-Route-Controller` was added to distingush between routes of old and new controller and is not used anymore.


AlexanderYastrebov:
:+1:

mikkeloscar:
:+1:

PR #5470, opened 2022-10-21
Adding kubenurse

myaser:
testing [kubenurse](https://github.com/postfinance/kubenurse), it monitors all network connections in a Kubernetes cluster. Kubenurse measures request durations, records errors, and exports those metrics in Prometheus format

[related to teapot/issues#1950](https://github.bus.zalan.do/teapot/issues/issues/1950)


cluster/manifests/z-kubenurse/service.yaml
```diff
@@ -0,0 +1,19 @@
...
+  namespace: kubenurse
+spec:
+  ports:
+    - name: http
+      port: 8080
+      protocol: TCP
+      targetPort: 8080
+  selector:
+    application: kubernetes
+    component: kubenurse
```
mikkeloscar: What is the purpose of Service (and Ingress)?

Service for a daemonset is generally not desired because it means we need to track ALL pods in the iptable rules of all nodes. So the bigger the cluster, the bigger per node iptable. We have hit memory limits in flannel because of this in the past, so if we don't strictly need the service, it's better to not have it.

myaser: kubenurse uses the service for one of its tests [(reach to self through service)](https://github.com/postfinance/kubenurse#me-service)

> Checks if the kubenurse is reachable at the /alwayshappy endpoint through the Kubernetes service. The address is provided by the environment variable KUBENURSE_SERVICE_URL that could look like http://kubenurse.mynamespace.default.svc:8080. This also verifies a working kube-proxy setup.

myaser: similarly, the ingress is used to verify it works upstream (DNS resolution and routing)

myaser: > if we don't strictly need the service (& ingress), it's better not to have it

I think the bare minimum requirement is that we check pod-to-pod communication, although these checks are useful, maybe we can omit them for now?

mikkeloscar: > I think the bare minimum requirement is that we check pod-to-pod communication, although these checks are useful, maybe we can omit them for now?

I agree. Better to start small especially when we know of the problems of the service for daemonset setup.

myaser: in progress
[postfinance/kubenurse/pull#54](https://github.com/postfinance/kubenurse/pull/54)

cluster/manifests/z-kubenurse/daemonset.yaml
```diff
@@ -0,0 +1,83 @@
...
+              value: {{.Cluster.ConfigItems.network_monitoring_check_api_server_direct}}
+            - name: KUBENURSE_CHECK_API_SERVER_DNS
+              value: {{.Cluster.ConfigItems.network_monitoring_check_api_server_dns}}
+            - name: KUBENURSE_CHECK_ME_SERVICE
+              value: {{.Cluster.ConfigItems.network_monitoring_check_kubenurse_service}}
+            - name: KUBENURSE_CHECK_ME_INGRESS
+              value: {{.Cluster.ConfigItems.network_monitoring_check_kubenurse_ingress}}
+            - name: KUBENURSE_CHECK_NEIGHBOURHOOD
+              value: {{.Cluster.ConfigItems.network_monitoring_check_neighborhood}}
+          image: "postfinance/kubenurse:latest"
```
myaser: we need to have the latest image also compliant and supports arm64

myaser: https://github.bus.zalan.do/teapot/kubenurse-pipeline/pull/1

cluster/manifests/z-kubenurse/daemonset.yaml
```diff
@@ -0,0 +1,83 @@
+{{if eq .Cluster.ConfigItems.network_monitoring_enabled "true" }}
```
linki: For completeness sake let's also remove the resources via `deletions.yaml` if it's turned off.

cluster/manifests/z-kubenurse/daemonset.yaml
```diff
@@ -0,0 +1,83 @@
...
+      serviceAccountName: "kubenurse"
+      containers:
+        - name: "kubenurse"
+          env:
+            - name: KUBENURSE_INGRESS_URL
+              value: https://kubenurse.{{ .Values.hosted_zone }}
+            - name: KUBENURSE_SERVICE_URL
+              value: "http://kubenurse.kubenurse.svc.cluster.local:8080"
+            - name: KUBENURSE_INSECURE
+              value: "true"
```
linki: If that's for calling `https://kubenurse.{{ .Values.hosted_zone }}` we might be able to set it to `false` since it serves a valid certificate. If it's for the internal API server we might be able to use `KUBENURSE_EXTRA_CA` to point it to the mounted certificate.

cluster/manifests/z-kubenurse/daemonset.yaml
```diff
@@ -0,0 +1,83 @@
...
+              scheme: HTTP
+            periodSeconds: 3
+            successThreshold: 1
+            timeoutSeconds: 1
+      tolerations:
+        - effect: NoSchedule
+          key: node.kubernetes.io/role
+          value: master
+        - operator: Exists
+          effect: NoSchedule
```
linki: This makes the other toleration redundant I believe. We can also use the pattern that we use for all other DaemonSets that should run everywhere, e.g. [flannel](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/flannel/daemonset.yaml#L127-L131):

```yaml
      tolerations:
      - operator: Exists
        effect: NoSchedule
      - operator: Exists
        effect: NoExecute
```

cluster/manifests/z-kubenurse/rbac.yaml
```diff
@@ -0,0 +1,64 @@
+{{ if eq .Cluster.ConfigItems.network_monitoring_enabled "true" }}
+{{ if eq .Cluster.ConfigItems.network_monitoring_check_unschedulable_nodes "false" }}
```
linki: Is that `if` correct or should it be `true`?

cluster/manifests/z-kubenurse/rbac.yaml
```diff
@@ -0,0 +1,64 @@
...
+---
+# Source: kubenurse/templates/rbac.yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: "kubenurse"
```
linki: Let's also add the application's labels to all resources:
```
  labels:
    application: kubernetes
    component: kubenurse
```

cluster/manifests/z-kubenurse/service.yaml
```diff
@@ -0,0 +1,21 @@
+{{ if and (eq .Cluster.ConfigItems.network_monitoring_enabled "true") (eq .Cluster.ConfigItems.network_monitoring_check_kubenurse_service "true")}}
```
linki: Since the Ingress needs the Service usually, we might need to create it when `network_monitoring_check_kubenurse_ingress == true` as well.

cluster/manifests/z-kubenurse/rbac.yaml
myaser: We only need this role in case we do *NOT* want to check unschedulable nodes. then we will create node watchers to know what are the unschedulable nodes to exclude them

if we want to check all nodes (flag == true) we then do not need this roles

cluster/manifests/z-kubenurse/daemonset.yaml
myaser: the kubenurse checker has a single http client for all checks (including pod to pod communication or calling api server or calling the me_ingress)

I was more thinking of calling the pod ip may fail

cluster/manifests/z-kubenurse/daemonset.yaml
```diff
@@ -0,0 +1,81 @@
...
+  labels:
+    application: kubernetes
+    component: kubenurse
+  name: kubenurse
+  namespace: kubenurse
+spec:
+  selector:
+    matchLabels:
+      application: kubernetes
+      component: kubenurse
```
mikkeloscar: let's use the `daemonset: kubenurse` label pattern for the matchLabels like we do for all other components.

cluster/manifests/z-kubenurse/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
+{{ if eq .Cluster.ConfigItems.network_monitoring_enabled "true" }}
```
linki: Does it need to be at the very end of the manifest list? Or can we rename the folder to `kubenurse` to avoid confusion?

myaser:
👍🏻

cluster/manifests/kubenurse/daemonset.yaml
```diff
@@ -0,0 +1,81 @@
...
+            - name: KUBENURSE_CHECK_NEIGHBOURHOOD
+              value: "{{.Cluster.ConfigItems.network_monitoring_check_neighborhood}}"
+          image: "container-registry.zalando.net/teapot/kubenurse:histogram-main-2"
+          resources:
+            requests:
+              cpu: "40m"
+              memory: "100Mi"
+            limits:
+              cpu: "40m"
+              memory: "100Mi"
```
linki, line 63: Does that play well with the current the DaemonSet reservations?

linki:
LGTM but did we update or check the DaemonSet resource reservations already?

cluster/manifests/kubenurse/daemonset.yaml
myaser: the defaults provided in the `config-defaults.yaml` are very high, so it fits there

every cluster overrides the defaults to different values, thus modifying them will be part of the rollout.
we need to adjust the reservations per cluster before enabling the feature flag

linki:
👍

PR #5485, opened 2022-11-01
Make kube-node-ready opt-in rather than opt-out

mikkeloscar:
By default `kube-node-ready` run on ALL nodes. This is fine when all nodes are managed by Autoscaling Groups (ASG) but is a problem if they are not like Karpenter #5136 where we would need to opt-out.

This PR does three things:

1. Adds a label `asg-lifecycle-hook=true` to worker nodes which has a profile based on ASGs.
2. Adds a node selector for `asg-lifecycle-hook=true` on `kube-node-ready` daemonset such that it only gets scheduled on nodes which indicate the nodes are managed by an ASG.
3. Makes the ASG Lifecycle Hook and label optional on control plane nodes. This way we can experiment turning it off for control plane nodes which is an idea to reduce time to recover the control plane nodes getting overloaded. It's still enabled by default.

Because this sets a label on worker nodes it will roll all nodes but could be combined with other changes.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -847,3 +847,6 @@ network_monitoring_check_kubenurse_ingress: "false"
 network_monitoring_check_neighborhood: "true"
 network_monitoring_check_unschedulable_nodes: "true"
 network_monitoring_check_interval: "1m"
+
+# specify if control plane nodes should rely on ASG Lifecycle Hook or not
+control_plane_asg_lifecycle_hook: "true"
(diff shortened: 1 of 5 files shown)


mikkeloscar:
This now targets `kube-1.22` branch so it can be introduced together with rolling the nodes for `kube-1.22`

mikkeloscar:
:+1:

mikkeloscar:
:+1:

demonCoder95:
👍

PR #5490, opened 2022-11-03
Update etcd AMI.

gargravarr:
New etcd AMI adds 2 collectors to prometheus node exporter: `ethtool` and `processes`. This way we can collect e.g. ENA metrics.

Will only have effect after rotating etcds.


gargravarr:
:+1:

linki:
👍

PR #5491, opened 2022-11-03
Add support for k8s-audit-metrics

mikkeloscar:
Introduce optional support for enabling https://github.com/giantswarm/k8s-audit-metrics such that we can expose metrics about clients connecting to the apiserver. This is helpful to understand where load is coming from and for finding misbehaving clients.

This is enabled with the config item: `auditlog_metrics` which is false by default in production.

This PR also cleans up a bit of the audit-policy configuration which we had from the past where we wanted to identify different things for migrations. It is now a bit more simplified keeping the same functionality.

One challenge we have is that we don't log many read events because we don't want to overload our internal audit-trail system. This however also means that we won't count those events in the k8s-audit-metrics. Read access events for services accounts not in `kube-system` can optionally be enabled via `auditlog_read_access` config item.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -639,13 +639,17 @@ rotate_service_account_tokens: "true"
 # issue tokens with a long expiration time in order to detect applications that don't refresh tokens.
 rotate_service_account_tokens_extended_expiration: "true"
 
-# enable auditlogging for read access such that we can identified clients using
-# the default service account to read from the API server.
+# enable auditlogging metrics such that we can identify clients calling the
+# apiserver.
 {{ if eq .Cluster.Environment "test" }}
-auditlog_read_access: "true"
+auditlog_metrics: "true"
 {{ else }}
-auditlog_read_access: "false"
+auditlog_metrics: "false"
 {{ end }}
+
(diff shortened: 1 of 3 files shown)


demonCoder95:
👍

demonCoder95:
We removed the previous audit policy from the userdata, what's the new one? 🤔 I don't see it in the patch.

demonCoder95:
Is it the case that we will be using the default policy and we got rid of the service-account read policy that we had already (for reasons you provide in the description)? The data available in the default policy is enough to have meaningful metrics in Prometheus? 🤔

mikkeloscar:
@demonCoder95 This is the one that will be effective: https://github.com/zalando-incubator/kubernetes-on-aws/blob/5db5b028f732034c1f19bfe2e6fa0bee3c5959fd/cluster/node-pools/master-default/userdata.yaml#L720-L833

The one I removed was a slimmed down version only tracking some read events. It was there to test this in test clusters long time ago. Now both test and prod has the same more elaborate policy.

demonCoder95:
Ah, that's perfect 🙂  Thanks!

So much history to learn 🙃 📚 📝

demonCoder95:
👍

mikkeloscar:
:+1:

PR #4590, opened 2021-09-02
ethtool metrics and qdisc metrics

szuecs:
about 220 metrics measured in teapot cluster skipper node
Signed-off-by: Sandor Szücs <sandor.szuecs@zalando.de>

PR #5496, opened 2022-11-04
Collect ethtool metrics in prometheus

mikkeloscar:
Follow up to #4590 to collect the `ethtool` metrics in prometheus if the collector is enabled.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #5498, opened 2022-11-04
Fix nil pointer crash in k8s-audit-metric

mikkeloscar:
Follow up to #5491

Fixes a `invalid memory address or nil pointer dereference` that made it crash.


mikkeloscar:
:+1:

linki:
👍

PR #5506, opened 2022-11-07
Update Admission Controller domain length config

katyanna:
- Allow `kubernurse` daemonset
- For clearer naming, rename `hostname_max_length` as
`subdomain_max_length`.

### DIFF
--- a/cluster/manifests/01-admission-control/config.yaml
+++ b/cluster/manifests/01-admission-control/config.yaml
@@ -7,7 +7,7 @@ data:
   daemonset.reserved.cpu: "{{ .Cluster.ConfigItems.teapot_admission_controller_daemonset_reserved_cpu }}"
   daemonset.reserved.memory: "{{ .Cluster.ConfigItems.teapot_admission_controller_daemonset_reserved_memory }}"
 
-  dns.default.hostname-max-length: "{{ .Cluster.ConfigItems.hostname_max_length }}"
+  dns.default.subdomain-max-length: "{{ .Cluster.ConfigItems.subdomain_max_length }}"
 
   pod.container-resource-control.min-memory-request: "25Mi"
   pod.container-resource-control.default-cpu-request: "{{ .Cluster.ConfigItems.teapot_admission_controller_default_cpu_request }}"
--- a/cluster/manifests/01-admission-control/config.yaml
+++ b/cluster/manifests/01-admission-control/config.yaml
@@ -8,6 +8,7 @@ data:
   daemonset.reserved.memory: "{{ .Cluster.ConfigItems.teapot_admission_controller_daemonset_reserved_memory }}"
 
   dns.default.subdomain-max-length: "{{ .Cluster.ConfigItems.subdomain_max_length }}"
+  dns.default.hostname-max-length: "{{ .Cluster.ConfigItems.hostname_max_length }}" # to be removed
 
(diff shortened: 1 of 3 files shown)


mikkeloscar:
I think this roll out needs to be done in two phases to be safe.

1. Add the new config option to the configmap, but also keep the old `dns.default.hostname-max-length` there.
2. Update the admission-controller image.

With this PR the image would first be updated and the image would assume the config value: `dns.default.subdomain-max-length` which would not yet be applied so I assume it would fall back to the default value which is not correct in all clusters.

cluster/manifests/01-admission-control/config.yaml
```diff
@@ -7,7 +7,8 @@ data:
   daemonset.reserved.cpu: "{{ .Cluster.ConfigItems.teapot_admission_controller_daemonset_reserved_cpu }}"
   daemonset.reserved.memory: "{{ .Cluster.ConfigItems.teapot_admission_controller_daemonset_reserved_memory }}"
 
-  dns.default.hostname-max-length: "{{ .Cluster.ConfigItems.hostname_max_length }}"
+  dns.default.subdomain-max-length: "{{ .Cluster.ConfigItems.subdomain_max_length }}"
+  dns.default.hostname-max-length: "{{ .Cluster.ConfigItems.hostname_max_length }}" # to be removed
```
linki, line 11: As @mikkeloscar already suggested I also believe this needs to be rolled out in a separate PR before this one.

This first thing that happens is updating the userdata.yaml below which will roll the master nodes. The masters will run `admission-controller:master-157` which is looking for the key `subdomain-max-length` in the admission config map. However, that config map hasn't been updated yet because that happens after. It's only a windows of like 20 minutes if you don't roll any workers, too, but that's too long.

Separating the userdata change into another PR should fix it. And I think you can also change this part here to
```
dns.default.subdomain-max-length: "{{ .Cluster.ConfigItems.subdomain_max_length }}"
dns.default.hostname-max-length: "{{ .Cluster.ConfigItems.subdomain_max_length }}" # to be removed
```
to safe one config item (but i'm not sure, your solution is definitely safer in that regard).

katyanna: Got it. I'll create another PR updating the config-items with the duplication and change this one to just delete the `hostname-max-length` and update the Admission Controller.

linki: Sounds good. Keep in mind that it needs to be split this way for every rollout to the other channels as well. This requires a little bit of coordination and we can easily forget it.

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -205,7 +205,7 @@ write_files:
             requests:
               cpu: 100m
               memory: 200Mi
-        - image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/admission-controller:master-155
+        - image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/admission-controller:master-157
```
linki: Let's already use the latest version here: `master-158`

linki:
@katyanna Given that this now contains unrelated changes, can we close this PR in favour of https://github.com/zalando-incubator/kubernetes-on-aws/pull/5510? Looks like they will be the same after all.

PR #5513, opened 2022-11-09
Use different images for each image policy test case

linki:
The image policy tests use images that are compliant in one namespace and non-compliant in another namespace to test that images really are rejected for image policy reasons and not something else.

However, this introduces non-compliant images into the image cache that interfere with the test cases. By using different images for each test case this should not happen.


mikkeloscar:
:+1:

linki:
👍

PR #5460, opened 2022-10-18
Add flags to overcome routesrv and skipper routes diff

MustafaSaber:
Updates https://github.bus.zalan.do/pg9/issues/issues/33

Signed-off-by: Mustafa Abdelrahman <mustafa.abdelrahman@zalando.de>

PR #5529, opened 2022-11-23
test: create-cluster fails on invalid template

AlexanderYastrebov:


szuecs:
Yes all pods are checked to be running before we run tests

AlexanderYastrebov:
I was reproducing https://github.com/zalando-incubator/kubernetes-on-aws/pull/5460#discussion_r999131532 to better understand it as pipeline logs are not available anymore.

The problem it takes long is because first it creates a cluster from the baseline and only then applies malformed templates from the PR.

We'd need to render/validate templates before attempting to create a cluster.

PR #5530, opened 2022-11-25
Make external-dns memory configurable

demonCoder95:
It was observed that for larger clusters like `fashion-store` external-dns can be OOM killed due to processing too much information (it has pod, node, endpoint informers enabled for our use case). Currently, the memory limit is not configurable.

This patch makes it possible to increase memory limit for external-dns.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -670,6 +670,9 @@ external_dns_policy: sync
 # legacy => v0.9.0-master-26
 external_dns_version: "current"
 
+# resource configuration
+external_dns_mem: "4Gi"
+
 # select which cache to use for Cluster DNS: unbound or dnsmasq.
 dns_cache: "dnsmasq"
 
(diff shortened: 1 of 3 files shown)


linki:
ExternalDNS has a VPA here: https://github.com/zalando-incubator/kubernetes-on-aws/blob/6aec328cca693f039cbe899ef2a50ed24273394b/cluster/manifests/external-dns/vpa.yaml

We should change the maximum allowed memory there, similar to https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/prometheus/prometheus-vpa.yaml#L24

demonCoder95:
Ah, got it. I'll update, thanks :)

demonCoder95:
Should I introduce a separate item for configuring the max for VPA? Like "external_dns_vpa_mem_max" 🤔

myaser:
:+1

myaser:
👍🏻

myaser:
👍🏻

myaser:
👍🏻

zaklawrencea:
👍

PR #5534, opened 2022-11-30
Block v1beta1 CRDs by default

demonCoder95:
Updates the sane default of `allow_v1beta1_crds` from `true` to `false` since the CRD rollout is almost finished. This config-item will stay in place until all clusters have been rolled out to 1.22. Afterwards, it will be removed.


linki:
👍

mikkeloscar:
👍

PR #5538, opened 2022-12-01
Add cloudformation definition for etcd key.

gargravarr:
This Pull Request adds the CloudFormation (CF) definition for the key used by CLM and etcd to exchange secrets on bootstrap of etcd.

For now CLM won't apply this CF definition. First we need to import the existing keys to CF.

The file name is prefixed with `pre`, to imply that CLM will apply it before the main stack.


gargravarr:
:+1:

demonCoder95:
👍

katyanna:
👍

PR #5541, opened 2022-12-05
Make autoscaler log level configurable

mikkeloscar:
This makes the autoscaler log level configurable per cluster so it's easier to temporally enable debug logging when trying to understand why a pod is not being scheduled on an expected instance type size.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #5542, opened 2022-12-05
Enhance audittrail adapter metrics

mikkeloscar:
This is an evolution of what was first introduced in #5491 

This enhances the audittrail-adapter component such that we can use the audit events and Prometheus metrics exposed to get detailed information about clients calling the Kubernetes API. This can be useful for the following cases and more:

* Detect clients not rotating service tokens: `authentication_stale_token` metric.
* Detect clients with high load on the API Server for certain calls.
* Detect client calling old api_version/api_groups for resources.

What changes is that the audittrail-adapter now can take a list of dimension to include in the event Prometheus metric exposed. This enables us to enrich the metric with information from the events.
The desired dimensions can be configured via the config-item `auditlog_metric_dimensions`. Possible values are: `authorization_decision,authentication_stale_token,user,user_agent,verb,code,resource,api_group,api_version` as also documented in the `config-defaults.yaml` file.

This change also enables `audittrail-adapter` to run in `test` clusters which was previously not the case. It runs in a mode where it only uses the events to enrich the Prometheus metrics, it doesn't forward the events anywhere so the behavior is similar to the past where audittrail-adapter was not running in `test`.

Lastly it removes the `k8s-audit-metrics` container previously introduced in #5491. This one has the same purpose as what is now implemented in `audittrail-adapter`, the new implementation is just more robust and easier to maintain being a project we own internally.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -646,16 +646,24 @@ rotate_service_account_tokens: "true"
 # issue tokens with a long expiration time in order to detect applications that don't refresh tokens.
 rotate_service_account_tokens_extended_expiration: "true"
 
-# enable auditlogging metrics such that we can identify clients calling the
-# apiserver.
-{{ if eq .Cluster.Environment "test" }}
-auditlog_metrics: "true"
-{{ else }}
-auditlog_metrics: "false"
-{{ end }}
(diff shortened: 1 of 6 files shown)


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #5544, opened 2022-12-05
Disable auditlog_read_access by default

mikkeloscar:
Follow up to #5542 to disable `auditlog_read_access` by default. We only want to enable it when investigating something.


mikkeloscar:
:+1:

katyanna:
:+1:

PR #5547, opened 2022-12-05
Revamp existing CSI implementation to latest updates

linki:
Revamp of our old CSI implementation. There are no significant changes here. It works similar to what we had back then.

There are three config items in this PR, based on the previous implementation:

* `enable_csi`: (new) This will deploy/remove the CSI system and create a StorageClass for it. When enabled, no migration of volumes is happening but CSI can be tested by using the dedicated StorageClass `ebs-sc`.
* `enable_csi_migration`: Turn on CSI migration which allows to use old PVC syntax and storage class but handle it via CSI internally. New volumes will use the CSI storage class directly. (Not part of this PR anymore but still left for context. CSI migration is handled [in a separate PR](https://github.com/zalando-incubator/kubernetes-on-aws/pull/5551))
* `prometheus_csi_ebs`: Deploy a replacement Prometheus that uses the new CSI-based StorageClass directly. It's mostly for testing as there's no migration happening and all data in prometheus will be lost.

The notable changes:
* Updated the versions and some flags to the latest state
* Added a config item to install/uninstall the whole CSI system. This allows to rollout CSI for individual clusters without opting everyone in via CSIMigration.
* The YAML files are reorganised to match [upstream](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/deploy/kubernetes/base) more closely, which should make it easier to keep it up to date.
* The CSI-Controller part has been changed from a DaemonSet on Master Nodes to a Deployment with two replicas on Master nodes to match the upstream files.

TODOs:
* [x] There are some TODOs in the files that need to be cleaned up
* [x] The old implementation didn't clean up all resources when we disabled it. Needs to be added to deletions.
* [x] The storageClass' topology settings doesn't work with CSIMigration (spaces). Needs to be handled before enabling CSIMigration (resolved as CSIMigration isn't part of the PR anymore).
* [x] When enabled, a new DaemonSet is added which needs to be added to reserved resources (resolved as this is part of the PR that enables the config item, similar to kubenurse)


cluster/config-defaults.yaml
```diff
@@ -625,7 +626,9 @@ cpu_manager_policy: "none"
 # sysctl names allowed to be used in security policies, comma-separated
 allowed_unsafe_sysctls: "net.ipv4.tcp_keepalive_time,net.ipv4.tcp_keepalive_intvl,net.ipv4.tcp_keepalive_probes,net.ipv4.tcp_syn_retries,net.ipv4.tcp_retries2"
 
-# enable CSIMigration feature flag
+# enable CSI Driver feature flag
+enable_csi: "true"
```
linki: TODO: turn off by default

gargravarr:
:+1:

gargravarr:
:+1:

linki:
👍

PR #5551, opened 2022-12-06
Allow to turn on CSIMigration and CSIMigrationAWS

linki:
The `CSIMigration` feature flag is `enabled` by default since Kubernetes `1.17` and will be GA in `1.25` when it cannot be turned off (like we do it currently).

This feature is not to be confused with `CSIMigrationAWS` which migrate the in-tree AWS EBS volumes to CSI. This feature gate will be enabled by default in Kubernetes `1.23` and GA in `1.25`.

~~This PR intends to use the default for `CSIMigration` (which is `true`) while actively keeping `CSIMigrationAWS` at `false` which is also the current value.~~

~~Effectively this enables `CSIMigration` which merely migrates in-tree volumes to the in-tree/core CSI handlers that don't require an out-of-tree CSI Driver, hence this should be safe to do.~~

~~The PR could also drop the `CSIMigrationAWS` feature flag entirely instead but keeping it toggleable will help with future CSI related work.~~

This PR intends to add a per-cluster switch to toggle "CSIMigration" for the volumes that actually matter (EBS). Doing it like it's done here will roll the nodes only when the feature is actually enabled. Merely merging this PR won't roll the nodes.


linki:
*blocked* Let's merge https://github.com/zalando-incubator/kubernetes-on-aws/pull/5568 first because this PR rolls all nodes 😢

linki:
We can also combine that with the actual activation of `CSIMigrationAWS` to avoid a double node roll. `CSIMigration`/`CSIMigrationAWS` isn't required to use CSI so we can merge https://github.com/zalando-incubator/kubernetes-on-aws/pull/5547 before to test it.

linki:
👍

gargravarr:
:+1:

PR #5552, opened 2022-12-06
Disable PCS for audittrail-adapter test

mikkeloscar:
Follow up to #5542 

This disables deployment of the PCS used by `audittrail-adapter` when the audittrail-api url is not set. It's only needed in production where audittrail-api URL is configured.

This should fix issues in e2e where the secret for the PCS is not created.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #5553, opened 2022-12-06
audittrail-adapter: Condition volume mount in all places

mikkeloscar:
Follow up to #5552 

Add the condition in all places so the manifest is valid.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #5554, opened 2022-12-06
Avoid using spaces when specifying volume zones

linki:
Using spaces when specifying volume zones will lead to an error when `CSIMigrationAWS` is enabled because it doesn't trim the space from the region name.

```
csi-provisioner E1205 13:43:11.085265       1 controller.go:957] error syncing claim "27ee3ab9-1f4a-490d-bd1c-a17583e4b581": failed to provision volume with StorageClass "standard": rpc error: code = Internal desc = Could not create volume "pvc-27ee3ab9-1f4a-490d-bd1c could not create volume in EC2: InvalidZone.NotFound: The zone ' eu-central-1b' does not exist.
```

While doing this I realised we could also switch to using [`allowedTopologies`](https://kubernetes.io/docs/concepts/storage/storage-classes/#allowed-topologies) because `zones` is [deprecated](https://kubernetes.io/docs/concepts/storage/storage-classes/#aws-ebs).

However, I wasn't able to identify which label to use for it to work as expected. There are the three options:
* `failure-domain.beta.kubernetes.io/zone`
* `topology.kubernetes.io/zone`
* `topology.ebs.csi.aws.com/zone`

Using the old approach allows us to defer that decision to whatever upstream does.

### DIFF
--- a/cluster/manifests/storageclass/storageclass.yaml
+++ b/cluster/manifests/storageclass/storageclass.yaml
@@ -11,5 +11,5 @@ provisioner: kubernetes.io/aws-ebs
 parameters:
   type: {{ if eq .Cluster.ConfigItems.storageclass_standard_gp3 "true" }}gp3{{else}}gp2{{end}}
   # TODO: this assumes 3 zones per region
-  zones: {{ .Cluster.Region }}a, {{ .Cluster.Region }}b, {{ .Cluster.Region }}c
+  zones: {{ .Cluster.Region }}a,{{ .Cluster.Region }}b,{{ .Cluster.Region }}c
 allowVolumeExpansion: true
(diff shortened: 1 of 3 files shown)


cluster/manifests/storageclass/storageclass.yaml
```diff
@@ -7,9 +7,16 @@ metadata:
   labels:
     kubernetes.io/cluster-service: "true"
     volume-type: {{ if eq .Cluster.ConfigItems.storageclass_standard_gp3 "true" }}gp3{{else}}gp2{{end}}
+    zones-sanitized: "{{ if eq .Cluster.ConfigItems.storageclass_standard_sanitized_zones "true" }}true{{else}}false{{end}}"
```
mikkeloscar, line 10: You would need to roll this change out first in a separate PR, otherwise the deletion would not match as no storage class would have the label.

linki: Thanks, I realised that too. But instead of a different PR I can just wait with flipping the config item to `"true"`.

Given that initially `storageclass_standard_sanitized_zones: "false"` merging this PR labels the SC with `zones-sanitized: "false"` and doesn't delete it as well. Flipping the config item will then delete the old SC and create the new one with `zones-sanitized: "true"` so it won't be deleted again.

linki:
👍

gargravarr:
:+1:

PR #5555, opened 2022-12-06
Correctly clean up previous CSI resources when it's disabled

linki:
The current implementation of CSI is guarded by a feature flag. However the feature flag isn't covering the RBAC resources and they and the CSIDriver are also not part of deletions. Hence, the resources are present in all clusters although CSI isn't used. This makes problems when rolling out [the revamp](https://github.com/zalando-incubator/kubernetes-on-aws/pull/5547) because some of them are immutable.


linki:
👍

demonCoder95:
👍

PR #5558, opened 2022-12-07
Make cpu/memory configurable for audittrail/kube2iam ds

mikkeloscar:
Makes cpu/memory configurable per cluster for `audittrail-adapter` and `kube2iam` daemonsets.

This enables reducing the resource reservation in low usage clusters like playground.


gargravarr:
:+1:

mikkeloscar:
:+1:

PR #5563, opened 2022-12-07
Disable old Ingress API by default

mikkeloscar:
This disables the old Ingress API versions by default. `allow_v1beta1_ingresses` is set to `true` in the clusters which still have not fully migrated away from the old API versions.

Once this is rolled out it will start to block deployments of ingresses still defining the old api versions, but the teams have automated PRs with the needed update so they can merge those to fix the deployments.


demonCoder95:
👍

gargravarr:
:+1:

PR #5565, opened 2022-12-07
log: change default access logs

szuecs:
log: change default access logs: disable 404 which is often the case by scanners and just produce log volume and 429 which is covered by metrics.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -86,7 +86,7 @@ skipper_suppress_route_update_logs: "true"
 skipper_validate_query: "true"
 skipper_validate_query_log: "false"
 
-skipper_default_filters: 'enableAccessLog(4,5) -> fifo(2000,20000,"3s")'
+skipper_default_filters: 'disableAccessLog(2,3,404,429) -> fifo(2000,20000,"3s")'
 skipper_disabled_filters: ""
 skipper_edit_route_placeholders: ""
 skipper_ingress_inline_routes: ""


cluster/config-defaults.yaml
```diff
@@ -86,7 +86,7 @@ skipper_suppress_route_update_logs: "true"
 skipper_validate_query: "true"
 skipper_validate_query_log: "false"
 
-skipper_default_filters: 'enableAccessLog(4,5) -> fifo(2000,20000,"3s")'
+skipper_default_filters: 'disableAccessLog(2,3,404,429) -> fifo(2000,20000,"3s")'
```
AlexanderYastrebov, line 89: Do we log 1xx by this config?

gargravarr:
:+1:

cluster/config-defaults.yaml
AlexanderYastrebov: This would disable all 404s but we may want to disable only those caused by route lookup failures.

szuecs: yes but 1xx is not very often and would be even interesting to see IMO.

AlexanderYastrebov:
:+1:

mikkeloscar:
@szuecs please describe in the PR description the motivation for this. It seems reasonable but it would be good with a description of why we choose to disable some 4xx type errors.

szuecs:
@mikkeloscar done

PR #5570, opened 2022-12-07
Deploy CSI-related components by default

linki:
Flips the switch and turns on CSI by default. Without `enable_csi_migration: true` as well it would be entirely opt-in for users by using a different StorageClass.

- [x] Based on https://github.com/zalando-incubator/kubernetes-on-aws/pull/5547 and must be merged after.

Precondition for enabling `CSIMigrationAWS` (i.e. `enable_csi_migration: true`) which requires a running CSI Driver (enabled by default in this PR).

Checklist:
- [x] This adds a DaemonSet to all clusters, so we need to increase DaemonSet reservations


linki:
👍

demonCoder95:
👍

PR #5602, opened 2022-12-16
external-dns: try experimental version that doesn't fail entire batch

linki:
This is https://github.com/kubernetes-sigs/external-dns/pull/1209 rebased on `v0.13.1` aka `v0.13.1-internal-*`.

Let's see if this version passes all tests.

If not:
* make sure that vanilla `v0.13.1` even passes (we used v0.12.2 so far)

If yes:
* add an e2e test containing a broken DNS record that would normally break everything, then check if this PR fixes it.


linki:
This can be closed in favour of https://github.com/zalando-incubator/kubernetes-on-aws/pull/5604 once it's green.

linki:
It's green.

PR #5604, opened 2022-12-19
Introduce a problematic DNS and ExternalDNS version with the fix

linki:
Let's add a Service of `Type: LoadBalancer` giving it a DNS name that ExternalDNS has problems with (as of `v0.13.1`, it might be fixed in later versions)

The idea is that DNS should be pretty broken when running e2e tests. Then we can see if introducing https://github.com/zalando-incubator/kubernetes-on-aws/pull/5602 makes sure to ignore the bad record.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -683,7 +683,7 @@ external_dns_excluded_domains: cluster.local
 external_dns_policy: sync
 
 # eternal-dns version for controlling roll-out, can be "current" or "legacy"
-# current => v0.13.1-internal-master-32
+# current => v0.13.2-12-ga18bf2b5-internal-master-34
 # legacy => v0.9.0-master-26
 external_dns_version: "current"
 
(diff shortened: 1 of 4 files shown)


linki:
This successfully blocks all ExternalDNS updates 💥

linki:
Now trying the same with an ExternalDNS version that should tolerate some broken records. This is hopefully green.

linki:
So this is green now. It means a single broken record doesn't block ExternalDNS and hence it passes all e2e tests. We should try to get https://github.com/kubernetes-sigs/external-dns/pull/1209 merged or run a fork.

linki:
As discussed let's go ahead with our internally released version from https://github.com/kubernetes-sigs/external-dns/commit/a18bf2b51258302403b2b5f081f236f8fb73e433 which includes https://github.com/kubernetes-sigs/external-dns/pull/1209. Dropping the `do-not-merge` label.

katyanna:
:+1:

linki:
👍

PR #5617, opened 2022-12-20
skipper: removes credentials path

AlexanderYastrebov:
`-credentials-path` is only used by `bearerinjector` filter that is no used in the ingress setting.


szuecs:
:+1:

gargravarr:
:+1:

PR #5650, opened 2023-01-05
Improve audit-policy

mikkeloscar:
Improves the audit-policy to log ALL read events if `auditlog_read_access=true`. 

We previously dropped events from service account users in kube-system, but this is now enabled if `auditlog_read_access=true`.

Also drops an old config supressing the `credentials-provider` identity which is no longer used (The user is called something like: `zalando-iam:zalando:service:k8sapi_credentials-provider`).


myaser:
👍🏻

myaser:
👍🏻

mikkeloscar:
:+1:

PR #5656, opened 2023-01-09
Update to cadvisor v0.47.0

mikkeloscar:
https://github.com/google/cadvisor/releases/tag/v0.47.0

In the past we ran v0.44.0 but downgraded to v0.38.0 to support ARM. Now we can upgrade to the latest version also running on ARM.


mikkeloscar:
:+1:

linki:
👍

PR #5657, opened 2023-01-09
Update e2e gomod to silence dependabot alert

linki:
There's a warning about using `require k8s.io/kubernetes` less than `v1.16.11`. Let's bump it from `v0.0.0` to our current version and see if it goes away.

Note the previous typos: `v0.21.14` doesn't event exist and the closest version `v0.21.4` is [more than seven years old](https://github.com/kubernetes/kubernetes/releases/v0.21.4) 🤣  These lines probably don't do anything.

### DIFF
--- a/test/e2e/go.mod
+++ b/test/e2e/go.mod
@@ -16,11 +16,11 @@ require (
 	github.com/zalando-incubator/kube-aws-iam-controller v0.1.2
 	gopkg.in/gcfg.v1 v1.2.3 // indirect
 	gopkg.in/warnings.v0 v0.1.2 // indirect
-	k8s.io/api v0.21.14
-	k8s.io/apimachinery v0.21.14
-	k8s.io/apiserver v0.0.0
-	k8s.io/client-go v0.21.14
-	k8s.io/kubernetes v0.0.0
+	k8s.io/api v1.21.14
+	k8s.io/apimachinery v1.21.14
+	k8s.io/apiserver v1.21.14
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

mikkeloscar:
> Note the previous typos: v0.21.14 doesn't event exist and the closest version v0.21.4 is [more than seven years old](https://github.com/kubernetes/kubernetes/releases/v0.21.4) rofl These lines probably don't do anything.

I don't think it's a typo. The separate modules like client-go etc. uses the version without v1 to fix some Go modules issue iirc. You can see the version tagged here: https://github.com/kubernetes/client-go/releases/tag/v0.21.14.

So I think only the `k8s.io/kubernetes` package needs to be updated from `v0.0.0` to `v1.21.14` and the others are correct.

linki:
Oh boy, thanks 😵‍💫  

I updated it such that it uses the `v1.*` tag if present, otherwise the `v0.*`. It then becomes:
```diff
        k8s.io/api v0.21.14
        k8s.io/apimachinery v0.21.14
-       k8s.io/apiserver v0.0.0
+       k8s.io/apiserver v0.21.14
        k8s.io/client-go v0.21.14
-       k8s.io/kubernetes v0.0.0
+       k8s.io/kubernetes v1.21.14
```

mikkeloscar:
:+1:

linki:
👍

PR #5678, opened 2023-01-12
Drop ZMON based scaling for skipper-ingress

mikkeloscar:
This feature is no longer used and thus we can remove the configuration.

## TODO

* [x] Remove usage from the last cluster.


mikkeloscar:
:+1:

zaklawrencea:
👍

PR #5705, opened 2023-01-24
reduce kubenurse collected data

myaser:
teapot/issues#1950

### DIFF
--- a/cluster/manifests/kubenurse/daemonset.yaml
+++ b/cluster/manifests/kubenurse/daemonset.yaml
@@ -53,7 +53,9 @@ spec:
               value: "{{.Cluster.ConfigItems.network_monitoring_check_kubenurse_ingress}}"
             - name: KUBENURSE_CHECK_NEIGHBOURHOOD
               value: "{{.Cluster.ConfigItems.network_monitoring_check_neighborhood}}"
-          image: "container-registry.zalando.net/teapot/kubenurse:v1.7.0-histogram-main-5"
+            - name: KUBENURSE_HISTOGRAM_BUCKETS
+              value: 0.005,0.05,0.5,1,5,10
+          image: "container-registry.zalando.net/teapot/kubenurse:v1.7.0-histogram-main-7"
           resources:
             requests:
               cpu: "{{.Cluster.ConfigItems.network_monitoring_daemonset_cpu}}"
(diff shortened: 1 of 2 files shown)


cluster/manifests/kubenurse/daemonset.yaml
```diff
@@ -53,7 +53,9 @@ spec:
               value: "{{.Cluster.ConfigItems.network_monitoring_check_kubenurse_ingress}}"
             - name: KUBENURSE_CHECK_NEIGHBOURHOOD
               value: "{{.Cluster.ConfigItems.network_monitoring_check_neighborhood}}"
-          image: "container-registry.zalando.net/teapot/kubenurse:v1.7.0-histogram-main-5"
+            - name: KUBENURSE_HISTOGRAM_BUCKETS
+              value: 0.005,0.05,0.5,1,5,10
```
demonCoder95, line 57: Can you explain the meaning of this variable? It appears to be the number of histogram buckets, from the way it's named now, but then how can it have a value of 0.005? Is it instead a time period, based on which we sample data in a bucket?

demonCoder95 (commented):
Just one comment. Rest looks good.

cluster/manifests/kubenurse/daemonset.yaml
myaser: when we measure kubenurse latency, Prometheus increament the counter of the bucket that is higher than the reading
lets assume the measured latency 4 milliseconds (0.004) then Prometheus will increment the value in the 0.005 bucket and every bucket higher than that

if the latency measured was 6 seconds, only the 10, +inf buckets will be incremented

the default prometheus buckets are `.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10` which is too granular and produce too much data (every bucket is a separate counter)

so came the need to reduce the number/size of buckets a bit

myaser: this variable is used for request latency, and the unit is seconds

so it is `0.005,0.05,0.5,1,5,10` 5ms, 50ms, 500ms, 1s, 5s, 10s

gargravarr:
:+1:

mikkeloscar:
:+1:

cluster/manifests/kubenurse/daemonset.yaml
demonCoder95: Ah, I see 🙂 makes perfect sense. Thanks for explaining!

PR #5722, opened 2023-01-31
Set apiserver memory limit based on instance-type

mikkeloscar:
This introduces a way to set the apiserver memory with a limit based on instance-type size such that we don't run apiserver without memory limits.

The motivation for this is that we have multiple times observed how an overrun apiserver instance becomes completely unresponsive. The theory is that if we have a memory limit set on the container then it will kill the process before it can use up all the memory of the instance and prevent it from dying completely.

This is set to 80% of the instance memory by default based on testing and the understanding that the kube-apiserver is the main process on the node using memory.
It can be disabled per cluster via the config-item: `apiserver_memory_limit_percent=0`


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #5383, opened 2022-09-21
Allow rolling back external-dns version

mikkeloscar:
We have observed issues related to the new version where we get the following errors in some clusters:

```
InvalidChangeBatch: FATAL problem: DomainLabelTooLong (Domain label is too long) encountered with 'cname
-cdp-e2e-zlawrence-f67f5f88d914191c18848b9b54e9957315a5a36c'\n\tstatus code: 400, request id: 04ab213b-74d3-4567-bebf-7bfb49b72645"
```

Allow rolling back via the config-item: `external_dns_version=legacy`

ref: #5354

PR #5735, opened 2023-02-07
Drop legacy external-dns version

mikkeloscar:
This reverts #5383 as there is no longer a need to run the legacy version of external-dns in any clusters.


mikkeloscar:
:+1:

linki:
👍

PR #5740, opened 2023-02-07
Update emergency access service to the latest version

linki:
Includes some code to retry calls to teams api.


linki:
Problem is not yet fixed in this version.

PR #5763, opened 2023-02-14
Update external-dns to multi-arch image

linki:
Update external-dns to multi-arch image in preparation for all-arm64 clusters.


linki:
👍

katyanna:
👍

PR #5764, opened 2023-02-15
Clean up unused configuration value `oidc-subject-prefix`

linki:
This isn't need anymore, replaced by `oidc-subject-key`

It was introduced for a short time but changed shortly after for `oidc-subject-key`. We left it for backwards compatibility but no one is using it anymore.


linki:
👍

demonCoder95:
👍

PR #5775, opened 2023-02-20
Update dependencies in various components

mikkeloscar:
Updates dependencies in various components.

Drops `version` label which we anyway forget to keep up to date. The version that matters is on the image itself.


mikkeloscar:
:+1:

katyanna:
👍

PR #5786, opened 2023-02-23
Build multi-arch e2e image

mikkeloscar:
As a follow up to https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/645 build and use a multi-arch image for running e2e.

This sets up everything for building a multi-arch image, but I limited it only to amd64 for now because otherwise it doubles the build time which we do for every PR. IMO we can merge as is and later enable arm64 if it makes sense.

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -32,9 +32,8 @@ pipeline:
         exit 1
       fi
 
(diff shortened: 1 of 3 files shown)


linki:
👍

mikkeloscar:
:+1:

szuecs:
:+1:

PR #5792, opened 2023-02-27
Update CDPDeploymentTask with source information fields

linki:
In preparation of rolling out source information we need to update the CRD. Tested backwards compatibility by rolling out cluster-registry to stups test while using the new CRD.


linki:
👍

katyanna:
👍

PR #5793, opened 2023-02-27
Update deployment-service components to log source information

linki:
Use the additional fields in the CRD to annotate resources with source file information.


linki:
👍

gargravarr:
:+1:

PR #5799, opened 2023-02-28
Remove kuberuntu v1.21 image

katyanna:
With the Kuberentes update to v1.22, the v1.21 image is no longer needed, so it's here removed from the cluster config.


mikkeloscar:
:+1:

katyanna:
👍

PR #5829, opened 2023-03-14
Vertically scale EBS CSI controller's memory

linki:
In two clusters CSI controller reached its maximum limit for one sidecar (provisioner). Let's use a VPA and a config item to be able to configure proper values quickly.

I left out the containers for the plugin-driver and the liveness probe in order to keep them consistent with the same values in the daemonset part. The issues we saw was only with provisioner so far, so this PR only deals with the "active" parts: provisioner, attacher, resizer and bumps it by 2x to 80Mi by default.


linki:
👍

gargravarr:
:+1:

PR #5834, opened 2023-03-15
Disable arm build for e2e image

mikkeloscar:
#5284 introduced a setup for multi-arch build but limited the container image build for now to amd64 as it takes much longer to build for both architectures.

However, it didn't actually disable the Go build of arm64 so we built for BOTH archs but only create an image for amd64. This disables the arm64 build to speed it up.


linki:
👍

mikkeloscar:
:+1:

PR #5844, opened 2023-03-17
Add support for PostgresAdmin

mikkeloscar:
This introduces support for a special `PostgresAdmin` role which in the `k8s-authnz-webhook` is configured to be allowed to `exec` to pods with `application: spilo` label and every other role is prevented from doing it.


mikkeloscar:
:+1:

katyanna:
👍

PR #5846, opened 2023-03-17
Drop TTLAfterFinished feature gate which is default now

mikkeloscar:
Drop `TTLAfterFinished` feature gate as it's set by default with Kubernetes v1.23

From [CHANGELOG](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md#api-change-6):

The `TTLAfterFinished` feature gate is now GA and enabled by default. (https://github.com/kubernetes/kubernetes/pull/105219, [@sahilvv](https://github.com/sahilvv))


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #5851, opened 2023-03-21
Remove cronjob-monitor

mikkeloscar:
This removes the `cronjob-monitor` component which is no longer needed since Kubernetes v1.22.

This component was setting the annotation `experimental.zalando.org/last-successful-completion-time` on successful cronjobs which was used in ZMON monitoring. Since Kubernetes v1.22 the same data is natively available in the `status.lastSuccessfulTime` field and ZMON is already relying on that.


zaklawrencea:
..removed to wait for e2e

linki:
👍

mikkeloscar:
:+1:

PR #5854, opened 2023-03-22
Add audittrail adapter timeout config item and reduce timeout value

zaklawrencea:
Add a config item for `audittrail-timeout` for the `audittrail-adapter`.

Note: the default config item value is set at 2s, which also lowers the default value from the current 5s.

---

After observing the `audittrail-adpater` being OOM killed when the `audittrail-api` is unavailable/slow, this reduction in timeout will allow the `audittrail-adapter` to fallback faster to the secondary approach of first sending events to s3 -- which will then be resent to the api once it is available. The goal of this is to reduce the load on the adapter *faster* which in turn will reduce the memory build up.

Teapot issue: 3406

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -576,6 +576,8 @@ audittrail_root_account_role: ""
 audittrail_adapter_cpu: "50m"
 audittrail_adapter_memory: "200Mi"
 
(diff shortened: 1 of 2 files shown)


zaklawrencea:
:+1:

mikkeloscar:
:+1:

PR #5857, opened 2023-03-22
Run cronjob-monitor cleanup in post-apply

mikkeloscar:
Trying to delete a `VerticalPodAutoscaler` resource in `pre_apply` breaks e2e because the `VerticalPodAutoscaler` resource type will not be available the very first time a cluster is set up. Therefore move it to `post_apply` to avoid this problem.

Follow up to #5851


mikkeloscar:
:+1:

demonCoder95:
👍

PR #5859, opened 2023-03-22
Update deployment-service to support special characters in source info

linki:
This propagates tags and annotations using sanitized values so that CloudFormation stacks can still be applied.


linki:
👍

mikkeloscar:
:+1:

PR #5869, opened 2023-03-24
Update AMI to containerd (Kube 1.22)

demonCoder95:
This updates the AMI to the latest version that has been tested to work. This is to test the new AMI against the e2e testing we have in this repository.

More details on [this PR](https://github.bus.zalan.do/teapot/kubernetes-on-ubuntu/pull/169) in `kubernetes-on-ubuntu` repo.


demonCoder95:
Containerd being a lower level runtime [doesn't support](https://github.com/containerd/containerd/issues/3351) `log-drivers` that Docker had. The log rotation is responsibility of the orchestrator, in our case the kubelet. 

The kubelet configuration schema [here](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/) provides the parameters for log rotation as `containerLogMaxSize` and `containerLogMaxFiles` parameters that we can set with same values as we had in `daemon.json` in our [Docker configuration](https://github.bus.zalan.do/teapot/kubernetes-on-ubuntu/blob/master/runtime/etc/docker/daemon.json#L10) in the AMI.

Adding a commit to configure these parameters in the kubelet configuration.

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -21,6 +21,8 @@ write_files:
       apiVersion: kubelet.config.k8s.io/v1beta1
       kind: KubeletConfiguration
       cgroupDriver: systemd
+      containerLogMaxSize: "50Mi"
+      conatinerLogMaxFiles: 2
```
linki: typo: conatiner -> container

demonCoder95: Oh God. Thanks for catching this 🤣

demonCoder95:
I just realised I've merged `kube-1.23` branch on an AMI patch for `1.22` version 😁 

I think since we're rolling out `1.23` and will roll it out to production next week, there's no need to have this `1.22` based PR here. I'll close this PR and will continue to work on the [1.23 PR we have](https://github.com/zalando-incubator/kubernetes-on-aws/pull/5871) and get that to 🟢  going forward.

PR #5872, opened 2023-03-24
Run load test after cluster has been fully updated

linki:
The current location of the load test script runs _after_ the initial cluster has been provisioned but _before_ the cluster that's actually being tested in the PR.

For example, let's say a PR introduces a performance regression, we would not notice that as part of the load test because what's being load tested is the previous cluster (before introducing the regression).

I believe the execution of the load test should come later.

@szuecs Can you check if that's still the behaviour that you expect?

### DIFF
--- a/test/e2e/run_e2e.sh
+++ b/test/e2e/run_e2e.sh
@@ -119,10 +119,6 @@ if [ "$create_cluster" = true ]; then
 
         # Wait for the resources to be ready
         ./wait-for-update.py --timeout 1200
-
-        # provision and start load test
-        echo "provision and start load test"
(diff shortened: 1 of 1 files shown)


szuecs:
@linki the purpose of the load test is not see perf regressions, but disconnects in case kube-ingress-aws-controller would mess up TG members because of some weird CF stack change. So the behavior what we want is to start the load test when cluster is ready and before update apply, until updates were done + a couple of minutes, such that kube-ingress-aws-controller changes would also update CF stacks and create the impact in case we did something wrong.

szuecs:
:-1:

linki:
@szuecs Thanks for the explanation. That makes sense 👍  I got confused by the name. I'm closing it.

PR #5879, opened 2023-03-28
Improve GPU e2e test by not restarting the pod

mikkeloscar:
The GPU e2e tests essentially works by running a single pod with the image `k8s.gcr.io/cuda-vector-add:v0.1` and checking that the container exists with status code 0 which indicates it successfully tested GPU capabilities on the node.

Because pods by default have the [`restartPolicy: Always`](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy), exiting with 0 means that the pod is automatically restarted again and again leading to the pod getting into a `CrashLoopBackOff` state.

I believe this can make the test flaky because it tries to read the logs after the pod has exited with 0 and if it's contentiously restarted the logs will change and the check could be wrong.

Avoid this by setting `restartPolicy: Never`


mikkeloscar:
:+1:

demonCoder95:
👍

PR #5884, opened 2023-03-31
feature: datadome poc preparation

szuecs:
configmap deploy via internal

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -94,7 +94,7 @@ spec:
         - name: HTTP_REDIRECT
           value: "true"
 {{ end }}
-{{ if eq .ConfigItems.datadome_enabled "true" }}
+{{ if eq .ConfigItems.skipper_lua_scripts_enabled "true" }}
         - name: LUA_PATH
           value: /etc/skipper/lua/?.lua
         - name: DATADOME_API_KEY
@@ -315,7 +315,7 @@ spec:
           - name: filters
             mountPath: /etc/config/default-filters
 {{ end }}
-{{ if eq .ConfigItems.datadome_enabled "true" }}
+{{ if eq .ConfigItems.skipper_lua_scripts_enabled "true" }}
           - mountPath: /etc/skipper/lua
             name: lua
             readOnly: true
@@ -342,7 +342,7 @@ spec:
             name: skipper-default-filters
             optional: true
 {{ end }}
-{{ if eq .ConfigItems.datadome_enabled "true" }}
+{{ if eq .ConfigItems.skipper_lua_scripts_enabled "true" }}
         - name: lua
           configMap:
             defaultMode: 420
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -94,6 +94,12 @@ spec:
         - name: HTTP_REDIRECT
           value: "true"
 {{ end }}
+{{ if eq .ConfigItems.datadome_enabled "true" }}
+        - name: LUA_PATH
+          value: /etc/skipper/lua/?.lua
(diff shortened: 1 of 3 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -309,6 +315,11 @@ spec:
           - name: filters
             mountPath: /etc/config/default-filters
 {{ end }}
+{{ if eq .ConfigItems.datadome_enabled "true" }}
+          - mountPath: /etc/skipper/lua
+            name: lua
```
AlexanderYastrebov: Lets use the same field order as for others:
```yaml
- name: skipper-ingress-lua
  mountPath: /etc/skipper/lua
  readOnly: true
```

cluster/manifests/skipper/deployment.yaml
```diff
@@ -93,6 +93,12 @@ spec:
 {{ if or (eq .ConfigItems.nlb_switch "pre") (eq .ConfigItems.nlb_switch "exec") }}
         - name: HTTP_REDIRECT
           value: "true"
+{{ end }}
+{{ if eq .ConfigItems.skipper_lua_scripts_enabled "true" }}
+        - name: LUA_PATH
+          value: /etc/skipper/lua/?.lua
+        - name: DATADOME_API_KEY
+          value: "{{ .ConfigItems.datadome_api_key }}"
```
AlexanderYastrebov: This should have its own `if`

szuecs: why? no

cluster/manifests/skipper/deployment.yaml
szuecs: boah you are seriously too picky

szuecs: The names above and below are `filters` and `hostname-credentials` so it does not follow your comment

cluster/manifests/skipper/deployment.yaml
AlexanderYastrebov: You may have lua scripts without datadome_api_key

cluster/manifests/skipper/deployment.yaml
```diff
@@ -331,6 +342,12 @@ spec:
             name: skipper-default-filters
             optional: true
 {{ end }}
+{{ if eq .ConfigItems.skipper_lua_scripts_enabled "true" }}
+        - name: lua
+          configMap:
+            name: skipper-ingress
+            optional: true
```
AlexanderYastrebov, line 352: Do we need optional here?

cluster/manifests/skipper/secret.yaml
```diff
@@ -7,5 +7,6 @@ metadata:
     application: skipper-ingress
 type: Opaque
 data:
+  datadome-token: "{{ .Cluster.ConfigItems.datadome_api_key | base64 }}"
```
AlexanderYastrebov: Why token, maybe `datadome-api-key`?

cluster/manifests/skipper/deployment.yaml
szuecs: `skipper-default-filters` has also `optional: true` so I stayed with that option. I guess it's better to have no lua but another instance than no new instance in case we need to scale-out and something happened why ever the configmap could not be mounted into the container.

cluster/manifests/skipper/secret.yaml
szuecs: changed as suggested.

AlexanderYastrebov:
:+1:

cluster/manifests/skipper/deployment.yaml
szuecs: but then an empty string datadome api key, does not matter, too

szuecs:
:+1:

katyanna:
👍

PR #5885, opened 2023-04-03
Enable CSI Migration which handles current PVCs through CSI

linki:
Switch the flip to enable CSI migration (default in 1.23). This will process all current and future EBS PVCs via the CSI driver. There's no need for users to change anything.


linki:
Closing, was just to run e2e at the moment.

PR #5889, opened 2023-04-03
skipper: add controller label to hostname credentials

AlexanderYastrebov:
Label PlatformCredentialsSets created from config item with a controller label to enable cooperation with another controller that would create credentials by processing Ingresses and RouteGroups.


mikkeloscar:
:+1:

AlexanderYastrebov:
Will update along with the addition of the hostname credentials controller

PR #5892, opened 2023-04-03
update AMI to move snapshot generation and drop compression

linki:
This should speed up boot time for instance storage disks.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #5900, opened 2023-04-05
Remove duplicate preprocessors

MustafaSaber:
* Move `-reverse-source-predicate` from skipper-ingress to RouteSRV in `exec` mode because RouteSRV will have the k8s dataclient at this point.

* Remove `apiUsageMonitoring` flags from RouteSRV since skipper-ingress will be responsible about validating/processing `defaultFIltersDir`

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -126,6 +126,7 @@ spec:
 {{ if ne .ConfigItems.skipper_routesrv_enabled "exec" }}
           - "-kubernetes-east-west-range-domains=ingress.cluster.local"
           - "-kubernetes-east-west-range-predicates=ClientIP(\"10.2.0.0/15\", \"{{ .Values.vpc_ipv4_cidr }}\")"
+          - "-reverse-source-predicate"
 {{ end }}
           - "-proxy-preserve-host"
           - "-compress-encodings={{ .ConfigItems.skipper_compress_encodings }}"
@@ -142,7 +143,6 @@ spec:
           - "-enable-ratelimits"
           - "-experimental-upgrade"
           - "-metrics-exp-decay-sample"
-          - "-reverse-source-predicate"
           - "-lb-healthcheck-interval=3s"
           - "-metrics-flavour=prometheus"
           - "-enable-connection-metrics"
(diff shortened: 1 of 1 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -447,10 +447,6 @@ spec:
           - "-kubernetes-east-west-range-domains=ingress.cluster.local"
           - "-kubernetes-east-west-range-predicates=ClientIP(\"10.2.0.0/15\", \"{{ .Values.vpc_ipv4_cidr }}\")"
           - "-reverse-source-predicate"
-          - "-enable-api-usage-monitoring"
-          - "-api-usage-monitoring-realm-keys=https://identity.zalando.com/realm"
-          - "-api-usage-monitoring-client-keys=https://identity.zalando.com/managed-id,sub"
-          - "-api-usage-monitoring-default-client-tracking-pattern=services[.].*"
```
AlexanderYastrebov, line 453: Why do we remove this?

MustafaSaber: I don't see it used in routesrv

AlexanderYastrebov: It should be consistent with -default-filters-dir

AlexanderYastrebov: I guess we need to enable it in skipper unconditionally then and remove `-default-filters-dir` and related stuff (configmap mount) for routesrv

MustafaSaber: `-default-filters-dir` is used in routesrv options but api usage isn't

MustafaSaber: but why? DefaultFilters will fill k8s data client and skipper will validate

szuecs: <s>Default filters should stay in routesrv and @MustafaSaber do not drop the options as it feeds into the same tool apimonitoring, which configures the apiMonitoring filters. We want to drop these from skipper-ingress and put it into routesrv only.</s>
@MustafaSaber is right. This is the filter configuration which needs to be passed to skipper-ingress. 
What I am not sure is if we can drop this in routesrv or if we loose the config.

AlexanderYastrebov: Ok, apiUsageMonitoring and defaultFIltersDir are related but I think it will work like you propose

AlexanderYastrebov:
:+1:

cluster/manifests/skipper/deployment.yaml
MustafaSaber: @szuecs options for apiMonitoring weren't in place the flag is provided in config but isn't collected from routesrv

szuecs: I would guess that we need this in both to make it work (so agree with @AlexanderYastrebov ) . The default-filters-dir is likely the one we can drop from skipper-ingress.
Routesrv likely requires the configuration, too, because it parses the kubernetes resources and creates the routes and then outputs eskip.
Skipper-ingress should only pick up the eskip with the same configuration to create the correct filter again.

szuecs:
:+1:

PR #5906, opened 2023-04-06
Enable StatefulSetAutoDeletePVC feature gate

katyanna:
[**`StatefulSetAutoDeletePVC`**](https://kubernetes.io/blog/2021/12/16/kubernetes-1-23-statefulset-pvc-auto-deletion/) is a feature gate made available on Kubernetes v1.23
that allows the usage of field `.spec.persistentVolumeClaimRetentionPolicy` to
manage if and how PVCs are deleted during the lifecycle of a StatefulSet.

This commit enables this feature by default.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -844,6 +844,10 @@ hyped_article_lifecycle_management: "false"
 # enable SizeMemoryBackedVolumes feature flag
 enable_size_memory_backed_volumes: "true"
 
+# enable StatefulSetAutoDeletePVC feature flag
+# https://kubernetes.io/blog/2021/12/16/kubernetes-1-23-statefulset-pvc-auto-deletion/
+enable_statefulset_autodelete_pvc: "true"
+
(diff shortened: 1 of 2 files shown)


cluster/config-defaults.yaml
```diff
@@ -844,6 +844,10 @@ hyped_article_lifecycle_management: "false"
 # enable SizeMemoryBackedVolumes feature flag
 enable_size_memory_backed_volumes: "true"
 
+# enable StatefulSetAutoDeletePVC feature flag
+# https://kubernetes.io/blog/2021/12/16/kubernetes-1-23-statefulset-pvc-auto-deletion/
+enable_statefulset_autodelete_pvc: "false"
```
mikkeloscar: If it works in our setup, let's enabled it by default with 1.23. Users would still have to opt-in to this behavior.

katyanna:
:+1:

mikkeloscar:
:+1:

PR #5917, opened 2023-04-12
Add config-item for disabling HPA field via flag

katyanna:
To split the rollout of the deprecation of the `horizontalPodAutoscaler` field on Stackset definitions for test and production clusters, the Admission Controller will validate Stacksets based on the `stackset_hpa_field_enabled` config-item.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -757,7 +757,7 @@ spot_allocation_strategy: "capacity-optimized"
 stackset_controller_sync_interval: "10s"
 stackset_controller_mem_min: "120Mi"
 stackset_controller_mem_max: "4Gi"
-stackset_hpa_field_enabled: "true"
+stackset_legacy_hpa_field_enabled: "false"
 
 # EBS settings for the root volume
 ebs_master_root_volume_size: "50"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -757,6 +757,7 @@ spot_allocation_strategy: "capacity-optimized"
 stackset_controller_sync_interval: "10s"
 stackset_controller_mem_min: "120Mi"
 stackset_controller_mem_max: "4Gi"
+stackset_hpa_field_enabled: "true"
 
 # EBS settings for the root volume
 ebs_master_root_volume_size: "50"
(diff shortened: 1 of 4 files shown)


cluster/manifests/01-admission-control/config.yaml
```diff
@@ -35,6 +35,7 @@ data:
   pod.env-inject.variable._PLATFORM_OBSERVABILITY_METRICS_PORT: "{{ .Cluster.ConfigItems.observability_metrics_port }}"
   pod.env-inject.variable._PLATFORM_OBSERVABILITY_ACCESS_TOKEN: "{{ .Cluster.ConfigItems.lightstep_token }}"
   pod.env-inject.variable._PLATFORM_OBSERVABILITY_COMMON_ATTRIBUTE_CLOUD__ACCOUNT__ID : "{{ .Cluster.Alias }}"
+  pod.env-inject.variable._STACKSET_HPA_FIELD_ENABLED: "{{ .Cluster.ConfigItems.stackset_hpa_field_enabled }}"
```
mikkeloscar: These are the environment variables that will be injected in pods. It should not be here, but rather a new config option similiar to e.g. `pod.parent-resource-hash.enable` above.

katyanna: @mikkeloscar Is `stackset.default.enable_hpa_field` ok? Or like `pod.stackset-def.enable_hpa_field`?

mikkeloscar: Maybe `stackset.legacy_hpa_field.enable` to follow the naming off the other fields.

katyanna:
:+1:

linki:
👍

cluster/config-defaults.yaml
```diff
@@ -757,6 +757,7 @@ spot_allocation_strategy: "capacity-optimized"
 stackset_controller_sync_interval: "10s"
 stackset_controller_mem_min: "120Mi"
 stackset_controller_mem_max: "4Gi"
+stackset_hpa_field_enabled: "true"
```
mikkeloscar: It could be a good idea to set it `false` by default and then set it `true` on _all_ clusters. Then deleting the config item per cluster is the same as disabling the feature. Benefit is that it's disabled by default in new clusters and we don't need to flip the default switch later.

Maybe also call this config item `stackset_legacy_hpa_field_enabled` to be consistent with the naming in the admission-controller config.

katyanna:
The config-item has been created set to `true` in all clusters.

mikkeloscar:
:+1:

katyanna:
👍

PR #5928, opened 2023-04-17
Address e2e test flakyness for complient image job tests

mikkeloscar:
This aims to address the flakyness of the tests:

* **Should create Job using non-compliant image [Image-Policy] [Non-Compliant] [Zalando]**
* **Should create Job with compliant image [Image-Policy] [Compliant] [Zalando]**

This is done by removing the step to wait for job pods to be running and ready as this step can be flaky if a pod finishes in between the check interval and thus is never detected to be running.
Instead we check for Job Completion which indirectly validates that the job container must have successfully run.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #5932, opened 2023-04-18
Remove flannel-awaiter and config to userdata

demonCoder95:
Removes the `flannel-awaiter` and moves the configuration from the `config-map` to `userdata`. This is done as this hack is no longer needed and the same job is doing by `kube-node-ready-controller`.

Relates-to: https://github.bus.zalan.do/teapot/kubernetes-on-ubuntu/pull/169

More context here: https://github.bus.zalan.do/teapot/kubernetes-on-ubuntu/pull/277


cluster/manifests/flannel/configmap.yaml
```diff
@@ -1,30 +0,0 @@
...
-        }
-      ]
-    }
-  net-conf.json: |
-    {
-      "Network": "10.2.0.0/15",
-      "Backend": {
-        "Type": "vxlan"
-      }
-    }
```
linki: Either in this PR or in a subsequent PR (depends on the impact) we need to add this to `deletions.yaml` to clean it up from the cluster.

cluster/manifests/flannel/daemonset.yaml
```diff
@@ -131,9 +108,6 @@ spec:
       - name: flannel-cfg
```
linki, line 126: All references to `flannel-cfg` can also be removed, can't they?

demonCoder95: Ah, good catch. Yes.

linki: Or at least you need to change its type away from `configMap` to something like `hostPath`. There should be other containers doing it already where you can peek.

linki: Because you need to mount the files from the host into container (previously this was done via mounting the configmap).

cluster/manifests/flannel/configmap.yaml
demonCoder95: Added in the next commit. Thanks!

cluster/manifests/flannel/daemonset.yaml
linki: Nevermind, no need to touch the host file system. That's the whole point 🤦  But the references to the configmap we can delete I guess.

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -864,3 +864,30 @@ write_files:
...
+    path: /etc/kubernetes/cni/net.d/10-flannel.conflist
+    content: |
+      {
+        "name": "podnet",
+        "cniVersion": "0.3.1",
+        "plugins": [
+          {
+            "type": "flannel",
+            "delegate": {
+            "isDefaultGateway": true,
```
linki: Let's fix the indentation for the final version.

cluster/manifests/flannel/daemonset.yaml
demonCoder95: Yes, deleted in the next commit. Thanks.

cluster/node-pools/master-default/userdata.yaml
demonCoder95: I've kept the indentation verbatim from the config-map, can you point out the problem?

demonCoder95:
Removed the `config/hash` annotation as the `userdata.yaml` based configuration will rotate the nodes upon a change and pod rotation through the annotation isn't required.

cluster/manifests/flannel/configmap.yaml
linki: I'm un-resolving this because we need to check if we can do that in this PR or have to do it in a subsequent PR. Especially with the recent karpenter changes that change the order of the rollout.

cluster/node-pools/master-default/userdata.yaml
linki: Maybe it's just here on GitHub but `isDefaultGateway` should be one level to the right.

demonCoder95: Oh. I didn't notice the bracket in the previous line. You're right. I'll fix that in the next commit. Thanks.

demonCoder95: Fixed.

cluster/manifests/flannel/configmap.yaml
demonCoder95: Let's merge the PR, then we can check in `dev` channel if the `deletions.yaml` worked. If not I can create a new PR with the changes 🙂

cluster/manifests/flannel/daemonset.yaml
```diff
@@ -36,29 +35,6 @@ spec:
             cpu: 1m
             memory: 50Mi
       containers:
-      - name: delayed-install-cni
-        image: container-registry.zalando.net/teapot/flannel-awaiter:master-9
-        command:
-        - /await-and-copy
```
mikkeloscar, line 42: The awaiter doesn't only copy the config, it also checks (and waits) that it can reach _all_ other nodes in the cluster. Without this flannel could become ready before it truly is ready, and user pods could be scheduled before flannel is able to route to all other nodes. 

We can only drop the _copy config_ part, not the _await_ part.

demonCoder95: Oh, I understood our discussion as we don't need the _await_ anymore because of `kube-node-ready`. I see. I'll bring back the container. Thanks!

demonCoder95: We might need to patch the `flannel-awaiter` to remove the copy functionality from the script and only keep the `await` part 🤔

demonCoder95: Since `flannel-awaiter` has been updated [here](https://github.bus.zalan.do/teapot/flannel-awaiter/pull/10), will proceed to add the container back with the updated version and desired readiness Probe.

demonCoder95: I've added the awaiter container back with the startup probe to probe the dummy file.

demonCoder95: The probe needs fixing. Working on that.

cluster/manifests/flannel/daemonset.yaml
```diff
@@ -37,20 +36,10 @@ spec:
             memory: 50Mi
       containers:
       - name: delayed-install-cni
-        image: container-registry.zalando.net/teapot/flannel-awaiter:master-9
+        image: container-registry.zalando.net/teapot/flannel-awaiter:master-10
```
linki: Can we make this an `initContainer` instead? That seems to better capture the use case and results in one less container that's running.

linki: It probably needs the local flannel running before the `await` can ever succeed, right?

mikkeloscar: > It probably needs the local flannel running before the await can ever succeed, right?

Yeah, that is the challenge and why it runs in this "interesting" way.

demonCoder95: Yes, we discussed making it an init container but then the parallel running is necessary, so kept it as it is.

cluster/manifests/flannel/daemonset.yaml
```diff
@@ -48,6 +48,9 @@ spec:
           limits:
             cpu: 25m
             memory: 50Mi
+        volumeMounts:
+        - name: run
+          mountPath: /run
```
mikkeloscar: This is only for the `flannel-ready` file right? In that case I would propose NOT to mount the host `/run` folder here and simply use the local container file system because the probe also runs in the context of the container.

demonCoder95: The local container filesystem isn't reachable by the `kube-flannel` container when it comes up and it ends up failing the readiness probe. I wanted a shared volume between both these containers to be able to provide the readiness probe to the other container, that already has this mounted.

demonCoder95: Oh, you mean the probe is needed for this container? I thought we needed the probe for `kube-flannel` container 👀

mikkeloscar: > Oh, you mean the probe is needed for this container?

yeah, I didn't realize it was on the other one. We just need it for this one because if any of the containers fail the probe it's enough to mark the pod not ready and fail the node from getting ready to run user pods.

demonCoder95: Oh, that simplifies a lot of things. Great. Thanks for catching this 🙂

demonCoder95: Fixed in the next commit.

cluster/manifests/flannel/daemonset.yaml
demonCoder95: Fixed the probe.

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -864,3 +864,30 @@ write_files:
...
+
+  - owner: root:root
+    path: /etc/kube-flannel/net-conf.json
+    content: |
+      {
+        "Network": "10.2.0.0/15",
+        "Backend": {
+          "Type": "vxlan"
+        }
+      }
```
linki: I wonder if we want to leave this config in the configMap in order to make flannel more self-contained. This is a [flag to flannel](https://github.com/flannel-io/flannel/blob/master/main.go#L131) itself and is not needed by kubelet (As far as I can tell). So from that perspective it can stay with flannel.

linki: I looked at this diff the last couple of days and I'm still confused. So please consider this a "nit".

demonCoder95: Fixed in the next commit. Thanks!

cluster/manifests/flannel/configmap.yaml
demonCoder95: I've removed the entry of `deletions.yaml` since we're only modifying the configmap now and not deleting it. The new one should effectively replace the old one.

demonCoder95:
👍

mikkeloscar:
:+1:

PR #5935, opened 2023-04-18
Avoid flakyness of other job based e2e tests

mikkeloscar:
Similar to #5928 aim to avoid flakyness of e2e tests using jobs, by waiting for the job to complete as opposed to wait for the job pod to be running which can be flaky if the check interval is between the pod running and completing.


demonCoder95:
👍

mikkeloscar:
👍

PR #5943, opened 2023-04-20
Give lower priority to rotation of spot replacement nodes

mikkeloscar:
The spot-node-rescheduler will now set a lower priority on the nodes that are marked for rotation such that nodes rolled as part of e.g. a Kubernetes update have higher priority.

This is done to avoid a problem where during a cluster upgrade the node rotation gets stuck because nodes rotated because of spot replacement keep being selected first by CLC.


mikkeloscar:
:+1:

linki:
👍

PR #5944, opened 2023-04-20
Add application/component tags to NAT gateway stack

mikkeloscar:
Adds `application` and `component` tags to the NAT stack such that NAT gateways are tagged with an application.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #5950, opened 2023-04-21
Use standard DNS settings as applied to all other core components

mikkeloscar:
Adds the standard DNS config to the karpenter deployment to align it with what we use for other components.


linki:
👍

mikkeloscar:
:+1:

PR #5954, opened 2023-04-25
Enable automatic CSI migration by default

linki:
Let's enable CSI Migration by default in the Kubernetes 1.23 branch so that we can roll it out to stable together.


mikkeloscar:
:+1:

katyanna:
👍

PR #5958, opened 2023-04-27
Increase max startup timeout for flannel-awaiter

mikkeloscar:
Follow up to #5932

Increases the StartUp probe to a maximum of  `30 x 10 = 300 seconds` for the `flannel-awaiter` container. The default is `failureThreshold=3` which means `3 x 10 = 30 seconds`. That is not enough as observed by the roll out to the `stups` cluster, hence the increase.


linki:
👍

mikkeloscar:
:+1:

PR #5959, opened 2023-04-27
node-exporter: Disable unused collectors

mikkeloscar:
Disable some node-exporter collector that we don't use:

* Disabling `wifi` and `hwmon` is inspired from how it's configured in https://github.com/prometheus-operator/kube-prometheus (EC2 don't have WIFI and we don't care about hardware monitoring if it even would work)
* Disabling `btrfs` is inspired by: https://twitter.com/MetalMatze/status/1648630535214825475. Not sure if it would have a positive impact for us but we use ext4 so btrfs monitoring definitely is not needed.


linki:
👍

mikkeloscar:
:+1:

PR #5966, opened 2023-04-28
fix: tee filter performance

szuecs:
fix: tee filter performance had a bottleneck because of default http client has only 2 connections to a single host
feature: tee filter has now a client that does also httptrace introspection like the main http client.


szuecs:
:+1:

demonCoder95:
👍

PR #5969, opened 2023-05-02
Remove unused explicit csi-based replacement prometheus

linki:
Remove obsolete CSI based replacement Prometheus to simplify the config.

Since `CSIMigration` the default storage class is automatically handled by CSI and we don't need to explicitly specify it. This makes the replacement Prometheus obsolete for testing CSI.

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -71,27 +71,6 @@ post_apply:
   namespace: kube-system
   kind: VerticalPodAutoscaler
 {{ end }}
-{{ if ne .ConfigItems.prometheus_csi_ebs "true" }}
-- name: prometheus-csi
-  namespace: kube-system
-  kind: StatefulSet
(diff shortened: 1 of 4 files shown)


cluster/manifests/deletions.yaml
```diff
@@ -71,27 +71,6 @@ post_apply:
...
-- name: prometheus
-  namespace: kube-system
-  kind: StatefulSet
-- name: prometheus-storage-volume-prometheus-0
-  kind: PersistentVolumeClaim
-  namespace: kube-system
-- name: prometheus-storage-volume-prometheus-1
-  kind: PersistentVolumeClaim
-  namespace: kube-system
-{{ end }}
```
linki, line 94: We can remove both branches because it's set to `false` in all clusters and therefore everything is cleaned up already.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #5973, opened 2023-05-04
Remove CSI related config items and rely on the defaults

linki:
With 1.23 `CSIMigration` and `CSIMigrationAWS` feature flags are enabled by default. Let's remove all configuration options that allow to disable CSI once it's rolled out everywhere.

Let's merge with the next required update of the AMI (worker node rotation).


demonCoder95:
👍

linki:
👍

PR #5978, opened 2023-05-04
Use autoscaling/v2 for HPAs

mikkeloscar:
With Kubernetes v1.23 the autoscaling API used for Horizontal Pod Autoscalers was promoted from `autoscaling/v2beta2` to `autoscaling/v2`.

Update it where we have HPAs


mikkeloscar:
:+1:

katyanna:
👍

PR #5980, opened 2023-05-04
Update StackSet to Kubernetes v1.23

mikkeloscar:
Updates the stackset-controller to Kubernetes v1.23 version of client-go and friend + updates to use `autoscaling/v2` for HPA resources.

The CRDs are now SO big that the cluster creation fails with:

```
 CustomResourceDefinition.apiextensions.k8s.io \"stacksets.zalando.org\" is invalid: metadata.annotations: Too long: must have at most 262144 bytes" module=stackset-controller
```

I choose a crude way to combat this by simply shorting descriptions longer than 100 chars using this script:

```py
#!/usr/bin/env python

import sys
import os
import yaml
import re

MAX_DESCRIPTION_LENGTH = 100


def slim_crd(crd):
    for version in crd.get("spec", {}).get("versions", []):
        slim_desc(version["schema"]["openAPIV3Schema"])


def slim_desc(obj):
    for key in obj:
        if key == "description" and len(obj[key]) > MAX_DESCRIPTION_LENGTH:
            obj[key] = re.sub("\..*", ".", obj[key])
            # obj[key] = obj[key][:MAX_DESCRIPTION_LENGTH] + "..."
        if key == "properties":
            for _, item in obj[key].items():
                slim_desc(item)


crd = {}

file = sys.argv[1]

with open(file, "r") as f:
    try:
        crd = yaml.safe_load(f)
    except yaml.YAMLError as exc:
        print(exc)
        os.exit(1)

slim_crd(crd)

with open(file, "w", encoding="utf8") as f:
    yaml.safe_dump(crd, f)
```

This _can_ affect UX when using `kubectl explain ...` because it may not show the full description, but it seems like a reasonable trade-off to not hit this stupid limit.


katyanna:
Would a _first sentence only_ approach where you cut out everything after the first "." be a viable option? It's a bit less disruptive to the `kubectl explain`

mikkeloscar:
@katyanna I changed it like you proposed (and updated the script in the description).

mikkeloscar:
:+1:

katyanna:
👍

AlexanderYastrebov:
I bet it is due to `kubectl.kubernetes.io/last-applied-configuration` that contains the whole previous version of CRD

mikkeloscar:
@AlexanderYastrebov yes, that is the problem :/

PR #5987, opened 2023-05-09
Allow disabling legacy autoscaling apis

mikkeloscar:
Make it possible to disable legacy autoscaling APIs so it's easier to phase them out cluster by cluster as users migrate.

They are enabled by default.


katyanna:
:+1:

gargravarr:
:+1:

PR #5989, opened 2023-05-09
skipper: update OAuth grant credentials name

AlexanderYastrebov:
Use special suffix to avoid credentials name clash with existing credentials having name with `-credentials` suffix.


RomanZavodskikh:
:+1:

gargravarr:
:+1:

PR #5992, opened 2023-05-10
Remove legacy HPA field from StackSet CRD by default

mikkeloscar:
This removes the legacy `horizontalPodAutoscaler` field from StackSet (and Stacks) by default. The idea of this is to gradually phase this out across clusters as usage drops.

I have set the config-item: `stackset_legacy_hpa_field_crd_enabled: true` in all clusters such that we simply drop the config-item per cluster to drop the field. This setting should only be changed after validation that no stacksets in the cluster uses the `horizontalPodAutoscaler` field.
This way it's also quick to get a list of clusters still needing migration (those with the config-item still set).


katyanna:
:+1:

gargravarr:
:+1:

PR #6009, opened 2023-05-15
Fix node priority for decommissioning

mikkeloscar:
This is a follow up to #5943

It changes the node priority such that the default is `100` and the priority for spot rescheduled nodes is `90`. Before it was `0` and `-10` respectively, but setting `-10` did not work because of:

```
Node \"ip-10-14-7-186.eu-central-1.compute.internal\" is invalid: metadata.labels: Invalid value: \"-10\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

### DIFF
--- a/cluster/manifests/cluster-lifecycle-controller/deployment.yaml
+++ b/cluster/manifests/cluster-lifecycle-controller/deployment.yaml
@@ -35,7 +35,7 @@ spec:
         operator: Exists
       containers:
       - name: cluster-lifecycle-controller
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6019, opened 2023-05-23
Update fabric-gateway Docker image to support additional filters per route

RomanZavodskikh:
Update fabric-gateway Docker image to support additional filters per route


AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

PR #6033, opened 2023-05-31
Widen status-service's permissions to readonly on all safe resources

linki:
`status-service`'s `/validate` endpoint needs to retrieve annotations of existing resources in the cluster. Since users can currently deploy most resources directly we need to widen `status-service`'s read-only access to these resources. The easiest and straight forward way is to give it the generic `readonly` cluster-wide role.

~Note, that it still can't and shouldn't have access to `Secret`s. There's currently no way to allow to read metadata, such as annotations, without also allowing read access to the secret data itself, hence we need to block it. This means that overwrite protection cannot be done for `Secret`s in the current implementation even after this PR.~

In addition to `readonly` it also requires access to Secret's which is added separately.

I removed the redundant read-only permissions of `status-service`'s specific role.

### DIFF
--- a/cluster/manifests/deployment-service/status-service-rbac.yaml
+++ b/cluster/manifests/deployment-service/status-service-rbac.yaml
@@ -27,6 +27,12 @@ rules:
     resources:
       - secrets
     verbs: ["get", "list", "watch"]
+  - apiGroups:
+      - zalando.org
+    resources:
+      - fabricgateways
+      - fabriceventstreams
+    verbs: ["get", "list", "watch"]
 ---
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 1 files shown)


linki:
👍

cluster/manifests/deployment-service/status-service-rbac.yaml
```diff
@@ -22,44 +22,6 @@ rules:
     resources:
       - cdpdeploymenttasks
     verbs: ["get", "list", "watch"]
-  - apiGroups:
```
demonCoder95, line 25: Why all these deletions were done? Just curious because I don't know.

linki: These permissions become redundant when we give status-service the `readonly` role.

demonCoder95: Ah, got it 🙂

demonCoder95:
👍

PR #6034, opened 2023-05-31
Scrap Akamai DataStream prometheus logs

MustafaSaber:

### DIFF
--- a/cluster/manifests/prometheus/configmap.yaml
+++ b/cluster/manifests/prometheus/configmap.yaml
@@ -232,6 +232,34 @@ data:
           regex: 'kubenurse_httpclient.*'
 {{ end }}
 {{ end }}
+    - job_name: 'akamai-datastreams'
+      scheme: http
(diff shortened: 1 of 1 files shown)


cluster/manifests/prometheus/configmap.yaml
```diff
@@ -232,6 +232,34 @@ data:
...
 {{ end }}
+    - job_name: 'akamai-datastreams'
+      scheme: http
+      metrics_path: '/metrics'
+      kubernetes_sd_configs:
+        - role: pod
+          namespaces:
+            names:
+              - akamai
+            relabel_configs:
```
demonCoder95, line 243: Need to fix the indenting here. Should be the same level as `kubernetes_sd_configs`

MustafaSaber: Thanks for the review @demonCoder95! Will close in favour of using the SDK with the service itself.

PR #6045, opened 2023-06-05
Enable Open Policy Agent filter in Skipper

mjungsbluth:
This PR introduces configuration to use Skipper with Open Policy Agent enabled as per https://github.com/zalando/skipper/pull/2364


cluster/manifests/skipper/configmap-open-policy-agent.yaml
```diff
@@ -0,0 +1,32 @@
...
+          s3_signing:
+            metadata_credentials:
+              aws_region: eu-central-1
+              iam_role: open-policy-agent-instance
+        name: styra-bundles
+        url: "{{ .ConfigItems.skipper_open_policy_agent_bundles_url }}"
+kind: ConfigMap
+metadata:
+  labels:
+    application: open-policy-agent-control-plane
```
mikkeloscar: Since this is the config read by skipper-ingress I think it's also fair to use `skipper-ingress` as the application label.

cluster/manifests/skipper/configmap-open-policy-agent.yaml
```diff
@@ -0,0 +1,32 @@
+{{ if eq .ConfigItems.skipper_open_policy_agent_enabled "true" }}
```
mikkeloscar, line 1: Please add an entry in `deletions.yaml` to remove the config-map in case it was previously enabled but is disabled again.

See example: https://github.com/zalando-incubator/kubernetes-on-aws/blob/e85e0b03498144f81e7d6488e727435764694a25/cluster/manifests/deletions.yaml#L115-L119

cluster/manifests/skipper/deployment.yaml
```diff
@@ -35,6 +35,9 @@ spec:
         prometheus.io/scrape: "true"
 {{- if eq .ConfigItems.skipper_topology_spread_enabled "true" }}
         zalando.org/topology-spread-timeout: 7m
+{{- end }}
+{{- if eq .ConfigItems.skipper_open_policy_agent_enabled "true" }}
+        iam.amazonaws.com/role: "{{ .ConfigItems.skipper_open_policy_agent_instance_role_arn }}"
```
mjungsbluth: This is sth that I'd like to discuss. This will be used to read bundles from S3. It would be better to use a service account IMO and assume a role bound to Skipper. Any preferences?

mikkeloscar: Yes, we should use service account (the OIDC based model) this is used in all other components in this repo. Ideally you can also provision a dedicated role as part of the cluster so it's all managed in one place. See how we do it for other roles here: https://github.com/zalando-incubator/kubernetes-on-aws/blob/d36fd80bc9a7671a9adb20dae588912baede3a45/cluster/cluster.yaml#L1520-L1654 we also have a role naming pattern including the LocalID, which you can follow as well. The idea is that we in theory can have more than one cluster per account (though we don't do it in practice at the moment).

mjungsbluth: Sounds good, I added a new role following the same pattern and assigned it to the existing `skipper-ingress` service account.

cluster/manifests/skipper/configmap-open-policy-agent.yaml
```diff
@@ -0,0 +1,40 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    application: skipper-ingress
```
szuecs, line 6: Please add also
`component: ingress`
or can we offload this to `component: routesrv` (I guess not)?

mjungsbluth: Yes, changed to `component: ingress`

cluster/manifests/skipper/secret.yaml
```diff
@@ -10,3 +10,7 @@ data:
   datadome-api-key: "{{ .Cluster.ConfigItems.datadome_api_key | base64 }}"
   lightstep-token: "{{ .Cluster.ConfigItems.lightstep_token | base64 }}"
   encryption-key: "{{ .Cluster.ConfigItems.skipper_ingress_encryption_key | base64 }}"
+  styra-token: "{{ `placeholder` | base64}}"
+#{{ if eq .Cluster.ConfigItems.skipper_open_policy_agent_enabled "true" }}
+#  styra-token: "{{ .Cluster.ConfigItems.skipper_styra_token | base64 }}"
```
szuecs: commented out secret

rishiraj88:
.

linki:
I rebased this branch on top of `kube-1.24` branch (no conflicts). This is basically the "dev" branch until all clusters are upgraded to 1.24.

cluster/config-defaults.yaml
```diff
@@ -947,3 +950,14 @@ sysctl_settings: ""
...
+# Set defaults values to be used by skipper ingress when the open policy agent is enabled
+{{if eq .Cluster.Environment "test"}}
+skipper_open_policy_agent_bucket_arn: "arn:aws:s3:::zalando-open-policy-agent-sandbox-bundles"
+skipper_open_policy_agent_observability_url: "https://open-policy-agent-sandbox-observability.corporate-iam.zalan.do/v1"
+skipper_open_policy_agent_bundles_url: "https://zalando-open-policy-agent-sandbox-bundles.s3.eu-central-1.amazonaws.com"
+{{else}}
+skipper_open_policy_agent_bucket_arn: "arn:aws:s3:::zalando-open-policy-agent-prod-bundles"
+skipper_open_policy_agent_observability_url: "https://open-policy-agent-observability.corporate-iam.zalan.do/v1"
+skipper_open_policy_agent_bundles_url: "https://zalando-open-policy-agent-prod-bundles.s3.eu-central-1.amazonaws.com"
+{end}}
```
linki: There's one `{` missing here.

cluster/config-defaults.yaml
```diff
@@ -947,3 +950,14 @@ sysctl_settings: ""
 
 # enables/disables the minDomains field for pod topology spread.
 min_domains_in_pod_topology_spread_enabled: "true"
+
+# Set defaults values to be used by skipper ingress when the open policy agent is enabled
+{{if eq .Cluster.Environment "test"}}
```
linki: Let's put this block of config items next to the related one `skipper_open_policy_agent_enabled`

cluster/config-defaults.yaml
```diff
@@ -947,3 +950,14 @@ sysctl_settings: ""
...
+skipper_open_policy_agent_bucket_arn: "arn:aws:s3:::zalando-open-policy-agent-sandbox-bundles"
+skipper_open_policy_agent_observability_url: "https://open-policy-agent-sandbox-observability.corporate-iam.zalan.do/v1"
+skipper_open_policy_agent_bundles_url: "https://zalando-open-policy-agent-sandbox-bundles.s3.eu-central-1.amazonaws.com"
+{{else}}
+skipper_open_policy_agent_bucket_arn: "arn:aws:s3:::zalando-open-policy-agent-prod-bundles"
+skipper_open_policy_agent_observability_url: "https://open-policy-agent-observability.corporate-iam.zalan.do/v1"
+skipper_open_policy_agent_bundles_url: "https://zalando-open-policy-agent-prod-bundles.s3.eu-central-1.amazonaws.com"
```
linki: Let's also define a "default" value for token as empty string `""`, similar to https://github.com/zalando-incubator/kubernetes-on-aws/blob/a06e8318aef6758a1d7a0987a3e8ca0b21fe7323/cluster/config-defaults.yaml#L861

This way all possible config items are documented in this file.

cluster/cluster.yaml
```diff
@@ -1229,6 +1229,40 @@ Resources:
...
+              Federated: !Sub "arn:aws:iam::${AWS::AccountId}:oidc-provider/{{.Cluster.LocalID}}.{{ .Values.hosted_zone }}"
+            Action:
+              - 'sts:AssumeRoleWithWebIdentity'
+            Condition:
+              StringEquals:
+                "{{ .Cluster.LocalID }}.{{ .Values.hosted_zone }}:sub": "system:serviceaccount:kube-system:skipper-ingress"
+        Version: 2012-10-17
+      Path: /
+      Policies:
+        {{ if eq .Cluster.ConfigItems.skipper_open_policy_agent_enabled "true" }}
```
linki: This `if` is superfluous as it's already within another `if` block.

cluster/config-defaults.yaml
```diff
@@ -288,6 +288,21 @@ skipper_ingress_backend_traffic_algorithm: "traffic-segment-predicate"
 # TODO: after a while we can remove this and hardcode (2023-06-30)
 skipper_ingress_default_lb_algorithm: "powerOfRandomNChoices"
 
+# Open Policy Agent
```
linki: Let's improve the comment of this config item. Currently it's not clear what it does.

cluster/config-defaults.yaml
```diff
@@ -946,4 +961,4 @@ control_plane_asg_lifecycle_hook: "true"
 sysctl_settings: ""
 
 # enables/disables the minDomains field for pod topology spread.
-min_domains_in_pod_topology_spread_enabled: "true"
+min_domains_in_pod_topology_spread_enabled: "true"
```
linki: This newline change looks unrelated. Let's not change it here.

cluster/manifests/skipper/configmap-open-policy-agent.yaml
```diff
@@ -0,0 +1,41 @@
...
+      resource: v1/bundles/zalando-apps/{{ `{{ .bundlename }}` }}/discovery.tgz
+      service: styra-bundles
+    labels:
+      environment: {{ .Cluster.Environment }}
+      kubernetes-cluster-alias: {{ .Cluster.Alias }}
+      aws-account-id: "{{ accountID .Cluster.InfrastructureAccount }}"
+    services:
+    - credentials:
+        bearer:
+          token: "{{ `{{ .Env.STYRA_TOKEN }}` }}"
```
linki, line 23: Since this is a secret value it should ideally be stored in a `Secret` object and not it a `ConfigMap`.

cluster/manifests/skipper/configmap-open-policy-agent.yaml
```diff
@@ -0,0 +1,41 @@
...
+      name: styra
+      url: "{{ .ConfigItems.skipper_open_policy_agent_observability_url }}"
+    - credentials:
+        s3_signing:
+          web_identity_credentials:
+            aws_region: eu-central-1
```
linki, line 29: Let's use `{{ .Cluster.Region }}` here.

linki: I'm assuming that `eu-west-1` would be the appropriate value if the cluster was in the `eu-west-1` region. If that's not the case, please disregard my comment.

cluster/manifests/skipper/secret.yaml
```diff
@@ -10,3 +10,6 @@ data:
   datadome-api-key: "{{ .Cluster.ConfigItems.datadome_api_key | base64 }}"
   lightstep-token: "{{ .Cluster.ConfigItems.lightstep_token | base64 }}"
   encryption-key: "{{ .Cluster.ConfigItems.skipper_ingress_encryption_key | base64 }}"
+{{ if eq .Cluster.ConfigItems.skipper_open_policy_agent_enabled "true" }}
```
szuecs: we can also drop the if, similar to other keys.

cluster/manifests/skipper/configmap-open-policy-agent.yaml
linki: We decided to skip this for now because it's not clear whether this has to be the same region as the cluster all the time.

linki:
👍

adetunjiakintundeakinde:
👍

demonCoder95:
👍

PR #6049, opened 2023-06-05
Allow to run containerd AMIs in existing clusters

linki:
We currently configure a single node pool in `teapot` that runs `containerd`. However, this doesn't work out of the box without also switching to the systemd cgroup driver. For a limited time, this config item can be set alongside the AMI to make this work and it should not lead to any node rotations for existing clusters.

This config item is short-lived and should be removed as soon as possible.

The intention is to configure a node pool, e.g. in teapot, like so:

```yaml
- config_items:
    kuberuntu_image_v1_23_amd64: ami-123456789 # AMI that runs containerd
    kuberuntu_image_v1_23_arm64: ami-987654321 # AMI that runs containerd
    support_containerd: "true"
  discount_strategy: none
  instance_type: m5.xlarge
  instance_types:
  - m5.xlarge
  max_size: 300
  min_size: 0
  name: containerd-node-pool
  profile: worker-splitaz
```


demonCoder95:
👍

linki:
👍

PR #6051, opened 2023-06-05
change: default fifo max queue

szuecs:
change: default fifo max queue. In order to tune MTTR for the good we drop out early in case we would queue up requests that would timeout anyways.


szuecs:
:+1:

katyanna:
👍

PR #6066, opened 2023-06-13
change: default fifo timeout to 1s

szuecs:
change: default fifo timeout to 1s

follow up step of https://github.com/zalando-incubator/kubernetes-on-aws/pull/6051

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -90,7 +90,7 @@ skipper_suppress_route_update_logs: "true"
 skipper_validate_query: "true"
 skipper_validate_query_log: "false"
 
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

mikkeloscar:
@szuecs Can you please add some more context on the motivation for this so it's more clear what we gain and potential trade-offs we make with this change?

szuecs:
It won't cause harm as this timeout is only counted on requests that are blocked in the queue. If we queue we normally drop requests quickly and reducing timeout will only help to drop MTTR.

mikkeloscar:
:+1:

PR #6067, opened 2023-06-14
fix: external-dns being redirected to virginia

szuecs:
fix: external-dns being redirected to virginia as we observed on 2023-06-13 22:20-22:50 CEST
      to mitigate this we use the `AWS_REGION` env var to stay in the same region


linki:
👍

szuecs:
:+1:

PR #6070, opened 2023-06-14
Add external RPS metric support

lucastt:
This change upgrades `kube-metrics-adapter` to version 0.2.1 to support External RPS metrics. It also adds E2E tests to ensure  External RPS metrics are working properly.

### DIFF
--- a/test/e2e/kube_metrics_adapter_test.go
+++ b/test/e2e/kube_metrics_adapter_test.go
@@ -134,7 +134,7 @@ var _ = describe("[HPA] Horizontal pod autoscaling (scale resource: Custom Metri
 		tc.Run()
 	})
 
-    It("should scale with external metric based on hostname RPS [CustomMetricsAutoscaling] [Zalando]", func() {
+	It("should scale with external metric based on hostname RPS [CustomMetricsAutoscaling] [Zalando]", func() {
 		hostName := fmt.Sprintf("%s-%d.%s", DeploymentName, time.Now().UTC().Unix(), E2EHostedZone())
 
 		initialReplicas := 2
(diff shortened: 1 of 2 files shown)


test/e2e/kube_metrics_adapter_test.go
```diff
@@ -133,6 +133,38 @@ var _ = describe("[HPA] Horizontal pod autoscaling (scale resource: Custom Metri
 		}
 		tc.Run()
 	})
+
+    It("should scale with external metric based on hostname RPS", func() {
```
mikkeloscar: Please add at least `[Zalando]` in the name similar to the style of other tests. This makes sure the test is in focus when we run the ginkgo test suite. Right now I think it's not being run.

lucastt: Oh, sorry I missed this detail.

mikkeloscar:
:+1:

test/e2e/kube_metrics_adapter_test.go
```diff
@@ -133,6 +133,38 @@ var _ = describe("[HPA] Horizontal pod autoscaling (scale resource: Custom Metri
 		}
 		tc.Run()
 	})
+
+	It("should scale with external metric based on hostname RPS [CustomMetricsAutoscaling] [Zalando]", func() {
+		hostName := fmt.Sprintf("%s-%d.%s", DeploymentName, time.Now().UTC().Unix(), E2EHostedZone())
+
+		initialReplicas := 2
+		scaledReplicas := 1
```
AlexanderYastrebov, line 141: Does this test metric-based _downscaling_?

mikkeloscar: Yes essentially, but from the test perspective it validates that metrics are changing the scaling which is what we want to prove. We are not testing the core HPA functionality of scaling up/down here, just about consuming the metrics and scaling to the right size.

This is modeled after existing tests related to kube-metrics-adapter, but I can't recall why testing scale down made more sense than scaling up test. Maybe it does not make sense/make a difference 🤔

AlexanderYastrebov:
:+1:

PR #6077, opened 2023-06-15
Remove ServiceQuota SLR from kube-1 stack

demonCoder95:
This removes the `ServiceLinkedRole` for `ServiceQuota` service. This is the first step to migration to AILM. The second step will be to deploy the stack that creates these roles through [this PR](https://github.bus.zalan.do/Base-Infrastructure/aws-base-infrastructure/pull/1062).

Basically, since all accounts are using the default value for the config-item, changing the value to false should remove the roles. I'm keeping the config-item in case we need to roll-back, we can do so by setting the item to `true` again. Later, we should remove the item + the resource from the CF template as well, once the migration is complete.

More context: https://github.bus.zalan.do/zooport/issues/issues/3517

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -2085,13 +2085,6 @@ Resources:
       Description: "AWS service role for application autoscaling DynamoDB"
     Type: "AWS::IAM::ServiceLinkedRole"
 {{- end }}
-{{- if eq .Cluster.ConfigItems.quotas_service_link_enabled "true" }}
-  ServiceLinkedRoleServiceQuotas:
-    Properties:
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

demonCoder95:
👍

demonCoder95:
All the `QuotaService` SLRs have been deleted from all k8s accounts. We can now proceed to merge this PR that removes this SLR completely from `kubernetes-on-aws`.

The SLR has been added in base infra, and change propagated through all channels! 🎉 The SLR is now managed by AILM 🙂

mikkeloscar:
:+1:

demonCoder95:
👍

PR #6080, opened 2023-06-16
stackset-contoller: update CRDs

AlexanderYastrebov:
Stack and StackSet CRDs were updated to not exceed annotation size limit, see https://github.com/zalando-incubator/stackset-controller/pull/498

Note that CRDs are not verbatim copies but contain conditional blocks (see https://github.com/zalando-incubator/kubernetes-on-aws/pull/5992). Templates were commented to make CRDs valid yaml.

Followup on https://github.com/zalando-incubator/kubernetes-on-aws/pull/5980

### DIFF
--- a/cluster/manifests/stackset-controller/01-stack-crd.yaml
+++ b/cluster/manifests/stackset-controller/01-stack-crd.yaml
@@ -1,4 +1,4 @@
-
+# This is an adjusted copy of https://github.com/zalando-incubator/stackset-controller/blob/master/docs/stack_crd.yaml
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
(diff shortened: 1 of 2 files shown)


cluster/manifests/stackset-controller/01-stack-crd.yaml
```diff
@@ -300,29 +328,37 @@ spec:
...
+                      replicas to which the autoscaler can scale down. It defaults
+                      to 1 pod.
                     format: int32
                     type: integer
                 required:
                 - maxReplicas
                 - metrics
                 type: object
-{{- if eq .Cluster.ConfigItems.stackset_legacy_hpa_field_crd_enabled "true" }}
+# {{ if eq .Cluster.ConfigItems.stackset_legacy_hpa_field_crd_enabled "true" }}
```
AlexanderYastrebov, line 339: Template yaml-commented here

cluster/manifests/stackset-controller/01-stackset-crd.yaml
```diff
@@ -536,29 +576,37 @@ spec:
...
                               of replicas to which the autoscaler can scale down.
+                              It defaults to 1 pod.
                             format: int32
                             type: integer
                         required:
                         - maxReplicas
                         - metrics
                         type: object
-{{- if eq .Cluster.ConfigItems.stackset_legacy_hpa_field_crd_enabled "true" }}
+# {{ if eq .Cluster.ConfigItems.stackset_legacy_hpa_field_crd_enabled "true" }}
```
AlexanderYastrebov, line 586: Template yaml-commented here

MustafaSaber: why comment not remove?

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

cluster/manifests/stackset-controller/01-stackset-crd.yaml
AlexanderYastrebov: We need this conditional block to remove part of the CRD when config item is false. I put it into yaml comment such that whole file is a valid yaml (see PR description). It is still evaluated because CLM template rendered does not care about yaml comments and treats the file just as plain text.

PR #6085, opened 2023-06-19
Enable `RouteSRV` as default in production

MustafaSaber:
Rollout done from `zregisrty` on multiple stages

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -167,7 +167,7 @@ datadome_api_key: ""
 # - "pre" - routesrv is deployed, skipper uses own k8s dataclient
 # - "exec" - routesrv is deployed, skipper uses routesrv
 {{if eq .Cluster.Environment "production"}}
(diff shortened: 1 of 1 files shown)


cluster/config-defaults.yaml
```diff
@@ -166,13 +166,12 @@ datadome_api_key: ""
 # - "false" - routesrv deployment is removed, skipper uses own k8s dataclient
 # - "pre" - routesrv is deployed, skipper uses own k8s dataclient
 # - "exec" - routesrv is deployed, skipper uses routesrv
+skipper_routesrv_enabled: "exec"
 {{if eq .Cluster.Environment "production"}}
-skipper_routesrv_enabled: "false"
 skipper_routesrv_cpu: "1000m"
 skipper_routesrv_memory: "1Gi"
```
szuecs: can you move memory also outside, as we have the same setting for both.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #6087, opened 2023-06-19
Fix e2e by generating service_account_key via aws-account-creator

mikkeloscar:
This removes the use of `openssl` to generate `service_account_private_key` as part of the e2e cluster and instead relies on `aws-account-creator` which is also used for all other certs/keys.

The motivation for this is that `openssl genrsa` changed behavior in the latest base image (`container-registry.zalando.net/library/python-3.11-slim`) and not generates a key PEM format incompatible with CLM.

We could also fix it by using `openssl genrsa -traditional` but better to not rely on `openssl` at all.

Note: This change will not work against `dev` because the e2e cluster setup depends on the `cluster_config.sh` of the target branch and this change is now incompatible. We need to merge this to `dev` and any new PR against `dev` will work.

It's validated that this change works in #6090


mikkeloscar:
:+1:

mikkeloscar:
:+1:

demonCoder95:
👍

PR #6088, opened 2023-06-19
ExternalDNS: allow to configure domain filter via config item

linki:
Allow to configure domain filter with a config item.

The default setting results in the current behaviour so there's no change when using the default config item value (empty string).

After that we can configure, e.g.:

```yaml
external_dns_domain_filter: teapot.zalan.do,teapot-e2e.zalan.do
```

to get the flags `--domain=filter=teapot.zalan.do --domain-filter=teapot-e2e.zalan.do` which includes all zones matching either of these two suffixes.


szuecs:
rebased on dev

szuecs:
:+1:

mikkeloscar:
:+1:

PR #6098, opened 2023-06-20
Update deployment-service to return empty response if no errors

linki:
If resources have no errors and warnings they won't be displayed at all in the response.


szuecs:
:+1:

linki:
👍

PR #6103, opened 2023-06-21
feature: fabric enable tracing filters

szuecs:
feature: fabric enable tracing filters

### DIFF
--- a/cluster/manifests/fabric-gateway/deployment.yaml
+++ b/cluster/manifests/fabric-gateway/deployment.yaml
@@ -1,4 +1,4 @@
-{{ $version := "master-200" }}
+{{ $version := "master-208" }}
 
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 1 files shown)


cluster/manifests/fabric-gateway/deployment.yaml
```diff
@@ -1,4 +1,4 @@
-{{ $version := "master-200" }}
+{{ $version := "master-208" }}
```
RomanZavodskikh: I think both 208 and 209 are ok, but would like it to be 209.

RomanZavodskikh:
:+1:

cluster/manifests/fabric-gateway/deployment.yaml
szuecs: yes because of e2e weirdness I agree

szuecs:
:+1:

mikkeloscar:
:+1:

PR #6109, opened 2023-06-22
Grant `list`, `get` and `watch` for `EndpointSlice` resources for `readonly` role.

1u0:
This is to enable the mentioned role to inspect `EndpointSlice` K8s resources for troubleshooting.
Some information about `EndpointSlice` K8s resource: https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/.

### DIFF
--- a/cluster/manifests/roles/readonly-role.yaml
+++ b/cluster/manifests/roles/readonly-role.yaml
@@ -169,6 +169,14 @@ rules:
   - list
   - watch
 - apiGroups:
+  - discovery.k8s.io
+  resources:
+  - endpointslices
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - extensions
   - networking.k8s.io
   resources:


cluster/manifests/roles/poweruser-role.yaml
```diff
@@ -165,6 +165,14 @@ rules:
...
   - patch
   - update
+- apiGroups:
+  - discovery.k8s.io
+  resources:
+  - endpointslices
+  verbs:
+  - get
+  - list
+  - watch
```
mikkeloscar: poweruser role implicitly depend on readonly, so it's not needed to add read actions to this role.

1u0: Thanks, I didn't know that.
The PR is updated to contain changes for `readonly` role only.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #6114, opened 2023-06-23
skipper: do not create HTTPS redirect routes when behind NLB

AlexanderYastrebov:
When Skipper runs behind AWS Network Load Balancer (NLB) it has a dedicated handler for HTTP traffic and therefore HTTPS redirect routes could be omitted to reduce routing table size.

See related https://github.com/zalando/skipper/issues/1914


cluster/manifests/skipper/deployment.yaml
```diff
@@ -483,6 +484,7 @@ spec:
 {{ if or (eq .ConfigItems.nlb_switch "pre") (eq .ConfigItems.nlb_switch "exec") }}
           - "-forwarded-headers=X-Forwarded-For,X-Forwarded-Proto=https,X-Forwarded-Port=443"
```
AlexanderYastrebov, line 485: This enforces `X-Forwarded-Proto: https` behind NLB but only except `-forwarded-headers-exclude-cidrs`.
Redirect routes never match such requests as they use `Header("X-Forwarded-Proto", "http")` predicate.

AlexanderYastrebov:
It is possible to have both ALB and NLB in the same cluster and for ALB we offer HTTP support.

PR #6123, opened 2023-06-27
Update Cluster Autoscaler to Kubernetes v1.23.17

mikkeloscar:
This updates our fork of the cluster-autoscaler to a version using Kubernetes v1.23.17. The purpose is to use `policy/v1` api for PodDisruptionBudgets to ensure that it will work with Kubernetes v1.25 where the `policy/v1beta1` API is no longer available.

Ref: https://github.com/zalando-incubator/autoscaler/pull/103

It also gives the autoscaler permissions to read namespaces. This is done because the scheduler code in v1.23.17 now needs this to check namespace labels: https://github.com/kubernetes/kubernetes/blob/v1.23.17/pkg/scheduler/framework/plugins/interpodaffinity/plugin.go#L130-L140

lastly it gives the autoscaler permission to read `csidrivers` and `csistoragecapacities` which have also been introduced in the scheduler in v1.23.17

### DIFF
--- a/cluster/manifests/kube-cluster-autoscaler/01-rbac.yaml
+++ b/cluster/manifests/kube-cluster-autoscaler/01-rbac.yaml
@@ -37,6 +37,7 @@ rules:
   resources:
   - "pods"
   - "services"
+  - "namespaces"
   - "replicationcontrollers"
   - "persistentvolumeclaims"
   - "persistentvolumes"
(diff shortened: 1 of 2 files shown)


linki:
👍

linki:
image

mikkeloscar:
:+1:

PR #6129, opened 2023-06-28
Allow setting custom sysctl settings

mikkeloscar:
This introduces a feature to allow setting custom `sysctl` settings per node pool. The motivation for this request is that we sometime need custom `sysctl` settings because of vendor tool requirements or simply because we want to test other settings. Having to make changes to the AMI is a lot of effort so it's useful to be able to set it per node pool so we can test it before making changes across all nodes.

To illustrate how this works we can look at changing [`aio-max-nr`](https://sort.veritas.com/public/documents/HSO/2.0/linux/productguides/html/hfo_admin_ubuntu/ch04s03.htm#:~:text=The%20aio%2Dmax%2Dnr%20parameter,wide%20number%20of%20asynchronous%20requests.) which is a concrete example from a vendor requirement.

On a node without a custom setting the value is this:

```sh
sudo cat /proc/sys/fs/aio-max-nr
65536
```

By setting the `sysctl_settings` config item we can change the value:

```yaml
- discount_strategy: spot
  instance_types:
  - m5.xlarge
  - m5.2xlarge
  - m5.4xlarge
  max_size: 99
  min_size: 0
  name: worker-with-aio-max-nr
  profile: worker-splitaz
  config_items:
    sysctl_settings: "fs.aio-max-nr=8388608"
```

Which results in the setting being set on a node of the pool:

```sh
sudo cat /proc/sys/fs/aio-max-nr
8388608
```

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -907,3 +907,20 @@ apiserver_memory_limit_percent: "80"
 
 # specify if control plane nodes should rely on ASG Lifecycle Hook or not
 control_plane_asg_lifecycle_hook: "true"
+
+# This allows setting custom sysctl settings. The config-item is intended to be
+# used on node-pools rather being set globally.
+#
+# The value is a comma seprated configuration of `sysctl_setting=value` as
+# illustrated in the example below.
+#
+# Example:
+#
+#   sysctl_settings: "fs.aio-max-nr=8388608,fs.inotify.max_user_watches=100000"
+#
+# Which translates to a file on the node:
+#
+#  cat /etc/sysctl.d/99-custom-sysctl-settings.conf
+#  fs.aio-max-nr = 8388608
+#  fs.inotify.max_user_watches = 100000
+sysctl_settings: ""
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -880,6 +880,15 @@ write_files:
           qps: {{ .Cluster.ConfigItems.event_rate_limit_config_qps }}
           burst: {{ .Cluster.ConfigItems.event_rate_limit_config_burst }}
 {{ end }}
+{{- if ne .NodePool.ConfigItems.sysctl_settings ""}}
+  - owner: root:root
+    path: /etc/sysctl.d/99-custom-sysctl-settings.conf
+    content: |
+      {{- range $i, $cfgs := split .NodePool.ConfigItems.sysctl_settings ","}}
+      {{- $cfg := split $cfgs "="}}
+      {{index $cfg 0}} = {{index $cfg 1}}
+      {{- end}}
+{{- end}}
 
   - owner: root:root
     path: /etc/kubernetes/cni/net.d/10-flannel.conflist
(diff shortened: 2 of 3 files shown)


demonCoder95:
Should we also create a config-defaults entry? If nothing else just a comment for documentation purposes?

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -883,6 +883,15 @@ write_files:
...
           burst: {{ .Cluster.ConfigItems.event_rate_limit_config_burst }}
 {{ end }}
+{{- if index .NodePool.ConfigItems "sysctl_settings"}}
+  {{- range $i, $cfgs := split .NodePool.ConfigItems.sysctl_settings ","}}
+  {{- $cfg := split $cfgs "="}}
+  - owner: root:root
+    path: /etc/sysctl.d/0{{$i}}-{{index $cfg 0}}.conf
+    content: |
+      {{index $cfg 0}} = {{index $cfg 1}}
+  {{- end}}
```
AlexanderYastrebov: Why not put them into a single file (i.e. move range into the `content`)?

mikkeloscar: Good idea. Will do.

mikkeloscar:
> Should we also create a config-defaults entry? If nothing else just a comment for documentation purposes?

Added

linki:
👍

cluster/config-defaults.yaml
```diff
@@ -907,3 +907,20 @@ apiserver_memory_limit_percent: "80"
...
+# The value is a comma seprated configuration of `sysctl_setting=value` as
+# illustrated in the example below.
+#
+# Example:
+#
+#   sysctl_settings: "fs.aio-max-nr=8388608,fs.inotify.max_user_watches=100000"
+#
+# Which translates to a file on the node:
+#
+#  cat /etc/sysctl.d/01-custom-sysctl-settings.conf
```
szuecs: I am not sure if we would like to put it last.
Like this a default in image will override the settings from the node-pool
Maybe better the filename should be `99-custom-sysctl-settings.conf`

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -883,6 +883,15 @@ write_files:
           qps: {{ .Cluster.ConfigItems.event_rate_limit_config_qps }}
           burst: {{ .Cluster.ConfigItems.event_rate_limit_config_burst }}
 {{ end }}
+{{- if ne .NodePool.ConfigItems.sysctl_settings ""}}
+  - owner: root:root
+    path: /etc/sysctl.d/01-custom-sysctl-settings.conf
```
AlexanderYastrebov: I don't know what we have in the image by default, on my system it says
```
~$ cat /etc/sysctl.d/README 
This directory contains settings similar to those found in /etc/sysctl.conf.
In general, files in the 10-*.conf range come from the procps package and
serve as system defaults.  Other packages install their files in the
30-*.conf range, to override system defaults.  End-users can use 60-*.conf
and above, or use /etc/sysctl.conf directly, which overrides anything in
this directory.

After making any changes, please run "service procps start" (or, from
a Debian package maintainer script "invoke-rc.d procps start").
```
so I am wondering should we use `30-` or `60-` prefix to enable override of defaults as well?..

AlexanderYastrebov: There is also a symlink `99-sysctl.conf -> ../sysctl.conf`

mikkeloscar: Changed to `99-` as it was intention to have it override other settings.

cluster/config-defaults.yaml
mikkeloscar: Changed to `99-` as it was intention to have it override other settings.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #6131, opened 2023-06-28
Update deployment service to support batch/v1 for CronJobs

mikkeloscar:
This updates the deployment-service to use `batch/v1` for CronJobs instead of `batch/v1beta1` which will not be available in Kubernetes v1.25.


mikkeloscar:
:+1:

linki:
👍

PR #6055, opened 2023-06-08
Add Audittrail IAM Role

zaklawrencea:
Due to the changes to the `audittrail-adapter` that will now send and store kubernetes events in a centralised S3 bucket, this PR adds an IAM role that will allow the adapter to sufficiently interact with the bucket via list and put actions.

[Relevant teapot issue](https://github.bus.zalan.do/teapot/issues/issues/3436)

PR #6133, opened 2023-06-28
Make auditrail role unique

mikkeloscar:
Follow up to #6055 to make the role unique when there are multiple clusters per account (e.g. in e2e tests).


zaklawrencea:
:+1:

mikkeloscar:
:+1:

PR #6141, opened 2023-06-28
all: simplify empty value checks

AlexanderYastrebov:
From https://pkg.go.dev/text/template#hdr-Actions
```
{{if pipeline}} T1 {{end}}
	If the value of the pipeline is empty, no output is generated;
	otherwise, T1 is executed. The empty values are false, 0, any
	nil pointer or interface value, and any array, slice, map, or
	string of length zero.
	Dot is unaffected.
```


mikkeloscar:
:+1:

linki:
👍

PR #6145, opened 2023-06-29
Consolidate Audittrail IAM roles

zaklawrencea:
Follow up to https://github.com/zalando-incubator/kubernetes-on-aws/pull/6055.

After testing the setup using the recently added centralised Audittrail IAM role, it was discovered that the approach of a single, extended role would be the simpler solution. This is due to the limitations of attaching multiple IAM roles to a single Service Account and the limitations of attaching multiple Service Accounts to a Kubernetes Pod. 

With a single role setup, the configuration of the `Audittrail-Adapter`'s Service Account can remain in it's current state, with only the policy of the centralised S3 Bucket needing to be edited to allow for access from this extended role.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -2067,39 +2067,6 @@ Resources:
                 Effect: Allow
                 Resource:
                   - !GetAtt AuditTrailBucket.Arn
-
-              - Action:
-                  - s3:PutObject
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #6149, opened 2023-06-29
skipper: update version to v0.16.125

RomanZavodskikh:
* dataclient/kubernetes: reduce amount of logs
* Introduce lockedSource type to avoid concurrent use of rand.Source

See https://github.com/zalando/skipper/compare/v0.16.118...v0.16.125

PR #6150, opened 2023-06-29
skipper: update version to v0.16.125

AlexanderYastrebov:
* dataclient/kubernetes: reduce amount of logs
* Introduce lockedSource type to avoid concurrent use of rand.Source

See https://github.com/zalando/skipper/compare/v0.16.118...v0.16.125


AlexanderYastrebov:
See #6149

PR #6157, opened 2023-06-30
feature: fabric gateway set ssl-policy to migrate to default cloud load balancers

szuecs:
feature: fabric gateway set ssl-policy to migrate to default cloud load balancers


AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #6193, opened 2023-07-17
Add config item to enable/disable traffic segment probing

linki:
This config instructs `cdp-steps` and `deployment-service` to expect TrafficSegments as the implementation for StacksetController's traffic switching functionality.

This config item can and should also be used to deploy or configure the corresponding StacksetController version such that they are aligned.


linki:
👍

gargravarr:
:+1:

PR #6199, opened 2023-07-19
Add config item for audit webhook batch max size

zaklawrencea:
Allow the audit webhook batch size to be configurable. We see batches of audit events timing out leading to entire batches being retried exponentially. By configuring (and specifically lowering) the batch size value from the default 400, we can ensure all events within the batch are processed within the timeout window.


zaklawrencea:
:+1:

demonCoder95:
👍

PR #5230, opened 2022-06-30
Add ClusterRole/ClusterRoleBinding and config-item for hyped-articles-lifecycle-management controller

zaklawrencea:
Related Teapot Issue: [Enable hyped article configmaps for application: hyped-articles-lifecycle-management](https://github.bus.zalan.do/teapot/issues/issues/3285).

> For hyped article releases the Purchase Risk Management team is responsible for distributing a list of articles which need to be handled (e.g. protected) in a certain way. This information is distributed via configmaps to various Kubernetes clusters and the relevant applications then mount the configmaps to read the information.

> In order to allow a controller to update the configmaps we need to give it access in dedicated clusters.

> The controller has the application ID: `hyped-articles-lifecycle-management` and needs permissions according to this [sheet](https://docs.google.com/spreadsheets/d/1Eql94dotMm8MW1chIptcnSr9iba4TpWXrvsDXHHg1ps/edit#gid=0).

This PR introduces a ClusterRole/ClusterRoleBinding providing the hyped-articles-lifecycle-management controller access to the relevant configmaps. It also adds a config-item which will then be enabled in the specific clusters which the controller requires access to.

PR #6201, opened 2023-07-19
Add missing stups prefix to zalando iam user

linki:
I believe the user name should have the `stups_` prefix. Please compare to a similar setup in `halm`: https://github.com/zalando-incubator/kubernetes-on-aws/pull/5230/files#diff-75b0f459088b8fce75a8f158f22d7cbd17b1b4491c5d5be8c00ebee19f1c567dR32

This is tentative. Maybe it works without the prefix these days.

### DIFF
--- a/cluster/manifests/roles/business-partner-rbac.yaml
+++ b/cluster/manifests/roles/business-partner-rbac.yaml
@@ -29,5 +29,5 @@ roleRef:
 subjects:
 - apiGroup: rbac.authorization.k8s.io
   kind: User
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
At least other configs in this folder use the prefix

linki:
Right, they all do. Thanks for pointing that out. Let's go ahead then.

linki:
👍

zaklawrencea:
:+1:

PR #6203, opened 2023-07-20
Update some skipper images

MustafaSaber:
Some of skipper images are outdated, and we also introduced some changes to routegroup admission webhook at https://github.com/zalando/skipper/pull/2431

I think the update is good anyway since we mitigated some CVE's in skipper, if you agree I can add it to the rest of skipper images here (I think only 1 left).

I changed `skipper-metrics` one because we also introduced couple of new metrics so I thought it makes sense to include here.

### DIFF
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -275,7 +275,7 @@ func createSkipperPodSpec(route string, port int32) corev1.PodSpec {
 		Containers: []corev1.Container{
 			{
 				Name:  "skipper",
-				Image: "container-registry.zalando.net/teapot/skipper:v0.15.23",
+				Image: "container-registry.zalando.net/teapot/skipper:latest",
 				Args: []string{
 					"skipper",
 					"-inline-routes",
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -275,7 +275,7 @@ func createSkipperPodSpec(route string, port int32) corev1.PodSpec {
 		Containers: []corev1.Container{
 			{
 				Name:  "skipper",
-				Image: "container-registry.zalando.net/teapot/skipper:latest",
+				Image: "container-registry.zalando.net/teapot/skipper:v0.16.154",
 				Args: []string{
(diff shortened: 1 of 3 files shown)


gargravarr:
:+1:

myaser:
I think you missed the sidecar named skipper-proxy

MustafaSaber:
Let's see if tests pass

test/e2e/util.go
```diff
@@ -275,7 +275,7 @@ func createSkipperPodSpec(route string, port int32) corev1.PodSpec {
 		Containers: []corev1.Container{
 			{
 				Name:  "skipper",
-				Image: "container-registry.zalando.net/teapot/skipper:v0.15.23",
+				Image: "container-registry.zalando.net/teapot/skipper:latest",
```
MustafaSaber: It makes sense to use `latest` tag here.

MustafaSaber:
What are the next steps here? Re-create the cluster and try to run test locally?

test/e2e/util.go
mikkeloscar: We can't use `latest` in e2e because we check that the images are production ready as part of e2e.

MustafaSaber: then that's the problem I think, will push and see. Thanks!

mikkeloscar:
:+1:

test/e2e/util.go
```diff
@@ -275,7 +275,7 @@ func createSkipperPodSpec(route string, port int32) corev1.PodSpec {
 		Containers: []corev1.Container{
 			{
 				Name:  "skipper",
-				Image: "container-registry.zalando.net/teapot/skipper:v0.15.23",
+				Image: "container-registry.zalando.net/teapot/skipper:v0.16.154",
```
MustafaSaber, line 278: Can't use `latest` tag because we disable it from `admission-controller`

MustafaSaber:
👍

PR #6211, opened 2023-07-24
Lower default audit webhook batch max size

zaklawrencea:
Following the reduction in errors in multiple clusters after lowering the audit batch max size from the default `400` value to `250`, this change will set this value to the default in all clusters.

More information in [Teapot Issue 3436](https://github.bus.zalan.do/teapot/issues/issues/3436) ([relevant comment](https://github.bus.zalan.do/teapot/issues/issues/3436#issuecomment-4651754))


gargravarr:
:+1:

linki:
👍

PR #6223, opened 2023-07-25
Configure ExternalDNS to cache the list of hosted zones for 1 hour

linki:
This _might_ help with ExternalDNS in `cloud`.

It doesn't reduce the number of requests for a single fetch of all hosted zones with pagination but it reduces the overall number of requests by only doing it once per hour instead of every minute.

Making it the default because hosted zones change so rarely that there's really no point in doing it more than once per hour.


linki:
👍

zaklawrencea:
:+1:

PR #6241, opened 2023-07-27
Run master nodes in test clusters on jammy

linki:
Let's switch master nodes to `jammy` in test clusters.

I picked `if eq .Cluster.Environment "test"` so that e2e tests still run with `focal` which then maps to what's actually running in production clusters. We'd see any issues in the dev channel first.


gargravarr:
:+1:

demonCoder95:
👍

PR #6249, opened 2023-07-28
fix operator for affinity on karpenter deployment

demonCoder95:
The operator value `Equals` is invalid. It should be `In`.

```
operator	<string> -required-
     Represents a key's relationship to a set of values. Valid operators are In,
     NotIn, Exists, DoesNotExist. Gt, and Lt.
```


myaser:
👍🏻

mikkeloscar:
:+1:

PR #6267, opened 2023-07-31
Reduce Karpenter CPU request

mikkeloscar:
Reduce Karpenter CPU request so it can fit on a `m6g.large` instance as we have in e2e.


mikkeloscar:
:+1:

myaser:
👍🏻

PR #6268, opened 2023-07-31
Reduce Karpenter CPU more

mikkeloscar:
Follow up to #6267 to reduce the CPU requests even more for karpenter to allow _all_ pods that need to run on master nodes to fit on `m6g.large` as we use in e2e.


myaser:
👍🏻

mikkeloscar:
:+1:

PR #6269, opened 2023-07-31
Reduce Karpenter CPU even more

mikkeloscar:
Follow up to #6268 again.

Reduce CPU on Karpenter even more so all components can fit on master nodes.


mikkeloscar:
:+1:

myaser:
👍🏻

PR #6275, opened 2023-08-01
Update Audittrail Adapter IAM Role to include new centralised S3 buckets 

zaklawrencea:
Update the Audittrail Adapter IAM Role to include the new centralised S3 buckets. As the production and test cluster events will be separated into two buckets, also update the referenced bucket in the role accordingly.


zaklawrencea:
:+1:

mikkeloscar:
:+1:

PR #6284, opened 2023-08-04
Prevent karpenter from scaling down e2e pod

mikkeloscar:
This is a quick-fix to prevent karpenter from scaling down e2e pods. Ideally it should not touch pods without a parent in the first place, that is to be understood in karpenter.


mikkeloscar:
:+1:

linki:
👍

PR #6285, opened 2023-08-04
Introduce a path to roll back to leagcy docker AMI

mikkeloscar:
This introduces a way that would enable us to roll back to the legacy docker AMI if needed. Can be done per node-pool and/or cluster. Not done for master-nodes as we don't see any issues there.


demonCoder95:
👍

mikkeloscar:
:+1:

PR #6293, opened 2023-08-07
Prevent scaling down orphaned or job pods with Karpenter

mikkeloscar:
This adds "prevent-scale-down" injection for pods that are from a Job or doesn't have a parent resource, when the provider is `karpenter`.

The motivation for this is to have the same behavior with karpenter as Cluster Autoscaler where those type of pods are handled in a special way and are not evicted when a node _could_ be scaled down.

In CA we implemented it directly: https://github.com/zalando-incubator/autoscaler/pull/29/files to avoid needing to implement this in Karpenter we can simply inject the respective prevent-scale-down annotation on those pods which Karpenter would respect.

Once rolled out it will allow to revert #6284


mikkeloscar:
:+1:

demonCoder95:
👍

PR #6307, opened 2023-08-09
Include StackSet ingress/routegroup definition in stack definition

gargravarr:
This Pull Request updates both StackSet and Stack CRD to include the Ingress and RouteGroup definition in the Stack.

With the Ingress/Routegroup included in the Stack definition, we can implement in the StackSet controller to base creation of the pre-stack and segment ingresses/routegroups directly from the Stack, and re-create a specific stacks' Ingress/RouteGroup in case of accidental deletion.

I put this feature behind a config item `stackset_ingress_in_stack_crd`, in case we need to rollback for a cluster.


gargravarr:
:+1:

mikkeloscar:
:+1:

PR #6304, opened 2023-08-08
Add config-item for Audittrail target S3 Bucket

zaklawrencea:
This PR makes the `audittrail-adapter` target S3 Bucket configurable via a config item. The default is set to the central S3 Bucket located in the `stups` cluster.

PR #6309, opened 2023-08-09
Restore adapter bucket to zalando-kubernetes-audit

demonCoder95:
This restores the bucket to `zalando-kubernetes-audit` which makes the rollback unnecessary, since the errors we observed were transient errors and they're gone.
 We would like to keep the setup running with the new bucket in other channels (beyond `dev`). So, "reverting" the change from #6304 before it goes further than `dev`.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #6310, opened 2023-08-09
Drop prevent downscale from e2e

mikkeloscar:
Reverts #6284 which is no longer needed as the admission-controller will correctly handle this as of #6293


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6335, opened 2023-08-14
fix: admissionControl filter in distributed call chains

szuecs:
fix: admissionControl filter in distributed call chains are not counting load shedding responses as errors, otherwise intermidiate services will shedd load because of leaf services that shedd load
feature: deletion of response Cookie is possible

https://github.com/zalando/skipper/compare/v0.16.166...v0.17.7


szuecs:
:+1:

mikkeloscar:
:+1:

PR #6341, opened 2023-08-15
Validation webhook for ingress

MustafaSaber:
Introducing validation for skipper `Ingress` annotations

Related https://github.com/zalando/skipper/pull/2493 & https://github.com/zalando/skipper/issues/1618

### DIFF
--- a/cluster/manifests/01-admission-control/routegroups-webhook.yaml
+++ b/cluster/manifests/01-admission-control/routegroups-webhook.yaml
@@ -3,7 +3,8 @@ kind: ValidatingWebhookConfiguration
 metadata:
   name: "skipper-admitter.teapot.zalan.do"
   labels:
-    application: skipper-admission-webhook
+    application: skipper-ingress
+    component: webhook
 webhooks:
 {{ if eq .Cluster.ConfigItems.routegroups_validation "enabled" }}
   - name: "routegroup-admitter.teapot.zalan.do"
--- a/cluster/manifests/01-admission-control/routegroups-webhook.yaml
+++ b/cluster/manifests/01-admission-control/routegroups-webhook.yaml
@@ -1,11 +1,11 @@
-{{ if eq .Cluster.ConfigItems.routegroups_validation "enabled" }}
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
-  name: "routegroup-admitter.teapot.zalan.do"
+  name: "skipper-admitter.teapot.zalan.do"
   labels:
-    application: routegroups-admission-webhook
+    application: skipper-admission-webhook
 webhooks:
+{{ if eq .Cluster.ConfigItems.routegroups_validation "enabled" }}
   - name: "routegroup-admitter.teapot.zalan.do"
     rules:
       - operations: ["CREATE", "UPDATE"]
@@ -18,6 +18,7 @@ webhooks:
     admissionReviewVersions: ["v1"]
     sideEffects: None
     timeoutSeconds: 5
+{{ end }}
   - name: "ingress-admitter.teapot.zalan.do"
     rules:
       - operations: ["CREATE", "UPDATE"]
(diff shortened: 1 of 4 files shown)


cluster/manifests/deletions.yaml
```diff
@@ -1,7 +1,7 @@
 # everything defined under here will be deleted before applying the manifests
 pre_apply:
 {{ if ne .Cluster.ConfigItems.routegroups_validation "enabled" }}
-- name: "routegroup-admitter.teapot.zalan.do"
+- name: "skipper-admitter.teapot.zalan.do"
```
AlexanderYastrebov: I think you'd need to have temporary unconditional deletion for `routegroup-admitter.teapot.zalan.do` due to rename.

cluster/manifests/01-admission-control/routegroups-webhook.yaml
```diff
@@ -1,11 +1,11 @@
-{{ if eq .Cluster.ConfigItems.routegroups_validation "enabled" }}
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
-  name: "routegroup-admitter.teapot.zalan.do"
+  name: "skipper-admitter.teapot.zalan.do"
   labels:
-    application: routegroups-admission-webhook
+    application: skipper-admission-webhook
```
AlexanderYastrebov: Should we take a chance and use the same application `skipper-ingress` as we use for skipper (and `webhook` component)?

cluster/manifests/deletions.yaml
MustafaSaber: Good catch

cluster/manifests/01-admission-control/routegroups-webhook.yaml
MustafaSaber: I think that wouldn't be a problem.
FYI those application IDs are not registered.

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -241,9 +241,8 @@ write_files:
             - mountPath: /etc/kubernetes/admission-controller-kubeconfig
               name: admission-controller-kubeconfig
               readOnly: true
-{{- if or (eq .Cluster.ConfigItems.routegroups_validation "provisioned") (eq .Cluster.ConfigItems.routegroups_validation "enabled") }}
```
MustafaSaber, line 244: Condition doesn't make sense if we always need ingress and sometimes routegroup

cluster/manifests/deletions.yaml
AlexanderYastrebov: Old `routegroup-admitter.teapot.zalan.do` should be put unconditionally and better into `post_apply` such that it is removed after `skipper-admitter.teapot.zalan.do` is created.

`skipper-admitter.teapot.zalan.do` should not be removed at all because `.Cluster.ConfigItems.routegroups_validation` condition only applies to the nested webhook inside of the `skipper-admitter.teapot.zalan.do`

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -491,7 +489,7 @@ write_files:
               -> disableAccessLog()
               -> setPath("/metrics")
               -> "http://127.0.0.1:8081";
-            routegroups_admission_webhook: Path("/routegroups-admission-webhook")
+            skipper_admission_webhook: Path("/skipper-admission-webhook")
```
AlexanderYastrebov, line 492: Check out where else `/routegroups-admission-webhook` is used, I think you'd need to update there as well.

AlexanderYastrebov:
:+1:

cluster/manifests/deletions.yaml
MustafaSaber: We will need also not to remove `skipper-admitter.teapot.zalan.do` in the first run since it's not there in the first place

gargravarr:
:+1:

PR #6342, opened 2023-08-15
[e2e] Update AMI to the latest version

linki:
This is just for e2e to validate that it can pull the new pause image's location.


linki:
Although this e2e passed, the Sandbox image in `master-281` gets garbage collected which breaks the nodes. This has been reverted for now.

PR #6343, opened 2023-08-15
Allow configuring max-requests-inflight

mikkeloscar:
This enables tweaking the `--max-requests-inflight` setting of the `kube-apiserver` which controls the control the concurrency limit of the [API Priority and Fairness](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/) feature.

This is done to be able to test different settings as it's likely that the default settings are too restrictive when scaling master nodes vertically.

The default is the same as the Kubernetes default for now.


mikkeloscar:
:+1:

linki:
👍

PR #6354, opened 2023-08-17
reconfigure node-lifecycle.provider per node-pool

myaser:
during karpenter rollout, clusters will run karpenter provisioners and CA pools at the same time
both scalers require different controls for the pod evection and for spot/on-demand scheduling

for pod evection:
- CA uses the annotation `cluster-autoscaler.kubernetes.io/safe-to-evict`
- karpenter uses the annotation `karpenter.sh/do-not-evict`

for the spot/on-demand scheduling
- CA uses this label `aws.amazon.com/spot` as node selector or affinity constraint
- karpneter uses this lable `karpenter.sh/capacity-type` as node selector or affinity constraint

in order to run both scalers in parallel, we need:
1.  to know the correct provider for each node-pool (karpenter vs Zalando). this is provided as configuration
2. to map the pod to the node-pool that it will land on. this is "guessed" by reading the "dedicated" toleration of the label

then we are able to inject the right controls

this PR configures the providers for each node-pool

to be merged after https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/692

### DIFF
--- a/cluster/manifests/01-admission-control/config.yaml
+++ b/cluster/manifests/01-admission-control/config.yaml
@@ -117,7 +117,7 @@ data:
   # keep legacy configs during rollouts
   pod.node-lifecycle.provider: "{{ .Cluster.ConfigItems.teapot_admission_controller_node_lifecycle_provider }}"
 {{- range $group, $provider := nodeLifeCycleProviderPerNodePoolGroup .Cluster.NodePools }}
-  pod.node-lifecycle.provider.{{ group }}: "{{ provider }}"
+  pod.node-lifecycle.provider.{{ $group }}: "{{ $provider }}"
 {{- end}}
 
   pod.runtime-policy.enable: "{{ .Cluster.ConfigItems.teapot_admission_controller_runtime_policy_enabled }}"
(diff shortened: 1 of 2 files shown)


cluster/manifests/01-admission-control/config.yaml
```diff
@@ -114,7 +114,11 @@ data:
   node.node-not-ready-taint.enable: "{{ .Cluster.ConfigItems.teapot_admission_controller_node_not_ready_taint }}"
   node.extended-node-restriction.enable: "true"
 
+  # keep legacy configs during rollouts
   pod.node-lifecycle.provider: "{{ .Cluster.ConfigItems.teapot_admission_controller_node_lifecycle_provider }}"
+{{- range $group, $provider := nodeLifeCycleProviderPerNodePoolGroup .Cluster.NodePools }}
+  pod.node-lifecycle.provider.{{ group }}: "{{ provider }}"
```
mikkeloscar:
```suggestion
  pod.node-lifecycle.provider.{{ $group }}: "{{ $provider }}"
```

mikkeloscar:
:+1:

myaser:
👍🏻

PR #6358, opened 2023-08-21
Reduce required CPU resources to run etcd backup job

linki:
Given our etcd setup, it's essential that etcd-backup runs at least once a day (retention time).

Since we don't implement CPU limits anyway, we only use them for making scheduling decisions. However, for etcd-backup there's really no decision to make since it already limits itself to run on master nodes.

[EDIT] This is very hacky. Looking for feedback on this.

### DIFF
--- a/cluster/manifests/etcd-backup/cronjob.yaml
+++ b/cluster/manifests/etcd-backup/cronjob.yaml
@@ -51,10 +51,10 @@ spec:
 {{ end }}
             resources:
               limits:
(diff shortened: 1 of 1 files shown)


myaser:
👍🏻

demonCoder95:
I think it's fine to have a hack for now, just to repair the issue, we can look into the behavior of the master node autoscaler later.

demonCoder95:
👍

PR #6360, opened 2023-08-21
pump admission-controller version

myaser:
Update admission-controller version.
The latest version supports multiple node-lifecycle providers.
It is needed to run the karpenter and cluster autoscaler concurrently, where every scaler expects a different provide

has to be rolled out after #6354


demonCoder95:
👍

myaser:
👍🏻

PR #6375, opened 2023-08-23
Make it possible to setup a custom etcd stack

mikkeloscar:
This makes the `etcd_stack_name` configurable such that it's possible to create a custom etcd cluster if needed for testing.

Apart from setting a custom value for `etcd_stack_name` you will also need to adjust:

* `etcd_endpoints` - This needs to follow a specific naming scheme based on `etcd_stack_name`. If `etcd_stack_name` is set to `my-etcd-<username>` then the etcd_endpoints` should be `https://etcd-server.<username>.<alias>.zalan.do:2479. This is logic hardcoded in the etcd.py and would need to be adjusted to make this more flexible.
* `etcd_s3_backup_bucket` - needs to be adjusted in order to create a unique bucket for this new stack to not overlap with the existing one.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -11,7 +11,7 @@ Resources:
     Properties:
       FromPort: 2379
       ToPort: 2379
-      GroupId: !ImportValue 'etcd-cluster-etcd:etcd-security-group-id'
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6381, opened 2023-08-24
Include externalIngress port in stack definition.

gargravarr:
Updates the Stack CRD in to include the `externalIngress` port. This field comes from the parent StackSet. Based from the generated CRD from [here](https://github.com/zalando-incubator/stackset-controller/blob/crd-stack-ingressrg/docs/stack_crd.yaml#L350-L361).


gargravarr:
:+1:

mikkeloscar:
:+1:

PR #6384, opened 2023-08-24
skipper: add config to enable ingress validation webhook

AlexanderYastrebov:
Similar to existing routegroups_validation (hence plural).

Renamed manifest from routegroups-webhook.yaml to skipper-webhook.yaml.

Tested that it works also when both are disabled:
```
~$ kubectl get ValidatingWebhookConfiguration skipper-admitter.teapot.zalan.do
NAME                               WEBHOOKS   AGE
skipper-admitter.teapot.zalan.do   0          27h
```

Followup on #6341


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #6385, opened 2023-08-24
skipper: downgrade webhook image version

AlexanderYastrebov:
Use version before #6341.
PR #6384 introduces config item for ingress validation.


RomanZavodskikh:
:+1:

AlexanderYastrebov:
Data shows there is no need to downgrade

PR #6388, opened 2023-08-25
Set maxPods on karpenter provisioner

mikkeloscar:
This sets the `maxPods` on the karpenter provisioner to the same value as we set directly in the kubelet configuration in userdata.
This ensures that Karpenter has the right expectation about how many pods a new instance can have.

After setting this we no longer see the errors:

```
check failed, expected 234 of resource pods, but found 110 (47.0% of expected)
```

Which was previously seen an indicates karpenter was overestimating the amount of pods on a node and thereby scaling less nodes than needed.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6395, opened 2023-08-29
Default to jammy distribution in dev channel

linki:
This makes `jammy` the default distribution for all node pools in the `dev` channel.

The condition is a little complicated and it's not good practice to switch on the channel like that (we don't do it anywhere else as far as I can see). It's only for easing the rollout, to not need to set the config item per cluster. The entire if clause and config items will be removed eventually.


linki:
👍

gargravarr:
:+1:

PR #6397, opened 2023-08-30
Update StackSet Controller

gargravarr:
* Updates StackSet Controller version. Changes for this Pull Request:
    * [1.4.13](https://github.com/zalando-incubator/stackset-controller/releases/tag/v1.4.13)
    * [1.4.12](https://github.com/zalando-incubator/stackset-controller/releases/tag/v1.4.12)
* Drops config items for legacy Horizontal Pod Autoscaler field
* Makes per-stack ingress definitions the default

PR #6406, opened 2023-09-01
Drop stackset legacy field check from admission-controller

mikkeloscar:
Drops the legacy stackset field check from admission-controller which is no longer needed since: #6397


demonCoder95:
👍

mikkeloscar:
:+1:

PR #6407, opened 2023-09-01
Update audittrail-adapter user group metric

mikkeloscar:
Adds a `user_group` dimension to the metrics. It's similar to `user` but groups nodes under one name `system:node` to reduce carnality. Could potentially do the same for other clients in the future if it makes sense.

This illustrates the problem that it solves:

All the individual node users would be treated as a single one reducing the number of time-series significantly.

image


mikkeloscar:
:+1:

szuecs:
:+1:

PR #6411, opened 2023-09-04
Make kube-proxy sync period configurable per cluster.

gargravarr:
Adds a new config item `kube_proxy_sync_period`, so that we can test the effect of reducing kube-proxy's `configSyncPeriod` in a cluster.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6417, opened 2023-09-05
Drop teapot_admission_controller_node_lifecycle_provider config-item

demonCoder95:
This config-item is no longer needed after the `admission-controller` patch made [here](https://github.bus.zalan.do/teapot/admission-controller/pull/194) which allows configuring `provider` per node-pool, which is the desired configuration for the `karpenter` rollout. So, a configuration per cluster is no longer needed.

This is also necessary since we wish to support both providers (cluster-autoscaler and karpenter) during the rollout phase.

### DIFF
--- a/cluster/manifests/01-admission-control/config.yaml
+++ b/cluster/manifests/01-admission-control/config.yaml
@@ -113,8 +113,6 @@ data:
   node.node-not-ready-taint.enable: "{{ .Cluster.ConfigItems.teapot_admission_controller_node_not_ready_taint }}"
   node.extended-node-restriction.enable: "true"
 
-  # keep legacy configs during rollouts
-  pod.node-lifecycle.provider: "{{ .Cluster.ConfigItems.teapot_admission_controller_node_lifecycle_provider }}"
 {{- range $group, $provider := nodeLifeCycleProviderPerNodePoolGroup .Cluster.NodePools }}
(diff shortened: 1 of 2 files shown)


demonCoder95:
I'm unsure if this is all the changes we need to make, maybe the config for the `admission-controller` needs to change to drop the cluster-wide parameter `pod.node-lifecycle.provider: "zalando"` altogether?

I think I'm missing something.

cluster/manifests/01-admission-control/config.yaml
```diff
@@ -113,8 +113,7 @@ data:
   node.node-not-ready-taint.enable: "{{ .Cluster.ConfigItems.teapot_admission_controller_node_not_ready_taint }}"
   node.extended-node-restriction.enable: "true"
 
-  # keep legacy configs during rollouts
-  pod.node-lifecycle.provider: "{{ .Cluster.ConfigItems.teapot_admission_controller_node_lifecycle_provider }}"
+  pod.node-lifecycle.provider: "zalando"
```
mikkeloscar: This setting is replaced by the per node pool setting below. There is no longer a global setting but every node pool, including the default, gets a setting.

mikkeloscar: I dropped it.

mikkeloscar:
:+1:

demonCoder95:
👍

PR #6418, opened 2023-09-05
Update audittrail-adapter to allow unset s3 audit bucket

zaklawrencea:
Currently we are sending too many requests to AWS S3 which is leading to a large cost increase. While we implement a better solution (e.g. request batching), this change will let us easily disable the centralised S3 feature.


zaklawrencea:
:+1:

linki:
👍

PR #6422, opened 2023-09-06
Tell prometheus to scrape kube-proxy metrics

linki:
Add kube-proxy to Prometheus' scraping targets. Kube-proxy doesn't use Pod IPs, but rather node IPs. An update to the firewall rules is needed for that.


linki:
👍

gargravarr:
:+1:

PR #6424, opened 2023-09-07
Update stackset-controller support multi metrics of single type

mikkeloscar:
This updates stackset-controller to support multiple autoscaler metrics of the same type. E.g. you can have two ZMON based metrics. Before it was only possible to have one per type.

Ref: https://github.com/zalando-incubator/stackset-controller/pull/365


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6432, opened 2023-09-08
mount efs with the additional noresvport option

linki:
For Zooport ticket https://github.bus.zalan.do/zooport/issues/issues/3965

This adds the recommended `noresvport` mount option. `vers=4.1` is default when nothing is specified, so we have to list that as well.

Note, this only applies to **new** EFS based PVCs.


mikkeloscar:
:+1:

katyanna:
👍

PR #6433, opened 2023-09-08
Allow status service to parse/decrypt Secret objects

linki:
Status Service uses Controller's `Parse` function which processes Secrets. As such it also needs access to the KMS key.

This was needed to make the `/validate` API work. In theory the payload of a Secret isn't needed to fulfil this functionality. 

Therefore, as a follow up we should check if we can avoid Status Service from needing access to Secret payloads.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -935,6 +935,11 @@ Resources:
                   - !Sub "arn:aws:s3:::${DeploymentServiceBucket}"
                 Effect: Allow
               - Action:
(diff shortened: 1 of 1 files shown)


mikkeloscar:
blocking for now as there should be a better way to do this without giving permissions

linki:
~~Can this be closed because we did https://github.com/zalando-incubator/kubernetes-on-aws/pull/6508?~~

Yes, this can be closed.

PR #6434, opened 2023-09-08
test: add shadow deployment for routesrv with enabled endpointslices

szuecs:
test: add shadow deployment for routesrv with enabled endpointslices

missing: skipper version that has endpointslices, so pipeline aborted


szuecs:
validated elsewhere

PR #6435, opened 2023-09-11
Avoid evicting kube-cluster-autoscaler on minimal disk usage

mikkeloscar:
If the master node comes under disk pressure the kube-cluster-autoscaler is one of the first pods to be kicked out because it uses a few KB of disk but doesn't request any. Request 256Mi to ensure it doesn't get kicked out right away.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6437, opened 2023-09-11
Add system/kubelet overhead to karpenter provisioner

mikkeloscar:
This adds `systemReserved` and `kubeReserved` to the karpenter provisioner such that karpenter knows the node overhead which we also configure on the kubelet.

We observed a case with a user pod requesting `3310m` CPU and karpenter spinning up a m6i.xlarge instance which has `4000m`. However, with daemonset overhead (`587m`) and system/kube overhead (`200m`) the pod didn't fit: `3310 + 587 + 200 = 4097m`


demonCoder95:
👍

mikkeloscar:
:+1:

PR #6448, opened 2023-09-12
[STABLE] Graceful node shutdown for master nodes

mikkeloscar:
Rolling out #6441 in stable.

**Note**: This does not rotate masters, because we set `control_plane_graceful_shutdown=false` in all clusters. After merging we will change to `control_plane_graceful_shutdown=true` to do the rotation as described in #6441


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6449, opened 2023-09-12
Allow changing kube-proxy verbose level

mikkeloscar:
Allow setting kube-proxy verbose level via config-item to get a better understanding of kube-proxy updates during master node rotation.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6450, opened 2023-09-12
Make skipper-proxy wait longer than apiserver during shutdown

mikkeloscar:
This is related to #6441 

This is an experiment which doesn't make things worse, but may also not improve anything.

This is an attempt to improve the network path in red shown in the diagram below:

image

The idea is that watches from e.g. kube-proxy going through the `NLB -> skipper-proxy -> kube-apiserver` can get stale during shutdown/rotation of a master node. By increasing the shutdown timeout of skipper-proxy, kube-apiserver which have a shutdown delay of 60s (10s less) is able to close the connection when it's still going through skipper-proxy and thereby allow for a more graceful closing of the connection.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #6468, opened 2023-09-14
Drop unused apiserver flag

mikkeloscar:
Since we run with the default `--endpoint-reconciler-type=lease`, the flag `--apiserver-count` has no effect and should be dropped.

Was found during investigation for #6441


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6469, opened 2023-09-15
Remove use of Values.node_labels

mikkeloscar:
Remove the use of `Values.node_labels` which is a [static value in CLM](https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/23f59e19ce033c8f7d804d9d73f14c6a70c1a847/provisioner/clusterpy.go#L267-L268) that might as well be defined inline in the manifests

This is a no-op because the rendered manifests will be the same.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #6477, opened 2023-09-18
Allowing dropping read-only events before audittrail-api

mikkeloscar:
This introduces a flag `--audittrail-drop-read-only` on the audittrail-adapter which allows dropping read-only events before sending to audittrail-api. The motivation for this is that we currently drop read-only events on the audit config (apiserver) level and therefore never see them in audittrail-adapter. Since audittrail-adapter also provide metrics about clients calling the apiserver it's helpful to have the read-only events in those metrics, but we still don't want to send to audittrail-api because of the volume, which was the motivation for dropping them in the first place.

The config-item `auditlog_read_access` is still set to `false` by default meaning the events are dropped at the apiserver level. The idea is to enable it in some cluster to see the effect before enabling it everywhere.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -632,6 +632,12 @@ audittrail_adapter_memory: "200Mi"
 audittrail_adapter_timeout: "2s"
 audittrail_adapter_bucket: "zalando-kubernetes-audit"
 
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #6484, opened 2023-09-19
Update audittrail-adapter to fix user_group metric label

mikkeloscar:
Fixes a bug in audittrail-adapter where it didn't group nodes under a single `user` `system:node` which was the whole point of the `user_group` label.


katyanna:
👍

mikkeloscar:
:+1:

PR #6491, opened 2023-09-19
Update vpa to v0.12.0

mikkeloscar:
This updates the VPA to v0.12.0 since the change is hard to test without running in a real cluster we keep the setup of `current|legacy` version. This allows us to quickly switch back to the previous version if the new VPA version doesn't behave as expected.

This is similar to what we did in #5361

The motivation for updating is to be compatible with Kubernetes v1.25 in the future.

ref: https://github.com/zalando-incubator/autoscaler/pull/101


mikkeloscar:
:+1:

szuecs:
:+1:

PR #6503, opened 2023-09-21
all: update kubectl cluster-info output

AlexanderYastrebov:
kubectl cluster-info displays addresses of the control plane and services with label kubernetes.io/cluster-service=true. It also uses kubernetes.io/name to name services but since label value must be a DNS name it can not be used to provide human readable name.

This change:
* removes kubernetes.io/cluster-service and kubernetes.io/name from resources not relevant for proxy access by cluster users
* adds kubernetes.io/cluster-service to skipper-ingress-routesrv service
* removes port name from routesrv service (also makes it consistent with kubernetes dashboard service)
* adds missing routesrv Service deletion

From https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster-services/#manually-constructing-apiserver-proxy-urls
> If you haven't specified a name for your port, you don't have to specify port_name in the URL.
> You can also use the port number in place of the port_name for both named and unnamed ports.

I.e. with named routesrv port proxy url would unnecessary need to be either /api/v1/namespaces/kube-system/services/skipper-ingress-routesrv:http/proxy/routes or /api/v1/namespaces/kube-system/services/skipper-ingress-routesrv:80/proxy/routes

### DIFF
--- a/cluster/manifests/coredns-local/daemonset-coredns.yaml
+++ b/cluster/manifests/coredns-local/daemonset-coredns.yaml
@@ -7,8 +7,6 @@ metadata:
     application: kubernetes
     component: coredns
     version: v1.9.3
-    kubernetes.io/cluster-service: "true"
-    kubernetes.io/name: "CoreDNS"
 spec:
   updateStrategy:
     type: OnDelete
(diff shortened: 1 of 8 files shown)


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #6505, opened 2023-09-21
fabric-gateway: update version to master-222

AlexanderYastrebov:
* require custom route headers
* use FabricGateway name as a value for the fabric-gateway label


RomanZavodskikh:
:+1:

AlexanderYastrebov:
:+1:

PR #6508, opened 2023-09-22
Update deployment-service to transparently validate secret resources

mikkeloscar:
Updates the `deployment-service` to a version that don't try to decrypt secret resources for the validation endpoint.

This is an alternative to #6433 so it doesn't need additional permissions.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #6534, opened 2023-10-05
Give poweruser permission to manage endpointslices

linki:
Add manage permissions for `endpointslices` for the poweruser role. This is consistent with `endpoints` (poweruser can manage them) and avoids any difference between the two, since `endpointslices` is the successor of `endpoints`. `endpointslices` has already been added to the readonly role in a previous change.

This should allow users to deploy (namespaced) roles and bindings targeting `endpointslices`.

This came out of https://github.bus.zalan.do/zooport/issues/issues/4070

### DIFF
--- a/cluster/manifests/roles/poweruser-role.yaml
+++ b/cluster/manifests/roles/poweruser-role.yaml
@@ -166,6 +166,16 @@ rules:
   - patch
   - update
 - apiGroups:
+  - discovery.k8s.io
+  resources:
+  - endpointslices
+  verbs:
+  - create
+  - delete
+  - deletecollection
+  - patch
+  - update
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
Or maybe we should revoke it for endpoints? What are the usecases for write access?

szuecs:
Endpoints are ok, because of postgres-operator/patroni needs to write it and so 24x7 seems reasonable to write it.
Endpointslices, I would rather wait and would rather disallow use.

linki:
The [original request](https://github.bus.zalan.do/zooport/issues/issues/4070) for it was closed. So I'm closing this as "not needed anymore".

PR #6535, opened 2023-10-05
logging: configure keyValue parser

AlexanderYastrebov:
Configure built-in keyValue parser for containers using https://github.com/sirupsen/logrus that logs in the format compatible with https://pkg.go.dev/github.com/kr/logfmt


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #6482, opened 2023-09-18
feature: skipper enable endpointslices to scale beyond 1000 endpoints

szuecs:
feature: skipper enable endpointslices to scale beyond 1000 endpoints

see also https://github.com/zalando/skipper/compare/v0.17.40...v0.18.0

PR #6538, opened 2023-10-06
skipper: allow use of endpointslices without routesrv

AlexanderYastrebov:
Followup on #6482


szuecs:
Why?

AlexanderYastrebov:
To avoid local `make test` passing without 100% coverage which requires one to amend pull request later after pipeline fails.

AlexanderYastrebov:
We have all other options in parallel such that we can disable routesrv and let skipper talk to k8s

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #4704, opened 2021-10-06
Skipper/routesrv2 [2/2]

szuecs:
needs to wait for https://github.com/zalando-incubator/kubernetes-on-aws/pull/4692 to be merged and then wait for [1/2] https://github.com/zalando-incubator/kubernetes-on-aws/pull/4702 to be in stable

PR #6540, opened 2023-10-09
skipper: remove route-cache volume

AlexanderYastrebov:
There is no need to mount empty volume to the /tmp.

This was added by #4704 initially


AlexanderYastrebov:
Container has readOnlyRootFilesystem

PR #6543, opened 2023-10-11
Use endpointslice discovery for prometheus

mikkeloscar:
Using `endpoints` discovery for Prometheus scaling is limited to 1000 endpoints. When running with more than 1000 nodes we only get metrics for 1000 of them for cadvisor and node-exporter.

The fix is to use `endpointslice` discovery which this PR introduces.

https://github.com/prometheus/prometheus/blob/ffa74eb12db3d46c98033cea42cfdb347e7d1e75/discovery/kubernetes/endpoints.go#L360


myaser:
👍🏻

mikkeloscar:
:+1:

PR #6547, opened 2023-10-12
skipper: rollback to v0.18.13

AlexanderYastrebov:
https://github.com/zalando/skipper/compare/v0.18.13...v0.18.24


AlexanderYastrebov:
Closing in favor of #6549

mikkeloscar:
re-opening as we may want to do this instead of moving forward.

szuecs:
We need to rebase anyways if so and the roll forward will unwire the endpointregistry which added a mutex in the hot path. I don't see that this rollback would be better than roll forward. Just think about a required emergency patch and we would use a complete untested version.

PR #6549, opened 2023-10-12
Update skipper image to avoid using endpointregistry in the hotpath

RomanZavodskikh:
This is needed because endpointregistry seems to show not so good performance when it is under lock contention.

FYI https://github.com/zalando/skipper/compare/v0.18.24...v0.18.32


RomanZavodskikh:
:+1:

AlexanderYastrebov:
:+1:

PR #6552, opened 2023-10-13
skipper: add change cause

AlexanderYastrebov:
Add kubernetes.io/change-cause with a value of the deployed version string.


AlexanderYastrebov:
Does not answer the question "what has changed".

PR #6561, opened 2023-10-19
Update to Karpenter v0.31.3

mikkeloscar:
This should reduce the chance of hitting the race condition bug outlined in: https://github.com/aws/karpenter/issues/4310

https://github.com/aws/karpenter/releases/tag/v0.31.0
https://github.com/aws/karpenter/releases/tag/v0.31.1
https://github.com/aws/karpenter/releases/tag/v0.31.2
https://github.com/aws/karpenter/releases/tag/v0.31.3

### DIFF
--- a/cluster/manifests/z-karpenter/karpenter.k8s.aws_awsnodetemplates.yaml
+++ b/cluster/manifests/z-karpenter/karpenter.k8s.aws_awsnodetemplates.yaml
@@ -197,52 +197,15 @@ spec:
               networkInterfaces:
                 items:
                   properties:
-                    associateCarrierIPAddress:
-                      description: "Associates a Carrier IP address with eth0 for
-                        a new network interface. \n Use this option when you launch
-                        an instance in a Wavelength Zone and want to associate a Carrier
-                        IP address with the network interface. For more information
-                        about Carrier IP addresses, see Carrier IP addresses (https://docs.aws.amazon.com/wavelength/latest/developerguide/how-wavelengths-work.html#provider-owned-ip)
-                        in the Wavelength Developer Guide."
-                      type: boolean
                     associatePublicIPAddress:
                       description: Associates a public IPv4 address with eth0 for
                         a new network interface.
-                      type: boolean
-                    deleteOnTermination:
-                      description: Indicates whether the network interface is deleted
-                        when the instance is terminated.
                       type: boolean
                     description:
                       description: A description for the network interface.
                       type: string
                     deviceIndex:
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

cluster/manifests/z-karpenter/karpenter.k8s.aws_awsnodetemplates.yaml
```diff
@@ -197,54 +197,17 @@ spec:
...
-                        an instance in a Wavelength Zone and want to associate a Carrier
-                        IP address with the network interface. For more information
-                        about Carrier IP addresses, see Carrier IP addresses (https://docs.aws.amazon.com/wavelength/latest/developerguide/how-wavelengths-work.html#provider-owned-ip)
-                        in the Wavelength Developer Guide."
-                      type: boolean
                     associatePublicIPAddress:
                       description: Associates a public IPv4 address with eth0 for
                         a new network interface.
                       type: boolean
-                    deleteOnTermination:
```
myaser, line 212: why these fields are removed?

mikkeloscar: I rebased the patch on your simplified PR where these fields were dropped: https://github.com/aws/karpenter-provider-aws/pull/4353

The other fields we don't use so it doesn't matter IMO, but `deleteOnTermination` I didn't think about if it breaks behavior :thinking:

myaser: Perhaps it may leak network interfaces when the node is deleted. I did not test it, but the default behavior is `deleteOnTermination=true`, so it should not be a problem. I just added it here for "just in case"

mikkeloscar: I will test it before we merge :) If it does delete them, then we're good IMO, if it leaks, then it's not ok.

myaser:
👍🏻

cluster/manifests/z-karpenter/karpenter.k8s.aws_awsnodetemplates.yaml
mikkeloscar: Succesfully tested, you're right the default is `deleteOnTermination=true`

image

So it's not leaking anything :tada:

mikkeloscar:
:+1:

myaser:
👍🏻

PR #6565, opened 2023-10-24
Categorise manual, poweruser, and emergency access as exempt from APF

zaklawrencea:
Do not apply Kubernetes API Priority and Fairness limits to users of the `Manual`/`PowerUser`/`Emergency` roles. Users responding to incidents should not have their requests affected by other applications/components so they can apply mitigations within a cluster as quickly and effectively as possible.

### DIFF
--- a/cluster/manifests/apf/manual-emergency-access.yaml
+++ b/cluster/manifests/apf/manual-emergency-access.yaml
@@ -1,0 +1,44 @@
+apiVersion: flowcontrol.apiserver.k8s.io/v1beta2
+kind: FlowSchema
+metadata:
+  name: manual-emergency-access
+spec:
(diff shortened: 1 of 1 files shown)


zaklawrencea:
@mikkeloscar Thoughts on changing this to only `Manual` /`Emergency` access? `PowerUser` could potentially allow users to abuse their access in test clusters to negatively impact the apiserver and the focus here is on incident response anyway. Or alternatively, we could only apply this for production clusters.

szuecs:
if it's ready please drop the do-not-merge label, but I always check all PRs that have not the label :)

zaklawrencea:
Closing for now, this can be revisited in the future if it's still relevant.

PR #6567, opened 2023-10-30
Add new okta based Administrator role to KMS keys

mikkeloscar:
With the switch to Okta based login we also switch the AWS role granted to Admin users. Therefore we need to also add that role to the respective kms keys/policies to continue to be able to manage those resources manually if needed in an emergency.


mikkeloscar:
:+1:

linki:
👍

PR #6573, opened 2023-11-02
Use a common name for the deployment secret across all clusters

linki:
In order to have a predictable reference name to the KMS key used to encrypt/decrypt secrets, let's create a new alias for it that doesn't include the local id. This way the KMS alias can always be referenced via `alias/deployment-secret` in every account.

The local ID was prepended to allow for multiple clusters to share the same AWS account. In practice, this is only the case in teapot and teapot-e2e where we could special case it.

Note, this doesn't solve the problem for teapot at the moment.


linki:
Closing this in favour of moving it to AILM.

Meanwhile created the missing two aliases manually.

PR #6576, opened 2023-11-07
Update Stackset controller with Stack owned ConfigMap feature

katyanna:
-> [Stackset Controller: Reference ConfigMap for per stack versioning](https://github.com/zalando-incubator/stackset-controller/pull/520)

### DIFF
--- a/cluster/manifests/stackset-controller/01-stackset-crd.yaml
+++ b/cluster/manifests/stackset-controller/01-stackset-crd.yaml
@@ -596,6 +596,25 @@ spec:
                         - maxReplicas
                         - metrics
                         type: object
+                      configurationResources:
+                        description: ConfigurationResources describes the ConfigMaps
+                          that will be created. Later Secrets and PlatformCredentialSets
+                          will also be defined on ConfigurationResources
+                        items:
+                          description: ConfigurationResourcesSpec makes it possible
+                            to defined the config resources to be created
+                          properties:
+                            configMapRef:
+                              description: ConfigMap to be versioned for Stack
+                              properties:
+                                name:
+                                  description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                                    TODO: Add other useful fields. apiVersion, kind,
+                                    uid?'
+                                  type: string
+                              type: object
+                          type: object
+                        type: array
                       minReadySeconds:
                         description: Minimum number of seconds for which a newly created
                           pod should be ready without any of its container crashing,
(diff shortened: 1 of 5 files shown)


cluster/manifests/stackset-controller/deployment.yaml
```diff
@@ -30,12 +30,15 @@ spec:
       serviceAccountName: stackset-controller
       containers:
       - name: stackset-controller
-        image: "container-registry.zalando.net/teapot/stackset-controller:{{ $version }}"
+        image: "container-registry-test.zalando.net/teapot/stackset-controller:pr-520-36"
```
linki: Let's keep `{{ $version }}` (We set it on the first line)

cluster/manifests/stackset-controller/01-stackset-crd.yaml
```diff
@@ -596,6 +596,25 @@ spec:
                         - maxReplicas
                         - metrics
                         type: object
+                      configurationResources:
```
linki, line 600: These CRD changes should also be guarded by the `stackset_configmap_support_enabled` config item.

cluster/manifests/stackset-controller/rbac.yaml
```diff
@@ -99,6 +99,18 @@ rules:
   - create
   - update
   - patch
+{{- if eq .Cluster.ConfigItems.stackset_configmaps_support_enabled "true" }}
```
linki: Typo: `stackset_configmap_support_enabled ` vs. `stackset_configmaps_support_enabled`

cluster/manifests/stackset-controller/rbac.yaml
```diff
@@ -99,6 +99,18 @@ rules:
...
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - create
+  - update
+  - delete
```
linki: I don't think we need `delete` anymore. Maybe let's just align it with the same definition as for events to keep it simple:
```
  - get
  - list
  - watch
  - create
  - update
  - patch
  ```

cluster/manifests/stackset-controller/deployment.yaml
katyanna: This is a temporary image, this line is never touching dev. We get back to the proper format once the the actual image is ready.

linki: But why did you change the `$version` variable on line 1 then?

katyanna: I did that after because I thought I may want to specify the test version on the labels. No obvious reason, just a "maybe it'd be useful".

linki:
👍

katyanna:
👍

PR #6579, opened 2023-11-13
prometheus: support optional metrics version label

AlexanderYastrebov:
Add version label derived from optional metrics/version pod label. A new label is used with metrics/ prefix to avoid adding version metrics label when it is not needed.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #6589, opened 2023-11-24
test: require IMDSv2 for default-karpenter nodetemplate

szuecs:
test result:
- new EC2 nodes come up with  `.MetadataOptions.HttpTokens == "required"`
- Kubernetes does not see them, because kubelet is not started
- EC2 node is accessible via SSH/SSM, but user template is not available.


szuecs:
Closing, because of the test results

PR #6590, opened 2023-11-27
skipper: manual canary

szuecs:
skipper: manual canary

requires: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/708

TODO:
- [ ] deletions
- [ ] reduce skipper-ingress hpa minReplicas by 1
- [ ] anything missing?

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -1,16 +1,17 @@
-{{- if eq .ConfigItems.skipper_canary_enabled "true" }}
 
 {{ range json `[
   {
   "name": "skipper-ingress",
-  "internal_version": "v0.18.42-691"
+  "internal_version": "v0.18.42-691",
   "version": "v0.18.42"
+{{- if eq .ConfigItems.skipper_canary_enabled "true" }}
   },
   {
   "name": "skipper-ingress-canary",
-  "internal_version": "v0.18.53-702"
+  "internal_version": "v0.18.53-702",
   "version": "v0.18.53",
-  "replicas": 1,
+  "replicas": 1
+{{ end }}
   }
 ]` }}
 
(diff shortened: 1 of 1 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -1,29 +1,48 @@
+{{- if eq .ConfigItems.skipper_canary_enabled "true" }}
```
AlexanderYastrebov: This does not look right because it conditions whole manifest on  skipper_canary_enabled.
I think we need something like:
```
{{ if eq .ConfigItems.skipper_canary_enabled "true" }}
  {{ $deployment_config := `[two items: main and canary]` }}
{{ else }}
  {{ $deployment_config := `[one item: main]` }}
{{ end }}

{{ range json $deployment_config }}
...
```

AlexanderYastrebov: I think there might be variable scope issue with the above though.

AlexanderYastrebov: I think it would need declaration and assignments like so 
```
{{ $deployment_config := "" }}
{{ if eq .ConfigItems.skipper_canary_enabled "true" }}
  {{ $deployment_config = `[two items: main and canary]` }}
{{ else }}
  {{ $deployment_config = `[one item: main]` }}
{{ end }}

{{ range json $deployment_config }}
...
```
https://stackoverflow.com/questions/68814385/how-do-i-conditionally-declare-variables-in-go-templates

cluster/manifests/skipper/deployment.yaml
```diff
@@ -1,29 +1,48 @@
...
+
+{{ range json `[
+  {
+  "name": "skipper-ingress",
+  "internal_version": "v0.18.42-691"
```
AlexanderYastrebov: missing comma

cluster/manifests/skipper/deployment.yaml
szuecs: What about the 2nd commit?
Is this possible?

cluster/manifests/skipper/deployment.yaml
szuecs: thanks fixed

PR #6591, opened 2023-11-27
Use IMDSv2 on AMI

mikkeloscar:
Test AMI using IMDSv2 (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-metadata-v2-how-it-works.html)

related to #6589


mikkeloscar:
Closing for now, we want to roll out Ubuntu 22.04 first

PR #6592, opened 2023-11-29
skipper: enable manual canary deployment

AlexanderYastrebov:
The change introduces parameterized `skipper-ingress` deployment template.
This template is used to define `main` and `canary` deployments.
A single replica `canary` deployment is enabled by a new `skipper_ingress_canary_enabled` config item.
When `canary` deployment is enabled number of `main` HPA min replicas is reduced by one.

To deploy a canary version update `$canary_internal_version` and set `skipper_ingress_canary_enabled`.
After checking that canary behaves as expected set `$internal_version` equal to `$canary_internal_version`.

I.e. in the cluster where canary is enabled there is always one canary
instance whose version is either above or equal to the main deployment version.

All config items are accessed through `.Cluster.ConfigItems` for clarity.

https://github.com/zalando-incubator/kubernetes-on-aws/labels/architectural

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -70,6 +70,14 @@ skipper_ingress_min_replicas: "2"
 {{end}}
 skipper_ingress_cpu: "1000m"
 skipper_ingress_memory: "1500Mi"
+
+# Enables deployment of canary version
+skipper_ingress_canary_enabled: "false"
+
+# Optional canary arguments separated by "[cf724afc]" to allow whitespaces, e.g.:
+# skipper_ingress_canary_args: "-foo=has a whitespace[cf724afc]-baz=qux"
+skipper_ingress_canary_args: ""
+
 # When set to true (and dedicated node pool for skipper is also true) the
 # daemonset overhead will be subtracted from the cpu settings such
 # that skipper will perfectly fit on the node.
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -1,27 +1,53 @@
 {{ $internal_version := "v0.18.42-691" }}
-{{ $version := index (split $internal_version "-") 0 }}
+{{ $canary_internal_version := "v0.18.42-691" }}
 
+{{ template "skipper-ingress" dict
+  "name" "skipper-ingress"
+  "internal_version" $internal_version
+
+  "Cluster" .Cluster
+  "Values" .Values
+}}
+
+{{ if eq .Cluster.ConfigItems.skipper_ingress_canary_enabled "true" }}
+{{ template "skipper-ingress" dict
+  "name" "skipper-ingress-canary"
+  "internal_version" $canary_internal_version
+  "replicas" 1
+  "args" (index .Cluster.ConfigItems "skipper_ingress_canary_args")
+
+  "Cluster" .Cluster
+  "Values" .Values
+}}
+{{ end }}
+
+{{ define "skipper-ingress" }}
+{{ $version := index (split .internal_version "-") 0 }}
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: skipper-ingress
+  name: "{{ .name }}"
   namespace: kube-system
   labels:
     application: skipper-ingress
     version: {{ $version }}
     component: ingress
 spec:
+{{ if index . "replicas" }}
+  replicas: {{ .replicas }}
+{{ end }}
   strategy:
     rollingUpdate:
       maxSurge: 1
       maxUnavailable: 0
   selector:
     matchLabels:
-      deployment: skipper-ingress
+      deployment: "{{ .name }}"
   template:
     metadata:
       labels:
-        deployment: skipper-ingress
+        deployment: "{{ .name }}"
         application: skipper-ingress
         version: {{ $version }}
         component: ingress
@@ -30,15 +56,15 @@ spec:
         kubernetes-log-watcher/scalyr-parser: |
           [{"container": "skipper-ingress", "parser": "skipper-access-log"}]
         config/hash: {{"secret.yaml" | manifestHash}}
-        logging/destination: "{{.Cluster.ConfigItems.log_destination_local}}"
+        logging/destination: "{{ .Cluster.ConfigItems.log_destination_local }}"
         prometheus.io/path: /metrics
         prometheus.io/port: "9911"
         prometheus.io/scrape: "true"
-{{- if eq .ConfigItems.skipper_topology_spread_enabled "true" }}
+{{- if eq .Cluster.ConfigItems.skipper_topology_spread_enabled "true" }}
         zalando.org/topology-spread-timeout: 7m
 {{- end }}
     spec:
-{{- if eq .ConfigItems.skipper_topology_spread_enabled "true" }}
+{{- if eq .Cluster.ConfigItems.skipper_topology_spread_enabled "true" }}
       topologySpreadConstraints:
         - maxSkew: 1
           topologyKey: topology.kubernetes.io/zone
@@ -56,13 +82,13 @@ spec:
       hostNetwork: true
       containers:
       - name: skipper-ingress
-        image: container-registry.zalando.net/teapot/skipper-internal:{{ $internal_version }}
+        image: container-registry.zalando.net/teapot/skipper-internal:{{ .internal_version }}
         ports:
         - name: ingress-port
           containerPort: 9999
           hostPort: 9999
           protocol: TCP
-{{ if or (eq .ConfigItems.nlb_switch "pre") (eq .ConfigItems.nlb_switch "exec") }}
+{{ if or (eq .Cluster.ConfigItems.nlb_switch "pre") (eq .Cluster.ConfigItems.nlb_switch "exec") }}
         - name: http-redirect
           containerPort: 9998
           hostPort: 9998
@@ -74,7 +100,7 @@ spec:
             secretKeyRef:
               name: skipper-ingress
               key: lightstep-token
-{{ if or (eq .ConfigItems.skipper_local_tokeninfo "production") (eq .ConfigItems.skipper_local_tokeninfo "bridge") }}
+{{ if or (eq .Cluster.ConfigItems.skipper_local_tokeninfo "production") (eq .Cluster.ConfigItems.skipper_local_tokeninfo "bridge") }}
         - name: LOCAL_TOKENINFO
           value: "true"
         - name: ENABLE_OPENTRACING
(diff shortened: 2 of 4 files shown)


cluster/config-defaults.yaml
```diff
@@ -70,6 +70,14 @@ skipper_ingress_min_replicas: "2"
 {{end}}
 skipper_ingress_cpu: "1000m"
 skipper_ingress_memory: "1500Mi"
+
+# Non empty value enables deployment of specified canary version
+skipper_ingress_canary_internal_version: ""
+
+# Optional canary arguments separated by "[cf724afc]" to allow whitespaces, e.g.:
```
RomanZavodskikh, line 77: Quite a counterintuitive separator :-)
From my side, it looks like it is possible to find something more simple. For example, I see no `;` used in the current skipper cmdline args.
Why don't we use `;` instead of `[cf724afc]` to make it much more readable?

AlexanderYastrebov: We already use `[cf724afc]` as a separator for two other flags that may contains whitespace or any other sequence of characters. `[cf724afc]` is chosen exactly because it is unlikely to appear in the flag value.

If you are curious where it comes from, it is:
```
~$ echo -n skipper | sha256sum | head -c 8
cf724afc
```

RomanZavodskikh: Okay, fair enough. :-)

RomanZavodskikh:
:+1:

cluster/manifests/skipper/deployment.yaml
```diff
@@ -1,35 +1,57 @@
...
 
+{{ template "skipper-ingress" dict
+  "ConfigItems" .ConfigItems
+  "name" "skipper-ingress"
+  "internal_version" $internal_version
+}}
+
+{{ if .ConfigItems.skipper_ingress_canary_internal_version }}
+{{ template "skipper-ingress" dict
+  "ConfigItems" .ConfigItems
```
mikkeloscar: ConfigItems seem equal in both cases, does it need to be defined here? 🤔

AlexanderYastrebov: Each template invocation receives its own `.` as a context, i.e. there is no global scope

AlexanderYastrebov: > A variable's scope extends to the "end" action of the control structure ("if", "with", or "range") in which it is declared, or to the end of the template if there is no such control structure. A template invocation does not inherit variables from the point of its invocation.

AlexanderYastrebov: Will update to pass .Values as well

cluster/manifests/skipper/deployment.yaml
```diff
@@ -1,35 +1,61 @@
 {{ $internal_version := "v0.18.42-691" }}
 {{ $version := index (split $internal_version "-") 0 }}
```
AlexanderYastrebov: This is used by routesrv that is not templated

cluster/manifests/skipper/deployment.yaml
```diff
@@ -429,7 +461,7 @@ spec:
         config/hash: {{"secret.yaml" | manifestHash}}
         kubernetes-log-watcher/scalyr-parser: |
           [{"container": "routesrv", "parser": "keyValue"}]
-        logging/destination: "{{.Cluster.ConfigItems.log_destination_local}}"
+        logging/destination: "{{ .ConfigItems.log_destination_local }}"
```
AlexanderYastrebov: Removed .Cluster prefix everywhere for consistency

cluster/manifests/skipper/deployment.yaml
```diff
@@ -1,35 +1,61 @@
...
+  "internal_version" $internal_version
+
+  "Values" .Values
+  "ConfigItems" .ConfigItems
+}}
+
+{{ if .ConfigItems.skipper_ingress_canary_internal_version }}
+{{ template "skipper-ingress" dict
+  "name" "skipper-ingress-canary"
+  "internal_version" .ConfigItems.skipper_ingress_canary_internal_version
```
szuecs: I understand the flexibility, but do we need this?
A version update would now mean to update two files and in one file 2 entries.

cluster/manifests/skipper/deployment.yaml
```diff
@@ -1,35 +1,61 @@
...
+        deployment: "{{ .name }}"
         application: skipper-ingress
         version: {{ $version }}
         component: ingress
       annotations:
         kubernetes-log-watcher/scalyr-parser: |
           [{"container": "skipper-ingress", "parser": "skipper-access-log"}]
         config/hash: {{"secret.yaml" | manifestHash}}
-        logging/destination: "{{.Cluster.ConfigItems.log_destination_local}}"
+        logging/destination: "{{ .ConfigItems.log_destination_local }}"
```
szuecs: We wanted to migrate to always use .Cluster scope for ConfigItems as far as I know.
This is also not relevant change so please clean it up

cluster/manifests/skipper/deployment.yaml
szuecs: no please not!

cluster/manifests/skipper/deployment.yaml
```diff
@@ -1,35 +1,61 @@
 {{ $internal_version := "v0.18.42-691" }}
 {{ $version := index (split $internal_version "-") 0 }}
 
+{{ template "skipper-ingress" dict
+  "name" "skipper-ingress"
+  "internal_version" $internal_version
+
+  "Values" .Values
+  "ConfigItems" .ConfigItems
```
szuecs: better to set it to `.Cluster.ConfigItems`

cluster/manifests/skipper/deployment.yaml
AlexanderYastrebov: We need to pass all parameters as arguments that are different between main and canary deployments.
Or do you mean something else?

cluster/manifests/skipper/deployment.yaml
AlexanderYastrebov: I can change it to .Cluster.ConfigItems but it is not consistent across the manifests anyway

AlexanderYastrebov: > This is also not relevant change so please clean it up

It is because .ConfigItems is a template parameter not the global variable, that is why it is passed during template invocation as `"ConfigItems" .ConfigItems`

cluster/manifests/skipper/deployment.yaml
AlexanderYastrebov: Why not? It should be either .ConfigItems or .Cluster.ConfigItems everywhere, I just picked .ConfigItems

cluster/manifests/skipper/deployment.yaml
szuecs: I mean the canary version as config item is very flexible but also leads to having to change the version not in the deployment.yaml, but in config-defaults and we can change it per cluster, which we might not want.

cluster/manifests/skipper/deployment.yaml
szuecs: true that makes sense. Then I would suggest to use something else as name like `CfgItems`, to be explicit that we are scoped.
wdyt?

cluster/manifests/skipper/deployment.yaml
szuecs: Let's keep the discussion in the other comment

cluster/manifests/skipper/deployment.yaml
```diff
@@ -1,43 +1,69 @@
...
+  "internal_version" $internal_version
+
+  "Cluster" .Cluster
+  "Values" .Values
+}}
+
+{{ if .Cluster.ConfigItems.skipper_ingress_canary_internal_version }}
+{{ template "skipper-ingress" dict
+  "name" "skipper-ingress-canary"
+  "internal_version" .Cluster.ConfigItems.skipper_ingress_canary_internal_version
```
szuecs: config-item skipper_ingress_canary_internal_version TBD. I would really like to have a fixed version.
I would suggest to have only a `skipper_ingress_canary_enable: "true"` cfg item to allow enable and delete, but a fixed version until we really need a dynamic settable config item version.
One safety net is that we need to roll through all channels and this would enable us to switch bypassing all channels.

cluster/manifests/skipper/deployment.yaml
AlexanderYastrebov: Changed to boolean config.

cluster/manifests/skipper/deployment.yaml
AlexanderYastrebov: Changed to boolean config.

cluster/config-defaults.yaml
```diff
@@ -70,6 +70,14 @@ skipper_ingress_min_replicas: "2"
 {{end}}
 skipper_ingress_cpu: "1000m"
 skipper_ingress_memory: "1500Mi"
+
+# Enables deployment of canary version
+skipper_ingress_canary_enabled: "false"
```
szuecs, line 75: Let's enable it for all non production clusters!
```
skipper_ingress_canary_enabled: "false"
{{else}}
skipper_ingress_canary_enabled: "true"
{{end}}
```

AlexanderYastrebov: Lets do this via cluster registry, then we can put it under condition

cluster/manifests/skipper/deployment.yaml
AlexanderYastrebov: Moved into routesrv block

cluster/manifests/skipper/deployment.yaml
```diff
@@ -193,122 +219,127 @@ spec:
...
+{{ end }}
+{{ if index . "args" }}
+          {{ range $arg := split .args "[cf724afc]" }}
+          - "{{ $arg }}"
+          {{ end }}
 {{ end }}
         resources:
 {{ if and (eq .Cluster.ConfigItems.enable_dedicate_nodepool_skipper "true") (eq .Cluster.ConfigItems.skipper_ingress_binpack "true") }}
-{{ $cpu_requests := sumQuantities .Cluster.ConfigItems.skipper_ingress_cpu (printf "-%s"  .Cluster.ConfigItems.teapot_admission_controller_daemonset_reserved_cpu) (printf "-%s" .Cluster.ConfigItems.kubelet_system_reserved_cpu) (printf "-%s" .Cluster.ConfigItems.kubelet_kube_reserved_cpu) }}
+{{ $cpu_requests := sumQuantities .Cluster.ConfigItems.skipper_ingress_cpu (printf "-%s" .Cluster.ConfigItems.teapot_admission_controller_daemonset_reserved_cpu) (printf "-%s" .Cluster.ConfigItems.kubelet_system_reserved_cpu) (printf "-%s" .Cluster.ConfigItems.kubelet_kube_reserved_cpu) }}
```
szuecs, line 309: Oh I wasn't aware of that. So we can actually just set the configuration item to the resource of the node and this calculates the resources that can be used by skipper-ingress.

szuecs:
:+1:

katyanna:
👍

cluster/manifests/skipper/deployment.yaml
mikkeloscar: This is only with `skipper_ingress_binpack=true` but this didn't work 100% so not enabled in most clusters yet. I was hoping to look into this with Karpenter as there seem to be a miscalculation with Cluster Autoscaler when I investigated it months ago.

PR #6593, opened 2023-11-29
Add kubelet-summary-metrics proxy

mikkeloscar:
Add a simple proxy that sits between kubelets and prometheus and translates the kubelet stats summary json into prometheus metric data.

The motivation for this is to provide per pod/container filesystem usage metrics which are not provided by cadvisor. Eventually more metric types could be provided this way and reduce the need for cadvisor (pending evaluation).

The architecture looks like this:

image

The benefit compared to cadvisor is that it runs centrally in the cluster instead of a daemonset and thereby can scale horizontally and not take up per node resources.


linki:
👍

mikkeloscar:
:+1:

PR #6604, opened 2023-12-04
set client.uid tag to caller uid

szuecs:
set client.uid tag to caller uid

- in case statebag entry exists it will set client.uid to uid of the statebag entry 
- if it does not exit in statebag the tag will not be created.

In case we need to rollback this change without rolling back the PR as incident mitigation:
Override the config-item  `skipper_default_filters_append` with value `dropResponseHeader("does-not-exist")`

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -101,6 +101,7 @@ skipper_validate_query: "true"
 skipper_validate_query_log: "false"
 
 skipper_default_filters: 'disableAccessLog(2,3,404,429) -> fifo(2000,20,"1s")'
+skipper_default_filters_append: 'stateBagToTag("auth-user", "client.uid")'
 skipper_disabled_filters: "lua,static,bearerinjector"
 skipper_edit_route_placeholders: ""
 skipper_ingress_inline_routes: ""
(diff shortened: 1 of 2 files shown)


cluster/config-defaults.yaml
```diff
@@ -93,6 +93,7 @@ skipper_validate_query: "true"
 skipper_validate_query_log: "false"
 
 skipper_default_filters: 'disableAccessLog(2,3,404,429) -> fifo(2000,20,"1s")'
+skipper_default_filters_append: 'stateBagToTag("uid", "client.uid")'
```
AlexanderYastrebov: I believe the state bag key is `auth-user` as described in https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#statebagtotag

Also, do we have other meanings for "uid"? I think no, so maybe we should just use it as a tag name: `stateBagToTag("auth-user", "uid") `

szuecs: I used https://github.com/zalando/skipper/blob/master/filters/auth/tokeninfo.go#L369 , which is of course only the parsed in-memory object, not stateBag. 
Thanks for pointing it out. I see in auth/auth.go on authorized() we actually set  "auth-user" stateBag as documented.

szuecs: > Also, do we have other meanings for "uid" - maybe we should just use it as a tag name: `stateBagToTag("auth-user", "uid") `

I don't know what users would think of "uid" tag. To be more specific I used the client.uid, because it's whatever the client sent to us.

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
We may want to wait until https://github.com/zalando/skipper/pull/2775

szuecs:
At this rate it does not really matter, but of course nice 10x optimization.

szuecs:
@AlexanderYastrebov I guess we need to wait for CLM fix

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #6607, opened 2023-12-05
Make nvidia device plugin ds run on Karpenter nodes

mikkeloscar:
* Adjusts the nodeAffinity so the nvidia device will also be scheduled on karpenter managed nvidia nodes
* Updates the nvidia device plugin to latest version
* This also removes the `version` label which was outdated and not used anyway.

### TODO:

* [x] Check if this is the best way to schedule on karpenter nodes
* [x] Update nvidia container


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6610, opened 2023-12-05
Update GPU instance types for e2e

mikkeloscar:
Update the set of GPU instance types used for e2e to include latest gen and remove the old p2 gen type.


linki:
👍

mikkeloscar:
👍

PR #6616, opened 2023-12-06
Update AMI to use default value for max_user_watches

linki:
Updates to AMI that uses the default value for `fs.inotify.max_user_watches`. More details in: https://github.bus.zalan.do/teapot/kubernetes-on-ubuntu/pull/312

For https://github.bus.zalan.do/zooport/issues/issues/4130

Note: combine with switch to Jammy in prodcution.


linki:
👍

mikkeloscar:
👍

PR #6624, opened 2023-12-08
skipper: remove skipper_ingress_canary_args config item

AlexanderYastrebov:
As we decided to not have a config item for canary version value it also does not make sense to have an item for canary args.

This change uses template variable for canary args to keep canary version and args close together.

Followup on #6592


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #6606, opened 2023-12-05
Update admission-controller with configmap protection feature

linki:
Update admission controller including [this feature](https://github.bus.zalan.do/teapot/admission-controller/pull/204) and enable it by default.

PR #6628, opened 2023-12-08
register configmap deletion events with admission webhook

linki:
We need to register the tracked event, otherwise nothing will happen.

Follow up from: https://github.com/zalando-incubator/kubernetes-on-aws/pull/6606


linki:
👍

mikkeloscar:
:+1:

linki:
👍

PR #6629, opened 2023-12-08
Create deployment event on merge to channel branches

mikkeloscar:
This introduces a "fake" deployment event on every merge to a channel branch. The purpose is to generate a CDP deployment event which can later be shown e.g. in Sunrise. This is similar to how Linus does for AILM.

See internal ticket: teapot/issues#3528 for more context

### DIFF
--- a/_deploy_event/configmap.yaml
+++ b/_deploy_event/configmap.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: "kubernetes-deploy-event"
+  labels:
+    application: "kubernetes"
+    component: "deploy-event"
+data:
+  empty-event: "deployment"
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -312,3 +312,16 @@ pipeline:
             requests:
               cpu: 500m
               memory: 1Gi
+
+- id: deploy-event
+  type: deploy
+  target: stups
+  when:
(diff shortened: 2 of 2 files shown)


mikkeloscar:
:+1:

_deploy_event/configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: "{{{APPLICATION}}}-{{{COMPONENT}}}"
```
AlexanderYastrebov: Any reason to not hardcode these values?

AlexanderYastrebov:
:+1:

_deploy_event/configmap.yaml
mikkeloscar: No particular reason. Changed it.

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

_deploy_event/configmap.yaml
AlexanderYastrebov: Hm, maybe we need APPLICATION var in delivery.yaml for CDP to figure out application...

mikkeloscar: Good point, I didn't actually validate what the event looks like :facepalm: Let me double check it.

PR #6630, opened 2023-12-11
Update AMI, max_user_instances, nvidia driver

mikkeloscar:
This updates the 22.04 based AMI with the following enhancements:

* Increase `fs.inotify.max_user_instances` to `8192`. This is needed because containerd uses more watches under cgroup v2 and there is not enough for other apps on nodes with many containers with the default limit of `128`.
* Upgrade nvidia-driver-510 => nvidia-driver-535


linki:
👍

mikkeloscar:
:+1:

PR #6639, opened 2023-12-12
Avoid overwrite for deployment event resource

mikkeloscar:
Follow up to #6629 which adds the branch name to the resource to avoid overwrite warnings on all PRs like this:

image


mikkeloscar:
:+1:

linki:
👍

PR #6640, opened 2023-12-12
Set DeletionPolicy to retain for deplyoment secret kms key

Shinzu:
Preparation to move kms key/alias creation to AILM


mikkeloscar:
@Shinzu Since we only can run e2e if the PR is created from a local branch I have take your commit and opened another PR here: https://github.com/zalando-incubator/kubernetes-on-aws/pull/6641

Will close this one in favor of the other one.

PR #6641, opened 2023-12-12
Set DeletionPolicy to retain for deplyoment secret kms key

mikkeloscar:
Preparation to move kms key/alias creation to AILM

This will allow us to remove the resources from the cluster.yaml stack while retaining them and importing in AILM config later on.

#6642 will enable the actual deletion from stack per cluster.

Opened as a local branch in favor of #6640


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6642, opened 2023-12-12
Allow disabling deployment secret key management in cluster stack

mikkeloscar:
Related to #6641 this allows removing the Deployment Secret key from the cluster stack via a config-item: `deployment_secret_key_managed`.

This will allow us to slowly migrate the KMS key to central management cluster by cluster.

To be 100% safe, this should only be merged once #6641 is fully rolled out.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6644, opened 2023-12-13
Also set UpdateReplacePolicy

Shinzu:
this is also needed to retain resource on update

```
cfn-lint: Both UpdateReplacePolicy and DeletionPolicy are needed to protect Resources/DeploymentSecretKey from deletion
```


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6654, opened 2023-12-14
Increase CPU resources for kubelet-summary-proxy

mikkeloscar:
I was a bit too optimistic setting `10m` as default. With `100m` only two pods are needed in a big cluster with `250` nodes, so that seems like a reasonable value.

ref: #6593


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6661, opened 2023-12-15
Update StackSet controller to support traffic segments.

gargravarr:
Summary of the changes here https://github.com/zalando-incubator/stackset-controller/pull/495

We control support of traffic segments via 2 config items:

* `stackset_enable_traffic_segments`: when set, enables traffic segment for all new and current stacksets in a cluster.
* `stackset_annotated_traffic_segments`: when set alongside `stackset_enable_traffic_segments`, enables traffic segments only in stacksets (new and current) conatining the special annotation `"stackset-controller.zalando.org/use-traffic-segments: true"` in their manifest.

By default, both config items are unset. The cluster for end-2-end tests will have both config items enabled, because stackset end 2 end tests already cover traffic segments.


mikkeloscar:
:+1:

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6664, opened 2023-12-15
Fix deployment event

mikkeloscar:
Follow up to #6629 

Use template variables so CDP can detect the application for which the event is relevant.


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #6665, opened 2023-12-15
Fix application association for deploy-event

mikkeloscar:
Follow up to #6664 and #6629

The other PRs did not work because CDP could not associate an application with the deployment. This is because the application is derived from labels/tags on the resources and configmap is not supported for deriving application.

This changes to use a CF stack instead of configmap which is supported.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6668, opened 2023-12-18
Only run kube2iam on worker nodes

mikkeloscar:
Run kube2iam only on worker nodes. This is motivated by two things:

1. Reduce the resources overhead on master nodes, relevant for very small master nodes in test clusters
2. There is no components depending on kube2iam on master nodes, so no reason to run it there.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6669, opened 2023-12-18
Force imdsv2 (master nodes)

mikkeloscar:
This adds an AMI which only uses [imdsv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) and it forces the use of imdsv2 on master nodes.

### DIFF
--- a/cluster/node-pools/master-default/stack.yaml
+++ b/cluster/node-pools/master-default/stack.yaml
@@ -49,6 +49,8 @@ Resources:
     Properties:
       LaunchTemplateName: '{{.Cluster.LocalID}}-{{ .NodePool.Name }}'
       LaunchTemplateData:
+        MetadataOptions:
+          HttpTokens: required
         TagSpecifications:
         - ResourceType: "volume"
           Tags:
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -709,10 +709,8 @@ tracing_coredns_local_zone_traces_endpoint: ""
 # AMI id given the image name and the Image AWS account owner.
 #
 # [0]: https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/8a9bd1cb2d094038a9e23e646421f8146b48886a/provisioner/template.go#L116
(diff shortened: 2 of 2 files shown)


mikkeloscar:
the secret encrypter on the apiserver is failing with IMDSv2 enforced.

mikkeloscar:
After #6668 kube-node-ready will also not work because it uses the metadata service to get region and instance ID.

cluster/node-pools/master-default/stack.yaml
```diff
@@ -49,6 +49,9 @@ Resources:
     Properties:
       LaunchTemplateName: '{{.Cluster.LocalID}}-{{ .NodePool.Name }}'
       LaunchTemplateData:
+        MetadataOptions:
+          HttpTokens: required
+          HttpPutResponseHopLimit: 2
```
szuecs: see also https://github.com/aws/elastic-beanstalk-roadmap/issues/242 for example

mikkeloscar: The idea was to _prevent_ that pods can call the metadata endpoint to avoid that they can get instance credentials, that's why I didn't set this to leave it at `1`. This currently does not work because kube-node-ready depend on metadata service, but this would give it access to the instance credentials from my understanding which is not intended.

My idea to make this work was rather to avoid `kube-node-ready` depending on metadata service (or not run kube-node-ready at all because of the questionable benefit nowadays).

szuecs: I thought that the delegation requires the hop limit to set to 2 anyways, but the definition is not clear to me (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-metadataoptions.html):

> HttpPutResponseHopLimit
    The desired HTTP PUT response hop limit for instance metadata requests. The larger the number, the further instance metadata requests can travel.

Article https://aws.amazon.com/blogs/security/get-the-full-benefits-of-imdsv2-and-disable-imdsv1-across-your-aws-infrastructure/

linki:
👍

szuecs:
:+1:

PR #6670, opened 2023-12-18
Fix reference to deploy key

mikkeloscar:
Follow up to #6641 and #6642 remove reference to cluster managed key if the key is not managed in cluster.yaml

When the key is managed account wide we give deployment-service access based on alias: https://docs.aws.amazon.com/kms/latest/developerguide/alias-authorization.html


mikkeloscar:
:+1:

demonCoder95:
👍

PR #6673, opened 2023-12-18
Only run ebs-csi-node on worker nodes

mikkeloscar:
Similar to #6668 we don't need volumes on master nodes, so only run ebs-csi-node daemonset on worker nodes to save some CPU resources on small master nodes.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #6676, opened 2023-12-18
Pass region to karpenter

mikkeloscar:
Pass region to Karpenter to avoid karpenter needing to call the ec2 instance metadata endpoint which we intend to block in the future.

Related to #6669


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6680, opened 2023-12-18
Revert "Only run kube2iam on worker nodes"

mikkeloscar:
Reverts zalando-incubator/kubernetes-on-aws#6668

Does not work because karpenter and kube-node-ready currently depend on EC2 metadata service endpoint


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #6683, opened 2023-12-19
admission-controller: Enhance daemonset resource check

mikkeloscar:
This adjusts the logic for selecting daemonsets in scope of the daemonset resource limit to _not_ exclude those with `node.kubernetes.io/role=worker` nodeSelector.

With https://github.com/zalando-incubator/kubernetes-on-aws/pull/6673 and https://github.com/zalando-incubator/kubernetes-on-aws/pull/6668 we scoped the daemonsets to _only_ run on worker nodes, however since the logic for checking daemonset resources excludes any daemonset with a nodeSelector, those are no longer in scope although they run on _all_ worker nodes, where the daemonset resource check is the most important.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6686, opened 2023-12-19
Only use alias when key centrally managed

mikkeloscar:
Follow up to #6670 only set decrypt-kms-alias-arn when the key is managed _outside_ the stack, otherwise the IAM role does not match.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6691, opened 2023-12-19
Add field for controlling traffic segment enablement.

gargravarr:
Adds the field `probe-for-traffic-segments` in the deployment-service configuration. Both `cdp-steps` and `deployment-service` will use this field to control enabling/diasbling traffic segment probling.

We use the already existing config item that the StackSet controller uses to control this feature.


gargravarr:
:+1:

mikkeloscar:
:+1:

PR #6709, opened 2024-01-02
Disable legacy APIs by default

mikkeloscar:
This disables legacy APIs by default which will no longer be available in Kubernetes v1.25 or v1.26.

It's implemented as a config-item so we can easily enable it for some clusters still not fully migrated away from using legacy APIs.

Note: ideally we also wanted to disable `policy/v1beta1` but this is not possible as the apiserver itself uses `policy/v1beta1` for setting up watches for PDBs. So this will be disabled _with_ Kubernetes v1.25

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -127,7 +127,7 @@ write_files:
           {{- end }}
           - --tls-cert-file=/etc/kubernetes/ssl/apiserver.pem
           - --tls-private-key-file=/etc/kubernetes/ssl/apiserver-key.pem
-          - --runtime-config=policy/v1beta1=true,authorization.k8s.io/v1beta1=true,scheduling.k8s.io/v1alpha1=true,admissionregistration.k8s.io/v1beta1=true,autoscaling/v2beta2={{ .Cluster.ConfigItems.autoscaling_v2beta2_enabled }},autoscaling/v2beta1={{ .Cluster.ConfigItems.autoscaling_v2beta1_enabled }}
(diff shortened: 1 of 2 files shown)


szuecs:
@mikkeloscar this will drop psp support, no?

```
kubectl get psp -o yaml
...
- apiVersion: policy/v1beta1
  kind: PodSecurityPolicy
  metadata:
```

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -127,7 +127,7 @@ write_files:
           {{- end }}
           - --tls-cert-file=/etc/kubernetes/ssl/apiserver.pem
           - --tls-private-key-file=/etc/kubernetes/ssl/apiserver-key.pem
-          - --runtime-config=policy/v1beta1=true,authorization.k8s.io/v1beta1=true,scheduling.k8s.io/v1alpha1=true,admissionregistration.k8s.io/v1beta1=true,autoscaling/v2beta2={{ .Cluster.ConfigItems.autoscaling_v2beta2_enabled }},autoscaling/v2beta1={{ .Cluster.ConfigItems.autoscaling_v2beta1_enabled }}
```
szuecs, line 130: This delete `--runtime-config=policy/v1beta1=true,` will likely drop psp support

mikkeloscar: It doesn't seem to do any difference as it still works without it listed there (I tested with a pet cluster). But I added it back now as it was not the intention to remove it.

I originally had a flag which set the value to `false` which did not work, and then I removed the whole thing by accident.

mikkeloscar:
> this will drop psp support, no?

Fixed the commit to not delete the `policy/v1beta1=true` config

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6710, opened 2024-01-02
deploy: use test cluster for deploy-event

AlexanderYastrebov:
Use test cluster to avoid Access denied error during deploy-event step.

Followup on #6629

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -315,7 +315,7 @@ pipeline:
 
 - id: deploy-event
   type: deploy
-  target: stups
+  target: stups-test
   when:
     event: push
     branch:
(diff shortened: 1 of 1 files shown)


delivery.yaml
```diff
@@ -324,4 +324,4 @@ pipeline:
     - beta
     - stable
   sources:
-    - dir: _deploy_event
+  - dir: _deploy_event
```
AlexanderYastrebov, line 327: Fixes CDP yaml linter warning

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #6712, opened 2024-01-02
Tag instance root volumes with application/component

mikkeloscar:
Add application/component tag to root volumes when instances are launched. This enables better allocation of cost for volumes.

## TODO:

* [x] Tags for karpenter instances.


ePaul:
:+1: (The change looks reasonable, though I'm not an expert in the context.)

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6713, opened 2024-01-02
Don't create metrics with zero timestamp

mikkeloscar:
Fixes a bug in kubelet-summary-metrics where it would expose metrics with zero timestamps which is not valid in systems consuming the metrics.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6716, opened 2024-01-03
Only tag worker node volumes when updating to jammy

mikkeloscar:
This is a follow up to #6712 to only add volume tags to worker nodes when rotating the nodes for the Ubuntu update. Otherwise the nodes will rotate because of the LaunchTemplate change which was not desired to do right now.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6723, opened 2024-01-03
fix: skipper-ingress route lookup edge case with fabric and stackset 1/2

szuecs:
fix: skipper-ingress route lookup edge case with fabric and stackset

step 1 canary only

https://github.com/zalando/skipper/compare/v0.18.82...v0.18.92


AlexanderYastrebov:
:+1:

szuecs:
e2e loadtest did not run, but I assume in the next iterations it will be successful. We don't change kube-ingress-aws-controller, which is the application that we test by the e2e loadtest.

szuecs:
:+1:

szuecs:
this looks good alpha channel

PR #6728, opened 2024-01-04
Expose more metrics via kubelet-summary-metrics

mikkeloscar:
Extends the proxy to expose the following additional metric types:

* CPU
* Memory
* Network

PR #6731, opened 2024-01-05
Add support for kubelet-cadvisor

mikkeloscar:
Related to #6728 

Currently we run cadvisor as a daemonset which takes up a lot of resources per node. With #6728 we work on migrating some of the main cadvisor metrics (CPU, Memory, Network) to the kubelet-summary-metrics proxy with the idea of eventually getting rid of cadvisor daemonset.

a few metrics that we depend on today are not supported by the kubelet-summary api, namely: `container_threads|container_file_descriptors`.
However kubelet _also_ has a cadvisor endpoint where it exposes the same metrics as the cadvisor daemonset, so we can simply get those few metrics from the kubelet instead of running a dedicated daemonset and collecting the same metrics twice.

This PR adds scraping of the kubelet cadvisor endpoint and only keep the metrics we care about: `container_threads|container_file_descriptors`. Scraping is enabled by default but can be turned off with a config-item.
Additionally it adds a config-item to turn of the cadvisor daemonset container so we can do it per cluster. This is still enabled by default as we need a period where we run both in order to safely migrate the metric collectors (ZMON/Otel collector).

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -333,6 +333,7 @@ event_rate_limit_config_burst: "1000"
 cadvisor_cpu: "150m"
 cadvisor_memory: "150Mi"
 cadvisor_profiling_enabled: "false"
+cadvisor_enabled: "true"
 
 # settings for enabling the kubelet-summary-metrics proxy and prometheus metric
 # collection.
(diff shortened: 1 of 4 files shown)


mikkeloscar:
:+1:

szuecs:
:+1:

PR #6732, opened 2024-01-05
Implement PSP in admission-controller

mikkeloscar:
This is an implementation of the proposed solution for migrating away from PSP ahead of v1.25 roll out.

As of Kubernetes v1.25 the native PodSecurityPolicy feature is no longer available. The admission-controller implements the small supset of PodSecurityPolicy rules which we depend on in Zalando.

`pod.pod-security-policy.enable: "true|false"` enables/disables the feature.

`pod.pod-security-policy.privileged-service-accounts.<service-account>: ""` allows to define service accounts which can use privileged features. The Service accounts are defined as `<namespace>_<service-account>`. It would normally just be for components running in `kube-system`/`visibility` namespace, but we could extend it via config-item if needed.

`pod.pod-security-policy.allowed-unsafe-sysctls.<sysctl>:` define a list of unsafe sysctls that are allowed to be defined.

`pod.pod-security-policy.allow-privilege-escalation: "true|false"` allow setting `allowPrivilegeEscalation: true` on pods in a cluster.

I have set:

```
teapot_admission_controller_pod_security_policy_enabled=false
pod_security_policy_enabled=true
```

In all clusters so merging this will not enable the feature, but we can do it per cluster/channel.

Changed the e2e tests in `test/e2e/psp.go` to test the admission-controller implementation. This also made it simpler to test the case where a restricted pod fails to use privileged features. A test we didn't have before :)


mikkeloscar:
:+1:

mikkeloscar:
e2e makes assumptions about PSP which needs to be addressed before this is ready.

mikkeloscar:
👍

cluster/manifests/psp/pod_security_policy.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Cluster.ConfigItems.pod_security_policy_enabled "true" }}
```
gargravarr, line 1: I don't understand this. The Pull Request disables by default PSPs via admission controller. But this change does _not_ apply by default this manifest. Can you clarify?

mikkeloscar: From the PR description:

---- 

I have set:

```
teapot_admission_controller_pod_security_policy_enabled=false
pod_security_policy_enabled=true
```

In all clusters so merging this will not enable the feature, but we can do it per cluster/channel.

----

So the idea was to remove the config-items per cluster in order to activate this (less cleanup afterwards).

If this is too confusing then I can also flip the logic to set the config-items to enable the new feature.

gargravarr: What it confuses me is the default of `pod_security_policy_enabled` is `false`: https://github.com/zalando-incubator/kubernetes-on-aws/pull/6732/files#diff-d2b7fb562825d0d0a4261e91eb9f214a1fefbc247b40666e0b976213473389c5R626

But I confirmed in the registry that the clusters have this config item set to `true`.

linki:
LGTM 👍

cluster/manifests/psp/pod_security_policy.yaml
linki: I think it's fine either way, i.e. the logic is correct: If `pod_security_policy_enabled: true` the PSP manifests _are_ applied and _not_ deleted, whereas when `pod_security_policy_enabled: false` the PSP manifest _are not_ applied and _are_ deleted.

We just need to remember to _delete_ the config item to enable this feature, rather than _set_ it.

gargravarr:
:+1:

PR #6747, opened 2024-01-09
Update Audittrail to allow Nakadi functionality

zaklawrencea:
This PR updates the Audittrail adapter version. The new version includes functionality which allows the adapter to send events to Nakadi. To enable this, a new config item `audittrail_nakadi_url` has been introduced which will be passed to the adapter to set the target Nakadi URL. In addition, the PlatformCredentialsSet for the adapter has been updated to allow write permission to Nakadi event streams.


cluster/manifests/audittrail-adapter/credentials.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Cluster.ConfigItems.audittrail_url }}
+{{- if or (ne .Cluster.ConfigItems.audittrail_url "") (ne .Cluster.ConfigItems.audittrail_nakadi_url "") }}
```
linki: Would that also work?
```
{{- if or .Cluster.ConfigItems.audittrail_url .Cluster.ConfigItems.audittrail_nakadi_url }}
```
Looking at the old version, empty string seems to be `false`.

zaklawrencea: Yes, that should also work (not 100% sure if the parentheses are still required for the `or` though, but I don't think so without the strict comparison).

szuecs:
if it's ready please drop the do-not-merge label, but I always check all PRs that have not the label :)

cluster/manifests/audittrail-adapter/credentials.yaml
zaklawrencea: Updated to use the simplified formatting.

zaklawrencea:
:+1:

mikkeloscar:
:+1:

PR #6754, opened 2024-01-09
prometheus: enable OAuth Grant flow for configured users

AlexanderYastrebov:
#0

### DIFF
--- a/cluster/manifests/prometheus/ingress.yaml
+++ b/cluster/manifests/prometheus/ingress.yaml
@@ -1,3 +1,4 @@
+# TODO: Remove
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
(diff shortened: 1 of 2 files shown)


cluster/manifests/prometheus/ingress.yaml
```diff
@@ -1,3 +1,4 @@
+# TODO: Remove
```
szuecs, line 1: ^

AlexanderYastrebov: I want to deploy RouteGroup and remove Ingress in the followup PR such that there is no down time (couple of seconds possible between deletion and Skipper discovering route group).

szuecs: There is nothing that depend on it that can not deal with a downtime, but if you want to make it 0-downtime it's also fine.

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #6756, opened 2024-01-10
etcd: move scalyr key to files.yaml

mikkeloscar:
This change fixes a bug where the userdata of the etcd launchtemplate always change when the etcd stack is rendered which causes a new launch template version leading to a need for rotating etcds more often than needed.

The issue happens because the scalyr key is encrypted and the encrypted value is stored in the userdata. The encrypted value is _always_ different although the underlying key doesn't change.

The fix is to move the scalyr config file to the files.yaml such that the encrypted value will be stored on s3 and pulled down at etcd instance startup like what happens for the certificate files. This way the userdata stays the same and launchtemplate doesn't change either.

Since we don't have easy access to the stackname in the files.yaml we remove the stack_name key from the scalyr config and instead pass `cluster_alias`. The stack is legacy from STUPS days and not very meaningful when querying in scalyr.

### DIFF
--- a/cluster/etcd/stack.yaml
+++ b/cluster/etcd/stack.yaml
@@ -51,11 +51,6 @@ Resources:
           Fn::Base64: !Sub |
             #cloud-config
             write_files:
-              - path: /etc/scalyr-agent-2/userdata.yaml
-                permissions: 0644
-                content: |
-                  scalyr_api_key: "{{.Values.etcd_scalyr_key}}"
-                  stack_name: ${AWS::StackName}
               - path: /etc/default/etcd
                 permissions: 0644
                 content: |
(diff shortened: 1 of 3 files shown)


cluster/etcd/stack.yaml
```diff
@@ -51,11 +51,6 @@ Resources:
           Fn::Base64: !Sub |
             #cloud-config
             write_files:
-              - path: /etc/scalyr-agent-2/userdata.yaml
-                permissions: 0644
-                content: |
-                  scalyr_api_key: "{{.Values.etcd_scalyr_key}}"
-                  stack_name: ${AWS::StackName}
```
gargravarr, line 58: I think we should delete this only after we switch to the new etcd AMI (the current etcd AMI still uses stack_name). Then we can do an additional PR to delete.

mikkeloscar: But the file path will be overwritten by the one defined in files.yaml (or it will fail if the path already exists, not sure).

In either case I don't see the benefit of keeping it, and requiring an additional PR?

mikkeloscar: I will update this PR to include the AMI change so it's done in one change.

gargravarr: You're right, it's the same file :facepalm:

mikkeloscar:
:+1:

demonCoder95:
👍

PR #6764, opened 2024-01-11
Update ensure-apiserver for optional ipv6 support

mikkeloscar:
Updates the ensure-apiserver container to a version that can handle apiservers with IPv6 addresses. This is not relevant on the short term but can be helpful in the future. As long as we have IPv4 it will continue to work as usual.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #6803, opened 2024-01-18
Synchronize annotation feature for StackSet.

gargravarr:
This Pull Request updates the Stackset Controller to a [new version](https://github.com/zalando-incubator/stackset-controller/releases/tag/v1.4.26), supporting ingress annotation synchronization among Ingress and/or RouteGroup segments.

The update configures the StackSet Controller to synchronize almost all annotations from [kube-ingress-aws-controller](https://github.com/zalando-incubator/kube-ingress-aws-controller/tree/master?tab=readme-ov-file#annotations). I opted to not synchronize `zalando.org/aws-load-balancer-shared`, because:
* When set to `true`, kube-ingress-aws-controller would create dedicated Load Balancers for _all_ stacks, which is a waste of resources. We also observed during testing that the Load Balancer where the requests go would be non-deterministic.
* Without synchronizing, the StackSet will use the Load Balancer from the oldest stack, which anyway won't break access to the application.    

This feature is only relevant when we enable traffic segments: flags `--enable-traffic-segments` and `--annotated-traffic-segments`.

Additionally, I moved the configuration for end-2-end testing to config.defaults file, which is easier to find.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -888,8 +888,13 @@ stackset_configmap_support_enabled: "false"
 {{end}}
 
 # enable/disable traffic segment support for stackset
+{{if eq .Cluster.Environment "e2e"}}
+stackset_enable_traffic_segments: "true"
+stackset_annotated_traffic_segments: "true"
+{{else}}
 stackset_enable_traffic_segments: "false"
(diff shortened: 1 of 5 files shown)


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6814, opened 2024-01-22
Revert faulty etcd AMI on dev.

gargravarr:
We already know that the AMI with ID "zalando-ubuntu-etcd-production-v3.5.9-amd64-main-27" fails when creating the ETCD cluster. Let's revert this change to `dev` as well, so `playground` gets created.


linki:
👍

mikkeloscar:
:+1:

PR #4923, opened 2022-01-24
Fix go templating in CoreDNS config

mikkeloscar:
Follow up to #4921 fixing a small issue with the template scoping.

PR #6818, opened 2024-01-22
coredns-local: fix variable scope

AlexanderYastrebov:
Follow up on #6808

See also #4923

### DIFF
--- a/cluster/manifests/coredns-local/configmap-local.yaml
+++ b/cluster/manifests/coredns-local/configmap-local.yaml
@@ -89,7 +89,7 @@ data:
         errors
         {{ if eq .Cluster.ConfigItems.tracing_coredns_route_traces_to_local_zone "true"}}
           {{- with $cluster := .Cluster }}
-          {{ range $src := split .Cluster.ConfigItems.tracing_coredns_global_traces_endpoint  "," }}
+          {{ range $src := split $cluster.ConfigItems.tracing_coredns_global_traces_endpoint  "," }}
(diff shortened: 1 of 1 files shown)


cluster/manifests/coredns-local/configmap-local.yaml
```diff
@@ -89,7 +89,7 @@ data:
         errors
         {{ if eq .Cluster.ConfigItems.tracing_coredns_route_traces_to_local_zone "true"}}
           {{- with $cluster := .Cluster }}
-          {{ range $src := split .Cluster.ConfigItems.tracing_coredns_global_traces_endpoint  "," }}
```
AlexanderYastrebov, line 92: Before #6808 template was
```
          {{- with $cluster := .Cluster }}
          {{ range $src := split .ConfigItems.tracing_coredns_global_traces_endpoint  "," }}
        rewrite name exact {{ $src }} {{ $cluster.ConfigItems.tracing_coredns_local_zone_traces_endpoint }}
          {{ end }}
          {{- end }}
```
The dot `.` in the range is actually `$cluster` due to `with` therefore CLM failed with
```
error rendering template main/cluster/manifests/coredns-local/configmap-local.yaml: template: main/cluster/manifests/coredns-local/configmap-local.yaml:92:41: executing "main/cluster/manifests/coredns-local/configmap-local.yaml" at <.Cluster.ConfigItems.tracing_coredns_global_traces_endpoint>: can't evaluate field Cluster in type *api.Cluster
```
after addition of `.Cluster` prefix.

gargravarr:
:+1:

AlexanderYastrebov:
:+1:

PR #6819, opened 2024-01-22
all: add .Values prefix

AlexanderYastrebov:
Add .Values prefix for .UserData and .S3GeneratedFilesPath access.

See previous #6808

### DIFF
--- a/cluster/node-pools/worker-splitaz/stack.yaml
+++ b/cluster/node-pools/worker-splitaz/stack.yaml
@@ -135,7 +135,7 @@ Resources:
 {{ end }}
   LaunchTemplate:
     Properties:
-      LaunchTemplateName: '{{ $data.Cluster.LocalID }}-{{ $data.NodePool.Name }}'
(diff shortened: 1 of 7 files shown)


cluster/node-pools/worker-splitaz/stack.yaml
```diff
@@ -135,7 +135,7 @@ Resources:
 {{ end }}
   LaunchTemplate:
     Properties:
-      LaunchTemplateName: '{{ $data.Cluster.LocalID }}-{{ $data.NodePool.Name }}'
```
AlexanderYastrebov, line 138: Removed `$data` for consistency.

Here `$data` is equivalent to dot `.`. Note also `{{- if eq .Cluster.ConfigItems.kuberuntu_distro_worker "jammy" }}` below that does not use `$data`

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #6820, opened 2024-01-23
Allow deployment service role to decrypt any secret.

gargravarr:
Allows the deployment service to decrypt any secret. The Pull Request puts this capability behind a feature flag, enabled by default. We will disable per cluster after AILM finishes apply the new deployment key.


mikkeloscar:
:+1:

Shinzu:
:+1:

gargravarr:
:+1:

PR #6828, opened 2024-01-24
bump deployment-service version

tcondeixa:
summary
- use the correct domain for validations response
- use rune instead of byte for tag validations


mikkeloscar:
Closing in favor of #6829

PR #6829, opened 2024-01-24
bump deployment-service version

mikkeloscar:
summary
- use the correct domain for validations response
- use rune instead of byte for tag validations

Replaces #6828


mikkeloscar:
:+1:

katyanna:
👍

PR #6840, opened 2024-01-26
Allow safe sysctl in new PSP implementation

mikkeloscar:
Follow up to  #6732 to allow safe sysctls as is the case for native PSP.


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #6843, opened 2024-01-26
Update Audittrail version with Nakadi functionality

zaklawrencea:
Follow up to https://github.com/zalando-incubator/kubernetes-on-aws/pull/6747.

> This PR updates the Audittrail adapter version. The new version includes functionality which allows the adapter to send events to Nakadi. To enable this, a new config item `audittrail_nakadi_url` has been introduced which will be passed to the adapter to set the target Nakadi URL. 

A new `PlatformCredentialsSet` for the adapter will be deployed when `audittrail_nakadi_url` is set -- utilising `token_version: v2` to allow write permission to Nakadi event streams.

### DIFF
--- a/cluster/manifests/audittrail-adapter/credentials.yaml
+++ b/cluster/manifests/audittrail-adapter/credentials.yaml
@@ -12,3 +12,20 @@ spec:
      audittrail:
        privileges: []
 {{- end }}
+{{- if .Cluster.ConfigItems.audittrail_nakadi_url }}
+---
+apiVersion: "zalando.org/v1"
+kind: PlatformCredentialsSet
+metadata:
+   name: "audittrail-adapter-nakadi"
+   namespace: kube-system
+   labels:
+     application: "audittrail-adapter"
+spec:
+   application: "audittrail-adapter"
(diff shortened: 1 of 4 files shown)


cluster/manifests/audittrail-adapter/credentials.yaml
```diff
@@ -12,3 +12,20 @@ spec:
      audittrail:
        privileges: []
 {{- end }}
+{{- if .Cluster.ConfigItems.audittrail_nakadi_url }}
+---
```
mikkeloscar, line 16: We should also add an entry in `cluster/manifest/deletions.yaml` Such that this PCS is deleted if we set `audittrail_nakadi_url=""` after having set an URL.

zaklawrencea: Added. I went with the `if eq ... ""` syntax as opposed to `if not ...` as it looked more in-line with the rest of the `deletions.yaml` file content.

mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #6852, opened 2024-01-29
Karpenter update

myaser:
in the latest version of karpenter >= v0.32, the CRDs have been renamed and their structure have slightly changed as the api graduated from alpha to beta
in addition, the couldformation templates security have been strengthened by adding limitations on the resources that can be manipulated by the controller

this PR adapts the manifests and cloudformation templates of the newer version


cluster/manifests/z-karpenter/webhooks.yaml
```diff
@@ -1,128 +0,0 @@
-{{ if eq .Cluster.ConfigItems.karpenter_pools_enabled "true"}}
```
mikkeloscar, line 1: note: we should also make sure to delete these resources via `cluster/manifests/deletions.yaml` file. Currently they would not be deleted with `karpenter_pools_enabled=true`

mikkeloscar: Same for the CRDs "renaming" there we should also delete the old ones via `deletions.yaml`

myaser: fixed, 
the only thing that could not be done in this iteration is the CRDs.
they can not be deleted before removing their instances
so I kept them under the `karpenter_pools_enabled=true` for now until being cleaned up later after migration is complete

test/e2e/Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN chmod +x /usr/bin/kubectl
 COPY --from=builder /go/bin/ginkgo /usr/local/bin/ginkgo
 
 # copy CLM
-COPY --from=container-registry.zalando.net/teapot/cluster-lifecycle-manager:latest /clm /usr/bin/clm
+COPY --from=container-registry-test.zalando.net/teapot/cluster-lifecycle-manager:pr-744-3 /clm /usr/bin/clm
```
myaser:
```suggestion
COPY --from=container-registry-test.zalando.net/teapot/cluster-lifecycle-manager:pr-744-5 /clm /usr/bin/clm
```

test/e2e/Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN chmod +x /usr/bin/kubectl
 COPY --from=builder /go/bin/ginkgo /usr/local/bin/ginkgo
 
 # copy CLM
-COPY --from=container-registry.zalando.net/teapot/cluster-lifecycle-manager:latest /clm /usr/bin/clm
+COPY --from=container-registry-test.zalando.net/teapot/cluster-lifecycle-manager:pr-744-5 /clm /usr/bin/clm
```
myaser:
```suggestion
COPY --from=container-registry.zalando.net/teapot/cluster-lifecycle-manager:latest /clm /usr/bin/clm
```

cluster/manifests/deletions.yaml
```diff
@@ -184,48 +220,31 @@ post_apply:
   kind: ClusterRole
 - name: karpenter-core
   kind: ClusterRoleBinding
-- name: karpenter-webhook
+- name: karpenter
   kind: ClusterRole
-- name: karpenter-webhook
+- name : karpenter
```
mikkeloscar: looks wrong with a space here :)

cluster/manifests/deletions.yaml
```diff
@@ -164,14 +164,50 @@ post_apply:
...
+  kind: Secret
+  namespace: kube-system
+- name: defaulting.webhook.karpenter.k8s.aws
+  kind: MutatingWebhookConfiguration
+- name: validation.webhook.karpenter.k8s.aws
+  kind: ValidatingWebhookConfiguration
+- name: validation.webhook.config.karpenter.sh
+  kind: ValidatingWebhookConfiguration
+- name: validation.webhook.karpenter.sh
+  kind: ValidatingWebhookConfiguration
```
mikkeloscar, line 197: Are all these legacy things that should be deleted for the update?

Maybe add a comment to make it clear? :)

mikkeloscar:
:+1:

myaser:
👍🏻

PR #6854, opened 2024-01-30
skipper: use redis 7

AlexanderYastrebov:
https://raw.githubusercontent.com/redis/redis/7.2/00-RELEASENOTES

The https://github.com/zalando/skipper/pull/2895 updates skipper tests to use redis-7

### DIFF
--- a/cluster/manifests/skipper/skipper-redis.yaml
+++ b/cluster/manifests/skipper/skipper-redis.yaml
@@ -3,7 +3,7 @@ kind: StatefulSet
 metadata:
   labels:
     application: skipper-ingress-redis
-    version: v6.2.7
+    version: v7.2.4
   annotations:
     zalando.org/update-using-hpa-replicas: skipper-ingress-redis
   name: skipper-ingress-redis
@@ -19,7 +19,7 @@ spec:
       labels:
         statefulset: skipper-ingress-redis
         application: skipper-ingress-redis
-        version: v6.2.7
+        version: v7.2.4
       annotations:
         cluster-autoscaler.kubernetes.io/safe-to-evict: "false"
         logging/destination: "{{.Cluster.ConfigItems.log_destination_infra}}"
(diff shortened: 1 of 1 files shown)


cluster/manifests/skipper/skipper-redis.yaml
```diff
@@ -3,7 +3,7 @@ kind: StatefulSet
 metadata:
   labels:
     application: skipper-ingress-redis
-    version: v6.2.7
+    version: "7.2.4"
```
szuecs: Why not prefix with v?
We do it basically for every container to omit the float issue :)

szuecs:
I don't like that "refactoring" no-op is mixed with image change.
Why do we need to have indentation and version "fixes"?

cluster/manifests/skipper/skipper-redis.yaml
AlexanderYastrebov: I checked other manifests and reverted to use v prefix

AlexanderYastrebov:
> I don't like that "refactoring" no-op is mixed with image change.
> Why do we need to have indentation and version "fixes"?

I thought to make it a valid yaml on the go (by putting template actions into yaml comments) but we can split this change in two so I reverted to version+image change only.

RomanZavodskikh:
:+1:

cluster/manifests/skipper/skipper-redis.yaml
```diff
@@ -3,7 +3,7 @@ kind: StatefulSet
 metadata:
   labels:
     application: skipper-ingress-redis
-    version: v6.2.7
+    version: v7.2.4
```
AlexanderYastrebov, line 6:
```
~$ docker run --rm container-registry.zalando.net/library/redis-7-alpine:7-alpine-20240226 redis-server --version
Redis server v=7.2.4 sha=00000000:0 malloc=jemalloc-5.3.0 bits=64 build=f17de61a0b604ef3
```

szuecs:
:+1:

PR #6857, opened 2024-01-30
Add option to inject NODE_OPTIONS envvar

mikkeloscar:
tThis optionally adds an option for injecting `NODE_OPTIONS=--max-old-space-size=` into containers to mitigate issues with nodejs 18 running on Ubuntu 22.04 (cgroup v2).

When enabled, this injects the environment variable into all containers if the feature is enabled. The idea is to enable this feature in all test clusters to validate that setting `NODE_OPTIONS=` in non-nodejs apps doesn't have a negative effect, and if that is successful to enable it in all production clusters _with_ nodejs apps.


mikkeloscar:
:+1:

linki:
👍

PR #6880, opened 2024-02-01
Update Stackset controller with Stack owned Secrets feature

katyanna:
-> [Reference Secrets for per stack versioning](https://github.com/zalando-incubator/stackset-controller/pull/565)

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -903,6 +903,9 @@ stackset_configmap_support_enabled: "true"
 stackset_configmap_support_enabled: "false"
 {{end}}
 
+# enable/disable secret support for stackset
+stackset_secret_support_enabled: "false"
+
 # enable/disable traffic segment support for stackset
 {{if eq .Cluster.Environment "e2e"}}
 stackset_enable_traffic_segments: "true"
--- a/cluster/manifests/stackset-controller/01-stack-crd.yaml
+++ b/cluster/manifests/stackset-controller/01-stack-crd.yaml
@@ -377,9 +377,9 @@ spec:
                             TODO: Add other useful fields. apiVersion, kind, uid?'
                           type: string
                       type: object
+{{ end }}
                   type: object
                 type: array
-{{ end }}
 {{ end }}
               externalIngress:
                 description: Stack specific ExternalIngress, based on the parent StackSet
--- a/cluster/manifests/stackset-controller/01-stack-crd.yaml
+++ b/cluster/manifests/stackset-controller/01-stack-crd.yaml
@@ -349,7 +349,7 @@ spec:
                 - maxReplicas
                 - metrics
                 type: object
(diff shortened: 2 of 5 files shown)


cluster/manifests/stackset-controller/01-stack-crd.yaml
```diff
@@ -349,7 +349,7 @@ spec:
                 - maxReplicas
                 - metrics
                 type: object
-{{- if eq .Cluster.ConfigItems.stackset_configmap_support_enabled "true" }}
+{{- if and (eq .Cluster.ConfigItems.stackset_configmap_support_enabled "true") (eq .Cluster.ConfigItems.stackset_secret_support_enabled "true") }}
```
mikkeloscar: This should be `or` right? one of them can be enabled and it will work. We have configMap enabled in some clusters I assume which would be disabled by default with this change.

katyanna: yes, thanks!

cluster/config-defaults.yaml
```diff
@@ -896,13 +896,19 @@ stackset_routegroup_support_enabled: "true"
 # E.g. switching from RouteGroup to Ingress or vice versa.
 stackset_ingress_source_switch_ttl: "5m"
 
+# enable/disable configuration resources support for stackset
+stackset_configuration_rsc_enabled: "false"
```
linki: We need to enable this in the clusters where we already use the configmap feature.

cluster/manifests/stackset-controller/01-stack-crd.yaml
```diff
@@ -349,7 +349,7 @@ spec:
                 - maxReplicas
                 - metrics
                 type: object
-{{- if eq .Cluster.ConfigItems.stackset_configmap_support_enabled "true" }}
+{{- if eq .Cluster.ConfigItems.stackset_configuration_rsc_enabled "true" }}
```
mikkeloscar: @katyanna initially you had this depend on either config-map or secret. I think that is a better approach because it's harder to keep in mind two different config-items have to be changed.

I would propose this:

```
{{- if or (eq .Cluster.ConfigItems.stackset_configmap_support_enabled "true") (eq .Cluster.ConfigItems.stackset_secret_support_enabled "true") }}
```

cluster/config-defaults.yaml
mikkeloscar: See also my comment below!

linki: Agree for @mikkeloscar's approach below.

cluster/manifests/stackset-controller/01-stack-crd.yaml
katyanna: I agree but it broke the pipeline. This was a test that seem to have worked but maybe it's related to the indentation, I'll try it again.

katyanna:
:+1:

gargravarr:
:+1:

PR #6887, opened 2024-02-05
Exclude c5d.large from default Karpenter instance-types

mikkeloscar:
Small optimization to exclude `c5d.large` from the default instance types for karpenter pools. The motivation is that the SSD on `c5d.large` is very slow, and combined with our AMI logic to move the `podruntime` folder to the SSD, this slows down startup of the instance and gives no disk space or performance improvement compared to the root EBS gp3 volume.


demonCoder95:
👍

gargravarr:
:+1:

PR #6898, opened 2024-02-06
add feature gate MaxUnavailableStatefulSet

demonCoder95:
This was requested in a zooport ticket: https://github.bus.zalan.do/zooport/issues/issues/4379

The use case was this somehow allows Redis pods for skipper to update faster. More details in the ticket above.

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -143,7 +143,7 @@ write_files:
           - --oidc-groups-claim=groups
           - "--oidc-groups-prefix=okta:"
 {{- end }}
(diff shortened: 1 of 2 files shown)


demonCoder95:
👍

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -143,7 +143,7 @@ write_files:
...
+          - --feature-gates=HPAScaleToZero={{ .Cluster.ConfigItems.enable_hpa_scale_to_zero }},EphemeralContainers={{ .Cluster.ConfigItems.enable_ephemeral_containers }},HPAContainerMetrics={{ .Cluster.ConfigItems.enable_hpa_container_metrics }},StatefulSetAutoDeletePVC={{ .Cluster.ConfigItems.enable_statefulset_autodelete_pvc }},TopologyAwareHints={{ .Cluster.ConfigItems.enable_topology_aware_hints }},MinDomainsInPodTopologySpread={{ .Cluster.ConfigItems.min_domains_in_pod_topology_spread_enabled }},CronJobTimeZone={{.Cluster.ConfigItems.cronjob_time_zone_enabled}},MaxUnavailableStatefulSet={{.Cluster.ConfigItems.max_unavailable_statefulset_enabled}}
```
mikkeloscar, line 146: Is it enough to set it here? I could imagine it should also be set on controller-manager. :thinking: 

Let's make sure to test this branch e.g. in playground to validate it's working before rolling out.

demonCoder95: The docs only mentioned adding it to the `api-server`. Let me test this on playground real quick, you might be right.

demonCoder95: I just checked in playground and it works. Also was able to create an STS resource with `rollingUpdate.maxUnavailable: 1`.

mikkeloscar:
:+1:

PR #6920, opened 2024-02-07
deployment-service: Inject source information tags in CF stacks

mikkeloscar:
Updates the deployment service to inject source info tags in CF stacks. This previously only happened if the template of the CF stack changed, but now it also happens if any of the source info tags change.

For test cluster we disable update of changes to the branch tag because this would lead to a change on every PR which is not desired.


mikkeloscar:
:+1:

katyanna:
👍

PR #6923, opened 2024-02-08
Update VPA CRD

mikkeloscar:
This updates the VPA CRD to the version related to the version of the VPA we run.

It still serves `v1beta1`, `v1beta2` and `v1`, but the `v1` spec is updated to include the field `controlledResources` which is used by the vpa-recommender but was not defined in the spec.

Updates all our VPAs to v1.


mikkeloscar:
:+1:

katyanna:
👍

PR #6929, opened 2024-02-08
Enable traffic segment probing when cluster supports either annotated segments or segments by default.

gargravarr:
We need to update the templating because of the change I applied in the flags for StackSet controller here https://github.com/zalando-incubator/stackset-controller/pull/575


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6931, opened 2024-02-08
Scale master node Daemonset VPAs dynamically relative to master node size

demonCoder95:
Size master node DaemonSet VPAs relative to the size of the master node instance. The following master node Daemonsets use a VerticalPodAutoscaler:

- [x] `karpenter`
- [x] `ebs-csi-controller`
- [x] `kube-aws-iam-controller`

The scaling factors used in templates were computed here: https://github.bus.zalan.do/teapot/issues/issues/3469#issuecomment-5134926

Reference Issue: https://github.bus.zalan.do/teapot/issues/issues/3469
Depends on the patch in `cluster-lifecycle-manager` that implements the used template functions: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/743

### DIFF
--- a/cluster/manifests/z-karpenter/vpa.yaml
+++ b/cluster/manifests/z-karpenter/vpa.yaml
@@ -24,6 +24,8 @@ spec:
         maxAllowed:
         {{ range $NodePool := .Cluster.NodePools}}
         {{ if eq $NodePool.Name "default-master" }}
+          # Scaling is relative to r6g.large (smallest master node)
+          # 0.016 -> ~250Mi memory, 0.027 -> ~50m CPU
           memory: {{ scaleQuantity (instanceTypeMemory (index .InstanceTypes 0)) 0.016 }}
           cpu: {{ scaleQuantity (instanceTypeCPU (index .InstanceTypes 0)) 0.027 }}
         {{ end }}
--- a/cluster/manifests/z-karpenter/vpa.yaml
+++ b/cluster/manifests/z-karpenter/vpa.yaml
@@ -19,6 +19,13 @@ spec:
     containerPolicies:
       - containerName: controller
         minAllowed:
-          memory: {{.Cluster.ConfigItems.karpenter_controller_memory}}
-          cpu: {{.Cluster.ConfigItems.karpenter_controller_cpu}}
+          memory: .Cluster.ConfigItems.karpenter_controller_memory
+          cpu: .Cluster.ConfigItems.karpenter_controller_cpu
+        maxAllowed:
+        {{ range $NodePool := .Cluster.NodePools}}
+        {{ if eq $NodePool.Name "default-master" }}
+          memory: {{ scaleQuantity (instanceTypeMemory (index .InstanceTypes 0)) 0.016 }}
+          cpu: {{ scaleQuantity (instanceTypeCPU (index .InstanceTypes 0)) 0.027 }}
+        {{ end }}
+        {{ end }}
 {{end}}
--- a/cluster/manifests/z-karpenter/vpa.yaml
+++ b/cluster/manifests/z-karpenter/vpa.yaml
@@ -19,8 +19,8 @@ spec:
     containerPolicies:
       - containerName: controller
         minAllowed:
-          memory: .Cluster.ConfigItems.karpenter_controller_memory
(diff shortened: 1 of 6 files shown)


demonCoder95:
Adds a working proof-of-concept to use 10% of resources for `karpenter` controller (scaling ratio of `0.1`). Has to be adjusted before the PR can be merged.

cluster/manifests/z-karpenter/vpa.yaml
```diff
@@ -19,6 +19,10 @@ spec:
     containerPolicies:
       - containerName: controller
         minAllowed:
-          memory: {{.Cluster.ConfigItems.karpenter_controller_memory}}
-          cpu: {{.Cluster.ConfigItems.karpenter_controller_cpu}}
+        {{ range $NodePool := .Cluster.NodePools}}
+        {{ if eq $NodePool.Name "default-master" }}
+          memory: {{ scaleQuantity (instanceTypeMemory (index .InstanceTypes 0)) 0.1 }}
+          cpu: {{ scaleQuantity (instanceTypeCPU (index .InstanceTypes 0)) 0.1 }}
```
mikkeloscar: Note, these should be for `maxAllowed` as we want to limit the max, not set the min :)

demonCoder95: Ah yes. Thanks, I'll update.

demonCoder95: Update in the next commit.

cluster/manifests/z-karpenter/vpa.yaml
```diff
@@ -19,6 +19,13 @@ spec:
...
         minAllowed:
-          memory: {{.Cluster.ConfigItems.karpenter_controller_memory}}
-          cpu: {{.Cluster.ConfigItems.karpenter_controller_cpu}}
+          memory: {{ .Cluster.ConfigItems.karpenter_controller_memory }}
+          cpu: {{ .Cluster.ConfigItems.karpenter_controller_cpu }}
+        maxAllowed:
+        {{ range $NodePool := .Cluster.NodePools}}
+        {{ if eq $NodePool.Name "default-master" }}
+          memory: {{ scaleQuantity (instanceTypeMemory (index .InstanceTypes 0)) 0.016 }}
+          cpu: {{ scaleQuantity (instanceTypeCPU (index .InstanceTypes 0)) 0.027 }}
```
mikkeloscar: Let's add a comment explaining that this means 50m on `r6g.large` :)

demonCoder95: Added in the next commit. Thanks!

demonCoder95:
I'm going to test the final PR once more on the pet cluster to make sure everything works as expected.

demonCoder95:
Everything works as expected. There is naturally a rounding error but it is acceptable (~5m CPU and ~7Mi memory), I think. Since we are already at 3 decimal places in our scaling factor which I believe is sufficient.

The PR is now ready.

demonCoder95:
`create-cluster` is failing because CLM hasn't been patched yet to include the new template functions. The PR works on the pet cluster where it was tested.

demonCoder95:
CLM has been updated, e2e should pass now.

demonCoder95:
The e2e fails because currently the default memory set for karpenter controller is `250Mi` and this is set as the `minAllowed` in the VPA. When the `maxAllowed` patch from this PR is applied, with `e2e` master node instance being `m6g.large` types, the computed limit turns out to be `~131Mi`. Since this `maxAlowed` is less than the fixed default `minAllowed` of `250Mi`, the patch fails.

```
time="2024-02-09T17:33:14Z" level=error msg="for: \"STDIN\": admission webhook \"vpa.k8s.io\" denied the request: max resource for memory is lower than min" module=z-karpenter
```

Will create a separate PR to first lower the default threshold that is used in `minAllowed`, so that this patch can succeed.

demonCoder95:
Created the PR to lower default karpenter threshold to sane values, that is a requirement for this patch to succeed:
https://github.com/zalando-incubator/kubernetes-on-aws/pull/6944

demonCoder95:
Rebased after merging the change to lower karpenter vpa resource thresholds.

demonCoder95:
👍

mikkeloscar:
:+1:

PR #6934, opened 2024-02-09
Enable StackSet's versioned Secret support in e2e tests

linki:
This is to make sure that e2e tests pass when this feature is enabled.


linki:
👍

gargravarr:
:+1:

PR #6944, opened 2024-02-09
lower karpenter resource thresholds

demonCoder95:
There is work underway to dynamically scale VPAs for master node DaemonSets, including karpenter. [See Issue](https://github.bus.zalan.do/teapot/issues/issues/3469).

For this patch to succeed, the current threshold used in `minAllowed` is too high. When the VPA is updated to add a `maxAllowed` value that is lower than this fixed higher threshold, it fails and the patch cannot succeed.

This fix should allow the patch to succeed, and is safe since as of now, karpenter VPA doesn't have any upper limits and can scale up if it needs to.

More details: https://github.com/zalando-incubator/kubernetes-on-aws/pull/6931#issuecomment-1936327793


demonCoder95:
👍

mikkeloscar:
:+1:

PR #6947, opened 2024-02-12
[WIP] External cloud provider [2/2]

mikkeloscar:
Run kubelet with external cloud provider (requirement from Kubernetes v1.27)

This changes the kubelet configuration in the following way so it's compatible with Kubernetes v1.27

It basically does the following changes in the AMI:

* Switch kubelet flag `--cloud-provider` from `aws` to `external`
* Add `--hostname-override` to set the EC2 internal DNS name as host. This was previously done by the `aws` cloud-provider logic.
* Set `--image-credential-provider` flags on kubelet
  * Add image credential provider config to tell kubelet how to authorize pulls from ECR
  * Add the `ecr-credential-provider` binary used to get ECR credentials (what was previously done by `--cloud-provider=aws`). We use the binary distributed for EKS which is available in a public S3 bucket managed by AWS.
* Introduces simple templating for the kubelet.yaml config file defined in [kubernetes-on-aws](https://github.com/zalando-incubator/kubernetes-on-aws/pull/6947/files#diff-fb712e77981d2e840a03b53964fd2fb37d9cd137e35b2db4d1dbb36f5170fb44R17-R76). This allows injecting `providerID` on instance startup, which has to be added manually and has to be in the `kubelet.yaml`. Same is done for clusterDNS which also can't be set as a flag from Kubernetes v1.27.

### Reference

https://cloud-provider-aws.sigs.k8s.io/prerequisites/
ECR: https://github.com/awslabs/amazon-eks-ami/commit/42c3f524cd74d82c7ea4aaecbc0d46c1c39d2095

## TODO

* [x] Describe changes
* [ ] Use latest valid image
* [x] Adjust resources

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -15,7 +15,7 @@ write_files:
       AWS_DEFAULT_REGION={{ .Cluster.Region }}
 
   - owner: root:root
-    path: /etc/kubernetes/config/kubelet.yaml
+    path: /etc/kubernetes/config/kubelet.yaml.template
     content: |
       # https://github.com/kubernetes/kubernetes/blob/v1.13.6/staging/src/k8s.io/kubelet/config/v1beta1/types.go
       apiVersion: kubelet.config.k8s.io/v1beta1
@@ -71,6 +71,9 @@ write_files:
       shutdownGracePeriod: "120s"
       shutdownGracePeriodCriticalPods: "80s"
 {{- end }}
+      # variables are replaced on instance start-up.
+      providerID: __PROVIDER_ID__
+      clusterDNS: [__CLUSTER_DNS__]
 
   - owner: root:root
     path: /etc/kubernetes/manifests/kube-apiserver.yaml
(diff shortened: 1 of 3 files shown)


cluster/node-pools/master-default/userdata.yaml
```diff
@@ -71,6 +71,9 @@ write_files:
       shutdownGracePeriod: "120s"
       shutdownGracePeriodCriticalPods: "80s"
 {{- end }}
+      # variables are replaced on instance start-up.
```
linki, line 74: How does that work?

linki: It's done in here: https://github.bus.zalan.do/teapot/kubernetes-on-ubuntu/pull/332

cluster/node-pools/worker-splitaz/userdata.yaml
```diff
@@ -107,6 +107,9 @@ write_files:
           cacheAuthorizedTTL: "5m"
           cacheUnauthorizedTTL: "30s"
       protectKernelDefaults: true
+      # variables are replaced on instance start-up.
+      providerID: __PROVIDER_ID__
```
linki, line 111: According to the [docs](https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/#running-cloud-controller-manager) `kubelet` should also specify `--cloud-provider=external`. Maybe implicit when setting `providerID` but something to double check.

linki: This is also [done in here](https://github.bus.zalan.do/teapot/kubernetes-on-ubuntu/pull/332) it seems.

mikkeloscar:
Closing this and making it part of v1.27 roll out to avoid multiple node rotations: #6997

PR #6950, opened 2024-02-12
Add aws-cloud-controller-manager [1/2]

mikkeloscar:
Add aws-cloud-controller-manager daemonset to run on master nodes.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -428,6 +428,12 @@ kubernetes_lifecycle_metrics_mem_min: "120Mi"
 kube_node_ready_controller_cpu: "50m"
 kube_node_ready_controller_memory: "200Mi"
 
+# Enable deployment of aws-cloud-controller-manager
+aws_cloud_controller_manager_enabled: "true"
+aws_cloud_controller_manager_cpu: "125m"
+aws_cloud_controller_manager_memory: "512Mi"
+
+
 # Kubernetes Downscaler (for non-production clusters)
 {{if eq .Cluster.Environment "test"}}
 downscaler_default_uptime: "Mon-Fri 07:30-20:30 Europe/Berlin"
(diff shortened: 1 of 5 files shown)


cluster/manifests/aws-cloud-controller-manager/rbac.yaml
```diff
@@ -0,0 +1,134 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cloud-controller-manager
+  namespace: kube-system
+  labels:
+    application: kubernetes
+    component: aws-cloud-controller-manager
```
linki, line 9: Don't we need an IAM role for it or how does it get permission to access AWS?

mikkeloscar: Good point!

It uses `hostNetwork` (taken from upstream example) so it's using the role of the node, but it makes sense to use a dedicated role ofc.

mikkeloscar: Changed to use dedicated IAM role and not `hostNetwork`.

mikkeloscar:
:+1:

cluster/config-defaults.yaml
```diff
@@ -428,6 +428,12 @@ kubernetes_lifecycle_metrics_mem_min: "120Mi"
 kube_node_ready_controller_cpu: "50m"
 kube_node_ready_controller_memory: "200Mi"
 
+# Enable deployment of aws-cloud-controller-manager
+aws_cloud_controller_manager_enabled: "true"
```
demonCoder95, line 432: Shouldn't we keep it `false` by default as we're testing it? 🤔  Or do we want to control the testing/rollout by channel and not by cluster?

mikkeloscar: We can flip the logic to false and enable per cluster. This is coming from the perspective, it does seem to work in e2e, so I assume it works and we can disable it if it does not.

But changing to false and enabling manually is also fine by me.

linki:
👍

cluster/config-defaults.yaml
demonCoder95: Makes sense to me then. Thanks for explaining! :)

demonCoder95:
Thanks for merging! 🙏

PR #6979, opened 2024-02-15
Fix runtime policy injection for prod

mikkeloscar:
Make sure to fall back to default runtime policy if nothing is set.

This fixes the issue that pods on karpenter nodes in production clusters would _not_ get the `karpenter.sh/capacity-type: "on-demand"` injected if they didn't specify a runtime policy explicitly.


mikkeloscar:
:+1:

myaser:
👍🏻

PR #6984, opened 2024-02-15
Enable versioned configuration resources support for all environments

linki:
Enable versioned configuration resources support for **all** environments.

If have set the config items to `false` in the registry so that we can opt-in clusters gradually by removing the config item.


linki:
👍

gargravarr:
:+1:

PR #6996, opened 2024-02-16
Give karpenter more memory by default

mikkeloscar:
100Mi is not quite enough for karpenter on a new cluster. This leads to karpenter pods OOMkilled for some time before VPA kicks in and this makes e2e unstable.

Set default to 256Mi and allow up to 1024Mi (6.4%) on `r6g.large` which is plenty of space. On `m6g.large` it would be 512Mi which is also not bad if we ever run smaller nodes and plenty for the other components as well.


myaser:
👍🏻

demonCoder95:
👍

PR #6147, opened 2023-06-29
Update to Kubernetes v1.26

mikkeloscar:
This updates to Kubernetes v1.26

This depends on #5995, #5871 and #6065 it's opened now to find potential issues early.

PR #6997, opened 2024-02-16
Update to Kubernetes v1.27

mikkeloscar:
Updates to Kubernetes v1.27

Depends on:

* #6147
* #6950 
* #6947 (This is now included here)

### Run kubelet with external cloud provider

Run kubelet with external cloud provider (requirement from Kubernetes v1.27)

This changes the kubelet configuration in the following way so it's compatible with Kubernetes v1.27

It basically does the following changes in the AMI:

* Switch kubelet flag `--cloud-provider` from `aws` to `external`
* Add `--hostname-override` to set the EC2 internal DNS name as host. This was previously done by the `aws` cloud-provider logic.
* Set `--image-credential-provider` flags on kubelet
  * Add image credential provider config to tell kubelet how to authorize pulls from ECR
  * Add the `ecr-credential-provider` binary used to get ECR credentials (what was previously done by `--cloud-provider=aws`). We use the binary distributed for EKS which is available in a public S3 bucket managed by AWS.
* Introduces simple templating for the kubelet.yaml config file defined in [kubernetes-on-aws](https://github.com/zalando-incubator/kubernetes-on-aws/pull/6947/files#diff-fb712e77981d2e840a03b53964fd2fb37d9cd137e35b2db4d1dbb36f5170fb44R17-R76). This allows injecting `providerID` on instance startup, which has to be added manually and has to be in the `kubelet.yaml`. Same is done for clusterDNS which also can't be set as a flag from Kubernetes v1.27.

### Reference

https://cloud-provider-aws.sigs.k8s.io/prerequisites/
ECR: https://github.com/awslabs/amazon-eks-ami/commit/42c3f524cd74d82c7ea4aaecbc0d46c1c39d2095


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #6998, opened 2024-02-16
Fix runtime policy handling

mikkeloscar:
Follow up to #6979 

This fixes the runtime policy injection in the following ways:

* Make sure to fall back to default runtime policy if nothing is set (from #6979).
* Exclude daemonset pods which should not be in scope of runtime policy as they should run on _all_ nodes.
* Don't inject if the default matches what the pod is requesting for the zalando provider (this was an issue introduced in #6979).
* Don't have runtime-policy logic in scope of exclude namespace (this means it will also apply to kube-system pods)
* Treat pods targeting the master nodes (via toleration) as part of the master pool which uses `zalando` lifecycle provider. This avoid injecting a karpenter relevant node selector for pods running on the master nodes, when the default pool is using `karpenter`.


mikkeloscar:
:+1:

myaser:
👍🏻

PR #6999, opened 2024-02-16
skipper: do not use service account with routesrv

AlexanderYastrebov:
Skipper does not need to access Kubernetes API when it uses routesrv.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #7007, opened 2024-02-19
feature: skipper-ingress redis hpa configure scale-down behavior

szuecs:
feature: skipper-ingress redis hpa configure scale-down behavior

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -160,6 +160,7 @@ skipper_ingress_redis_target_average_utilization_memory: "60"
 skipper_ingress_redis_min_replicas: "1"
 skipper_ingress_redis_max_replicas: "100"
 skipper_ingress_redis_cluster_scaling_schedules: ""
+skipper_ingress_redis_hpa_scale_down_wait: "600"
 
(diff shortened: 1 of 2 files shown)


cluster/config-defaults.yaml
```diff
@@ -160,6 +160,7 @@ skipper_ingress_redis_target_average_utilization_memory: "60"
 skipper_ingress_redis_min_replicas: "1"
 skipper_ingress_redis_max_replicas: "100"
 skipper_ingress_redis_cluster_scaling_schedules: ""
+skipper_ingress_redis_hpa_scale_down_wait: "600"
```
AlexanderYastrebov, line 163: Do we need `_hpa_`? Also maybe it should be named like the field it sets, i.e. `skipper_ingress_redis_stabilization_window_seconds`

szuecs: I prefer this because it's basically what it does on high level.
We also have the same name in skipper-ingress hpa.

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #7005, opened 2024-02-19
Fix focal v1.26 amd64 AMI

mikkeloscar:
follow up to #6994 

Update the AMI as there was a bug where the image: `zalando-ubuntu-focal-20.04-kubernetes-production-v1.26.14-amd64-master-310` was promoted as `zalando-ubuntu-jammy-22.04-kubernetes-production-v1.26.14-amd64-master-310`.

PR #7008, opened 2024-02-19
Fix arch on focal 1.26 build

mikkeloscar:
Follow up to #7005

Also fix the arch in the focal 1.26 amd64 AMI as it was wrongly set to arm64.


mikkeloscar:
:+1:

linki:
👍

PR #7032, opened 2024-02-21
Reduce ephemeral storage default

mikkeloscar:
Reduce ephemeral storage default to `25Mi`

(This feature is not enabled in any cluster _yet_)


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #7036, opened 2024-02-21
Add admission for ephemeralcontainers

mikkeloscar:
This adds admission of ephemeral containers enabled since Kubernetes v1.25.

This admission will replace ZECR vanity image names with the ecr URL, validate production readiness of images and it will inject the standard envar as are injected in normal pod containers.

The MutatingWebhookConfiguration resource is called `pods/ephemeralcontainers` which is triggered via an update to the pods/ephemeralcontainers api of an existing pod. The request that the admission-controller gets is identical to a pods request, e.g. old and new pod where the new includes the ephemeral container being added.

See https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/


mikkeloscar:
:+1:

szuecs:
:+1:

PR #7048, opened 2024-02-22
Update Admission Controller.

gargravarr:
Updates the Admission Controller. Main change: deny using the annotation `zalando.org/aws-load-balancer-shared: "false"` in StackSets.


gargravarr:
:+1:

RomanZavodskikh:
:+1:

PR #7067, opened 2024-02-26
Tag Network Interfaces with application/component

mikkeloscar:
Since public IPs now have a cost, it makes sense to tag Network interfaces created for instances with the application/component tags such that the cost can be attributed in AWS Cost explorer.

Similar to for volumes #6716

### DIFF
--- a/cluster/node-pools/worker-combined/stack.yaml
+++ b/cluster/node-pools/worker-combined/stack.yaml
@@ -132,6 +132,14 @@ Resources:
             Value: kubernetes
           - Key: component
             Value: "shared-resource"
+        - ResourceType: "network-interface"
+          Tags:
+          - Key: application
+            Value: kubernetes
+          - Key: component
+            Value: "shared-resource"
+          - Key: Name
+            Value: "{{ .NodePool.Name }} ({{ .Cluster.ID }})"
 {{- end }}
         BlockDeviceMappings:
         - DeviceName: /dev/sda1
--- a/cluster/etcd/stack.yaml
+++ b/cluster/etcd/stack.yaml
@@ -36,6 +36,14 @@ Resources:
             Value: kubernetes
           - Key: component
             Value: etcd-cluster
+        - ResourceType: "network-interface"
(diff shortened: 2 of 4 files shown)


mikkeloscar:
:+1:

cluster/node-pools/worker-combined/stack.yaml
```diff
@@ -132,6 +132,14 @@ Resources:
             Value: kubernetes
           - Key: component
             Value: "shared-resource"
+        - ResourceType: "network-interface"
+          Tags:
+          - Key: application
+            Value: kubernetes
+          - Key: component
+            Value: "shared-resource"
```
demonCoder95, line 140: Is this a known `component` value? Or are we introducing this only for interfaces?

demonCoder95: Ah, I see we used a similar value for volumes [here](https://github.com/zalando-incubator/kubernetes-on-aws/pull/6712/files#diff-e46dbad970101fad1d15cfac46cf072cd7ed6fb5ebb4d160a485a58cb62700cfR133).

demonCoder95:
👍

cluster/node-pools/worker-combined/stack.yaml
mikkeloscar: Yeah, we use it for all "worker node resources" just to indicate this is not totally owned by the `kubernetes` component especially in terms of cost.

szuecs: @mikkeloscar do we use the same for ALB/NLB/ELB?

mikkeloscar: @szuecs for ALB/NLB from ingress the `application` tag is `kube-ingress-aws-controller` and there is no component defined. The ENIs don't seem to be tagged at all when created for a load balancer :/

For service Type=LoadBalancer users can set the application and component labels on the service resource and we add those as tags on the LB.

PR #7070, opened 2024-02-26
Fix calculating karpenter vpa max CPU

mikkeloscar:
Follow up to https://github.com/zalando-incubator/kubernetes-on-aws/pull/6931

The scale should be based on the full CPU availble on the instance type and not based on the allocatable CPU. Eg. before we calculated `50m / 1800m ~= 0.027` But the right calculation is `50m / 2000m = 0.025` otherwise the max becomes 54m instead of 50m.

This is to make room for all things as we have packed all master components to fit on 1799 (1803 because of this issue which doesn't fit with 1800m allocatable)


mikkeloscar:
:+1:

demonCoder95:
👍

PR #7074, opened 2024-02-26
Disable kube-node-ready on master nodes in test

mikkeloscar:
`kube-node-ready` is becoming less and less relevant and has the following issues in the current setup:

* Slows down recovery of master nodes in case of issues with the control plane
* Depends on ec2 instance metadata endpoint which we want to restrict _eventually_
* The ASG lifecycle hook signal is not relevant anymore since we have pod disruption budgets etc.

For this reason, first disable it in test clusters on master nodes to validate that it indeed is not an issue to remove it and eventually remove it from masters in production as well. Removing this from masters will also let us get rid of kube2iam on master nodes.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #7083, opened 2024-02-27
Disable kube2iam on master nodes in test

mikkeloscar:
With #7074 we can also exclude kube2iam on master nodes because nothing depend on it running there.

This would further enable running master nodes with IMDSv2 enforced (#6669)


mikkeloscar:
:+1:

demonCoder95:
👍

PR #7086, opened 2024-02-28
Don't make resource tagging depending on Jammy

mikkeloscar:
In #6716 we made tagging of worker node volumes a feature enabled together with the update to Ubuntu 22.04 because it requires the rotation of nodes.

Since then we had to roll back Ubuntu 22.04 in several clusters which then removed the volume tagging for new nodes.

Instead of combining it with Ubuntu 22.04, combine this with the roll out of Kubernetes v1.26. This will roll nodes in clusters with focal (Ubuntu 20.04 nodes).


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #7095, opened 2024-02-29
Use hostNetwork for EC2 metadata access for aws-cloud-controller-manager

mikkeloscar:
With #7083 and later #6669 the aws-cloud-controller-manager can't connect to the ec2 metadata to discover some initial information.

Enable hostNetwork such that it has direct access without kube2iam.

PR #7101, opened 2024-02-29
[ALPHA] Use hostnetwork for aws-cloud-controller-manager to access ec2 metadata

mikkeloscar:
Adding #7095 directly to alpha as e2e is broken without this.


mikkeloscar:
:+1:

linki:
👍

PR #7104, opened 2024-02-29
Fix condition for maxPods

mikkeloscar:
In #7093 the condition was missing the else clause, fix that to either set a fixed `maxPods` value or a automatic one.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #7117, opened 2024-03-04
Make canary-only configuration for skipper-ingress possible

RomanZavodskikh:
When "skipper_ingress_canary_enabled" and "skipper_ingress_test_single_pod" are both "true"

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -39,6 +39,8 @@ metadata:
 spec:
 {{ if index . "replicas" }}
   replicas: {{ .replicas }}
+{{ else if eq .Cluster.ConfigItems.skipper_ingress_test_single_pod "true" }}
+  replicas: 0
 {{ end }}
   strategy:
     rollingUpdate:
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -77,6 +77,7 @@ skipper_ingress_memory: "1500Mi"
 
 # Enables deployment of canary version
 skipper_ingress_canary_enabled: "true"
+skipper_ingress_test_single_pod: "false"
 
 # When set to true (and dedicated node pool for skipper is also true) the
 # daemonset overhead will be subtracted from the cpu settings such


szuecs:
@RomanZavodskikh that looks super complex. Why not something like a custom bool config item?
Another way could be also if max_replica set to 0 then delete hpa, but I guess a bool would be simple and cover what we need.

RomanZavodskikh:
Yes, I tried this idea, and I agree the configuration becomes much simpler. However, what if "skipper_ingress_canary_enabled" is set to "exclusive" in some important production cluster? Such accidential configuration will make all traffic go through single skipper-ingress pod.
I have considered such a scenario and added additional safety net.

szuecs:
@RomanZavodskikh you can also damage it by setting max replica to 1...

RomanZavodskikh:
Okay, I see, we cannot take so much responsibility here. Reverted to the most simple option.

szuecs:
Looks much better now!

AlexanderYastrebov:
I think you can add `replicas: 0` to the deployment conditionally - HPA does not scale deployment that has zero replicas

AlexanderYastrebov:
I also do not like that we change boolean (note _enabled suffix) config into multivalue but alternative is another config item which might be even worse.

cluster/manifests/deletions.yaml
```diff
@@ -59,7 +59,7 @@ post_apply:
   kind: ClusterRoleBinding
 {{- end }}
 
-{{ if ne .Cluster.ConfigItems.skipper_ingress_canary_enabled "true" }}
+{{ if and (ne .Cluster.ConfigItems.skipper_ingress_canary_enabled "true") (ne .Cluster.ConfigItems.skipper_ingress_test_single_pod "true") }}
```
szuecs: We don't need this change. The controller to do load tests can also just set canary to false or in case canary should be tested then we don't change the canary flag..

cluster/manifests/skipper/deployment.yaml
```diff
@@ -12,7 +12,7 @@
   "Values" .Values
 }}
 
-{{ if eq .Cluster.ConfigItems.skipper_ingress_canary_enabled "true" }}
+{{ if or (eq .Cluster.ConfigItems.skipper_ingress_canary_enabled "true") (eq .Cluster.ConfigItems.skipper_ingress_test_single_pod "true") }}
```
szuecs: this we can also drop

cluster/manifests/deletions.yaml
RomanZavodskikh: I see your idea: loadtest controller must set both "skipper_ingress_canary_enabled" and "skipper_ingress_test_single_pod" to "true". Okay, it makes sense.

RomanZavodskikh:
:+1:

szuecs:
:+1:

cluster/manifests/skipper/deployment.yaml
```diff
@@ -39,6 +39,8 @@ metadata:
 spec:
 {{ if index . "replicas" }}
   replicas: {{ .replicas }}
+{{ else if eq .Cluster.ConfigItems.skipper_ingress_test_single_pod "true" }}
```
AlexanderYastrebov, line 42: Hm, I think it would be cleaner to do
```yaml
{{ if eq .Cluster.ConfigItems.skipper_ingress_test_single_pod "true" }}
{{ template "skipper-ingress" dict
  "name" "skipper-ingress"
  "internal_version" $internal_version
  "replicas": 0
  "Cluster" .Cluster
  "Values" .Values
}}
{{ else }}
{{ template "skipper-ingress" dict
  "name" "skipper-ingress"
  "internal_version" $internal_version
  "Cluster" .Cluster
  "Values" .Values
}}
{{ end }}
```

AlexanderYastrebov: Also skipper_ingress_test_single_pod implies _test_ and _single_, but it is single canary pod which in the future might become several pods.
Maybe it should be just `skipper_ingress_replicas` that is empty by default:
```yaml
{{ if ne .Cluster.ConfigItems.skipper_ingress_replicas "" }}
{{ template "skipper-ingress" dict
  "name" "skipper-ingress"
  "internal_version" $internal_version
  "replicas": .Cluster.ConfigItems.skipper_ingress_replicas
  "Cluster" .Cluster
  "Values" .Values
}}
{{ else }}
{{ template "skipper-ingress" dict
  "name" "skipper-ingress"
  "internal_version" $internal_version
  "Cluster" .Cluster
  "Values" .Values
}}
{{ end }}
```

PR #7130, opened 2024-03-06
refactor: skipper svcaccount and split routesrv to have its own svcaccount [1/2]

szuecs:
refactor: skipper svcaccount and split routesrv to have its own svcaccount [1/2]
prepare: deletion for step 2
cleanup: routesrv enable/disable and config items

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -81,9 +81,7 @@ spec:
               parent-resource-hash: 71556441059f2d033fb06b1e73df03598c7ecaa6
 {{- end }}
       priorityClassName: "{{ .Cluster.ConfigItems.system_priority_class }}"
-{{ if ne .Cluster.ConfigItems.skipper_routesrv_enabled "exec" }}
       serviceAccountName: skipper-ingress
-{{ end }}
       terminationGracePeriodSeconds: {{ .Cluster.ConfigItems.skipper_termination_grace_period }}
       dnsPolicy: ClusterFirstWithHostNet
       hostNetwork: true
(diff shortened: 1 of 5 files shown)


cluster/manifests/skipper/rbac.yaml
```diff
@@ -5,12 +5,26 @@ metadata:
...
     application: skipper-ingress
+    component: ingress
 {{ if eq .Cluster.ConfigItems.skipper_open_policy_agent_enabled "true" }}
   # Note: if the role extends beyond OPA use, this condition can be removed
   annotations:
     iam.amazonaws.com/role: "{{ .Cluster.LocalID }}-app-skipper-ingress"
 {{ end }}
 ---
+apiVersion: v1
+kind: ServiceAccount
```
AlexanderYastrebov, line 16: Why do we need this instead of existing ServiceAccount (see also #6999)?

cluster/manifests/skipper/deployment.yaml
```diff
@@ -495,7 +445,7 @@ spec:
               parent-resource-hash: abd943226b6885f66785592be28bdf303863fbac
 {{- end }}
       priorityClassName: "{{ .Cluster.ConfigItems.system_priority_class }}"
-      serviceAccountName: skipper-ingress
+      serviceAccountName: skipper-ingress-routesrv
```
szuecs, line 448: @linki does it break CLM?
Below `rbac.yaml` has the creation of the new serviceaccount `skipper-ingress-routesrv`.

cluster/manifests/skipper/rbac.yaml
AlexanderYastrebov: Lets also split such changes in two PRs

szuecs: because least privileges and you broke OPA

The problem of https://github.com/zalando-incubator/kubernetes-on-aws/pull/6999 is that if "exec", then skipper-ingress does not use serviceaccount skipper-ingress which has https://github.com/zalando-incubator/kubernetes-on-aws/blob/beta/cluster/manifests/skipper/rbac.yaml#L8-L12 that if OPA is enabled leads to no permission to fetch the policy from S3.

AlexanderYastrebov: Existing `skipper-ingress` ServiceAccount permissions should be reduced to minimum necessary to fetch the policy from S3.

AlexanderYastrebov: Lets revert #6999 by https://github.com/zalando-incubator/kubernetes-on-aws/pull/7131 and work from there

szuecs: > Existing `skipper-ingress` ServiceAccount permissions should be reduced to minimum necessary to fetch the policy from S3.

no we should not change the running serviceaccount that is assigned to a workload!

PR #7138, opened 2024-03-06
Karpenter setup modifications

myaser:
- the instance profile that karpenter controller uses  for the workers it spawns is not under a feature flag. this guarantees it ALWAYS exists in the cluster even if not used. it has 0 costs and can prevent issues when the feature flag is turned off while there is already nodes running
- fix minor configuration issue

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -28,6 +28,10 @@ cluster_autoscaler_max_graceful_termination_sec: "1209600" # 2 weeks
 cluster_autoscaler_max_usnchedulable_pods_considered: "1000"
 
 # karpenter settings
+# DO NOT SET TO FALSE IF THE CLUSTER HAS KARPENTER POOLS OR NODES. REFER TO TEAPOT DOCS  FOR HOW TO ROLLBACK KARPENTER
+# https://teapot.docs.zalando.net/howtos/karpenter-operations/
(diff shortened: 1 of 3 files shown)


cluster/config-defaults.yaml
```diff
@@ -28,6 +28,9 @@ cluster_autoscaler_max_graceful_termination_sec: "1209600" # 2 weeks
 cluster_autoscaler_max_usnchedulable_pods_considered: "1000"
 
 # karpenter settings
+# DO NOT SET TO FALSE IF THE CLUSTER HAS KARPENTER POOLS OR NODES. REFER TO TEAPOT DOCS FOR HOW TO ROLLBACK KARPENTER
```
demonCoder95: Do we already have the docs created? I know that the scripting is in progress, but was wondering if there's any doc PR. We should link that here.

myaser: yea, here
https://github.bus.zalan.do/teapot/docs/pull/438

will add the link to the docs page

myaser:
👍🏻

cluster/config-defaults.yaml
demonCoder95: Let's add the doc link here and then we can merge :)

demonCoder95:
👍

myaser:
👍🏻

PR #7140, opened 2024-03-07
Drop config items for configuration resources (enabled by default)

linki:
These config items are enabled everywhere by default. Let's drop the config items to clean up.


linki:
👍

mikkeloscar:
:+1:

PR #7142, opened 2024-03-07
Update worker node AMI

demonCoder95:
This updates the AMI to include the optimisation where we clean the APT cache after package installation. This saves a lot of disk space. Also this helps address the support issue about increased disk space usage in a cluster.

Reference: https://github.bus.zalan.do/zooport/issues/issues/4473


demonCoder95:
Closing this in favor of [this PR](https://github.bus.zalan.do/teapot/kubernetes-on-ubuntu/pull/332) to update the AMI to add external cloud provider. The updated AMI will be rolled out as part of Kubernetes `1.27` rollout so this way we avoid rolling out many times.

PR #7154, opened 2024-03-12
Fix string matching in e2e test

mikkeloscar:
The reported error has changed in kube-1.27 so change the regex.


demonCoder95:
👍

mikkeloscar:
:+1:

PR #7160, opened 2024-03-13
Allow users to use coordination.k8s.io api for Leader election

mikkeloscar:
This allows the poweruser role access to use the `coordination.k8s.io` API which is used by controllers for leader election. Users wanting to deploy operators that use this API should be allowed to.

Ref: https://kubernetes.io/docs/concepts/architecture/leases/

### DIFF
--- a/cluster/manifests/roles/poweruser-role.yaml
+++ b/cluster/manifests/roles/poweruser-role.yaml
@@ -232,4 +232,11 @@ rules:
   - deletecollection
   - patch
   - update
-
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - create
+  - update
+  - patch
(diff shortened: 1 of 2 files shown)


linki:
Is there a use case for allowing deletion as well?

linki:
👍

mikkeloscar:
> Is there a use case for allowing deletion as well?

Not sure didn't find a case where `delete` is defined, but looking at: https://github.com/zalando-incubator/kubernetes-on-aws/blob/4d2be6071f8ca9e5d5a9cb9b5cce4b4ab0468ef3/cluster/manifests/z-karpenter/02-role.yaml#L19-L21 we may want `patch` as well.

mikkeloscar:
:+1:

katyanna:
:+1:

PR #7171, opened 2024-03-14
Increase EBS root volume size.

gargravarr:
Since kubernetes 1.26, many clusters have Pods being Evicted or Erroring because of disk pressure in the root volume. This Pull request mitigates this problem by increaseing the default for EBS root volume size.


gargravarr:
:+1:

demonCoder95:
👍

PR #7184, opened 2024-03-18
Only schedule coredns based on known taints

mikkeloscar:
Before CoreDNS tolerated _any_ taint. To avoid CoreDNS landing on nodes in Kubernetes v1.27 which are not initialized (taint: `key: node.cloudprovider.kubernetes.io/uninitialized, Value:true`) We enumerate _only_ known taints that should be tolerated.


mikkeloscar:
:+1:

gargravarr:
:+1:

linki:
👍

PR #7114, opened 2024-03-04
feature: ingress enable cert tag filter [1/2]

szuecs:
Enabled for test as first step. Test certs are all tagged so nothing should happen. It is also enabled for e2e clusters
All new created certs have to have the ACM tag `kubernetes=enabled` to make sure that our ingress controller will pick them up to use in cloud load balancers. It's a safety feature to make sure manually created certificates are not automatically picked up, but there is an active choice to do so.

PR #7187, opened 2024-03-19
feature: ingress enable cert tag filter [2/2]

szuecs:
production clusters were already changed via cluster-registry. This will now set it as default for all clusters such that new clusters will have this behavior from the beginning.

see also https://github.com/zalando-incubator/kubernetes-on-aws/pull/7114


szuecs:
:+1:

gargravarr:
:+1:

PR #7196, opened 2024-03-22
Enable inplace pod resize featuregate

mikkeloscar:
Try out in place pod resize feature: https://kubernetes.io/blog/2023/05/12/in-place-pod-resize-alpha/

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -1102,3 +1102,8 @@ sysctl_settings: ""
 
 # enables/disables the minDomains field for pod topology spread.
 min_domains_in_pod_topology_spread_enabled: "true"
+
+# enable InPlacePodVerticalScaling
+#
+# https://kubernetes.io/blog/2023/05/12/in-place-pod-resize-alpha/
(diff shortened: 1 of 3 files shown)


cluster/config-defaults.yaml
```diff
@@ -1092,3 +1092,8 @@ sysctl_settings: ""
 
 # enables/disables the minDomains field for pod topology spread.
 min_domains_in_pod_topology_spread_enabled: "true"
+
+# enable InPlacePodVerticalScaling
+#
+# https://kubernetes.io/blog/2023/05/12/in-place-pod-resize-alpha/
+enable_in_place_pod_vertical_scaling: "true"
```
AlexanderYastrebov, line 1109: What do you plan to use it for (and how)?
I think it might be interesting for Skipper deployment to use whole dedicated node resources.

mikkeloscar: Not sure if the feature makes sense today, there are missing support in e.g. the Vertical pod autoscaler so it's maybe less interesting today.

mikkeloscar:
Closing this for now.

PR #7211, opened 2024-03-26
Update VPA to v1.0.0

mikkeloscar:
Update the vertical-pod-autoscaler components to a version based on upstream v1.0.0: https://github.com/zalando-incubator/autoscaler/pull/108

The version is updated via the config-item: `vertical_pod_autoscaler_version=current|legacy` such that we run the new version by default, but can easily switch back by setting `vertical_pod_autoscaler_version=legacy`. This is done as it's hard to validate the VPA without running it for some time.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #7235, opened 2024-04-02
[TEST E2E] Validate AMI works with MACAddressPolicy=none

mikkeloscar:
This is just to test that an AMI with `MACAddressPolicy=none` works in our setup.

This setting would be needed for AWS VPC CNI and if it works we don't need to run a patched AMI version to test that in connection with EKS.


mikkeloscar:
Closing after successful test

PR #7236, opened 2024-04-02
skipper: replace parent-resource-hash selector

AlexanderYastrebov:
Use `kind: name` labels introduced for skipper-ingress application
consolidation (see https://github.com/zalando-incubator/kubernetes-on-aws/pull/4855 and https://github.com/zalando-incubator/kubernetes-on-aws/pull/4847).

This is a less obscure approach compared to use of `parent-resource-hash`
label (see https://github.com/zalando-incubator/kubernetes-on-aws/pull/4758) added by admission controller which is simply a hash of pod
owner kind and name:
```console
~$ echo -n Deployment-skipper-ingress | sha1sum
71556441059f2d033fb06b1e73df03598c7ecaa6  -
~$ echo -n Deployment-skipper-ingress-routesrv | sha1sum
abd943226b6885f66785592be28bdf303863fbac  -
~$ echo -n StatefulSet-skipper-ingress-redis | sha1sum
97bcb33ef5bafb09bdbf83fc09c11e5f5fc84dad  -
```

PR #7241, opened 2024-04-03
skipper: use name template in topology selector

AlexanderYastrebov:
Follow up on #7236


AlexanderYastrebov:
Maybe its better that canary instance shares selector with main fleet.

PR #7248, opened 2024-04-03
rename configmap in config-provider-service rbac

demonCoder95:
It was requested in a support ticket that a renaming is required of a ConfigMap resource in the ClusterRole for the `config-provider-service`. 

Reference: https://github.bus.zalan.do/zooport/issues/issues/4550


demonCoder95:
👍

mikkeloscar:
:+1:

PR #7249, opened 2024-04-04
Disable kube-node-ready on master nodes

mikkeloscar:
Similar to #7074, but for production

`kube-node-ready` is becoming less and less relevant and has the following issues in the current setup:

* Slows down recovery of master nodes in case of issues with the control plane
* Depends on ec2 instance metadata endpoint which we want to restrict _eventually_
* The ASG lifecycle hook signal is not relevant anymore since we have pod disruption budgets etc.

For this reason, after running with it disabled in test for a month also disable it from production clusters by default. Removing this from masters will also let us get rid of kube2iam on master nodes.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #7256, opened 2024-04-04
Updates cluster autoscaler to use k8s v1.27.12 client

mikkeloscar:
Update the Cluster Autoscaler to use v1.27.12 k8s client. The version is still called v1.18.2 because that is the CA version it's based on.

### DIFF
--- a/cluster/manifests/kube-cluster-autoscaler/daemonset.yaml
+++ b/cluster/manifests/kube-cluster-autoscaler/daemonset.yaml
@@ -36,7 +36,7 @@ spec:
         effect: NoSchedule
       containers:
       - name: cluster-autoscaler
-        image: container-registry.zalando.net/teapot/kube-cluster-autoscaler:v1.18.2-internal.44
+        image: container-registry.zalando.net/teapot/kube-cluster-autoscaler:v1.18.2-internal.47
         command:
           - ./cluster-autoscaler
           - --v={{.Cluster.ConfigItems.autoscaling_autoscaler_log_level}}


mikkeloscar:
:+1:

cluster/manifests/kube-cluster-autoscaler/daemonset.yaml
```diff
@@ -36,7 +36,7 @@ spec:
         effect: NoSchedule
       containers:
       - name: cluster-autoscaler
-        image: container-registry.zalando.net/teapot/kube-cluster-autoscaler:v1.18.2-internal.44
+        image: container-registry.zalando.net/teapot/kube-cluster-autoscaler:v1.27.5-internal.46
```
demonCoder95: Just a nit, but shouldn't this say `v1.27.12` instead of `v1.27.5` here? 🤔

mikkeloscar: It's a good point and actually it maybe need to be something like `v1.18.2-v1.27.12` because the CA version this is based on is v1.18.2, but with patches to update client-go to v1.27.12. I put `v1.27.5` when updating CA because that is the CA version I looked at to make it build, but there are 100s if not 1000s of commits from v1.27.5 CA not in our fork :/

mikkeloscar: @demonCoder95 I propose to revert the version numbering here: https://github.com/zalando-incubator/autoscaler/pull/110

mikkeloscar: Image is now named after the CA version it's based on v1.18.2

mikkeloscar:
:+1:

szuecs:
:+1:

PR #7267, opened 2024-04-08
Document the open_sg_ingress_ranges config-item

mikkeloscar:
The config-item `open_sg_ingress_ranges` can be used when we need to add custom ingress rules to the worker security group e.g. if an application is deploying a service type load balancer with an NLB.

The config-item is used in a few places, but the format was not documented here, making it hard to know about. Add documentation and reference to the implementation in CLM.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #7270, opened 2024-04-08
Disable kube2iam on master nodes in production

mikkeloscar:
With #7249 we can also exclude kube2iam on master nodes because nothing depend on it running there.

This would further enable running master nodes with IMDSv2 enforced (#6669)

Similar to #7083 but for production clusters.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #7275, opened 2024-04-09
e2e: Make decomission step more graceful

mikkeloscar:
Don't fail the `decomission-cluster` step if the stack is already deleted. We do this by checking if the response of `describe-stacks` is `<stack> does not exist`.


szuecs:
:+1:

RomanZavodskikh:
:+1:

PR #7279, opened 2024-04-09
skipper: update dependencies and prepare poc

szuecs:
skipper: update dependencies and prepare poc

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -1,5 +1,5 @@
 {{ $internal_version := "v0.21.40-868" }}
-{{ $canary_internal_version := "v0.21.40-868" }}
+{{ $canary_internal_version := "v0.21.48-876" }}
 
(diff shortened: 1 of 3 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -133,6 +133,11 @@ spec:
             secretKeyRef:
               name: skipper-ingress
               key: datadome-api-key
+        - name: KASADA_API_KEY
+          valueFrom:
+            secretKeyRef:
+              name: skipper-ingress
+              key: kasada-api-key
```
mikkeloscar, line 140: I assume this will only be relevant for some clusters for PoC, will skipper-ingress handle this correctly if the secret value is `""` (empty string?)

szuecs: Yes that's not a problem. This is only loaded in case of a lua filter that will be deployed via internal. The same way we did with datadome.

RomanZavodskikh:
:+1:

szuecs:
:+1:

PR #7294, opened 2024-04-11
Update deployment service version.

gargravarr:
Updates deployment-controller and deployment-status-service to the latest version. This version removes support for probing central ingress and routegroup resources in StackSets. We don't need this probing, because we migrated all StackSets to use traffic segments.


gargravarr:
:+1:

mikkeloscar:
:+1:

PR #7305, opened 2024-04-12
remove okta_auth_enabled config item because it's enabled by default

linki:
Remove the `okta_auth_enabled` config item because it's enabled by default everywhere.

For https://github.bus.zalan.do/teapot/issues/issues/3055


mikkeloscar:
:+1:

linki:
👍

PR #7333, opened 2024-04-16
authnz: Add support for okta postgres-admin group

mikkeloscar:
Updates the `k8s-authnz-webhook` to add support for the `okta:common/postgres-admin` group which is the group name when authentication via okta compared to `CollaboratorPostgresAdmin` as it's called in ZACK.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #7349, opened 2024-04-18
skipper: setup lua environment for operators

szuecs:
skipper: setup lua environment such that operators can create lua files via configmap to be used in routes

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -109,7 +109,8 @@ skipper_validate_query_log: "false"
 
 skipper_default_filters: 'disableAccessLog(2,3,404,429) -> fifo(2000,20,"1s")'
 skipper_default_filters_append: 'stateBagToTag("auth-user", "client.uid")'
-skipper_disabled_filters: "lua,static,bearerinjector"
+skipper_disabled_filters: "static,bearerinjector"
+skipper_lua_sources: "file"
 skipper_edit_route_placeholders: ""
 skipper_ingress_inline_routes: ""
 skipper_ingress_refuse_payload: ""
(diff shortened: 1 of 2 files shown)


cluster/config-defaults.yaml
```diff
@@ -109,7 +109,8 @@ skipper_validate_query_log: "false"
 
 skipper_default_filters: 'disableAccessLog(2,3,404,429) -> fifo(2000,20,"1s")'
 skipper_default_filters_append: 'stateBagToTag("auth-user", "client.uid")'
-skipper_disabled_filters: "lua,static,bearerinjector"
+skipper_disabled_filters: "static,bearerinjector"
+skipper_lua_sources: "file"
```
AlexanderYastrebov, line 113: Maybe we can reuse existing `skipper_lua_scripts_enabled` instead?

szuecs: 1) why?
2) I don't really understand your proposal

AlexanderYastrebov: We already have `skipper_lua_scripts_enabled` flag that we can check instead of  adding another `skipper_lua_sources`.
Or do you want to have a way to allow lua inline script selectively as well?

AlexanderYastrebov:
:+1:

cluster/config-defaults.yaml
szuecs: I need only in 3 clusters lua file and whatever we need in some other edge case.
It feels wrong to me to use `skipper_lua_scripts_enabled` as config.

szuecs:
:+1:

szuecs:
load test has ok enough results

PR #7359, opened 2024-04-22
Zmon role can list tags for IAM certificates

diegoroccia:
As discussed, I need Zmon AWS agent to be able to collect tags for certificates, so that I can be more specific when targeting ACM Certificates with checks


vetinari:
:+1:

szuecs:
Please push to origin, thanks

PR #7368, opened 2024-04-23
Fix VPA limit for ebs-csi-controller

mikkeloscar:
In https://github.com/zalando-incubator/kubernetes-on-aws/pull/6931 we introduced dynamic scaling for control plane components based on the master node instance type size.

For the ebs-csi-controller we did not consider that it has more than 3 containers and we applied the total limit to each container making the CPU limit go above what is possible on the smallest instance type `r6g.large`. Fix this in the following way:

1. Don't autoscale the containers we never intended to autoscale
2. Divide the limit by 3 such that the total CPU limit on the smallest node becomes: `10 + 10 + (3*11) ~= 53` (target was 55).

### DIFF
--- a/cluster/manifests/03-ebs-csi/vpa.yaml
+++ b/cluster/manifests/03-ebs-csi/vpa.yaml
@@ -15,12 +15,16 @@ spec:
     updateMode: Auto
   resourcePolicy:
     containerPolicies:
+    - containerName: ebs-plugin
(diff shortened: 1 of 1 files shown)


linki:
👍

linki:
👍

mikkeloscar:
:+1:

PR #7372, opened 2024-04-23
e2e: enable access via Okta

mikkeloscar:
Enable access to e2e clusters via Okta by using a static client id: `kubernetes.cluster.teapot-e2e` for all e2e clusters. This will allow people with a role for the `teapot-e2e` account access to all e2e clusters in this account.


mikkeloscar:
:+1:

linki:
👍

PR #7374, opened 2024-04-23
Make jammy the default for worker nodes

mikkeloscar:
Make `jammy` (Ubuntu 22.04) the default for worker nodes as we're progressing with the roll out.

All clusters that has to run `focal` already have the config-item set.

This will ensure new clusters start with `jammy` as default.


mikkeloscar:
:+1:

linki:
👍

PR #7376, opened 2024-04-23
Increase timeout for gpu e2e test

mikkeloscar:
The GPU e2e test is often flaky because a GPU node is not created in time and therefore the pod doesn't get scheduled.

The default timeout is 5 min. but because GPU nodes have local storage, where we have a special setup, it sometimes take several minutes to get a GPU node ready to run a pod. Increase the timeout to reduce the chance of a node not getting ready in a reasonable time.


mikkeloscar:
:+1:

linki:
👍

PR #7407, opened 2024-04-29
Make etcd DNS record configurable

mikkeloscar:
This makes the etcd DNS record configurable and default to `etcd-server.<region>.<hosted-zone>` instead of `etcd-server.etcd`.

The AMI is updated to accept the `ETCD_DNS_RECORD_PREFIXES` environment variable which enables it to configure the etcd DNS record name based on that instead of relying on the CF stack name as it was before.

## TODO

* [x] Ensure all existing clusters have `etcd_dns_record_prefixes=etcd-server.etcd` set to avoid changes for those.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -650,6 +650,7 @@ blocked_availability_zone: ""
 
 # etcd cluster
 etcd_stack_name: "etcd-cluster-etcd"
+etcd_dns_record_prefixes: "etcd-server.{{.Cluster.Region}}"
 
 {{if eq .Cluster.Environment "production"}}
 etcd_instance_count: "5"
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

demonCoder95:
👍

cluster/config-defaults.yaml
```diff
@@ -650,6 +650,7 @@ blocked_availability_zone: ""
 
 # etcd cluster
 etcd_stack_name: "etcd-cluster-etcd"
+etcd_dns_record_prefixes: "etcd-server.{{.Cluster.Region}}"
```
demonCoder95, line 653: nit: why do we have a plural name? `prefixes` instead of `prefix`? Just curious.

mikkeloscar: It's because we allow multiple for migration. Let me add a comment in a follow up to make this more clear

PR #7421, opened 2024-04-30
Add support for additional cluster regions

zaklawrencea:
This PR contains additions for enabling the creation of a Kubernetes cluster in additional regions:
- update the node pool mappings to be region specific
- rename the etcd IAM role name to include the region
- update kube-metrics-adapter to use the current region

### DIFF
--- a/cluster/manifests/kube-metrics-adapter/deployment.yaml
+++ b/cluster/manifests/kube-metrics-adapter/deployment.yaml
@@ -40,7 +40,7 @@ spec:
         - --scaling-schedule-default-scaling-window={{.Cluster.ConfigItems.kube_metrics_adapter_default_scaling_window}}
         - --scaling-schedule-ramp-steps={{.Cluster.ConfigItems.kube_metrics_adapter_scaling_schedule_ramp_steps}}
         - --aws-external-metrics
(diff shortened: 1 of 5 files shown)


cluster/manifests/kube-metrics-adapter/deployment.yaml
```diff
@@ -42,6 +42,7 @@ spec:
         - --aws-external-metrics
         - --aws-region=eu-central-1
         - --aws-region=eu-west-1
+        - --aws-region=eu-north-1
```
mikkeloscar: At some point we enabled support for scaling based on SQS queues in `eu-west-1` even if the cluster was in `eu-central-1`. That is why the config looks like this. We should double check if this is used at all and only enable the local region. I would propose that for now we change `--aws-region=eu-central-1` to `--aws-region={{.Cluster.Region}}` and drop the flag for `--aws-region=eu-north-1` such that we only have local region + `eu-west-1` which may go away again.

zaklawrencea: Sure, lets do that for now. Updated.

katyanna:
:+1:

zaklawrencea:
:+1:

PR #7437, opened 2024-05-03
Introduce separate config item for pod factory configmap deletion protection

linki:
This will allow us to toggle configmap deletion protection for `running pods` and `pod factory objects` separately.

The former is widely accepted but the latter doesn't work everywhere. In order to make progress, let's separate them and enable the former everywhere while we think what we do with the latter.


mikkeloscar:
:+1:

linki:
👍

PR #7454, opened 2024-05-07
Update VPA to v1.1.1

mikkeloscar:
Updates the VPA to v1.1.1: https://github.com/kubernetes/autoscaler/releases/tag/vertical-pod-autoscaler-1.1.1

The tag changed because the VPA is now built differently with patches applied instead of from a fork.

Keeping the logic to switch back to the previous version via `vertical_pod_autoscaler_version=legacy`.


mikkeloscar:
:+1:

katyanna:
👍

PR #7473, opened 2024-05-08
Add 7i instances as default option in Karpenter

mikkeloscar:
AWS has _some_ availability of the latest generation intel instances under the name `7i`. Add those instance types to karpenter default pool setup such that we can get some of those, if available. If no available it will just fall back to 5th gen and 6i as before.


myaser:
👍🏻

mikkeloscar:
:+1:

PR #7474, opened 2024-05-08
Remove Kubernetes-on-Ubuntu v1.28 images

zaklawrencea:
This PR removes the default Kubernetes on Ubuntu `v1.28` AMIs which are no longer in use.


linki:
These four entries aren't in-use anymore because we don't run 1.28 anymore: https://github.com/zalando-incubator/kubernetes-on-aws/commit/4b5adcb76584a99ccdb80f8c3c03925cb5acb685#diff-76b81dccc88262acd16b1c68d1917eb58feef214f836b995f9cc7a140362e56e

We should remove them.

zaklawrencea:
@linki I've updated this PR to remove the images instead.

mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #7511, opened 2024-05-14
deployment-service: use new status-service url [2/2]

mikkeloscar:
Follow up to #7509 

Switch deployment-service status-service URL. This is used both by the controller and the central deployment-service.

#7523 must be fully rolled out before this can be merged to avoid short downtime.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #7523, opened 2024-05-15
Fix deployment-status-service dns name length issue

mikkeloscar:
Follow up to #7509

We have one cluster where the alias is too long for the DNS limit of 63. Do this hack to reduce the length of the record :shrug:


linki:
👍

mikkeloscar:
:+1:

PR #7544, opened 2024-05-17
deployment-service: restrict access to healthcheck endpoint

AlexanderYastrebov:
Add a RouteGroup with a route that restricts access to /healthz.

### DIFF
--- a/cluster/manifests/deployment-service/status-service-ingress.yaml
+++ b/cluster/manifests/deployment-service/status-service-ingress.yaml
@@ -1,19 +1,27 @@
-apiVersion: networking.k8s.io/v1
-kind: Ingress
+apiVersion: zalando.org/v1
+kind: RouteGroup
 metadata:
-  name: "deployment-service-status-service"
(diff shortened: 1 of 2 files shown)


cluster/manifests/deployment-service/status-service-ingress.yaml
```diff
@@ -1,3 +1,4 @@
+# TODO: Remove after RouteGroup rollout
```
szuecs: You can already delete it, because we only do `apply` style. Maybe also add a deletions.yaml entry in [post apply](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/deletions.yaml#L4-L5) to delete the ingress after routegroup apply

AlexanderYastrebov: Makes sense, changed.

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #7545, opened 2024-05-21
Patch kube-scheduler to fix crash

mikkeloscar:
This adds a version of the `kube-scheduler` with this patch applied: https://github.com/kubernetes/kubernetes/pull/124933 to avoid crashes as outlined in: https://github.com/kubernetes/kubernetes/issues/124930

Note: this is simply updating the image which will rotate only master nodes as opposed to update the whole AMI as we usually do since only the kube-scheduler needs an update. We will build a new AMI once the upstream fix is merged and released.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #7593, opened 2024-05-29
Enable configmap deletion protection for e2e

linki:
Let's see which tests actually fail here and check whether we can make them pass.


delivery.yaml
```diff
@@ -31,6 +31,9 @@ pipeline:
         echo "Running go mod tidy modified go.mod and/or go.sum"
         exit 1
       fi
+      for patch in patches/*.patch; do
+        cd e2e_modules/kubernetes && git apply "../../$patch" && cd -
```
mikkeloscar, line 35: The usage of `e2e_modules` was dropped in https://github.com/zalando-incubator/kubernetes-on-aws/pull/7704

linki:
Closing this for now.

PR #7594, opened 2024-05-29
explicitly set 1 replica for external-dns

mikkeloscar:
This is a simple way to ensure that external-dns is started after CLM applies in case it was previously scaled to 0.


linki:
👍

szuecs:
:+1:

PR #7589, opened 2024-05-28
Update etcd-proxy to use static base image

mikkeloscar:
Updates the etcd-proxy image to a version using a static base image + updates the Go dependencies.

PR #7602, opened 2024-05-30
Update etcd-proxy base image

mikkeloscar:
Alternative version of #7589 which doesn't depend on `/bin/sh` which is no longer in the image.


mikkeloscar:
:+1:

katyanna:
👍

PR #7609, opened 2024-05-31
Update to kube2iam 0.11.2 + static base image

mikkeloscar:
https://github.com/jtblin/kube2iam/releases/tag/0.11.2

Updates kube2iam to a version that is custom built with patches to go modules to avoid any CVEs. The build is using the static base image.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #7612, opened 2024-05-31
Set default prefix for external-dns

mikkeloscar:
All clusters have been migrated to use this prefix for external-dns records. Simply make it the default.


mikkeloscar:
:+1:

linki:
👍

PR #7617, opened 2024-06-03
flex karpenter node-pool configurations

myaser:
- allow not specifying instance-types
- allow custom requirements by users
- configure the admission controller with the flags necessary for the scheduling controls feature

related to https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/775

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -1146,3 +1146,9 @@ control_plane_graceful_shutdown: "true"
 #  fs.aio-max-nr = 8388608
 #  fs.inotify.max_user_watches = 100000
 sysctl_settings: ""
+
+
+
+# scheduling_controls
+teapot_admission_controller_scheduling_controls_enabled: "false"
+teapot_admission_controller_scheduling_controls_default_architecture: "amd64"
--- a/cluster/node-pools/worker-karpenter/provisioners.yaml
+++ b/cluster/node-pools/worker-karpenter/provisioners.yaml
@@ -155,6 +155,18 @@ spec:
             - "c7in"
             - "m7in"
             - "r7in"
+#{{ else if (gt (len .NodePool.InstanceTypes) 0) }}
+        - key: "node.kubernetes.io/instance-type"
+          operator: In
+          values:
+#      {{ range $type := .NodePool.InstanceTypes }}
+          - "{{ $type }}"
+#      {{ end }}
+#{{ end }}
+
+# safety guards to prevent the use of unwanted instance types in case the user has not specified any specific instance types
+#{{ if or (eq .NodePool.KarpenterInstanceTypeStrategy "default-for-karpenter") (eq .NodePool.KarpenterInstanceTypeStrategy "not-specified") }}
+        # exclude unwanted sizes
         - key: "karpenter.k8s.aws/instance-size"
           operator: "NotIn"
           values:
@@ -166,14 +178,19 @@ spec:
             - "c5d.large"
             - "m5d.large"
             - "r5d.large"
-#{{ else }}
-        - key: "node.kubernetes.io/instance-type"
-          operator: In
+#{{end}}
+
+#{{ if (index .NodePool.ConfigItems "requirements") }}
+#    {{ range $requirement := .NodePool.KarpenterRequirements }}
+        - key: "{{ $requirement.Key }}"
+          operator: "{{ $requirement.Operator }}"
           values:
-#      {{ range $type := .NodePool.InstanceTypes }}
-          - "{{ $type }}"
-#      {{ end }}
+#         {{ range $value := $requirement.Values }}
+            - "{{ $value}}"
+#         {{ end }}
+#    {{ end }}
 #{{ end }}
+        # other configuration
         - key: "karpenter.sh/capacity-type"
           operator: In
           values:
--- a/cluster/manifests/01-admission-control/config.yaml
+++ b/cluster/manifests/01-admission-control/config.yaml
@@ -52,6 +52,8 @@ data:
 {{- end}}
 
   pod.env-inject.node-options.enable: "{{ .Cluster.ConfigItems.teapot_admission_controller_inject_node_options_environment_variable }}"
+  pod.scheduling-controls.enable: "{{ .Cluster.ConfigItems.teapot_admission_controller_scheduling_controls_enabled }}"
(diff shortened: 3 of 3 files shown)


cluster/node-pools/master-default/userdata.yaml
```diff
@@ -203,7 +203,7 @@ write_files:
             limits:
               memory: {{ .Values.InstanceInfo.MemoryFraction (parseInt64 .Cluster.ConfigItems.apiserver_memory_limit_percent)}}
 {{- end }}
-        - image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/admission-controller:master-206
+`        - image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/staging_namespace/teapot/admission-controller:pr-203-7
```
myaser: testing. 
do not merge

cluster/config-defaults.yaml
```diff
@@ -1091,3 +1091,8 @@ sysctl_settings: ""
 
 # enables/disables the minDomains field for pod topology spread.
 min_domains_in_pod_topology_spread_enabled: "true"
+
+
+# scheduling_controls
+teapot_admission_controller_scheduling_controls_enabled: "false"
+teapot_admission_controller_scheduling_controls_default_architecture: "amd64"
```
demonCoder95: nit: Missing new line 🙂

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -112,7 +112,23 @@ spec:
       # Operators { In, NotIn, Exists, DoesNotExist, Gt, and Lt } are supported.
       # https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#operators
       requirements:
-#{{ if and (eq (len .NodePool.InstanceTypes) 1) (eq (index .NodePool.InstanceTypes 0) "default-for-karpenter") }}
+#{{ if eq .NodePool.KarpenterInstanceTypeStrategy "default-for-karpenter"  "not-specified" }}
```
demonCoder95: Do we want two values here? I mean, I see `default-for-karpenter` is being handled below anyway, maybe this should be `not-specified` only?

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -143,23 +159,27 @@ spec:
...
+#    {{ range $requirement := .NodePool.KarpenterRequirements }}
+        - key: "{{ $requirement.Key }}"
+          operator: "{{ $requirement.Operator }}"
+          values:
+#         {{ range $value := $requirement.Values }}
+            - "{{ $value}}"
+#         {{ end }}
+#    {{ end }}
+#{{ end }}
+        # other configuration
```
demonCoder95: Are these commented lines here for documentation purpose? It looks a little confusing, maybe we can just put an example configuration in a large comment block above this? just to seperate out the documenting comments and the actual configuration making it easier to read?

cluster/node-pools/worker-karpenter/provisioners.yaml
myaser: no, it is meant for both cases
in this block we exclude certain sizes and instance types when the configurations are very broad
this happen in 2 cases, when we use the default set of instance-families (default for karpenter) , and when we do not restrict instance types at all  (not specified)
then we exclude certain sizes
when the user provides their own set of instance-types. we use them as is with out further limitation

cluster/node-pools/worker-karpenter/provisioners.yaml
myaser: things like
        # other configuration
        # user's custom configurations
are definately comments for cotext

however, `#    {{ range $requirement := .NodePool.KarpenterRequirements }}` and so on, are no comments. they are functional go templates.

they are comments in the yaml, but go templates can render them without a problem. this way I can get the IDE to parse the yaml file correctly

cluster/node-pools/worker-karpenter/provisioners.yaml
demonCoder95: I see, makes sense.
So are these conditions `OR`ed together? Or `AND`ed? I don't see an operator. Because it will only work like you suggest in case of an `AND`.

cluster/node-pools/worker-karpenter/provisioners.yaml
demonCoder95: Just looks a bit confusing, maybe we can remove them later when we merge, so it doesn't look confusing like this?

cluster/node-pools/worker-karpenter/provisioners.yaml
myaser: I am trying to do "if KarpenterInstanceTypeStrategy in {"default-for-karpenter"  "not-specified"}
but maybe I am doing it wrong. will check again

szuecs:
Please squash rebase the commits into one to keep the history clean

cluster/config-defaults.yaml
```diff
@@ -1146,3 +1146,11 @@ control_plane_graceful_shutdown: "true"
 #  fs.aio-max-nr = 8388608
 #  fs.inotify.max_user_watches = 100000
 sysctl_settings: ""
+
+# enables/disables the minDomains field for pod topology spread.
+min_domains_in_pod_topology_spread_enabled: "true"
```
mikkeloscar: This field doesn't seem to be used anywhere?

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -155,25 +155,43 @@ spec:
...
           values:
             - "metal"
-        # exclude instance-types with slow SSD
+        # exclude unwanted instance-types
         - key: "node.kubernetes.io/instance-type"
           operator: "NotIn"
           values:
-            - "c5d.large"
+            - "c5d.large" # too slow SSD
+        # user's custom configurations
```
mikkeloscar: This looks like it wasn't merged/rebased quite right since: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8406

cluster/config-defaults.yaml
myaser: seem to be a leftover from rebasing. removed it now

demonCoder95: This feature became GA in `1.31` so we got rid of this config-item

mikkeloscar:
:+1:

myaser:
👍🏻

PR #7618, opened 2024-06-03
Don't bind collaborator to PowerUser access

mikkeloscar:
With okta we have accidentially given the `collaborator` poweruser by default. The intention is that they still need to request manual access (engineer role) to be able to do poweruser actions.

Drop binding from `poweruser` to `okta:common/collaborator`


mikkeloscar:
:+1:

katyanna:
👍

PR #7622, opened 2024-06-03
Update stackset-controller to use a static base image

mikkeloscar:
Update stackset-controller to use a static base image.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #7626, opened 2024-06-03
fabric-gateway: update to master-254

AlexanderYastrebov:
* update version to use static base image
* remove version config item as it did not turn out useful


RomanZavodskikh:
:+1:

AlexanderYastrebov:
:+1:

PR #7628, opened 2024-06-04
Make use of kube-scheduler's config file

linki:
Scheduler can make scheduling decision by taking [podTopologySpreads into account globally](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#cluster-level-default-constraints) (without users or admission-controller injecting anything). This needs to be configured in `kube-scheduler`'s config file. This PR introduces the config file so that it can be used later.

I'm getting a better result on a simple deployment. Although it's hard to test on two different clusters because their available capacity is different.

`teapot` cluster

Screenshot 2024-06-17 at 16 03 43

This e2e cluster

Screenshot 2024-06-17 at 16 04 21

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -867,6 +867,19 @@ write_files:
       kind: KubeSchedulerConfiguration
       clientConnection:
         kubeconfig: /etc/kubernetes/scheduler-kubeconfig
+      profiles:
+        - schedulerName: default-scheduler
+          pluginConfig:
+            - name: PodTopologySpread
+              args:
+                defaultConstraints:
+                  - maxSkew: 1
+                    topologyKey: kubernetes.io/hostname
+                    whenUnsatisfiable: ScheduleAnyway
+                  - maxSkew: 1
+                    topologyKey: topology.kubernetes.io/zone
+                    whenUnsatisfiable: ScheduleAnyway
+                defaultingType: List
 
 {{ if eq .Cluster.ConfigItems.event_rate_limit_enable "true"}}
   # Enabling EventRateLimit admission plugin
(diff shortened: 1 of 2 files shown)


cluster/node-pools/master-default/userdata.yaml
```diff
@@ -853,6 +859,25 @@ write_files:
...
+      profiles:
+        - schedulerName: default-scheduler
+          pluginConfig:
+            - name: PodTopologySpread
+              args:
+                defaultConstraints:
+                  - maxSkew: 1
+                    topologyKey: topology.kubernetes.io/zone
+                    whenUnsatisfiable: ScheduleAnyway
+                defaultingType: List
```
mikkeloscar, line 880: Let's put it behind a config-item. Can be on by default but let us have a quick way to turn it off if it's not behaving as expected.

cluster/node-pools/master-default/userdata.yaml
```diff
@@ -666,7 +666,7 @@ write_files:
         - name: kube-scheduler
           image: nonexistent.zalan.do/teapot/{{if eq .Cluster.ConfigItems.kubernetes_scheduler_image "zalando" }}kube-scheduler-internal{{else}}kube-scheduler{{end}}:fixed
           args:
-          - --kubeconfig=/etc/kubernetes/scheduler-kubeconfig
```
mikkeloscar, line 669: Why can this be removed? 🤔

linki: This flag became part of the config file itself: https://github.com/zalando-incubator/kubernetes-on-aws/pull/7628/files#diff-fb712e77981d2e840a03b53964fd2fb37d9cd137e35b2db4d1dbb36f5170fb44R869

There are a handful of other potentially interesting knobs we can turn in the scheduler config. Full reference is here: https://kubernetes.io/docs/reference/config-api/kube-scheduler-config.v1/#kubescheduler-config-k8s-io-v1-KubeSchedulerConfiguration

image

mikkeloscar: Ah got it, missed that it was defined below :+1:

mikkeloscar:
:+1:

linki:
Will need another 👍  I changed the default value to `true` now. Wanted to see if it passes with `false` as well.

linki:
👍

mikkeloscar:
:+1:

PR #7642, opened 2024-06-06
`journald-reader`: mount `/usr` to access `libsystemd` and `python`

linki:
This allows to run `journald-reader` with a minimal/static image.

Requires the latest AMI (>=master-336), so let's merge this after/with the next AMI update.

DO NOT MERGE: AMI update and DaemonSet update have to be executed in two different rollouts.


linki:
👍

linki:
This needs to be rolled out separately from the AMI update otherwise it won't work (journald-reader updates before worker nodes are rolled). Hence this will break when we merge `kube-1.30` with `stable` eventually. Let's just wait until the needed AMI is used in all channels.

mikkeloscar:
:+1:

linki:
👍

PR #7653, opened 2024-06-06
Update default PHC configuration items

RomanZavodskikh:
It is completely safe to do so because the same configuration is already applied to all clusters.


RomanZavodskikh:
:+1:

MustafaSaber:
👍

PR #7662, opened 2024-06-10
Revert "Update skipper version, step 1/2"

RomanZavodskikh:
This reverts commit 7ca2d495f394466dad6322f9ea78603cc7a1eb3a because of skipper pod seems to not work https://github.com/zalando-incubator/kubernetes-on-aws/pull/7655#issuecomment-2157603705 .


RomanZavodskikh:
It seems root cause of the problem may be reverted by this https://github.com/zalando-incubator/kubernetes-on-aws/pull/7663, not by the current one.

PR #7666, opened 2024-06-10
Remove legacy Runtime policy toggle

mikkeloscar:
This config is not used in the admission-controller, so remove it to avoid confusion about the feature being enable or not. It is enabled by node pool by default.


myaser:
👍🏻

mikkeloscar:
:+1:

PR #7670, opened 2024-06-11
Update admission-controller drop runtime policy

mikkeloscar:
Follow up to #7666 

Removes the legacy runtime policy config from admission-controller code.

#7666 must be fully rolled out before merging.


linki:
👍

mikkeloscar:
:+1:

PR #7680, opened 2024-06-11
collect node profile, will be used for enriching metrics on Karpenter

myaser:
collecting the node-profile of a node. from which we can know if the node is karpenter node or CA node.
this can help generating meaningful metrics on how karpenter perform VS CA


mikkeloscar:
:+1:

myaser:
👍🏻

PR #7683, opened 2024-06-11
Allow readonly and poweruser roles access to configmaps/status

linki:
Not sure if these sub-resources really exist but `kuberay` defines them [here](https://github.com/ray-project/kuberay/blob/40a946a57f9aae943d7deae13a54bb1ec6288621/helm-chart/kuberay-operator/templates/leader_election_role.yaml#L23-L27). This allows `deployment-service` to deploy RBAC permissions that access `configmaps/status` which should be fairly safe.

For https://github.bus.zalan.do/zooport/issues/issues/4764

### DIFF
--- a/cluster/manifests/roles/poweruser-role.yaml
+++ b/cluster/manifests/roles/poweruser-role.yaml
@@ -19,6 +19,7 @@ rules:
   - ''
   resources:
   - configmaps
+  - configmaps/status
   - endpoints
   - limitranges
   - namespaces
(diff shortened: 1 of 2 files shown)


linki:
Let's not do this. It's a mistake in kuberay's template files.

Upstream issue in kubebuilder: https://github.com/kubernetes-sigs/kubebuilder/issues/1473
and the fix: https://github.com/kubernetes-sigs/kubebuilder/pull/1511/files

PR #7695, opened 2024-06-14
fix ebs-csi-node RBAC for ebs-csi-driver

demonCoder95:
I notice a lot of RBAC errors reported by the `ebs-csi-driver` in many clusters. Example from `teapot`

```
E0614 08:52:23.098313       1 node.go:875] "Unexpected failure when attempting to remove node taint(s)" err="isAllocatableSet: failed to get CSINode for ip-10-149-93-57.eu-central-1.compute.internal: csinodes.storage.k8s.io \"ip-10-149-93-57.eu-central-1.compute.internal\" is forbidden: User \"system:serviceaccount:kube-system:ebs-csi-node-sa\" cannot get resource \"csinodes\" in API group \"storage.k8s.io\" at the cluster scope: access undecided system:serviceaccount:kube-system:ebs-csi-node-sa/[system:serviceaccounts system:serviceaccounts:kube-system system:authenticated]"
```

This adds the missing RBAC to resolve these issues. I'm assuming the driver will `get` the node and then `patch` it (if needed) so adding both permissions. Can monitor the cluster once we merge to `dev` and see if that if that fixes the problem.

### DIFF
--- a/cluster/manifests/03-ebs-csi/clusterrole-csi-node.yaml
+++ b/cluster/manifests/03-ebs-csi/clusterrole-csi-node.yaml
@@ -9,3 +9,6 @@ rules:
   - apiGroups: [""]
     resources: ["nodes"]
     verbs: ["get", "patch"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["csinodes"]
+    verbs: ["get", "patch"]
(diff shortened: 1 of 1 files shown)


linki:
It's best to align with https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/deploy/kubernetes/base.

Looks like they only give `get` permission: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/deploy/kubernetes/base/clusterrole-csi-node.yaml#L16-L18. We probably only need `patch` on the node objects, which is already configured.

demonCoder95:
Hmm, good point. I'll change it only to add `get`.

linki:
👍

demonCoder95:
👍

PR #7696, opened 2024-06-14
Align deployment-service config key with reality

linki:
The key over at `deployment-service` is actually different: https://github.bus.zalan.do/teapot/deployment-service/blob/454182013bae7d1b906cef9241e1a2cd1c3d28fc/pkg/deployment_config/config.go#L29


linki:
👍

gargravarr:
:+1:

PR #7647, opened 2024-06-06
Update kube-dns images to latest version 1.23.1

mikkeloscar:
Update kube-dns images to latest version 1.23.1

PR #7702, opened 2024-06-18
Add simple DNS e2e test

mikkeloscar:
This adds a simple DNS e2e test which simply runs a pod on a _new_ node such that we test against an update to CoreDNS daemonset.

The test is very basic and only tries to do `nslookup` to `google.com` with a timeout of 1 second (you can't do lower timeout in `nslookup`).
This should be enough to catch the delay we saw with the roll out of #7647

* [ ] Check updating similar as #7647 will be caught by e2e.


mikkeloscar:
Closing for now in favor of #7725

PR #7704, opened 2024-06-18
Improve go.mod for e2e tests

mikkeloscar:
In the past we needed a hack to setup kubernetes repository with Go modules. This is no longer needed and we can simple depend on e.g. `k8s.io/kubernetes v1.29.4` for building. This PR does so and removes the legacy script and configuration needed for the custom git clone of kubernetes/kubernetes.

This should have the side effect of removing false positive dependabot security warnings.


mikkeloscar:
:+1:

myaser:
👍🏻

PR #7705, opened 2024-06-18
update kubernetes-lifecycle-metrics version

myaser:
related:
https://github.bus.zalan.do/teapot/kubernetes-lifecycle-metrics/pull/19

"collecting the node-profile of a node. from which we can know if the node is karpenter node or CA node.
this can help making the SLO (pod startup time) visualize what improvement we have due to karpenter"


myaser:
👍🏻

mikkeloscar:
:+1:

PR #7714, opened 2024-06-19
flip the order of nvidia driver's nodeAffinities to mitigate a karpenter bug

myaser:
it seems that karpenter has a bug where it only considers the first affinity of a daemonset when calculating daemonset resources of a node pool

[here](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/scheduling/requirements.go#L104) it is reading only the first of the affinities relying on an outer loop to remove the affinity and continue with the next one

this loop is not happening for daemonset calculation as seen [here](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/provisioning/scheduling/scheduler.go#L344)


mikkeloscar:
:+1:

zaklawrencea:
👍

PR #7725, opened 2024-06-21
Rotate nodes with old daemonset pods before e2e

mikkeloscar:
The coredns daemonset has update strategy `OnDelete` meaning the pods will only be updated if the node is replaced and a new pod is created for new nodes. This is done to avoid downtime of coredns which would break our DNS availability temporary on a node.

This has the problem that updates to the coredns daemonset is not properly tested during e2e, because most of the nodes will be running the previous version.

This PR adds a small tool `check-daemonset-updated` which is run after CLM during cluster creation for e2e, this ensure that if coredns is updated all the nodes are also rotated until there are no old coredns pods. This increases the time to create an e2e cluster, but only for cases where coredns is updated.

### DIFF
--- a/test/e2e/.gitignore
+++ b/test/e2e/.gitignore
@@ -5,3 +5,4 @@ base_config/
 internal_config.sh
 .kube/
 stackset-e2e
+check-daemonset-updated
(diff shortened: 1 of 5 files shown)


mikkeloscar:
:+1:

linki:
👍

PR #7726, opened 2024-06-21
[RUN e2e] Update kube-dns images to latest version 1.23.1

mikkeloscar:
Test that the rotation of daemonset during e2e introduced in #7725  will catch the broken dnsmasq update in this change.


mikkeloscar:
Closing as this was just validating that #7725 catches the broken dnsmasq update in this PR.

PR #7676, opened 2024-06-11
skipper: add config to enable prometheus start label

AlexanderYastrebov:
See https://github.com/zalando/skipper/pull/3089 for details.

PR #7738, opened 2024-06-24
skipper: set skipper_prometheus_start_label_enabled by default

AlexanderYastrebov:
The config `skipper_prometheus_start_label_enabled` is enabled in all clusters so make it true by default.

See #7676

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -122,7 +122,8 @@ skipper_endpointslices_enabled: "true"
 
 skipper_compress_encodings: "gzip,deflate,br"
 
-skipper_prometheus_start_label_enabled: "false"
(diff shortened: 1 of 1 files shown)


RomanZavodskikh:
:+1:

szuecs:
This needs to show up green. I retrigger the whole pipeline (retrigger the step won't work, because we need to make sure load test is done while executing the cluster update.

 process/stups-test/e2e-load-test-result — Process step failed. 

@RomanZavodskikh I deleted you +1 comment, because of the step needs to be successful, not sure if zappr will re-evaluate if someone adds another +1 but I guess not :D

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

AlexanderYastrebov:
:+1:

PR #7745, opened 2024-06-25
ingress-controller: configure keyValue log parser

AlexanderYastrebov:
Configure built-in keyValue parser for containers using https://github.com/sirupsen/logrus that logs in the format compatible with https://pkg.go.dev/github.com/kr/logfmt

See previous #6535


AlexanderYastrebov:
:+1:

MustafaSaber:
👍

PR #7746, opened 2024-06-25
fix: routesrv metrics port was changed to 9911 a while ago

szuecs:
fix: routesrv metrics port was changed to 9911 a while ago


AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #7758, opened 2024-06-27
Disable cadvisor by default

mikkeloscar:
Disable cadvisor by default.

This is done because the same metrics are provided by the kubelet embedded cadvisor, so there is no need to run/maintain a custom one.

The current clusters have `cadvisor_enabled=true` set via cluster-config-manager so this change is a no-op for existing clusters.


linki:
👍

mikkeloscar:
:+1:

PR #7762, opened 2024-06-28
admission-controller: Native sleep preStop hook

mikkeloscar:
This version of admission-controller injects a preStop hook into pods using the native `preStop.sleep` featue: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-handler-implementations

That is, it's injecting:

```yaml
lifecycle:
  preStop:
    sleep:
      seconds: 20
```

Instead of what it previously did which was:

```yaml
lifecycle:
  preStop:
    exec:
      command: ["sleep", "20"]
``` 

This way a `sleep` binary is no longer a requirement in container images.

This update also updating dependencies to Kubernetes v1.30.2


mikkeloscar:
:+1:

linki:
👍

PR #7763, opened 2024-06-28
Don't set Cluster.ID on etcd LT tags

mikkeloscar:
Don't include `Cluster.ID` in the Name tag of the etcd launch template.

With the tag we create a new launch template on _every_ e2e cluster which is not desirable. Having the cluster.ID is also not meaningful for the etcd launch template.


mikkeloscar:
:+1:

linki:
👍

PR #7774, opened 2024-07-01
Update to Prometheus v2.53.0

mikkeloscar:
* Update to Prometheus v2.53.0 https://github.com/prometheus/prometheus/releases/tag/v2.53.0
* Update init container to use amazonlinux-2023-slim instead of alpine as it has no CVEs.


mikkeloscar:
:+1:

katyanna:
👍

PR #7792, opened 2024-07-02
update deployment-service deps for k8s v1.30

demonCoder95:
As a follow-up for updating dependencies in the deployment-service to use Kubernetes v1.30 based libraries, this PR is for rolling out the new version.


demonCoder95:
👍

katyanna:
👍

PR #7796, opened 2024-07-03
fabric: enable on all clusters by default

MustafaSaber:
- [ ] clean config items
- [ ] clean fabric docs

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -335,14 +335,8 @@ skipper_ingress_routegroup_crd_require_hosts: "true"
 skipper_open_policy_agent_enabled: "false"
 skipper_open_policy_agent_styra_token: ""
 
-#
 # FabricGateway controller config
-#
-# fabric_gateway_controller_mode:
-#   - disabled: scales controller to zero replicas
-#   - production: runs the controller
-#
-fabric_gateway_controller_mode: "production"
+
 fabric_gateway_controller_cpu: "50m"
 fabric_gateway_controller_memory: "150Mi"
 fabric_gateway_controller_allow_all_filters: "false"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -338,11 +338,11 @@ skipper_open_policy_agent_styra_token: ""
 #
 # FabricGateway controller config
 #
-# fabric_gateway_controller_mode:
-#   - disabled: scales controller to zero replicas
-#   - production: runs the controller
+# fabric_gateway_controller_enabled:
+#   - false: scales controller to zero replicas
+#   - true: runs the controller
 #
-fabric_gateway_controller_mode: "production"
+fabric_gateway_controller_enabled: "true"
 fabric_gateway_controller_cpu: "50m"
 fabric_gateway_controller_memory: "150Mi"
 fabric_gateway_controller_allow_all_filters: "false"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -335,8 +335,14 @@ skipper_ingress_routegroup_crd_require_hosts: "true"
 skipper_open_policy_agent_enabled: "false"
 skipper_open_policy_agent_styra_token: ""
 
+#
 # FabricGateway controller config
-
+#
+# fabric_gateway_controller_mode:
+#   - disabled: scales controller to zero replicas
+#   - production: runs the controller
+#
+fabric_gateway_controller_mode: "production"
 fabric_gateway_controller_cpu: "50m"
 fabric_gateway_controller_memory: "150Mi"
 fabric_gateway_controller_allow_all_filters: "false"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -342,10 +342,9 @@ skipper_open_policy_agent_styra_token: ""
 #   - disabled: scales controller to zero replicas
 #   - production: runs the controller
 #
-fabric_gateway_controller_mode: "disabled"
+fabric_gateway_controller_mode: "production"
 fabric_gateway_controller_cpu: "50m"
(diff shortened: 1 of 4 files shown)


cluster/config-defaults.yaml
```diff
@@ -342,10 +342,10 @@ skipper_open_policy_agent_styra_token: ""
 #   - disabled: scales controller to zero replicas
 #   - production: runs the controller
 #
-fabric_gateway_controller_mode: "disabled"
+fabric_gateway_controller_mode: "production"
 fabric_gateway_controller_cpu: "50m"
 fabric_gateway_controller_memory: "150Mi"
-fabric_gateway_crd_v1_enabled: "false"
+fabric_gateway_crd_v1_enabled: "true"
```
szuecs: Let's drop all config items that are only to "enable" it.
https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/fabric-gateway/fabricgateway_crd.yaml#L1

cluster/config-defaults.yaml
```diff
@@ -342,10 +342,10 @@ skipper_open_policy_agent_styra_token: ""
 #   - disabled: scales controller to zero replicas
 #   - production: runs the controller
 #
-fabric_gateway_controller_mode: "disabled"
+fabric_gateway_controller_mode: "production"
```
szuecs: Let's drop all config items that are only to "enable" it.
https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/fabric-gateway/deployment.yaml#L11-L17

MustafaSaber: I think this is nice to have, so we have the crd everywhere but can control the controller

MustafaSaber: decided to also remove, and introduce  `fabric_gateway_controller_replicas` when we need

cluster/manifests/fabric-gateway/deployment.yaml
```diff
@@ -8,13 +8,7 @@ metadata:
     application: gateway-operator
     component: fabric-gateway-controller
     version: "{{ $version }}"
-    mode: "{{ .Cluster.ConfigItems.fabric_gateway_controller_mode }}"
 spec:
-{{ if eq .Cluster.ConfigItems.fabric_gateway_controller_mode "disabled" }}
-  replicas: 0
-{{ else }}
-  replicas: 1
```
RomanZavodskikh, line 16: How many replicas there will be if spec.replicas is not specified?

MustafaSaber: 1, I also mentioned this in the commit message

```
➜ kubectl explain deployments.spec.replicas
KIND:     Deployment
VERSION:  apps/v1

FIELD:    replicas <integer>

DESCRIPTION:
     Number of desired pods. This is a pointer to distinguish between explicit
     zero and not specified. Defaults to 1.
```

szuecs: 0

MustafaSaber: changed to have the field but it actually defaults to 1

cluster/config-defaults.yaml
```diff
@@ -335,17 +335,10 @@ skipper_ingress_routegroup_crd_require_hosts: "true"
...
 skipper_open_policy_agent_styra_token: ""
 
-#
 # FabricGateway controller config
-#
-# fabric_gateway_controller_mode:
-#   - disabled: scales controller to zero replicas
-#   - production: runs the controller
-#
-fabric_gateway_controller_mode: "disabled"
```
AlexanderYastrebov, line 345: Should we keep this to disable controller just in case. The same could be achieved by  scaling to 0 while cluster updates are blocked.
OTOH we never used it after rollout.

AlexanderYastrebov:
:+1:

cluster/config-defaults.yaml
szuecs: That was my internal suggestion to have a config item for the number of replicas that we set here to 1 as default.

AlexanderYastrebov: Then lets keep `fabric_gateway_controller_mode` as is and set to "production" by default.
We do not want more than one controller running anyway and controller may have multiple worker threads if needed (not used currently).

cluster/config-defaults.yaml
```diff
@@ -335,8 +335,14 @@ skipper_ingress_routegroup_crd_require_hosts: "true"
 skipper_open_policy_agent_enabled: "false"
 skipper_open_policy_agent_styra_token: ""
 
+#
 # FabricGateway controller config
-
+#
+# fabric_gateway_controller_enabled:
+#   - false: scales controller to zero replicas
+#   - ttue: runs the controller
```
AlexanderYastrebov: typo

cluster/manifests/fabric-gateway/deployment.yaml
```diff
@@ -8,9 +8,8 @@ metadata:
     application: gateway-operator
     component: fabric-gateway-controller
     version: "{{ $version }}"
-    mode: "{{ .Cluster.ConfigItems.fabric_gateway_controller_mode }}"
 spec:
-{{ if eq .Cluster.ConfigItems.fabric_gateway_controller_mode "disabled" }}
+{{ if eq .Cluster.ConfigItems.fabric_gateway_controller_enabled "false" }}
```
AlexanderYastrebov: minor: usually we check if it equals "true"

AlexanderYastrebov:
:+1:

cluster/manifests/fabric-gateway/deployment.yaml
MustafaSaber: done

szuecs:
:+1:

MustafaSaber:
👍

PR #7798, opened 2024-07-03
Update preStop hooks

AlexanderYastrebov:
Use builtin sleep hook instead of running sleep binary.

### DIFF
--- a/cluster/manifests/skipper/skipper-redis.yaml
+++ b/cluster/manifests/skipper/skipper-redis.yaml
@@ -74,8 +74,8 @@ spec:
             memory: {{ .Cluster.ConfigItems.skipper_redis_memory }}
         lifecycle:
           preStop:
-            exec:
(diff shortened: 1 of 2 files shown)


cluster/manifests/skipper/skipper-redis.yaml
```diff
@@ -75,7 +75,8 @@ spec:
         lifecycle:
           preStop:
             exec:
-              command: ["sleep", "30"]
+              sleep:
```
mikkeloscar: The `exec` should be replaced by `sleep`

Thanks for following up on this!

AlexanderYastrebov: Late night PR :facepalm: 
Fixed!

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #7800, opened 2024-07-04
added additional pod.pod-security-policy.allowed-restricted-capabilities

doggopadre:
added additional `pod.pod-security-policy.allowed-restricted-capabilities`


mikkeloscar:
This is not the right place to add these, it should be done via the config-item: `teapot_admission_controller_pod_security_policy_additional_restricted_capabilities: "<comma separated list of new capabilities>"`

PR #7805, opened 2024-07-05
eks: Use known ipv4 service CIDR range

mikkeloscar:
* Use the same service cidr range as we use in non-eks clusters: `10.5.0.0/16` (This is not compatible with existing clusters, so recreation is needed). The motivation for this is:
   1. Things like `ingress.cluster.local` endpoints will _just work_
   2. It's helpful to use a known range dedicated for Kubernetes service CIDR so we don't make new assumptions about what private ipv4 ranges are available.
* Update the aws-node daemonset to the latest version which is provided by default in eks v1.30


mikkeloscar:
:+1:

cluster/manifests/01-aws-node/01-sa.yaml
mikkeloscar, line 1: This was renamed to try and reduce the chance of this bug: https://github.com/aws/amazon-eks-pod-identity-webhook/issues/174

gargravarr:
:+1:

PR #7808, opened 2024-07-05
[EKS] Change Kubernetes control plane to EKS (the `eks` branch)

mikkeloscar:
This branch contains a Proof of Concept setup for running a Zalando Kubernetes cluster using EKS as the control plane. All the EKS specific configuration is behind feature flags: `{{ if eq .Cluster.Provider "zalando-eks"}}` making it compatible with non-eks setup.

This PR is open to facilitate merging changes from `dev` into the branch regularly and to get it into a state that makes it mergeable.

~~**Note**: This branch currently only works with a modified version of CLM: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/777~~


cluster/cluster.yaml
```diff
@@ -2580,22 +3335,37 @@ Resources:
...
+  # EKSServiceIpv6Cidr:
+  #   Export:
+  #     Name: "{{.Cluster.ID}}:eks-service-cidr"
+  #   Value: !GetAtt EKSCluster.KubernetesNetworkConfig.ServiceIpv6Cidr
+  EKSOpenIdConnectIssuerUrl:
     Export:
-      Name: '{{.Cluster.ID}}:worker-iam-role'
-    Value: !Ref WorkerIAMRole
-  RemoteFilesEncryptionKey:
+      Name: "{{.Cluster.ID}}:eks-oidc-url"
```
linki, line 3349: Should we use `.Cluster.Name` instead of `.ID` here (and other places?). PR that introduced the field: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/829/files#diff-61dbe200b05c4fa62230dc5e9efedf4ac8d35088d3dfb7a1cde9163b79b41066

cluster/manifests/01-coredns-local/daemonset-coredns.yaml
```diff
@@ -58,7 +66,11 @@ spec:
             command:
             - dig
             - "+short"
+{{- if and (eq .Cluster.Provider "zalando-eks") (eq .Cluster.ConfigItems.eks_ip_family "ipv4") }}
             - "@127.0.0.1"
+{{- else }}
+            - "::1"
```
linki, line 72: Looks like we enter this branch for legacy clusters (IPv4) as well. Does this make sense?

linki: And should it be `@::1`?

cluster/manifests/01-coredns-local/daemonset-coredns.yaml
```diff
@@ -80,9 +92,14 @@ spec:
...
       - name: unbound-telemetry
+{{- if eq .Cluster.Provider "zalando-eks" }}
+        image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/unbound-telemetry:master-5
+{{- else }}
         image: container-registry.zalando.net/teapot/unbound-telemetry:master-5
+{{- end }}
         args:
         - tcp
+        # TODO: ipv6
         - --bind=0.0.0.0:9054
```
linki, line 103: Should this bind to all interfaces (for IPv6)?

cluster/manifests/01-coredns-local/daemonset-coredns.yaml
```diff
@@ -123,8 +144,8 @@ spec:
         - --neg-ttl=60
         # send requests to the last server first, only fallback to the previous ones if it's unreachable
         - --strict-order
-        - --server=10.5.0.11#53
-        - --server=127.0.0.1#9254
+        - --server=10.5.0.11#53 # TODO: fix this for ipv6
```
linki, line 147: Should this be an IPv6 address?

cluster/manifests/02-admission-control/teapot.yaml
```diff
@@ -290,11 +519,16 @@ webhooks:
     matchConditions:
       - name: 'exclude-privileged-groups'
         expression: 'request.userInfo.groups.all(g, !(g in ["okta:common/administrator", "zalando:administrator"]))'
-      - name: 'exclude-postgres-admins'
-        expression: 'request.userInfo.groups.all(g, !(g in ["okta:common/postgres-admin"]))'
```
linki, line 294: Should we keep that for non-eks clusters?

linki: It doesn't apply to non-eks. So let's keep it like that for now.

cluster/manifests/external-dns/deployment.yaml
```diff
@@ -52,6 +53,7 @@ spec:
         - --aws-zones-cache-duration={{ .Cluster.ConfigItems.external_dns_zones_cache_duration }}
         - --annotation-filter=external-dns.alpha.kubernetes.io/exclude notin (true)
         - --policy={{ .Cluster.ConfigItems.external_dns_policy }}
+        - --log-level=debug
```
linki, line 56: Let's drop this at some point.

cluster/manifests/02-admission-control/teapot.yaml
linki: Nevermind, this is fine.

linki:
👍

cluster/manifests/01-coredns-local/daemonset-coredns.yaml
mikkeloscar: This is unbound which I didn't invest in making ipv6 compatible yet. We use dnsmasq in eks clusters for now so we can defer fixing this to later.

cluster/manifests/01-coredns-local/daemonset-coredns.yaml
mikkeloscar: I think in practice this address is never used as it's a fallback if the local CoreDNS is not available. Hence we didn't notice issues in IPv6 clusters.

We should fix this, but I would say we don't have to fix it before merging this.

cluster/manifests/01-coredns-local/daemonset-coredns.yaml
mikkeloscar: This is unbound which is not yet being activated in EKS clusters. Let's defer fixing this to later.

mikkeloscar:
:+1:

PR #7810, opened 2024-07-05
EKS branch cleanup

mikkeloscar:
* Drop the `eks` config-item no longer used since we use `Provider` to determine eks or not.
* Use the constant service IPs for coreDNS and skipper-internal service in ipv4 mode, possible since: #7805 
* Cleanup changes to external-dns which is no longer needed


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #7818, opened 2024-07-05
eks: PoC: Zalando IAM AWS Proxy

mikkeloscar:
This adds a Proof of Concept proxy which can translate from Zalando IAM tokens to AWS IAM credentials which gives access in EKS.

The AWS IAM credentials are mapped to a username in Kubernetes: `zalando-iam:zalando:service:{{SessionName}}` Where `SessionName` will be the `UID` of the Zalando IAM token. Such that a user for a service becomes e.g.: `zalando-iam:zalando:service:stups_<app_id>` which is what we have in place today.

The idea is that the proxy works as a secondary API-endpoint to be called by services with Zalando IAM tokens: `https://<local-id>-zalando-iam-aws-proxy.<account>.zalan.do`

It may even make sense to set this URL as the `api_server_url` in cluster registry for EKS clusters, such that existing clients like CDP will just transparently work.

### DIFF
--- a/cluster/manifests/zalando-iam-aws-proxy/deployment.yaml
+++ b/cluster/manifests/zalando-iam-aws-proxy/deployment.yaml
@@ -1,0 +1,117 @@
+# {{ if eq .Cluster.Provider "zalando-eks" }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: zalando-iam-aws-proxy
+  namespace: kube-system
+  labels:
+    application: kubernetes
+    component: zalando-iam-aws-proxy
+spec:
+  selector:
+    matchLabels:
+      deployment: zalando-iam-aws-proxy
+  template:
+    metadata:
+      labels:
+        application: kubernetes
+        component: zalando-iam-aws-proxy
+        deployment: zalando-iam-aws-proxy
+      annotations:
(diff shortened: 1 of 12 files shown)


cluster/manifests/zalando-iam-aws-proxy/deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+        logging/destination: "{{.Cluster.ConfigItems.log_destination_infra}}"
+    spec:
+      dnsConfig:
+        options:
+          - name: ndots
+            value: "1"
+      serviceAccountName: zalando-iam-aws-proxy
+      containers:
+      - name: proxy
+        image: mikkeloscar/zalando-iam-aws-proxy:2
```
mikkeloscar: Needs to use a properly built image

linki: Done in https://github.com/zalando-incubator/kubernetes-on-aws/pull/7818/commits/98c4f8d315cc2b227ab985a3620c57bb8af58bdb

linki:
👍

linki:
👍

linki:
👍

gargravarr:
:+1:

PR #7811, opened 2024-07-05
update deployment-service to add cleanup task endpoints

demonCoder95:
Follow-up to #7804 to add the new service endpoints to create cleanup task objects

PR #7822, opened 2024-07-08
update deployment-service with resource cleanup task handler

demonCoder95:
Follow-up to adding #7811 , this update the deployment-service to add the resource cleanup parts in the controller to do the resource cleanup, also a few bugfixes in the code to make the cleanup work.


linki:
👍

demonCoder95:
👍

PR #7827, opened 2024-07-08
Give deployment-service permission to handle resource cleanup tasks

linki:
Both controller and API components need access to the resource cleanup tasks.


linki:
👍

demonCoder95:
👍

PR #7832, opened 2024-07-10
kube-aws-iam-controller: Don't assume worker role on eks

mikkeloscar:
Since kube-aws-iam-controller runs on a worker node on eks based cluster, it doesn't need to assume the worker role.


mikkeloscar:
:+1:

gargravarr:
:+1:

PR #7841, opened 2024-07-11
Include alias in Prometheus endpoint

mikkeloscar:
When running multiple clusters in a single AWS account, the prometheus endpoint would overlap between the clusters. Add a hostname which includes the alias such that it can be by cluster without overlap.

Similar to #7509

### DIFF
--- a/cluster/manifests/prometheus/ingress.yaml
+++ b/cluster/manifests/prometheus/ingress.yaml
@@ -9,6 +9,7 @@ metadata:
 spec:
   hosts:
     - system-prometheus.{{ .Values.hosted_zone }}
+    - system-prometheus-{{ .Cluster.Alias }}.{{ .Values.hosted_zone }}
   backends:
     - name: prometheus
       type: service
--- a/cluster/manifests/prometheus/ingress.yaml
+++ b/cluster/manifests/prometheus/ingress.yaml
@@ -23,6 +23,7 @@ spec:
         - HeaderRegexp("Authorization", "^Bearer .+")
       filters:
         - oauthTokeninfoAnyKV("realm", "/employees", "realm", "/services")
+        - enableAccessLog()
 
     # {{ if .Cluster.ConfigItems.prometheus_ui_users }}
     #   {{ $uidKVs := "" }}
@@ -34,4 +35,5 @@ spec:
       filters:
         - oauthGrant()
         - oauthTokeninfoAnyKV({{ $uidKVs }})
+        - enableAccessLog()
     # {{ end }}


mikkeloscar:
:+1:

MustafaSaber:
👍

cluster/manifests/prometheus/ingress.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
 spec:
   hosts:
     - system-prometheus.{{ .Values.hosted_zone }}
+    - system-prometheus-{{.Cluster.Alias}}.{{ .Values.hosted_zone }}
```
AlexanderYastrebov: For the most cases `.Cluster.Alias` is the prefix of the `.Values.hosted_zone`.
Maybe we can make it 
```
if .Values.hosted_zone startswith (.Cluster.Alias + ".") then
  system-prometheus.{{ .Values.hosted_zone }}
else
  system-prometheus-{{ .Cluster.Alias }}.{{ .Values.hosted_zone }}
end
```
WDYT?

AlexanderYastrebov: to avoid stuttering for the common case `system-prometheus-foo.foo.zalan.do`

AlexanderYastrebov: OTOH this will require conditional logic on the client side to build up the hostname so maybe its not worth it.

AlexanderYastrebov: minor: I'd add spaces between template separators for consistency like `{{ .Cluster.Alias }}`

AlexanderYastrebov: Could be done during old hostname cleanup.

AlexanderYastrebov:
:+1:

cluster/manifests/prometheus/ingress.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
 spec:
   hosts:
     - system-prometheus.{{ .Values.hosted_zone }}
+    - system-prometheus-{{ .Cluster.Alias }}.{{ .Values.hosted_zone }}
```
szuecs, line 12: a bit ugly name TBH

szuecs: Do we plan to run more than one cluster in each account in the future?

szuecs:
:+1:

cluster/manifests/prometheus/ingress.yaml
szuecs: Talked with @mikkeloscar about it and if this is the best way to solve it go for it, but I would assume we could also check @AlexanderYastrebov 's jwtMetrics reports to figure out who is using these targets before deploying this change to all clusters.

AlexanderYastrebov: jwtMetrics won't help (we can discuss why separately) but we can enable 2xx access log and it will contain token uid (token is present and required for both service and oauthGrant routes)

mikkeloscar: I have added `enableAccessLog()` now.

mikkeloscar:
:+1:

cluster/manifests/prometheus/ingress.yaml
```diff
@@ -21,6 +22,7 @@ spec:
       predicates:
         - HeaderRegexp("Authorization", "^Bearer .+")
       filters:
+        - enableAccessLog()
```
AlexanderYastrebov: Maybe move it after auth filters to reduce spam in the logs for unauthenticated requests.

AlexanderYastrebov:
:+1:

cluster/manifests/prometheus/ingress.yaml
mikkeloscar: Fixed!

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #7842, opened 2024-07-12
Update tokeninfo base image to use static base image

mikkeloscar:
Update tokeinfo image to a version that uses a static base image.


mikkeloscar:
:+1:

linki:
👍

PR #7853, opened 2024-07-15
Skip mustache rendering by default

linki:
This configuration was rolled out to all channels. Let's switch the default in order to remove the individual configs here: https://github.bus.zalan.do/teapot/cluster-config-manager/blob/main/channels.yaml


mikkeloscar:
:+1:

linki:
👍

PR #7876, opened 2024-07-18
[For E2E] Test AMI with ipv6 support

mikkeloscar:
This updates the AMI to a version which has optional support for ipv6 and flexible number of pods based on instance-type/ENIs which is relevant for EKS.

This PR is to run e2e to verify that the AMI will still work in non ipv6/flexible pod setup which would allow us to merge the changes already.


linki:
e2e tests passed. Closing this.

PR #7915, opened 2024-07-30
[e2e] test that newer CLM version works with existing setup

linki:
This runs the [current, most recent](https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/777) version of CLM against our e2e setup.

This is to validate that the newly added functionality doesn't affect the existing clusters negatively.


linki:
Works fine.

PR #7916, opened 2024-07-31
Add Session Manager logging CF

zaklawrencea:
Add CloudFormation resources to enable logging of SSM Sessions. Logs will first be collected in a Log Group in each account and then forwarded to a centralised destination via a Subscription Filter.


zaklawrencea:
:+1:

gargravarr:
:+1:

PR #7918, opened 2024-08-01
Update fabric-gateway-controller

RomanZavodskikh:
Changelog:

+ Create ControllerRevision if stackVersion is set
+ Added some follow-up small enhancenents

### DIFF
--- a/cluster/manifests/fabric-gateway/deployment.yaml
+++ b/cluster/manifests/fabric-gateway/deployment.yaml
@@ -1,4 +1,4 @@
-# {{ $version := "master-263" }}
+# {{ $version := "master-265" }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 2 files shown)


cluster/manifests/fabric-gateway/deployment.yaml
```diff
@@ -47,6 +47,9 @@ spec:
             {{ if eq .Cluster.ConfigItems.fabric_gateway_controller_allow_all_filters "true" }}
             - -allow-all-filters=true
             {{ end }}
+            {{ if eq .Cluster.ConfigItems.fabric_gateway_controller_enable_versioning "true" }}
+            - -enable-versioning
```
MustafaSaber, line 51: I think `=true` is missing

RomanZavodskikh: `-enable-versioning` and `-enable-versioning=true` work the same way.

RomanZavodskikh:
:+1:

cluster/manifests/fabric-gateway/deployment.yaml
MustafaSaber: I think explicit is better is this case like `-allow-all-filters=true` but that's a minor

MustafaSaber:
👍

PR #7921, opened 2024-08-02
Use `-enable-versioning=true` instead of `-enable-versioning`

RomanZavodskikh:
The new cmdline options causes less confusion and has no way it could be misinterpreted.


RomanZavodskikh:
:+1:

MustafaSaber:
👍

PR #7924, opened 2024-08-02
Add fabric-gateway-controller snapshotting support

RomanZavodskikh:
There is a plan to enable this feature first gradually on Zalando clusters, so there is some snapshot for every FabricGateway being created together with the StackSet.


MustafaSaber:
👍

RomanZavodskikh:
:+1:

PR #7930, opened 2024-08-06
Update fabric-gateway-controller

RomanZavodskikh:
+ Use time.UnixMilli() to set unversioned ControllerRevision name
+ Apply new RBAC so fabric-gateway-controller has permissions to operate with ControllerRevisions


RomanZavodskikh:
:+1:

MustafaSaber:
👍

PR #7938, opened 2024-08-08
Update RBAC to correspond to fabric-gateway-controller code

RomanZavodskikh:
Just a small change for fabric-gateway-controller not to fail because of lack of permissions.


RomanZavodskikh:
:+1:

MustafaSaber:
👍

PR #7959, opened 2024-08-20
[TEST] use AMI with containerd 1.7

demonCoder95:
## DO NOT MERGE

Creating this PR with an experimental AMI to test the version bump to the containerd runtime `1.6.32` -> `1.7.20`. Will only use this PR to run e2e tests for now.

Reference: https://github.bus.zalan.do/teapot/kubernetes-on-ubuntu/pull/369


demonCoder95:
We didn't rollout the `v1.30.3` after updating the AMI, so this change is rolling that out as well and running into issues. Will work on resolving those.

demonCoder95:
Going to test a previous version of the AMI just to verify that the issues are not related to this change, and already existed in the AMI. We never discovered them as we never rolled it out.

demonCoder95:
Seems like the rollback worked and e2e is green for the AMI! I'm going to close this PR now.

PR #7960, opened 2024-08-20
Make Session Manager logging resources unique per cluster

zaklawrencea:
Add a unique Cluster ID to Session Manager resources to avoid resource conflicts in accounts with multiple clusters


demonCoder95:
👍

zaklawrencea:
:+1:

PR #7963, opened 2024-08-20
Do not create account level Session Manager preferences document during e2e

zaklawrencea:
Due to conflicting ownership of the AWS Account level Session Manager preferences document, only create a session-level `Document` during e2e testing.


zaklawrencea:
:+1:

gargravarr:
:+1:

PR #7964, opened 2024-08-20
reduce the disruption budget of the karpenter node pools node pools

myaser:
many users reported that karpenter is aggressive in disrupting pods. this budget reduces the maximum number of nodes to be disrupted at a time from 10% to the minimum of 5% or 10 nodes


szuecs:
:+1:

myaser:
👍🏻

PR #7965, opened 2024-08-20
Update to Kubernetes 1.31

demonCoder95:
https://kubernetes.io/blog/2024/08/13/kubernetes-v1-31-release/

## Summary of e2e Changes
1. In `apiserver.go`, the function `job.WaitForJobComplete` now also takes a `reason` parameter  which is `*string` type. Looking at [upstream k8s](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/apps/job.go#L85) which uses `ptr.To(batchv1.JobReasonCompletionsReached)` as value, so I've used the same as it seems most appropriate.
2. In `node_test.go`, `framework.BusyBox` is no longer available. This has been refactored out of the framework into `imageutils` package. Following the way it's [done in upstream k8s](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/kubectl/kubectl.go#L813), I replaced it with `imageutils.GetE2EImage(imageutils.BusyBox)`.
3. In `audit.go`, updated the pause image to `pause:3.7-master-21`.

## Other Changes

1. `HPAContainerMetrics` feature gate is now GA, so it's removed now.
2. `MinDomainsInPodTopologySpread` feature gate is now GA, so it's also removed.
3. `--cloud-provider` CLI flag on kube-apiserver is no longer needed, so it's removed.

Reference: https://github.bus.zalan.do/teapot/issues/issues/3681


demonCoder95:
Rebased to include the patch for image validation https://github.com/zalando-incubator/kubernetes-on-aws/pull/7973.
Seems like the patch worked! 🎉

demonCoder95:
Updated the pause image used to be consistent in all e2e tests `pause:3.7-master-21`.
Admission control requires pause image to run as non-root. I think 3.7 has non-root support, so the container doesn't get forced to run as root.

test/e2e/audit.go
```diff
@@ -44,14 +44,26 @@ var _ = describe("Audit", func() {
...
+					Image: "container-registry.zalando.net/teapot/pause:3.7-master-21",
+					// Drop all capabilities, run as non-root, and use the default seccomp profile.
+					// This is required as pods created are assumed to be restricted by default since k8s v1.31.
+					SecurityContext: &apiv1.SecurityContext{
+						Capabilities: &apiv1.Capabilities{
+							Drop: []apiv1.Capability{"ALL"},
+						},
+						RunAsNonRoot: &runAsNonRoot,
+						SeccompProfile: &apiv1.SeccompProfile{
+							Type: "RuntimeDefault",
```
linki: Let's use the Golang constant instead of a string, e.g. `types.SeccompProfileTypeRuntimeDefault`.

test/e2e/audit.go
```diff
@@ -44,14 +44,26 @@ var _ = describe("Audit", func() {
...
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "audit-pod",
 			},
 			Spec: apiv1.PodSpec{
 				Containers: []apiv1.Container{{
 					Name:  "pause",
-					Image: "container-registry.zalando.net/teapot/pause:3.4.1-master-18",
+					Image: "container-registry.zalando.net/teapot/pause:3.7-master-21",
+					// Drop all capabilities, run as non-root, and use the default seccomp profile.
+					// This is required as pods created are assumed to be restricted by default since k8s v1.31.
```
linki: I don't fully understand. If they are restricted by default, why do we need to add these instructions? Do you have a link to the docs of the change that was made in 1.31?

test/e2e/audit.go
```diff
@@ -44,14 +44,26 @@ var _ = describe("Audit", func() {
...
+					SecurityContext: &apiv1.SecurityContext{
+						Capabilities: &apiv1.Capabilities{
+							Drop: []apiv1.Capability{"ALL"},
+						},
+						RunAsNonRoot: &runAsNonRoot,
```
linki: Let's try to use `ptr.To(true)` using the `k8s.io/utils/ptr` package like above.

test/e2e/audit.go
demonCoder95: I just tried by removing the whole securityContext stuff and the test passed. So, I guess we can drop it altogether 🙂

test/e2e/audit.go
demonCoder95: I dropped it and the test worked 😅

test/e2e/audit.go
demonCoder95: I just dropped the context and the test worked, so we don't need this I guess.

linki:
After merging, let's reset the channel setup: https://github.bus.zalan.do/teapot/k8s-on-aws-manager/pull/150

linki:
👍

mikkeloscar:
:+1:

demonCoder95:
👍

PR #7975, opened 2024-08-22
deploy: pod deletion cost controller

szuecs:
it takes care of application=skipper-ingress,component=ingress to have a proper deletion cost as written down in https://github.com/kubernetes/kubernetes/issues/124149

### DIFF
--- a/cluster/manifests/skipper/pod-deletion-cost-controller.yaml
+++ b/cluster/manifests/skipper/pod-deletion-cost-controller.yaml
@@ -1,0 +1,58 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    application: skipper-ingress
+    component: pod-deletion-cost-controller
+  name: pod-deletion-cost-controller
+  namespace: kube-system
+spec:
+  replicas: 1
+  revisionHistoryLimit: 3
+  selector:
+    matchLabels:
+      deployment: pod-deletion-cost-controller
+  strategy:
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 0
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        application: skipper-ingress
+        component: pod-deletion-cost-controller
+        deployment: pod-deletion-cost-controller
+    spec:
+      containers:
+      - args:
+        - /pod-deletion-cost-controller
+        - -address=:8080
+        - -log-level=INFO
+        image: container-registry.zalando.net/gwproxy/pod-deletion-cost-controller:main-18
+        name: skipper-ingress-pod-deletion-cost-controller
+        ports:
+        - containerPort: 8080
+          name: metrics
+          protocol: TCP
+        readinessProbe:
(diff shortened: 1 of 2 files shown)


cluster/manifests/skipper/pod-deletion-cost-controller.yaml
```diff
@@ -0,0 +1,58 @@
...
+  replicas: 1
+  revisionHistoryLimit: 3
+  selector:
+    matchLabels:
+      deployment: pod-deletion-cost-controller
+  strategy:
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 0
+    type: RollingUpdate
```
AlexanderYastrebov: I think [Recreate strategy](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#recreate-deployment) fits better to ensure single instance like we use for fabric-gateway-controller

cluster/manifests/skipper/pod-deletion-cost-controller.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: apps/v1
```
AlexanderYastrebov, line 1: Do we want a feature toggle for it?

szuecs: Why?
It does not harm to run it

cluster/manifests/skipper/pod-deletion-cost-controller.yaml
szuecs: It should not matter and I do not think that "ensure single" is happening anyways, regardless of the config. 
Think of it more that "most often" it will be like you expect.
I will change as you suggest.

szuecs: done

AlexanderYastrebov:
:+1:

szuecs:
:+1:

rtomadpg:
If I understand the purpose of the "pod-deletion-cost-controller" correctly, it is the first of its kind.

Are there any plans for making it available outside Zalando? Much appreciated.

szuecs:
@rtomadpg first we have to check if it works as we expect, but normally I am one of the people that would OSS it. It's great to see if there are already folks asking for this so there is public demand for such a controller.
In the next weeks we will gather information and if it's good enough I have to polish the code such that backends will also get the pod deletion cost settings, right now it focuses only on our ingress implementation [skipper http proxy](https://opensource.zalando.com/skipper).

rtomadpg:
We're using Skipper, so 🚀 !

zhuoyang:
Any updates on this? I am interested to use this pod deletion cost controller to solve AZ balancing issue during scale down

szuecs:
> Any updates on this? I am interested to use this pod deletion cost controller to solve AZ balancing issue during scale down

I am on parental leave and I am not sure if the project is currently going on or paused until I come back. We will let you know when it is good enough to share. We got some more experience and will share it with the public later.

PR #7976, opened 2024-08-23
fix: missing image label that we need for versioned metrics

szuecs:
fix: missing image label that we need for versioned metrics


szuecs:
:+1:

gargravarr:
:+1:

PR #7979, opened 2024-08-23
fix: pod-deletion-cost-controller use also patch under the hood

szuecs:
fix: pod-deletion-cost-controller use also patch under the hood


szuecs:
:+1:

zaklawrencea:
:+1:

PR #7981, opened 2024-08-23
Improve code quality for e2e tests by fixing linter errors

demonCoder95:
1. remove unused parameters from functions
2. rename unused variables to _
3. uncapitalise error msg text
4. remove duplicate imports

Also, a routegroup e2e test had a bug where it was using a response value from a previous request, it got caught because of the linter complaining that the new response value was never used. (`routegroup.go` line 393)

### DIFF
--- a/test/e2e/routegroup.go
+++ b/test/e2e/routegroup.go
@@ -390,7 +390,9 @@ rBackend: Path("/backend") -> inlineContent("%s") -> <shunt>;
 			return code == http.StatusTooManyRequests
 		}, false)
 		Expect(err).NotTo(HaveOccurred())
-		Expect(resp).To(Equal(expectedResponse))
+		s, err = getBody(resp)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(s).To(Equal(expectedResponse))
 	})
 
 	It("Should create blue-green routes [RouteGroup] [Zalando]", func() {
--- a/test/e2e/routegroup.go
+++ b/test/e2e/routegroup.go
@@ -269,14 +269,14 @@ rBackend4: Path("/router-response") -> inlineContent("NOT OK") -> <shunt>;
 
 		// response for / is from our backend
 		By("checking the response code of a request without required request header, we can check if predicate match works correctly")
-		req, err := http.NewRequest("GET", "https://"+hostName+"/backend", nil)
+		req, _ := http.NewRequest("GET", "https://"+hostName+"/backend", nil)
 		resp, err = waitForResponseReturnResponse(req, 10*time.Minute, isNotFound, false)
 		Expect(err).NotTo(HaveOccurred())
 		resp.Body.Close()
 
 		// checking backend route with predicates and filters
 		By("checking the response status code for a request to /backend without correct headers we should get 404")
-		err = waitForResponse("https://"+hostName+"/backend", "https", 10*time.Minute, isNotFound, false)
+		waitForResponse("https://"+hostName+"/backend", "https", 10*time.Minute, isNotFound, false)
 		By("checking the response for a request to /backend with the right header we know if we got the correct route")
 		req, err = http.NewRequest("GET", "https://"+hostName+"/backend", nil)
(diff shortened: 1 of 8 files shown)


test/e2e/routegroup.go
```diff
@@ -390,7 +390,7 @@ rBackend: Path("/backend") -> inlineContent("%s") -> <shunt>;
 			return code == http.StatusTooManyRequests
 		}, false)
 		Expect(err).NotTo(HaveOccurred())
-		Expect(s).To(Equal(expectedResponse))
+		Expect(resp).To(Equal(expectedResponse))
```
demonCoder95: This fixes the test case for rate limiting, it was never checking the response value of the right request, but the one for the previous tase.

cc @MustafaSaber I see you added this test case, maybe you can check if my assumption is correct here 🙂

demonCoder95:
Also, we have tons of unused functions in the code that we can just get rid of, as we haven't used them for many years.

test/e2e/routegroup.go
MustafaSaber: Great catch! Thanks for letting me know.

do we need to get the body to  compare?

MustafaSaber: I think my plan was to override `s` but for some reason forgot to.


```diff
		Expect(err).NotTo(HaveOccurred())
-		Expect(resp).To(Equal(expectedResponse))
+		s, err = getBody(resp)
+               Expect(err).NotTo(HaveOccurred())
+		Expect(s).To(Equal(expectedResponse))
```

demonCoder95: Makes sense, I'll add the change to update `s`.

demonCoder95: I updated as suggested, thanks 🙂

MustafaSaber:
👍

demonCoder95:
Will wait for e2e to pass and then we can merge 🙂

szuecs:
:+1:

demonCoder95:
seems like the e2e failed @MustafaSaber 😅 can you debug on why it fails now?

Backend returns status code 429

test/e2e/routegroup.go
```diff
@@ -390,6 +390,8 @@ rBackend: Path("/backend") -> inlineContent("%s") -> <shunt>;
 			return code == http.StatusTooManyRequests
 		}, false)
 		Expect(err).NotTo(HaveOccurred())
+		s, err = getBody(resp)
```
szuecs: this is wrong, because test is:
```
By("checking the response is for a request to /backend with the right header we know if we got the correct route but get ratelimited")
```
and the response is:
```
resp, err = waitForResponseReturnResponse(req, 5*time.Minute, func(code int) bool {
			return code == http.StatusTooManyRequests
		}, false)
```

so we wait for status code 429 and then you compare the body with the initial expected body if we are not rate limited.

demonCoder95: Oh, so you mean there's no need to check response body, and we can just check if the returned code is 429, that implies rate limiting?

test/e2e/routegroup.go
MustafaSaber: Apparently overriding `s` was wrong 😅  according to @szuecs comment. Will also double check the code locally to confirm.

test/e2e/routegroup.go
MustafaSaber: I added a commit to only check status code, testing local takes forever so I pushed it.

MustafaSaber: still failing 😅 but not because of this test case I think. Or I need to restart the full pipeline not only e2e step.

demonCoder95: Thanks! I retriggered the pipeline 😅

demonCoder95: It passed now 🙂

demonCoder95:
👍

MustafaSaber:
👍

demonCoder95:
👍

PR #7986, opened 2024-08-26
update AMI with k8s 1.30.3 and ipv6

demonCoder95:
Update k8s AMI to include k8s v1.30.3 and optional ipv6 support.

Reference: https://github.bus.zalan.do/teapot/kubernetes-on-ubuntu/pull/365


demonCoder95:
Seems like e2e passes here, so up til this change, the AMI works fine. I won't roll this change out because it will rotate all the nodes. Since we have to update to `1.31`, we can roll nodes just once. 

I'm going to close this PR now.

PR #7988, opened 2024-08-26
test rollback image

demonCoder95:
Need to test that e2e passes on the AMI with rolled back version for `nvidia-container-toolkit`.


demonCoder95:
The rollback seems to make `create cluster` step succeed. So, will close this PR now.

PR #7989, opened 2024-08-26
[e2e] fix flake tests with image validation API errors

demonCoder95:
E2e tests often fail because the image validation API return 502 return status, which causes the image admission to fail and the test fails. 

```
Reason:FailedCreate Message:admission webhook \"pod-admitter.teapot.zalan.do\" denied the request: image check for registry.opensource.zalan.do/teapot/skipper:v0.14.0 failed: unexpected HTTP response status: 502 Bad Gateway
```

Looking at docker-meta logs we see the base image has expired
```
INFO:meta.opa:'registry.opensource.zalan.do/teapot/skipper:v0.17.2' base image is not allowed (['base image registry.opensource.zalan.do/library/alpine-3:3-20230731 is expired.']). OPA Input:
```

This PR updates the compliant images so that the check doesn't fail.

### DIFF
--- a/test/e2e/apiserver.go
+++ b/test/e2e/apiserver.go
@@ -49,17 +49,17 @@ const (
 	compliantImage7 = "container-registry.zalando.net/teapot/skipper:v0.21.6"
 	compliantImage8 = "container-registry.zalando.net/teapot/skipper:v0.21.7"
 
-	// these are non-compliant because of expired base image
-	nonCompliantImage1  = "container-registry.zalando.net/teapot/skipper:v0.16.0"
-	nonCompliantImage2  = "container-registry.zalando.net/teapot/skipper:v0.16.1"
-	nonCompliantImage3  = "container-registry.zalando.net/teapot/skipper:v0.16.2"
-	nonCompliantImage4  = "container-registry.zalando.net/teapot/skipper:v0.16.3"
-	nonCompliantImage5  = "container-registry.zalando.net/teapot/skipper:v0.16.4"
-	nonCompliantImage6  = "container-registry.zalando.net/teapot/skipper:v0.16.5"
-	nonCompliantImage7  = "container-registry.zalando.net/teapot/skipper:v0.16.6"
-	nonCompliantImage8  = "container-registry.zalando.net/teapot/skipper:v0.16.7"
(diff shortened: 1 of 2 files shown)


demonCoder95:
Adding the compliance namespace to the softfail list for the admission-controller doesn't seem to fix the issue for now, will investigate further.

Update: Decided to update image references, let's see if it resolves the issue.

MustafaSaber:
👍

test/e2e/apiserver.go
```diff
@@ -37,24 +37,27 @@ import (
...
-	nonCompliantImage10 = "registry.opensource.zalan.do/teapot/skipper-test:pr-2080-13"
+	// these are several compliant images
+	compliantImage1 = "container-registry.zalando.net/teapot/skipper:v0.21.0"
+	compliantImage2 = "container-registry.zalando.net/teapot/skipper:v0.21.1"
+	compliantImage3 = "container-registry.zalando.net/teapot/skipper:v0.21.2"
+	compliantImage4 = "container-registry.zalando.net/teapot/skipper:v0.21.3"
+	compliantImage5 = "container-registry.zalando.net/teapot/skipper:v0.21.4"
+	compliantImage6 = "container-registry.zalando.net/teapot/skipper:v0.21.5"
+	compliantImage7 = "container-registry.zalando.net/teapot/skipper:v0.21.6"
+	compliantImage8 = "container-registry.zalando.net/teapot/skipper:v0.21.7"
```
linki, line 50: These must be real images (i.e. the tag must exist and the image must be considered compliant or non-compliant (see images below) by docker-meta). The reason why there's so many test images is to ensure that the tests using them aren't affected by any caching because that can lead to wrong results.

Meaning we need 8 compliant and 10 non-compliant test images for the test cases or refactor the test cases.

demonCoder95: I can look into this today, thanks for the review 🙏

demonCoder95:
Rebased on latest `dev`, will try working on this and resolving the issue so we can no longer rely on the `softfail` configuration.

test/e2e/apiserver.go
demonCoder95: I'll use the production registry images that are compliant and test registry images that will naturally be non-compliant.

mikkeloscar:
Closing as not relevant

PR #8008, opened 2024-08-29
Create SSM Documents per Cluster

zaklawrencea:
Create an SSM Document per cluster instead of utilising the account wide `SSM-SessionManagerRunShell` Document to manage SSM system preferences. Due to the retention policies introduced in https://github.com/zalando-incubator/kubernetes-on-aws/pull/8024, the default Document will remain in the accounts while a new `SSM-SessionManager-<cluster-alias>` Document holding session configuration will be created.


linki:
👍

zaklawrencea:
:+1:

PR #8009, opened 2024-08-29
double the memory of karpenter controller

myaser:
during load test, it turns out that karpenter needs more memory during peak times (spikes). we checked and there is enough capacity


demonCoder95:
👍

myaser:
👍🏻

PR #8010, opened 2024-08-29
config: init sync for pod-cost-del-ctl

szuecs:
config: init sync for pod-cost-del-ctl

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -220,6 +220,9 @@ skipper_pod_deletion_cost_controller_memory: 200Mi
 skipper_pod_deletion_cost_controller_cpu: 50m
 # klog style -v=0
 skipper_pod_deletion_cost_controller_log_v: "0"
+# pod informer sync config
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

szuecs:
:+1:

cluster/config-defaults.yaml
```diff
@@ -220,6 +220,9 @@ skipper_pod_deletion_cost_controller_memory: 200Mi
 skipper_pod_deletion_cost_controller_cpu: 50m
 # klog style -v=0
 skipper_pod_deletion_cost_controller_log_v: "0"
+# pod informer sync config
+skipper_pod_deletion_cost_controller_poll_interval: "10s"
```
RomanZavodskikh, line 224: As far as I remember, interval should be several times smaller than timeout, shouldn't it?

szuecs: Why?
It's also doesn't matter so much because it happens only at startup and test if initial sync was done.

RomanZavodskikh:
:+1:

PR #8016, opened 2024-08-30
[e2e] use bigger instance for `default-worker-splitaz` to avoid flake

demonCoder95:
We have a flake e2e test where a pod remains pending for too long and times out. This increases the instance size so we can avoid that flakey test failure.

Reference: https://github.bus.zalan.do/teapot/issues/issues/3619


demonCoder95:
👍

linki:
👍

PR #8019, opened 2024-08-30
[e2e] re-enable skipped e2e tests

demonCoder95:
We skip some e2e tests in our setup and this is to debug those to see if they're easily fixable to improve our e2e coverage.

## Summary of Changes
1. Add `dns-*` namespace to ignore list of the admission controller so it can allow the test pod to use `hostNetwork`. See the e2e test [here](https://github.com/kubernetes/kubernetes/blob/v1.31.0/test/e2e/network/dns.go#L642).
2. Add `sched-pred-*` namespace to ignore list to allow `hostPort` based test pod. See the e2e test [here](https://github.com/kubernetes/kubernetes/blob/v1.31.0/test/e2e/scheduling/predicates.go#L705)
3. Update route-group gradual traffic test where assertions are missing annotations making it hard to debug. Also fix a typo.
4. Allow some skipped tests to run that now pass in our infrastructure. (detail below)
5. Update the comments in `test/e2e/run_e2e.sh` with the current status of the skipped tests.

## Tests Re-Enabled
The following tests have been re-enabled and they now pass in our setup
1. should resolve DNS of partial qualified names for the cluster
2. should resolve DNS of partial qualified names for services
3. should be able to change the type from ExternalName to NodePort
4. should be able to create a functioning NodePort service
6. should have session affinity work for NodePort service
7. should have session affinity timeout work for NodePort service
8. should be able to switch session affinity for NodePort service
9. should create gradual traffic routes
10. should create blue-green routes

## Tests We Continue to Skip

1. validates that there is no conflict between pods with same hostPort but different hostIP and protocol. This didn't work. Because it attempts to access a hostPort from a public endpoint, which we don't yet support.
2. [Serial] e2e tests.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -601,7 +601,7 @@ teapot_admission_controller_namespace_delete_protection_enabled: "false"
 teapot_admission_controller_resolve_vanity_images: "true"
 
 {{if eq .Cluster.Environment "e2e"}}
-teapot_admission_controller_ignore_namespaces: "^kube-system|((downward-api|kubectl|projected|statefulset|pod-network|scope-selectors|resourcequota|limitrange|sysctl|node-tests|e2e-kubelet-etc-hosts|csiinlinevolumes|dns)-.*)$"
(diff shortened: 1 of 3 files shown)


demonCoder95:
Will retrigger the e2e steps a few times once it's green so we have some sense of flakiness as well and will try debugging that.

demonCoder95:
All e2es pass now! Let's continue to use the new tests for now.

demonCoder95:
👍

szuecs:
:+1:

PR #8022, opened 2024-09-02
Add permissions in zmon role for lambda functions

Shinzu:
allow zmon role to list lambda functions and their tags


Shinzu:
pipeline seems not to run here, anything i can do?

PR #8024, opened 2024-09-02
Add retention policies to the SSM-SessionManagerRunShell Document

zaklawrencea:
Update the `DeletionPolicy` / `UpdateReplacePolicy` for the default `SSM-SessionManagerRunShell` SSM Document. Due to upcoming changes to create Documents per cluster, we want to retain the default Document in each account but remove the Document content.


zaklawrencea:
:+1:

linki:
👍

PR #8025, opened 2024-09-02
[experiment] Switch from LocalID to Alias

linki:
This is an attempt to overcome the sometimes cryptic use of `Cluster.LocalID`.

In the past we didn't parametrize our cluster-wide resources (such as IAM roles) because there was no need for it since the Kubernetes stack was only created once per account. The introduction of multiple e2e clusters in the very same e2e account changed that and let to the need to occasionally add the `Cluster.LocalID` parameter to the resource names. However, the list grew and with the outlook of having multiple clusters in many accounts will only intensify.

Unfamiliar users, such as new team members or other platform teams, are even more surprised by the need of this parameter as they haven't been exposed to multiple clusters in the accounts up until now.

In order to make this concept more clear and understandable, this PR changes it to use the `Cluster.Alias`. The alias is globally unique and hence will also be unique in the same account, like `Cluster.LocalID`. In addition, the more familiar alias value is easier to understand and unforeseen future name conflicts are avoided proactively by making the alias part of each AWS resource name.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -556,7 +556,7 @@ Resources:
                 Resource: '*'
             Version: 2012-10-17
           PolicyName: root
-      RoleName: "{{.Cluster.LocalID}}-app-autoscaler"
+      RoleName: "cluster-autoscaler-{{ .Cluster.Alias }}"
     Type: 'AWS::IAM::Role'
   KarpenterNodeInstanceProfile: # instance profile for worker nodes spawn by karpenter controller
     Type: "AWS::IAM::InstanceProfile"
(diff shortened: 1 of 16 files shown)


cluster/cluster.yaml
```diff
@@ -556,7 +556,7 @@ Resources:
                 Resource: '*'
             Version: 2012-10-17
           PolicyName: root
-      RoleName: "{{.Cluster.LocalID}}-app-autoscaler"
+      RoleName: "cluster-autoscaler-{{ .Cluster.Alias }}"
```
linki, line 559: We should think about keeping the `app-` prefix as a kind of namespace to avoid clashing with other resources. Let's say there's another role related to the autoscaler for a different purpose, it could still lead to clashes although less likely because it's in the same project.

linki:
Alright, roll out is actually not that difficult: Instead of modifying the IAM roles in place, we can just make duplicate roles. Then change the annotations on the Kubernetes resources in a separate PR. Once everything is rolled out, remove the old IAM roles from the Stack and AWS will clean them up.

linki:
Closing for now.

PR #8034, opened 2024-09-02
[e2] refactor external-dns test to use `service.TestJig`

demonCoder95:
This is to resolve a TODO in `test/e2e/external_dns.go` where we have to refactor the code to use `service.TestJig` from `k8s.io/kubernetes/test/e2e/framework/service` pkg to create the Service.
I used [this upstream e2e test](https://github.com/kubernetes/kubernetes/blob/v1.31.0/test/e2e/network/service.go#L3829) as a reference.

Also added annotations to the error assertions.

### DIFF
--- a/test/e2e/external_dns.go
+++ b/test/e2e/external_dns.go
@@ -18,32 +18,38 @@ import (
 	"fmt"
 	"time"
 
+	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

demonCoder95:
👍

PR #8039, opened 2024-09-03
Add permissions to list lambda for zmon role

Shinzu:
adds permissions to list lambda functions and their tags for zmon role

with this zmon-aws-agent can query information to create entities used in zmon


Shinzu:
:+1:

gargravarr:
:+1:

PR #8045, opened 2024-09-04
introduce config-items to control kubelet image qps

demonCoder95:
## Background
It was reported recently in a support ticket that there are intermittent ImagePullBackoff errors for various images.
Upon looking deeper into this, it turns out the backoff is triggered by the kubelet, and not the ECR service.
```
E0903 10:30:03.291420    1088 pod_workers.go:1298] "Error syncing pod, skipping" err="failed to \"StartContainer\" for \"<CONTAINER_NAME>\" with ErrImagePull: \"pull QPS exceeded\"" pod="<POD_NAME>" podUID="<POD_ID>"
```
The default value of the QPS is 5. This is not sufficient, as we often initiate to run more than 5 containers on a node at a time, and so we run into this limit quite often.

## Summary
This introduces the config-items in the kubelet config userdata so we configure this value to a higher value, alongside a burst that can help temporarily increase this value to avoid QPS issues. The configuration items can be seen [here](https://github.com/kubernetes/kubernetes/blob/v1.30.0/staging/src/k8s.io/kubelet/config/v1beta1/types.go#L200-L212).

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -787,6 +787,22 @@ ebs_csi_controller_sidecar_cpu: "10m"
 # pull images in parallel
 serialize_image_pulls: "false"
 
+# rate of image pull in the kubelet, see
+# see https://github.com/kubernetes/kubernetes/blob/v1.30.0/staging/src/k8s.io/kubelet/config/v1beta1/types.go#L200-L212
+#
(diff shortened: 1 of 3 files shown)


demonCoder95:
👍

zaklawrencea:
:+1:

PR #8092, opened 2024-09-15
[1/2] remove k8s dashboard resources

demonCoder95:
As the next step to decommissioning the Kubernetes Dashboard UI from our clusters, this sets a toggle which we can use to reliably remove it from our clusters step by step.

As a final clean-up item, we will remove this along with any other config-items related to the dashboard.

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -295,3 +295,36 @@ post_apply:
 - name: deployment-service-status-service
   kind: Ingress
   namespace: kube-system
+
+{{ if ne .Cluster.ConfigItems.k8s_dashboard_enabled "true" }}
+- name: kubernetes-dashboard
+  namespace: kube-system
+  kind: Deployment
+- name: kubernetes-dashboard
+  namespace: kube-system
+  kind: Service
+- name: dashboard-metrics-scraper
+  namespace: kube-system
+  kind: Service
+- name: dashboard-metrics-scraper
+  namespace: kube-system
+  kind: Deployment
+- name: kubernetes-dashboard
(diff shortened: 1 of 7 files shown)


demonCoder95:
`ServiceAccount` deletion is failing as long as deployment pods remain, so we can do the deletion in two steps.

demonCoder95:
Testing the patch on `playground` cluster to ensure it runs as expected. Will try toggling the config-item and see if it deletes/restores resources as expected.

demonCoder95:
The toggle works just fine. It can add and remove resources as expected. Let's move forward with merging the PR.

demonCoder95:
👍

linki:
👍

cluster/manifests/deletions.yaml
```diff
@@ -295,3 +295,37 @@ post_apply:
...
+  kind: ClusterRoleBinding
+- name: kubernetes-dashboard-readonly
+  kind: ClusterRoleBinding
+- name: dashboard-metrics-scraper-vpa
+  namespace: kube-system
+  kind: VerticalPodAutoscaler
+# clean this up in another step after the above resources have been cleaned-up
+# - name: kubernetes-dashboard
+#   namespace: kube-system
+#   kind: ServiceAccount
```
linki, line 330: It's ok to keep it. Yes, it will initially fail but CLM automatically retries and eventually all pods are gone and this succeeds.

demonCoder95: Ah I see. I can clean it up in another PR anyway, since e2e fails otherwise 😅 Thanks for the hint!

PR #8094, opened 2024-09-16
Remove region conditional for SSM resource creation

zaklawrencea:
A `Destination` resource has been created in `eu-north-1` allowing SSM logs from this region to be centrally collected. This PR removes the cluster region conditional that restricted the Log Group's Subscription Filter and relevant IAM Role from only be created for clusters in `eu-central-1`.


zaklawrencea:
:+1:

demonCoder95:
👍

PR #8105, opened 2024-09-17
[2/2] remove k8s dashboard resources

demonCoder95:
A follow-up to #8092 , this cleans up the `ServiceAccount` for the dashboard. This was done separately as running pods in the e2e cluster was causing e2e to fail complaining that the `ServiceAccount` is in use. This should now just work because dashboard pods won't be running anymore.

The final step will be to remove the toggle alongside the entry in `deletions.yaml`.


demonCoder95:
👍

szuecs:
:+1:

PR #8108, opened 2024-09-17
Use AccountName field to infer default Okta client ID

linki:
With [CLM adding the account name to the cluster struct](https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/816), we can use it to calculate the right Client ID for Okta.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -974,7 +974,7 @@ stackset_pcs_support_enabled: "true"
 enable_control_plane_profiling: "false"
 
 okta_auth_issuer_url: ""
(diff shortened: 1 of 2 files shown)


linki:
I'm dropping this approach.

Although it works fine, it doesn't allow clients to _detect_ whether the Client ID differs from the default (since there's no value stored in cluster registry). This makes it difficult for clients like `kube-web-view` or `zkubectl` to [detect the correct client ID](https://github.bus.zalan.do/teapot/kube-web-view/blob/main/kube_web/web.py#L1386-L1394) if they only have "cluster" context.

So, I think it's better to keep "Alias" approach (because it can be replicated by clients) and overwrite it when needed. Ultimately, we might set the value explicitly on all clusters.

PR #8120, opened 2024-09-19
[test] use updated flannel CNI plugin

demonCoder95:
Updates the AMI to use the latest flannel CNI plugin version. This is to avoid version skew and solve the flannel issues we're observing.

Will test this in a test setup.


demonCoder95:
I tested by rolling `playground` cluster with the override it seemed fine. Not sure if it's enough to replicate the issue we see in production but at least the AMI itself seems to not have caused any more issues.

mikkeloscar:
We concluded this is not related to the issues we observed. So let's close this one and just include this CNI update in the next AMI update we do :)

PR #8140, opened 2024-09-23
Allow Powerusers to patch pods/ephemeralcontainers

mikkeloscar:
Extend the poweruser role to have access to `patch` `pods/ephemeralcontainers` which is needed for attaching a debug container via `kubectl debug`.


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #8152, opened 2024-09-23
clean up leftover k8s dashboard RBAC resources

demonCoder95:
These RBAC resources were missed during cleanup in #8092 and #8105 . This puts these resources behind the toggle and add `deletions.yaml` entry to clean them up from all clusters.


demonCoder95:
👍

mikkeloscar:
:+1:

PR #8186, opened 2024-09-24
ingress: log into both cluster and infra accounts

AlexanderYastrebov:
It makes sense for ingress components to log into
both cluster and infra accounts:
* cluster logs enable correlation with skipper-ingress access logs and user application logs
* infra logs enable correlation across clusters
* ingress components typically do not produce large amount of logs (unlike skipper-ingress access logs)

For pod-deletion-cost-controller save last chunk of logs into container status on error to simplify debugging.


szuecs:
:+1:

demonCoder95:
👍

PR #8210, opened 2024-09-26
Update Tokeninfo to latest version

mikkeloscar:
Similar to #8165 but adds configuration of METRICS_LISTEN_ADDRESS for the sandbox tokeninfo to avoid it overlapping with the port used by the tokeninfo container.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #8193, opened 2024-09-25
skipper: automate canary version update

AlexanderYastrebov:
Define canary image variable to enable automatic updates by image-updater-bot.

PR #8216, opened 2024-09-26
skipper: disable image-updater-bot for main image

AlexanderYastrebov:
image-updater-bot replaces all matching image urls in the manifest so use print template function to disable it for main image.

Follow up #8193

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -1,6 +1,6 @@
-{{/* image-updater-bot detects *image variables so use __ suffux to disable it for main image */}}
+{{/* image-updater-bot detects *image variables so use print to disable it for main image */}}
 
-{{ $main_image__ := "container-registry.zalando.net/teapot/skipper-internal:v0.21.198-1017" }}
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
> image-updater-bot replaces all matching image urls in the manifest so use print template function to disable it for main image.

See example https://github.com/zalando-incubator/kubernetes-on-aws/pull/8205

AlexanderYastrebov:
> image-updater-bot replaces all matching image urls in the manifest so use print template function to disable it for main image.

Here is an example when it worked properly because main image url was different from canary
https://github.com/zalando-incubator/kubernetes-on-aws/pull/8220/commits/71a794b8bb7ca2e08a00b7680bcfe9a231aecbea

RomanZavodskikh:
https://github.com/zalando-incubator/kubernetes-on-aws/pull/8220/files advances only canary image, so I guess we can move on like this and find some way to make the description of it more informative.

RomanZavodskikh:
:+1:

AlexanderYastrebov:
:+1:

PR #8227, opened 2024-09-26
Make ssm session logging optional

mikkeloscar:
Make ssm session logging configuration optional based on the config-item: `session_manager_destination_arn`.

This way we don't need to enable it in pet clusters. It will be enabled in all other clusters which have a value set for `session_manager_destination_arn`


mikkeloscar:
:+1:

myaser:
👍🏻

PR #8232, opened 2024-09-27
Make eks okta identity provider optional

mikkeloscar:
This makes the okta identity provider, configured for the EKS cluster, optional. It's enabled by default, but it's possible to turn it off.

The main purpose is to allow disabling it for pet clusters which significantly speeds up the cluster creation. Login is still allowed via Administrator AWS role and generating a kubeconfig like:

```
aws eks --region <region> update-kubeconfig --name "<cluster-name>"
```

The reason it takes longer to create a cluster with an identity provider is that EKS/Cloudformation works like this:

1. Create EKS control plane, without identity provider
2. Add the entity provider and rotate all EKS control plane nodes to configure the apiserver with the OIDC config (this almost doubles the creation time).


mikkeloscar:
:+1:

linki:
👍

PR #8233, opened 2024-09-27
e2e: Allow non ready nodes when starting e2e run

mikkeloscar:
When running e2e tests the test framework always wait for all nodes in a cluster to be ready before starting. Since our clusters autoscale it's not uncommon to have some nodes not yet ready and this can slow down the start of the e2e test eventhough it shouldn't matter on the test itself, the system is dynamic in nature and should still succeed the e2e tests.

This PR allows to run e2e even if some nodes are not ready. This should be fine as we already verify all kube-system components are healthy as part of the cluster creation step, so it should be ready even if some nodes rotate during.


mikkeloscar:
:+1:

demonCoder95:
Nice!

demonCoder95:
👍

PR #8243, opened 2024-09-27
Provide AWS credentials to aws-node via eks-pod-identity

mikkeloscar:
Proof of Concept using the newer [eks-pod-identity feature](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html) to provide AWS credentials to the aws-node daemonset.

This both showcases how to use it, and it potentially resolves a race-condition on cluster creation where aws-node pods would start before the aws-node serviceacount had the aws role annotation (via [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html)) leading to the pod starting without AWS permissions and failing to get ready. Via eks-pod-identity the race should not happen.

~This currently does _not_ work with eks v1.31. I have created a support case with AWS as this seem like an EKS bug. Successfully tested on eks v1.30.~
This was fixed by AWS.


linki:
👍

mikkeloscar:
:+1:

PR #8249, opened 2024-09-30
Experimental support for ipv6

mikkeloscar:
This PR adds some missing features to get a fully working EKS ipv6 cluster.

1. Custom version of Karpenter which has support for setting `ipv6AddressCount` on instances at launch
2. Provide IPv6 subnet addresses to skipper-ingress such that skipper-ingress can detect internal vs. external traffic.
3. Configure a static IPv6 address (based on EKS ipv6 service CIDR) for the skipper-internal service and corresponding coredns config to make `ingress.cluster.local` hostnames work.

These changes depend on the related changes in CLM: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/822

### DIFF
--- a/cluster/manifests/skipper/service-internal.yaml
+++ b/cluster/manifests/skipper/service-internal.yaml
@@ -14,10 +14,9 @@ metadata:
     component: ingress
 spec:
   type: ClusterIP
-{{- if ne .Cluster.Provider "zalando-eks"}}
-# TODO: how to do internal-ingress?
-# function to derive IP from range?
-# Can be hardcoded for ipv4, must be dynamic for ipv6
+{{- if and (eq .Cluster.Provider "zalando-eks") (eq .Cluster.ConfigItems.eks_ip_family "ipv6") }}
+  clusterIP: {{ addressNFromIPv6CIDR .Cluster.ConfigItems.service_ipv6_cidr 50 }}
+{{- else}}
   clusterIP: 10.5.99.99
 {{- end}}
   ports:
--- a/cluster/manifests/z-karpenter/07-karpenter.k8s.aws_ec2nodeclasses.yaml
+++ b/cluster/manifests/z-karpenter/07-karpenter.k8s.aws_ec2nodeclasses.yaml
@@ -242,6 +242,11 @@ spec:
                 enum:
                 - RAID0
                 type: string
+              ipv6AddressCount:
+                description: IPv6AddressCount controls the number of IPv6 addresses
+                  assigned to instances that are launched with the nodeclass.
+                format: int64
+                type: integer
               metadataOptions:
                 default:
                   httpEndpoint: enabled
--- a/cluster/manifests/z-karpenter/deployment.yaml
+++ b/cluster/manifests/z-karpenter/deployment.yaml
@@ -51,7 +51,7 @@ spec:
                 - ALL
             readOnlyRootFilesystem: true
           {{if eq .Cluster.ConfigItems.karpenter_version "current"}}
-          image: "container-registry.zalando.net/teapot/karpenter:0.37.0-main-26.patched"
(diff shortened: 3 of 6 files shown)


cluster/manifests/z-karpenter/07-karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -242,6 +242,11 @@ spec:
                 enum:
                 - RAID0
                 type: string
+              ipv6AddressCount:
+                description: IPv6AddressCount controls the number of IPv6 addresses
+                  are assigned to instances that are launched with the nodeclass.
```
demonCoder95: nit: `IPv6 addresses _that_ are assigned`

cluster/manifests/skipper/service-internal.yaml
```diff
@@ -14,10 +14,9 @@ metadata:
     component: ingress
 spec:
   type: ClusterIP
-{{- if ne .Cluster.Provider "zalando-eks"}}
-# TODO: how to do internal-ingress?
-# function to derive IP from range?
-# Can be hardcoded for ipv4, must be dynamic for ipv6
+{{- if and (eq .Cluster.Provider "zalando-eks") (eq .Cluster.ConfigItems.eks_ip_family "ipv6") }}
+  clusterIP: {{ addressNFromIPV6CIDR .Cluster.ConfigItems.service_ipv6_cidr 50 }}
```
demonCoder95: Why `50`? I'm assuming based on how the template function works that we're going to reserve the 50th IP address of the v6 CIDR in each  subnet for internal ingress?

And I suppose 50 is an arbitrary number we chose? 😅

demonCoder95: Will this also somehow help us know if an ingress is internal; i.e. having the 50th address in the subnet will tell us this, or is that not the intention here?

demonCoder95: Okay I think the (3) part of the PR description explains it, we need this static setup to make local DNS work.

mikkeloscar: 50 is a random number, it just has to be not right in the beginning of the range to avoid overlapping with other services that may be assigned an IP automatically when created before this one. We could also use 42 or 131 😂

mikkeloscar: The IP is the one you will get when you resolve `*.ingress.cluster.local` addresses inside the cluster and the IP will point to skipper.

demonCoder95: Got it. Thanks 🙂

linki:
👍

cluster/manifests/z-karpenter/deployment.yaml
```diff
@@ -51,7 +51,8 @@ spec:
                 - ALL
             readOnlyRootFilesystem: true
           {{if eq .Cluster.ConfigItems.karpenter_version "current"}}
-          image: "container-registry.zalando.net/teapot/karpenter:0.37.0-main-26.patched"
+          # patched version with ipv6adresscount support
+          image: "container-registry-test.zalando.net/teapot/karpenter:0.37.0-pr-26-3.patched"
```
mikkeloscar: Using a PR image here is not desired. We need to integrate this change in the prod image for this to be mergable.

mikkeloscar: fixed

mikkeloscar:
:+1:

mikkeloscar:
:+1:

demonCoder95:
👍

demonCoder95:
Retried e2e so it succeeds, just one test had failed.

mikkeloscar:
:+1:

demonCoder95:
👍

PR #8254, opened 2024-10-01
fix: skipper topologySpreadConstraints.labelSelector

szuecs:
we have 2 deployments and component=ingress identifies the data-plane

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -79,7 +79,8 @@ spec:
           whenUnsatisfiable: DoNotSchedule
           labelSelector:
             matchLabels:
-              deployment: skipper-ingress
+              component: ingress
+              application: skipper-ingress
 {{- end }}
       priorityClassName: "{{ .Cluster.ConfigItems.system_priority_class }}"
       serviceAccountName: skipper-ingress


AlexanderYastrebov:
:+1:

cluster/manifests/skipper/deployment.yaml
```diff
@@ -79,7 +79,7 @@ spec:
           whenUnsatisfiable: DoNotSchedule
           labelSelector:
             matchLabels:
-              deployment: skipper-ingress
```
RomanZavodskikh, line 82: Should we add application label as well?

AlexanderYastrebov:
Does it matter though? The second deployment is a single canary instance.

cluster/manifests/skipper/deployment.yaml
szuecs: good idea!

szuecs:
> Does it matter though? The second deployment is a single canary instance.

The problem is that we set `maxSkew 1`, but with canary deployment we really have 1 not counted so we can assume always  `maxSkew 2` and we can fix it by changing the selector.

cluster/manifests/skipper/deployment.yaml
szuecs: fixed

RomanZavodskikh:
:+1:

AlexanderYastrebov:
:+1:

PR #8260, opened 2024-10-01
Allow powerusers to also delete leases

mikkeloscar:
The `leases` API can be useful for building leader election/distributed locks via the Kubernetes API.

Also allow powerusers and thereby services to `delete` leases` if needed (delete is not much different than updating).


demonCoder95:
👍

mikkeloscar:
:+1:

PR #8281, opened 2024-10-02
bugfix: add dedicated label to requirements

myaser:
the dedicated label was not added to the node pool requirements

### DIFF
--- a/cluster/node-pools/worker-karpenter/provisioners.yaml
+++ b/cluster/node-pools/worker-karpenter/provisioners.yaml
@@ -112,6 +112,18 @@ spec:
       # Operators { In, NotIn, Exists, DoesNotExist, Gt, and Lt } are supported.
       # https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#operators
       requirements:
+#{{ if $taints }}
+#  {{ range $taints }}
+#    {{ $taint := . }}
+#    {{ if eq $taint.Key "dedicated"}}
+        - key: dedicated
+          operator: In
+          values:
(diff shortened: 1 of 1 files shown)


cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -112,6 +112,18 @@ spec:
...
+#  {{ range $taints }}
+#    {{ $taint := . }}
+#    {{ if eq $taint.Key "dedicated"}}
+        - key: dedicated
+          operator: In
+          values:
+            - "{{ $taint.Value}}"
+#    {{ end}}
+#  {{ end}}
+#{{ end}}
```
mikkeloscar, line 125: I don't fully understand why the taints set above doesn't cover this? Can you elaborate on the problem that it solves?

myaser: pods targeting a dedicated node-pool uses a selector label dedicated=my-app

so this label needs to be added to all nodes. this is the way to do that, to be part of the requirements 

as we do for cluster autoscaler, we need both a taint and a selector label for it to work

myaser:
👍🏻

demonCoder95:
👍

cluster/node-pools/worker-karpenter/provisioners.yaml
mikkeloscar: I see. I thought having it under the label section enough would be enough to indicate that the node will have a particular label.

Thanks for elaborating! 🙏

PR #8303, opened 2024-10-04
[test] Dynamic Role creation for reading Secrets in Namespaces

demonCoder95:
Using this to test the dynamic RBAC for reading Secret by using the test image for the admission-controller.

### DIFF
--- a/cluster/manifests/02-admission-control/deployment.yaml
+++ b/cluster/manifests/02-admission-control/deployment.yaml
@@ -54,6 +54,8 @@ spec:
           - --address=:8085
           - --tls-cert-file=/etc/tls-certs/admission-controller.pem
           - --tls-key-file=/etc/tls-certs/admission-controller-key.pem
+          # TODO: remove debug flag
+          - --debug
 {{- if index .Cluster.ConfigItems "application_registry_url" }}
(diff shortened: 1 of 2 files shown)


demonCoder95:
e2e on the `dev` branch passes. Will switch the base branch to `eks`, to be able to set the override in an EKS cluster to test.

cluster/manifests/02-admission-control/deployment.yaml
```diff
@@ -33,7 +33,7 @@ spec:
       priorityClassName: system-cluster-critical
       containers:
       - name: admission-controller
-        image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/admission-controller:master-206
```
demonCoder95: This is the only related change, the rest should go away once we merge `dev-to-eks` branch.

demonCoder95: Rebased on latest `eks` after the merge, and now it's as expected.

demonCoder95:
We found two issues with this approach:
1. Roles cannot aggregate into ClusterRoles.
2. admission-controller cannot create Roles in a Namespace that it's currently admitting. 

We decided to implement this otherwise. So, will close this PR.

PR #8304, opened 2024-10-07
Fix ingress.cluster.local feature for eks

mikkeloscar:
This changes CIDR ranges config used for the skipper-ingress east-west feature based on the cluster setup.

For EKS the CIDR range is either the VPC ipv4 or ipv6 CIDR range and for non EKS it stays as it is.

@AlexanderYastrebov Let me know if these conditions are too complex or if they are readable 😅

* [x] Depends on `vpc_ipv6_cidrs` exposed in CLM: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/825

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -3,9 +3,17 @@
 {{ $main_image := print "container-registry.zalando.net/teapot/skipper-internal:" "v0.21.211-1033" }}
 {{ $canary_image := "container-registry.zalando.net/teapot/skipper-internal:v0.21.216-1038" }}
 
-
 {{/* Optional canary arguments separated by "[cf724afc]" to allow whitespaces, e.g. "-foo=has a whitespace[cf724afc]-baz=qux" */}}
 {{ $canary_args := "" }}
+
+{{ $cluster_internal_cidrs := stringSlice "10.2.0.0/15" .Values.vpc_ipv4_cidr }}
+{{ if eq .Cluster.Provider "zalando-eks" }}
+{{ if eq .Cluster.ConfigItems.eks_ip_family "ipv6" }}
+{{ $cluster_internal_cidrs := .Values.vpc_ipv6_cidrs }}
+{{ else }}
+{{ $cluster_internal_cidrs := stringSlice .Values.vpc_ipv4_cidr }}
+{{ end }}
+{{ end }}
 
 {{ template "skipper-ingress" dict
   "name" "skipper-ingress"
@@ -13,6 +21,7 @@
 
   "Cluster" .Cluster
   "Values" .Values
+  "cluster_internal_cidrs" $cluster_internal_cidrs
 }}
 
 {{ if eq .Cluster.ConfigItems.skipper_ingress_canary_enabled "true" }}
@@ -24,6 +33,7 @@
 
   "Cluster" .Cluster
   "Values" .Values
+  "cluster_internal_cidrs" $cluster_internal_cidrs
 }}
 {{ end }}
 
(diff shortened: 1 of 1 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -172,7 +172,7 @@ spec:
           - "-wait-for-healthcheck-interval={{ .Cluster.ConfigItems.skipper_wait_for_healthcheck_interval }}"
 {{ if ne .Cluster.ConfigItems.skipper_routesrv_enabled "exec" }}
           - "-kubernetes-east-west-range-domains=ingress.cluster.local"
-          - "-kubernetes-east-west-range-predicates=ClientIP(\"10.2.0.0/15\", \"{{ .Values.vpc_ipv4_cidr }}\")"
+          - "-kubernetes-east-west-range-predicates=ClientIP({{ if eq .Cluster.Provider "zalando-eks" }}{{if eq .Cluster.ConfigItems.eks_ip_family "ipv4"}}\"{{ .Values.vpc_ipv4_cidr }}\"{{else}}\"{{join .Values.vpc_ipv6_cidrs `\",\"`}}\"{{end}}{{else}}\"10.2.0.0/15\", \"{{ .Values.vpc_ipv4_cidr }}\"{{end}})"
```
AlexanderYastrebov: Would it be possible to define `cluster_internal_cidrs` variable to use it like:
```yaml
- "-kubernetes-east-west-range-predicates=ClientIP(\"{{ join $cluster_internal_cidrs `\",\"` }}\")"
...
- "-forwarded-headers-exclude-cidrs={{ join $cluster_internal_cidrs "," }}"
```

mikkeloscar: Good idea, will do!

mikkeloscar: Updated. This also needs a few changes in CLM: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/825

cluster/manifests/skipper/deployment.yaml
```diff
@@ -3,16 +3,25 @@
...
 {{ $canary_image := "container-registry.zalando.net/teapot/skipper-internal:v0.21.212-1034" }}
 
-
 {{/* Optional canary arguments separated by "[cf724afc]" to allow whitespaces, e.g. "-foo=has a whitespace[cf724afc]-baz=qux" */}}
 {{ $canary_args := "" }}
 
+{{ $cluster_internal_cidrs := stringSlice "10.2.0.0/15" .Values.vpc_ipv4_cidr }}
+{{ if eq .Cluster.Provider "zalando-eks" }}
+{{ if eq .Cluster.ConfigItems.eks_ip_family "ipv6" }}
+{{ $cluster_internal_cidrs := .Values.vpc_ipv6_cidrs }}
```
AlexanderYastrebov: This should use assignment as `cluster_internal_cidrs` already exists.
Should it include `10.2.0.0/15` or we just use what's in the value?

cluster/manifests/skipper/deployment.yaml
```diff
@@ -3,16 +3,25 @@
 {{ $main_image := print "container-registry.zalando.net/teapot/skipper-internal:" "v0.21.211-1033" }}
 {{ $canary_image := "container-registry.zalando.net/teapot/skipper-internal:v0.21.212-1034" }}
 
-
 {{/* Optional canary arguments separated by "[cf724afc]" to allow whitespaces, e.g. "-foo=has a whitespace[cf724afc]-baz=qux" */}}
 {{ $canary_args := "" }}
 
+{{ $cluster_internal_cidrs := stringSlice "10.2.0.0/15" .Values.vpc_ipv4_cidr }}
```
AlexanderYastrebov: We can also make it local inside `{{ define "skipper-ingress" }}` template and then we also don't need to pass it as template argument.

AlexanderYastrebov: There already exists `list` template function, I think it should work instead of `stringSlice`

AlexanderYastrebov: See https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/825#discussion_r1790531050

cluster/manifests/skipper/deployment.yaml
mikkeloscar: `10.2.0.0/15` is only relevant for non EKS clusters.

cluster/manifests/skipper/deployment.yaml
mikkeloscar: I'm not totally sure how you would like to structure it. You're welcome to push a commit to change it as you suggest.

AlexanderYastrebov: Ah, ok, we need to for routesrv so it must be top-level.

AlexanderYastrebov:
:+1:

mikkeloscar:
:+1:

AlexanderYastrebov:
@mikkeloscar Shouldn't this be merged to dev branch?

mikkeloscar:
We don't yet have things like `.ConfigItems.eks_ip_family` in `dev` so it was more fitting for `eks` for now. Eventually that should end in `dev` though.

The prometheus change I made in `dev` because it was generic enough.

PR #8308, opened 2024-10-07
fix: use resync as default in pod deletion cost controller

szuecs:
fix: use resync as default in pod deletion cost controller


AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #8321, opened 2024-10-08
skipper-canary-controller: add preStop for the cronjob

MustafaSaber:
Give time for the controller to push metrics to lightstep


cluster/manifests/skipper-canary-controller/canary-cronjob.yaml
```diff
@@ -30,6 +30,10 @@ spec:
           - name: skipper-canary-controller
             terminationMessagePolicy: FallbackToLogsOnError
             image: container-registry.zalando.net/gwproxy/skipper-canary-controller:main-29
+            lifecycle:
+              preStop:
+                sleep:
+                  seconds: 30
```
AlexanderYastrebov, line 36: If we need any delay it should be added into the application shutdown hook.
Also value of 30 equals default termination grace period so this leaves no time for pod to shutdown.

MustafaSaber:
close in favour of https://github.com/zalando-incubator/kubernetes-on-aws/pull/8331

PR #7917, opened 2024-08-01
Configure zalando-eks as provider during e2e tests

linki:
Configure zalando-eks as provider during e2e tests.

PR #8326, opened 2024-10-08
EKS: Make e2e work

mikkeloscar:
Follow up to #7917 making e2e for EKS work with a bit less hacks.

When this is merged it will configure the `eks` branch like this:

Only run e2e for `zalando-eks`. Only create the cluster based on the branch, don't update cluster from the target branch `dev`. This would not work as it would try to update a cluster from non-eks to eks. We later need to adjust this before merging into `dev` to avoid disabling non-eks e2e tests. We can only do this shortly before we're ready to merge `eks` branch into `dev`.

The load-test step has also been disabled as this doesn't work when: `E2E_SKIP_CLUSTER_UPDATE=true`. Will test this with a PR against `eks` branch once this is merged to validate that it works.

I also dropped the `toggle-scaledown.py` script as it seems meaningless now-a-days for our e2e and then we don't have to port it to work both with eks and non-eks.


mikkeloscar:
:+1:

test/e2e/cluster_config.sh
```diff
@@ -47,19 +47,37 @@ clusters:
...
+  $(if [ "${CLUSTER_PROVIDER}" == "zalando-eks" ]; then
+cat <<EOFF
+- config_items:
+      labels: dedicated=cluster-seed
+      taints: dedicated=cluster-seed:NoSchedule
+    discount_strategy: none
+    instance_types:
+    - "m6i.xlarge"
+    max_size: 99
+    min_size: 2
```
demonCoder95, line 67: will CLM not complain here, keeping minsize not a multiple of 3? (it usually does when creating pools)

test/e2e/run_e2e.sh
```diff
@@ -215,9 +244,6 @@ if [ "$e2e" = true ]; then
           junit_reports/ "s3://$RESULT_BUCKET/$TARGET_DIR/"
     fi
 
-    # enable cluster downscaling after running e2e
-    ./toggle-scaledown.py enable
-
```
demonCoder95, line 220: why don't we need this anymore?

test/e2e/cluster_config.sh
mikkeloscar: ` profile: worker-combined` does not have this restriction because it has an ASG spanning all zones as opposed to one ASG per zone.

It's actually not intended to autoscale as what fits on the seed node should not. It's kind of like a master node in non-eks setup as concept.

demonCoder95: Ah, got it 🙂

test/e2e/run_e2e.sh
mikkeloscar: Updated the description to mention this.

demonCoder95: Thanks! 🙂

demonCoder95:
Probably need to drop `e2e-load-test-result` step from dependencies of other steps as well.

mikkeloscar:
@demonCoder95 Yes, I just changed the way to disable to work better and be easier to toggle.

mikkeloscar:
:+1:

demonCoder95:
👍

PR #8328, opened 2024-10-08
prometheus: ipv6 pod support

mikkeloscar:
Handle both ipv4 and ipv6 addresses when scraping for the `kubernetes-pods` job.

The motivation for this is to make it work with EKS ipv6 clusters, but it should be backwards compatible with non-eks/ipv4 clusters.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #8358, opened 2024-10-10
Introduce S3 csi driver (disabled by default)

mikkeloscar:
This introduces support for the S3 CSI driver: https://github.com/awslabs/mountpoint-s3-csi-driver

It's disabled by default, but in clusters where it's enabled it will allow creation of PVs pointing to an S3 bucket and mount the PVC between multiple pods which can then write to a volume backed by S3.

It's not configured with an IAM role for the driver, meaning users need to define an iam role on the service account of the pods which will be used for the s3 access: https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/docs/CONFIGURATION.md#pod-level-credentials

This implementation has one caveat in our setup. It relies on the EKS annotation on serviceAccounts: `eks.amazonaws.com/role-arn` as opposed to the one we use today: `iam.amazonaws.com/role`. If the pod needs IAM roles for more than s3 driver access then adding both annotations is a workaround.

The intention is to roll this out to select clusters first and evaluate if this should be a promoted GA feature. It comes with daemonset overhead when enabled.

### DIFF
--- a/cluster/manifests/s3-csi-driver/daemonset.yaml
+++ b/cluster/manifests/s3-csi-driver/daemonset.yaml
@@ -1,0 +1,180 @@
+{{- if eq .Cluster.ConfigItems.s3_csi_driver "true" }}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    application: kubernetes
+    component: aws-mountpoint-s3-csi-driver
+    app.kubernetes.io/name: aws-mountpoint-s3-csi-driver
+  name: s3-csi-node
+  namespace: kube-system
+spec:
+  selector:
+    matchLabels:
+      daemonset: s3-csi-node
+  template:
+    metadata:
+      labels:
+        daemonset: s3-csi-node
+        application: kubernetes
+        component: aws-mountpoint-s3-csi-driver
+    spec:
+      priorityClassName: system-node-critical
+      serviceAccountName: s3-csi-driver
+      nodeSelector:
+        node.kubernetes.io/role: worker
+{{- if eq .Cluster.ConfigItems.s3_csi_driver_node_selector_restricted "true" }}
+        s3-csi-driver: enabled
+{{- end }}
+      tolerations:
+      - operator: Exists
+      containers:
+      - args:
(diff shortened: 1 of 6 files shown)


cluster/manifests/s3-csi-driver/daemonset.yaml
```diff
@@ -0,0 +1,180 @@
...
+        volumeMounts:
+        - mountPath: /csi
+          name: plugin-dir
+      initContainers:
+      - command:
+        - /bin/install-mp
+        env:
+        - name: MOUNTPOINT_INSTALL_DIR
+          value: /target
+        image: container-registry.zalando.net/teapot/aws-mountpoint-s3-csi-driver:v1.9.0-master-21
```
demonCoder95, line 137: One TODO for later maybe would be to refactor this so the image-updater can update it.

cluster/manifests/s3-csi-driver/daemonset.yaml
```diff
@@ -0,0 +1,180 @@
...
+            memory: 50Mi
+        volumeMounts:
+        - mountPath: /csi
+          name: plugin-dir
+        - mountPath: /registration
+          name: registration-dir
+      - args:
+        - --csi-address=/csi/csi.sock
+        - --health-port=9810
+        image: container-registry.zalando.net/teapot/livenessprobe:v2.14.0-eks-1-31-4-master-20
```
demonCoder95, line 117: Same here

cluster/manifests/s3-csi-driver/daemonset.yaml
demonCoder95: (same for any other images used)

mikkeloscar: What do you mean? It can update it or?

mikkeloscar:
:+1:

cluster/manifests/s3-csi-driver/daemonset.yaml
demonCoder95: I just misunderstood if bot can update images without the `$image` variables 😅

demonCoder95:
👍

PR #8364, opened 2024-10-11
Allow higher memory limit for flannel and kube-proxy

mikkeloscar:
This introduces new config-items to control requests and limits of memory for `kube-proxy` and `flannel` independently. The idea behind this is to easy operations of components that scale vertically with the cluster size and let them sometime spike above the requests while still having a limit to prevent complete uncontrolled memory usage.

Apart from this change we also need a change to admission-controller to allow setting this different for daemonset pods in certain namespaces.

([Internal design issue](https://github.bus.zalan.do/teapot/issues/issues/3720
))

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -402,16 +402,19 @@ node_exporter_experimental_metrics: "false"
 
 # kube-proxy settings
 kube_proxy_cpu: "50m"
-kube_proxy_memory: "200Mi"
+kube_proxy_memory_request: "200Mi"
+kube_proxy_memory_limit: "200Mi"
 kube_proxy_sync_period: "15m0s"
 kube_proxy_verbose_level: "2"
 
 # flannel settings
 flannel_cpu: "25m"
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

demonCoder95:
👍

demonCoder95:
What's our plan for the existing config-items? Do we use them in `requests` now?

mikkeloscar:
> What's our plan for the existing config-items? Do we use them in `requests` now?

Yes, maybe I should rename them? This just requires a migration step because the current values are defined in cluster-config-manager. This also reminds me I need to add the limit config-item to cluster-config-manager for those clusters where the requests are set differently.

demonCoder95:
Yes, renaming sounds better. I removed my +1 just in case.

mikkeloscar:
@demonCoder95 Config items were renamed and the clusters which has a non default value has gotten a compatible update.

mikkeloscar:
:+1:

linki:
👍

PR #8391, opened 2024-10-17
[WIP] Prepare for Custom Networking

katyanna:
Prepare AMI to adopt VPC CNI Custom Networking.

### DIFF
--- a/cluster/node-pools/worker-splitaz/userdata.yaml
+++ b/cluster/node-pools/worker-splitaz/userdata.yaml
@@ -46,6 +46,9 @@ write_files:
                - get-token
                - --cluster-name
                - "{{.Cluster.ID}}"
+               - --network-plugin=cni
+               - --cni-config-dir
+               - --cni-bin-dir
       contexts:
       - context:
           cluster: local
--- a/cluster/node-pools/worker-splitaz/userdata.yaml
+++ b/cluster/node-pools/worker-splitaz/userdata.yaml
@@ -47,8 +47,8 @@ write_files:
                - --cluster-name
                - "{{.Cluster.ID}}"
                - --network-plugin=cni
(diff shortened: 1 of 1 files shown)


cluster/node-pools/worker-splitaz/userdata.yaml
```diff
@@ -46,6 +46,9 @@ write_files:
                - get-token
                - --cluster-name
                - "{{.Cluster.ID}}"
+               - --network-plugin=cni
+               - --cni-config-dir
+               - --cni-bin-dir
```
myaser:
```suggestion
               - --cni-conf-dir=/etc/cni/net.d
               - --cni-bin-dir=/opt/cni/bin
```

cluster/node-pools/worker-splitaz/userdata.yaml
```diff
@@ -46,6 +46,9 @@ write_files:
                - get-token
                - --cluster-name
                - "{{.Cluster.ID}}"
+               - --network-plugin=cni
+               - --cni-conf-dir=/etc/cni/net.d
+               - --cni-bin-dir=/opt/cni/bin
```
linki, line 51: Are these flags defined at the right place?

linki: Given that we already run CNI for the EKS setup, we should be able to [peek here](https://github.com/zalando-incubator/kubernetes-on-aws/pull/7808/files#diff-fbbc8221d3bad7fc60e7c368f7da00a60c7e56a4060b29196a89e0834e42755c) for some inspiration on how to set it up.

myaser: no, that is not the right place. this PR is wrong and should have been closed

PR #8395, opened 2024-10-18
Fix podCIDR handling in kube-controller-manager

mikkeloscar:
This switches to a patched version of `kube-controller-manager` which includes a fix for podCIDR handling as described here: https://github.com/kubernetes/kubernetes/issues/127792#issuecomment-2419735724

Once there is an upstream fix we can revert from the patched image.


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #8406, opened 2024-10-21
Exclude *5d.large instances in karpenter

mikkeloscar:
Similar to https://github.com/zalando-incubator/kubernetes-on-aws/pull/6887 add also m and r instance types to be excluded from Karpenter default instance types because they are slow to start with instance storage.

For some reason these instance types are often the first choice on spot so it's common to get these slow to startup nodes in pet clusters which is annoying.


mikkeloscar:
:+1:

katyanna:
👍

PR #8407, opened 2024-10-21
Make it possible to use IRSA for ZMON IAM role [1/2]

mikkeloscar:
Extend the trust policy of the ZMON IAM role in order to switch ZMON components to IRSA/OIDC based flow

This is one of two PRs. We need to clean up the old trust policy once ZMON components are fully migrated to IRSA.


mikkeloscar:
:+1:

linki:
👍

PR #8426, opened 2024-10-22
Make kube-node-ready optional

mikkeloscar:
`kube-node-ready` has lost its usefulness for several reasons:

* Karpenter nodes don't run it which is the majority of nodes in most clusters
* `kube-node-ready` was introduced at a time where we had a problem that ASGs could be automatically AZ reblanced, terminating nodes before the replacements had a chance to startup. Having nodes only marked ready in the ASG once the kube-node-ready pod was running slowed this rebalancing down. We no longer have AZ rebalance enabled, so not relevant.
* We used it as a pseudo way to check pod network functionality, but we now have kubenurse which does a better job at this.

This PR introduces a config-item `kube_node_ready_enabled` which lets us disable it. The intention is to first disable in test clusters and later in all clusters.

This is indirectly motivated by switching to IMDSv2 for which `kube-node-ready` is not compatible with.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -484,6 +484,9 @@ kubernetes_lifecycle_metrics_mem_min: "120Mi"
 kube_node_ready_controller_cpu: "50m"
 kube_node_ready_controller_memory: "200Mi"
 
+# Enable kube-node-ready ASG lifecycle hook feature.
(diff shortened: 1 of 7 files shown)


mikkeloscar:
:+1:

demonCoder95:
👍

PR #8432, opened 2024-10-22
Set AWS_DEFAULT_REGION for kube2iam

mikkeloscar:
Follow up to #8427 

Since kube2iam was upgraded to use aws-go-sdk-v2 we need to set `AWS_DEFAULT_REGION` or we get errors like:

```
"Error assuming role operation error EC2: DescribeRegions, failed to resolve service endpoint, endpoint rule error, Invalid Configuration: Missing Region"
```


linki:
👍

mikkeloscar:
:+1:

PR #8436, opened 2024-10-23
admission-control: validate that Ingress and RouteGroup hosts in hosted zone domain

AlexanderYastrebov:
Add ValidatingAdmissionPolicy that validates Ingress and RouteGroup hosts from hosted zone parent domain are in hosted zone domain.

E.g. for hosted zone `foo.bar.test` its parent domain is `bar.test` and therefore Ingress and RouteGroup hosts from `bar.test` domain must also be in `foo.bar.test` domain.

### DIFF
--- a/cluster/manifests/01-admission-control/host-policy.yaml
+++ b/cluster/manifests/01-admission-control/host-policy.yaml
@@ -1,0 +1,93 @@
+# {{ $hosted_zone_parent_domain := slice (split .Values.hosted_zone ".") 1 | join "." }}
+
+# {{ if eq .Cluster.ConfigItems.ingresses_validation "enabled" }}
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: ingress-host-policy.teapot.zalan.do
+  annotations:
+    kubernetes.io/description: |
+      Validates that Ingress hosts from {{ $hosted_zone_parent_domain }} domain are in {{ .Values.hosted_zone }} domain.
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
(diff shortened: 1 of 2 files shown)


cluster/manifests/01-admission-control/hostname-policy.yaml
```diff
@@ -0,0 +1,56 @@
+# {{ if or (eq .Cluster.ConfigItems.ingresses_validation "enabled") (eq .Cluster.ConfigItems.routegroups_validation "enabled") }}
+# {{ $hosted_zone_parent_domain := slice (split .Values.hosted_zone ".") 1 | join "." }}
```
AlexanderYastrebov: This assumes `.Values.hosted_zone` is a three-level domain like `foo.bar.test`

cluster/manifests/01-admission-control/host-policy.yaml
```diff
@@ -0,0 +1,93 @@
+# {{ $hosted_zone_parent_domain := slice (split .Values.hosted_zone ".") 1 | join "." }}
+
+# {{ if eq .Cluster.ConfigItems.ingresses_validation "enabled" }}
```
AlexanderYastrebov, line 3: Do we need another new config to toggle theses policies off?

mikkeloscar:
Maybe this stuff deserves an e2e test?

AlexanderYastrebov:
> E.g. for hosted zone foo.bar.test its parent domain is bar.test and therefore Ingress and RouteGroup hosts from bar.test domain must also be in foo.bar.test domain.

There could be multiple hosted zones (e.g. foo.bar.test and qux.bar.test) in the same cluster using the same parent domain (bar.test) so this validation logic will not accept *.qux.bar.test because it only accepts *.foo.bar.test

PR #8440, opened 2024-10-25
Allow to separate subnets for LBs and nodes

mikkeloscar:
* [x] Depends on https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/827

Allows to have separate subnets for LBs and for nodes which enables different rules for ingress and egress.

Existing clusters will continue to use the same subnets for both. A migration is needed to split it.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #8481, opened 2024-10-30
Use local endpoint for deployment-controller to status-service

mikkeloscar:
This adds a cluster local endpoint for status-service-url which can be used by the deployment-controller so it doesn't have to connect via the ingress NLB as illustrated below

image

Before it was using the red path and with this change it communicates internally via the green path.
The `status-service-url` is still kept in the config-map because this is used by the central `deployment-service` to get the public address.

### DIFF
--- a/cluster/manifests/deployment-service/01-config.yaml
+++ b/cluster/manifests/deployment-service/01-config.yaml
@@ -15,6 +15,7 @@ data:
   oidc-subject-key: "{{.Cluster.LocalID}}.{{.Cluster.Alias}}.zalan.do:sub"
   s3-bucket-name: "zalando-deployment-service-{{accountID .Cluster.InfrastructureAccount}}-{{.Cluster.LocalID}}"
   status-service-url: "https://deployment-status-svc-{{.Cluster.Alias}}.{{.Values.hosted_zone}}"
+  status-service-url-local: "http://deployment-status-service.ingress.cluster.local."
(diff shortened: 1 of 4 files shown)


mikkeloscar:
:+1:

cluster/manifests/deployment-service/01-config.yaml
```diff
@@ -15,6 +15,7 @@ data:
   oidc-subject-key: "{{.Cluster.LocalID}}.{{.Cluster.Alias}}.zalan.do:sub"
   s3-bucket-name: "zalando-deployment-service-{{accountID .Cluster.InfrastructureAccount}}-{{.Cluster.LocalID}}"
   status-service-url: "https://deployment-status-svc-{{.Cluster.Alias}}.{{.Values.hosted_zone}}"
+  status-service-url-local: "http://deployment-status-service.ingress.cluster.local."
```
demonCoder95, line 18: Do we need the `.` at the end?

mikkeloscar: This is just a small optimization to tell the DNS resolver to not add search domains as this is the full name (this could ofc. be added in a lot of places).

demonCoder95: I see. Thanks 🙏

demonCoder95:
👍

PR #8482, opened 2024-10-30
add role-sync-controller cronjob

demonCoder95:
This introduces the CronJob we need for RBAC resource synchronisation required by the new RBAC setup to replace the Authz webhook in the api-server.

This is introduced behind a feature flag with the default value of `false`.

### DIFF
--- a/cluster/manifests/role-sync-controller/cronjob.yaml
+++ b/cluster/manifests/role-sync-controller/cronjob.yaml
@@ -1,0 +1,21 @@
+{{ if .Cluster.ConfigItems.role_sync_controller_enabled "true" }}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: role-sync-controller
+  labels:
+    application: kubernetes
+    component: role-sync-controller
+spec:
+  schedule: "*/1 * * * *"
+  jobTemplate:
+    spec:
+      backoffLimit: 3
+      template:
+        spec:
+          restartPolicy: OnFailure
+          containers:
+          - name: role-sync-controller
+            image: container-registry.zalando.net/teapot/role-sync-controller:main-1
+            imagePullPolicy: IfNotPresent
+{{ end }}
--- a/cluster/manifests/role-sync-controller/cronjob.yaml
+++ b/cluster/manifests/role-sync-controller/cronjob.yaml
@@ -3,6 +3,7 @@ apiVersion: batch/v1
 kind: CronJob
 metadata:
   name: role-sync-controller
+  namespace: kube-system
   labels:
     application: kubernetes
     component: role-sync-controller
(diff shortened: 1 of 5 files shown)


cluster/manifests/role-sync-controller/cronjob.yaml
```diff
@@ -0,0 +1,21 @@
+{{ if .Cluster.ConfigItems.role_sync_controller_enabled "true" }}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: role-sync-controller
+  labels:
+    application: kubernetes
+    component: role-sync-controller
```
mikkeloscar, line 9: should be in `kube-system`

Needs also RBAC role to read namespaces/create roles

cluster/manifests/role-sync-controller/cronjob.yaml
```diff
@@ -0,0 +1,21 @@
...
+  jobTemplate:
+    spec:
+      backoffLimit: 3
+      template:
+        spec:
+          restartPolicy: OnFailure
+          containers:
+          - name: role-sync-controller
+            image: container-registry.zalando.net/teapot/role-sync-controller:main-1
+            imagePullPolicy: IfNotPresent
```
mikkeloscar: nit: this is the default, let's just drop it from the manifest

cluster/manifests/role-sync-controller/cronjob.yaml
demonCoder95: I was wondering about using `kube-system` 😅 
I'll add the RBAC as well. Thanks 🙏

cluster/manifests/role-sync-controller/cronjob.yaml
demonCoder95: Dropped in the next commit. Thanks.

cluster/manifests/role-sync-controller/cronjob.yaml
demonCoder95: Added in the next commit.

cluster/manifests/role-sync-controller/cronjob.yaml
```diff
@@ -0,0 +1,21 @@
+{{ if .Cluster.ConfigItems.role_sync_controller_enabled "true" }}
```
mikkeloscar: Since this is behind a feature flag, we should also have entries in `deletions.yaml` to clean it up if first enabled and then disabled again.

demonCoder95: ah, good point. I'll add.

demonCoder95: Added in the next commit

linki:
👍

demonCoder95:
👍

mikkeloscar:
:+1:

PR #8487, opened 2024-10-31
Set correct cluster tag on EKS

mikkeloscar:
This sets the correct cluster on EKS which is `kubernetes.io/cluster/<eks-cluster-name>: owned`. This fixes support for Service Type LoadBalancer because the EKS cloud-controller-manager was not able to find and attach nodes as they had the wrong cluster tag.

This is a bit messy because we use `kubernetes.io/cluster/<cluster-id>` on non-EKS and `kubernetes.io/cluster/<eks-cluster-name>` on EKS. So right now it's adding both tags on EKS to be compatible e.g. with e2e cleanup.

I propose to create an internal tech-depth issue to discuss how to clean up this mess with cluster ID vs. Cluster Name.

### Depends on

* [x] https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/829

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -5,9 +5,7 @@ Metadata:
     InfrastructureComponent: "true"
     application: "kubernetes"
     cluster-provider: "{{ .Cluster.Provider }}"
-# {{ if eq .Cluster.Provider "zalando-eks" }}
-    "kubernetes.io/cluster/{{.Cluster.ID | eksID}}": "owned"
-# {{ end }}
+    "kubernetes.io/cluster/{{.Cluster.Name}}": "owned"
 
 Resources:
 {{ if eq .Cluster.Provider "zalando-eks" }}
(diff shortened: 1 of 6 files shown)


katyanna:
:+1:

cluster/cluster.yaml
```diff
@@ -666,9 +623,6 @@ Resources:
       GroupId: !Ref WorkerSecurityGroup
       IpProtocol: tcp
       SourceSecurityGroupId: !Ref MasterSecurityGroup
-      Tags:
-        - Key: 'kubernetes.io/cluster/{{.Cluster.ID}}'
-          Value: owned
```
mikkeloscar, line 671: All these cases are dropped because we set the tag at stack level already

mikkeloscar:
:+1:

cluster/cluster.yaml
```diff
@@ -27,18 +29,11 @@ Resources:
   EKSControlPlaneSecurityGroup:
     Type: AWS::EC2::SecurityGroup
     Properties:
-      # GroupDescription: "{{ .Cluster.Alias }}-eks-control-plane"
       GroupDescription: "{{ .Cluster.Alias }}-control-plane"
-      # TODO:
-      # SecurityGroupIngress:
```
demonCoder95, line 33: Was this TODO about adding the cluster tag on the ingress security group? I'm curious why it's removed.

cluster/cluster.yaml
```diff
@@ -1173,9 +1104,9 @@ Resources:
                   ],
                   "Condition": {
                     "StringEquals": {
-                      "aws:ResourceTag/kubernetes.io/cluster/{{.Cluster.ID}}": "owned",
+                      "aws:ResourceTag/kubernetes.io/cluster/{{.Cluster.Name}}": "owned",
+                      "aws:RequestTag/kubernetes.io/cluster/{{.Cluster.Name}}": "owned",
```
demonCoder95, line 1108: ~duplicate line?~ nevermind 😅 
Why did we add this request tag now?

mikkeloscar: It's just because I moved line 1178 up here instead. It made more sense in the original PR where I used if guards. I can also revert this part so the lines don't jump.

cluster/cluster.yaml
mikkeloscar: This was some left over comments from when I initially did the EKS setup. I removed them now, but it's ofc. not related to this particular change so I can also move it to a separate PR.

cluster/cluster.yaml
demonCoder95: If it's just moving the lines then it's fine. I thought we changed something otherwise.

cluster/cluster.yaml
demonCoder95: If the TODO is complete, it's fine to drop it here, I guess.

demonCoder95:
👍

PR #8488, opened 2024-10-31
fix RBAC for role-sync-controller

demonCoder95:
The CronJob itself needs `poweruser` role to be able to bind the role to `okta:common/engineer`.

I deleted the `deletions.yaml` entry as well since we haven't rolled the feature out yet, so there are no `ClusterRoles` in place.


demonCoder95:
👍

linki:
👍

PR #8491, opened 2024-10-31
Test RBAC e2e for EKS

demonCoder95:
This enables the role-sync-controller for e2e clusters on EKS. Also, re-enables the RBAC e2e test.

## Summary of Changes
1. Read the `kubecofig` from the local file system and use it to authenticate with the `api-server` for the RBAC e2e tests. This is needed since the authentication model with the api-server in EKS is different from our current setup, it requires a CA Cert in addition to the token. Using the `kubeconfig` we already build is an easier method. There's a [programmatic way of getting the token](https://docs.aws.amazon.com/eks/latest/best-practices/identity-and-access-management.html#_controlling_access_to_eks_clusters) but that requires knowing the name of the cluster, which is not easy to provide. 

An alternate solution is to use `client-go` and rely on that to perform the handshake under the hood without having to deal with these unnecessary details in the e2e tests, I'll attempt that if this gets too messy.

We decided to close this PR and rewrite all the RBAC e2e tests.


demonCoder95:
As we've decided to rewrite the RBAC e2e tests in #8532 , will close this PR in favor of that one. Once the refactor is complete, will rebase that PR on `eks` branch to test this.

PR #8504, opened 2024-11-04
Register `/rolebinding` admitter in order to reject certain rolebindings

linki:
This will catch creations and updates of `rolebindings` and `clusterrolebindings` so that they can be inspected by admission-controller and rejected in certain cases.


linki:
👍

mikkeloscar:
:+1:

linki:
👍

PR #8507, opened 2024-11-04
Optional internal control plane LB

mikkeloscar:
This adds optional support for adding an internal Load balancer for the control plane nodes. That is, it will allow workers to connect to the control plane inside the VPC and not go through the public LB used by external users. This is illustrated in the diagram below:

image

This is helpful if the cluster/account has custom egress routing to the internet which doesn't work well if something inside the VPC calls an NLB in the VPC via the internet.

The feature is controller via the config-item: `control_plane_load_balancer_internal` and is disabled by default.


diegoroccia:
:+1:

linki:
👍

PR #8510, opened 2024-11-05
eks: Optional support for AWS VPC CNI Custom Networking

mikkeloscar:
This adds optional support for using [Custom Networking](https://docs.aws.amazon.com/eks/latest/best-practices/custom-networking.html) as a way to get more IPv4 space for pods.

This assumes that the target account has a VPC with an additional VPC CIDR for pods as well as subnets for pods tagged with `kubernetes.io/role/pod: enabled`.

* It also depends on a small change in CLM to discover the pod subnets via the tag: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/830


mikkeloscar:
:+1:

linki:
👍

mikkeloscar:
:+1:

PR #8516, opened 2024-11-06
Add config item to require FabricGateway versioning

RomanZavodskikh:
The versioned FabricGateway is tested enough to be sure it is safe to use for newly created FabricGateway's with external service provider. So this config item is introduced to eventually require such FabricGateway versioning on all clusters.


RomanZavodskikh:
:+1:

MustafaSaber:
Are we sure that this template condition will work in the CRD template?

I don't think we should merge this PR

MustafaSaber:
We will wait a bit for this one, will close for now.

PR #8517, opened 2024-11-06
[EKS] Register generic admitter for write protection behind a feature flag

linki:
This adds an admitter that catches unprivileged write requests to the API.


cluster/manifests/prometheus/rbac.yaml
```diff
@@ -8,6 +8,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: prometheus
+  labels:
+    admission.zalando.org/infrastructure-component: "true"
```
linki, line 12: Example of how global resources can be opted-in.

linki:
👍

demonCoder95:
👍

PR #8518, opened 2024-11-08
Update admission-controller to master-224

linki:
This updates admission-controller to the latest version.

All changes are behind feature flags so this is just to make sure we have the most recent version and it works.

PR #8519, opened 2024-11-08
Register generic admitter for write protection behind a feature flag

linki:
This is the same as https://github.com/zalando-incubator/kubernetes-on-aws/pull/8517 for dev (there are little conflicts that I'll fix in the `dev-to-eks` merge)

This supercedes https://github.com/zalando-incubator/kubernetes-on-aws/pull/8518

The goal is to deploy the latest version of admission-controller in dev, so that we have an up-to-date version. The new features are entirely turned off so **this is a no-op**.

### DIFF
--- a/cluster/manifests/prometheus/rbac.yaml
+++ b/cluster/manifests/prometheus/rbac.yaml
@@ -8,6 +8,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: prometheus
+  labels:
+    admission.zalando.org/infrastructure-component: "true"
 rules:
 - apiGroups: [""]
   resources:
@@ -37,6 +39,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: prometheus
+  labels:
+    admission.zalando.org/infrastructure-component: "true"
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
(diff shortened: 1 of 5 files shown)


cluster/manifests/prometheus/rbac.yaml
```diff
@@ -8,6 +8,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: prometheus
+  labels:
+    admission.zalando.org/infrastructure-component: "true"
```
demonCoder95, line 12: Is there some particular reason we're doing this specifically for prometheus only? I'm assuming given the rules of admission specified above, we will need this for all other infrastructure resources as well?

(Like say, the RBAC resources in the `roles` directory)

I know the feature is disabled by default so it's not relevant, but I'm curious, in terms of enabling this long term

demonCoder95:
👍

cluster/manifests/prometheus/rbac.yaml
linki: Yes, it will be needed on all resources but I'm not sure yet if it's the right approach. I didn't want to find and attach it to everything yet to keep the diff small. On the other hand I want something I can test with.

demonCoder95: Makes sense. I also saw in your [comment on the other PR](https://github.com/zalando-incubator/kubernetes-on-aws/pull/8517/files#r1831487728) that's to trigger an opt-in where we can test. Sounds good 🙂

linki:
👍

PR #8524, opened 2024-11-08
make kube-janitor opt-in for production clusters

demonCoder95:
This is required to address some support requests where we need to enable kube-janitor in some production clusters. This keeps the default behavior intact, so it's a no-op.

In a follow-up, we probably need to expose some kube-janitor configuration as config-items as well, to better control its behavior per-cluster.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -1170,7 +1170,13 @@ control_plane_load_balancer_internal: "none"
 #  fs.inotify.max_user_watches = 100000
 sysctl_settings: ""
 
-
+# kube-janitor configuration
+{{if eq .Cluster.Environment "production"}}
+# This makes kube-janitor opt-in for production clusters
+kube_janitor_enabled: "false"
+{{else}}
+kube_janitor_enabled: "true"
+{{end}}
 
 # scheduling_controls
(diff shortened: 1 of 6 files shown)


cluster/config-defaults.yaml
```diff
@@ -1170,7 +1170,13 @@ control_plane_load_balancer_internal: "none"
 #  fs.inotify.max_user_watches = 100000
 sysctl_settings: ""
 
-
+# kube-janitor configuration
+{{if eq .Cluster.Environment "production"}}
+# This makes kube-janitor opt-in for production clusters
```
linki, line 1175: Let's add a visible note here that before enabling this in production, one should check whether there are any currently running applications that make use of the annotation. There might be cases where people don't limit the annotation to the test environment and it propagates to production where it doesn't have an effect _so far_. Enabling it could then trigger a deletion of their application at any time.

demonCoder95: Hmm I agree. I'll add a note, and also double-check in the opted cluster to make sure we don't have anything  side-effects. Thanks!

demonCoder95: I added the note.

linki:
👍

demonCoder95:
👍

PR #8529, opened 2024-11-19
[EKS] switch to generic deny-all admitter, match conditions and fail policy

linki:
Using `matchConditions` we can prevent the webhooks from being called for privileged components in the first place. This should solve the chicken and egg problem.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -677,9 +677,9 @@ teapot_admission_controller_configmap_deletion_protection_factories_enabled: "tr
 # enable the rolebinding admission-controller webhook which validates rolebindings and clusterrolebindings
 teapot_admission_controller_enable_rolebinding_webhook: "true"
 
-# enable the generic admission-controller webhook which catches all resources
-teapot_admission_controller_enable_generic_webhook: "true"
-# prevent write operations for non-admin users in protected namespaces
+# enable the generic deny-all admission webhook which rejects all requests it receives
+teapot_admission_controller_enable_write_protection_webhook: "true"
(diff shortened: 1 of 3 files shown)


linki:
👍

cluster/config-defaults.yaml
```diff
@@ -677,9 +677,9 @@ teapot_admission_controller_configmap_deletion_protection_factories_enabled: "tr
 # enable the rolebinding admission-controller webhook which validates rolebindings and clusterrolebindings
 teapot_admission_controller_enable_rolebinding_webhook: "true"
 
-# enable the generic admission-controller webhook which catches all resources
-teapot_admission_controller_enable_generic_webhook: "true"
-# prevent write operations for non-admin users in protected namespaces
+# enable the generic deny-all admission webhook which rejects all requests it receives
+teapot_admission_controller_enable_write_protection_webhook: "true"
+# configure the behaviour of the deny-all admission webhook, `true` blocks everything, `false` allows everything
 teapot_admission_controller_prevent_write_operations: "true"
```
demonCoder95, line 683: Do we still need `teapot_admission_controller_prevent_write_operations` config-item now that the webhook is deny-all or allow-all? If the webhook is deny-all, then this config item isn't necessary, and if it's allow-all then it can cause problems. No?

linki: It was mostly to turn it off quickly in case unwanted things are blocked. But I guess the other config item (that removes the webhook entirely) is equally quick (or changing the conditions). I would probably just leave it for now. It's not hurting to have it.

demonCoder95: Sure. As long as we have considered it 🙂

demonCoder95:
👍

PR #8531, opened 2024-11-19
Drop unused eks config-item

mikkeloscar:
Drops the unused `eks` config-item which we no longer use as we use `.Cluster.Provider == "zalando-eks"` to enable EKS specific configuration.


mikkeloscar:
:+1:

myaser:
👍🏻

PR #8532, opened 2024-11-19
[e2e] Refactor RBAC tests to use client-go instead of raw HTTP

demonCoder95:
## Background
The current RBAC e2e test implementation breaks with stricter CA Cert validation enforcement when doing TLS handshake with the api-server in EKS clusters.

This rewrites the RBAC test cases to use client-go, in order to avoid handling raw HTTP calls, and also creates a cleaner and more readable Gingko suite for RBAC e2e tests. This change for now, simply retains the existing tests (in `authorisation_test.go`) and writes these in addition to those. We can drop the older ones later.

## Summary of Changes
1. Drop `PodSecurityPolicy` related e2e tests as PSPs have been removed now.
2. Use `Context` and `When` containers for better organization of our RBAC e2e specs. The current implementation puts all the test cases in a single spec.
3. The utility functions are moved in `authorization_utils.go` file to keep the main `authorization.go` file cleaner.
4. Certain permission functionality from the Authz webhook has been moved to the admission-controller and this test suite being based on `SubjectAccessReview` can't verify that. These tests will be moved to the admission-controller e2e tests and removed from here.

I decided not to generalise the attribute expansion logic and make the code too brittle and complicated. The code for expansion right now is very verbose, but it's clearer to follow, understand and debug.


test/e2e/authorization_utils.go
```diff
@@ -0,0 +1,38 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func createSubjectAccessReview(ctx context.Context, cs kubernetes.Interface, sar authv1.SubjectAccessReview) (*authv1.SubjectAccessReview, error) {
+	return cs.AuthorizationV1().SubjectAccessReviews().Create(ctx, &sar, metav1.CreateOptions{})
+}
+
+func createLocalClient() (kubernetes.Interface, error) {
```
mikkeloscar: The e2e framework already provides a way to get a client (see other e2e tests in this repo). Doing it this way would make it more complicated to run locally as you would need to setup a hardcoded path: `/workdir/test/e2e/kubeconfig`

demonCoder95: I'll check. Thanks!

demonCoder95: Turns out that's much easier than this 😆   Just need to do
```golang
f := framework.NewDefaultFramework("framework-name")
cs := f.ClientSet
```

Done!

test/e2e/authorization_utils.go
```diff
@@ -0,0 +1,278 @@
...
+						Groups:             group,
+					},
+				}
+				sars = append(sars, sar)
+			}
+		}
+	}
+	return sars
+}
+
```
demonCoder95: This is a temporary fix to get the tests running, will be refactored into expansions.

demonCoder95: This is now done.

demonCoder95:
This is now ready for review 🎉

cluster/config-defaults.yaml
```diff
@@ -1194,4 +1194,8 @@ teapot_admission_controller_scheduling_controls_enabled: "false"
 teapot_admission_controller_scheduling_controls_default_architecture: "amd64"
 
 # role-sync-controller configs
+{{ if eq .Cluster.Environment "e2e" }}
+role_sync_controller_enabled: "true"
```
linki, line 1199: This would enable the sync controller for non-EKS clusters. Do we want that?

test/e2e/authorization.go
```diff
@@ -0,0 +1,572 @@
...
+					"apps/deployments",
+					"apps/statefulsets",
+					"apps/deployments/scale",
+					"apps/statefulsets/scale",
+					"services",
+					"persistentvolumes",
+					"persistentvolumeclaims",
+					"configmaps",
+				}
+				tc.data.verbs = []string{"create", "update", "delete", "patch"}
```
linki, line 212: We could extract these "static" lists as vars similar to `allGroups` in the top and then re-use them. For example:

```go
	var (
		namespacedResource = []string{
			"pods",
			"apps/deployments",
			"apps/statefulsets",
			"apps/deployments/scale",
			"apps/statefulsets/scale",
			"services",
			"persistentvolumes",
			"persistentvolumeclaims",
			"configmaps",
		}
		globalResources = []string{
			"namespaces",
			"nodes",
			"rbac.authorization.k8s.io/clusterroles",
			"storage.k8s.io/storageclasses",
			"apiextensions.k8s.io/customresourcedefinitions",
		}
		readOperations  = []string{"get", "list", "watch"}
		writeOperations = []string{"create", "update", "delete", "patch"}
	)
```

test/e2e/authorization.go
```diff
@@ -0,0 +1,572 @@
...
+			})
+			// These should be covered by the admission-controller tests.
+			// They're written here for completeness.
+			g.It("should deny write access in kube-system and visibility namespaces", func() {})
```
linki, line 216: Can we implement them but `skip` them? This way we can validate them if we later manage to implement them with RBAC.

test/e2e/authorization.go
```diff
@@ -0,0 +1,572 @@
...
+			g.It("should allow write access in namespaces other than kube-system and visibility", func() {})
```
linki, line 217: Similar to `"should allow write access in namespaces other than kube-system"` (line 314) we can add the implemetation here 🤷

test/e2e/authorization.go
```diff
@@ -0,0 +1,572 @@
...
+			})
+			g.It("should allow to update DaemonSet finalizers", func() {
+				tc.data.resources = []string{"apps/daemonsets/finalizers"}
+				tc.data.verbs = []string{"update"}
+				tc.run(context.TODO(), cs, true)
+				gomega.Expect(tc.output.passed).To(gomega.BeTrue(), tc.output.String())
+			})
+			// TODO: Need to verify in the original tests if this is a permission on
+			// the controller-manager or the daemonset-controller.
+			// g.It("should allow to create Pods", func() {})
```
linki, line 378: Is this clarified?

test/e2e/authorization.go
```diff
@@ -0,0 +1,572 @@
...
+		})
+
+		g.When("the default service account is in default namespace", func() {
```
linki, line 381:
```suggestion
		g.When("the service account is the default service account in the default namespace", func() {
```

test/e2e/authorization.go
```diff
@@ -0,0 +1,572 @@
...
+			g.BeforeEach(func() {
+				tc.data.users = []string{"system:serviceaccount:default:default"}
+			})
+			g.It("should deny to list StatefulSets", func() {
```
linki, line 385: Interesting

test/e2e/authorization.go
```diff
@@ -0,0 +1,572 @@
...
+
+		g.When("the default service account is in non-default namespace", func() {
+			g.BeforeEach(func() {
+				tc.data.users = []string{"system:serviceaccount:non-default:default"}
+			})
+			g.It("should deny to list StatefulSets", func() {
+				tc.data.resources = []string{"apps/statefulsets"}
+				tc.data.verbs = []string{"list"}
+				tc.run(context.TODO(), cs, false)
+				gomega.Expect(tc.output.passed).To(gomega.BeTrue(), tc.output.String())
```
linki, line 401: This is the same behaviour as the previous test, right? Can we combine them?

test/e2e/authorization.go
```diff
@@ -0,0 +1,572 @@
...
+					tc.data.names = []string{"skipper-default-filters"}
+					tc.run(context.TODO(), cs, true)
+					gomega.Expect(tc.output.passed).To(gomega.BeTrue(), tc.output.String())
+				})
+				g.It("should deny to update any other ConfigMap", func() {
+					tc.data.resources = []string{"configmaps"}
+					tc.data.verbs = []string{"update"}
+					// Technically, this should result in access undecided because we allow
+					// access to 'skipper-default-filters' ConfigMap only and we haven't
+					// specified a resource name in the test case.
```
linki, line 459: Can we just specify another resource name (like `foo`)?

test/e2e/authorization.go
```diff
@@ -0,0 +1,572 @@
...
+					gomega.Expect(tc.output.passed).To(gomega.BeTrue(), tc.output.String())
+				})
+			})
+		})
+
+		g.When("the user is k8sapi_credentials-provider", func() {
+			g.BeforeEach(func() {
+				tc.data.users = []string{"zalando-iam:zalando:service:k8sapi_credentials-provider"}
+			})
+			g.It("should allow to get Secrets in kube-system namespace", func() {
```
linki, line 471: (Allow-listing) writing secrets is again done in `admission-controller` 😠 Hopefully we find a better solution in the future.

linki: Actually RBAC also gives credentials provider read/write to secret. Maybe it makes sense to add a test for it as well?

test/e2e/authorization.go
```diff
@@ -0,0 +1,572 @@
...
+				tc.data.namespaces = []string{"kube-system"}
+			})
+			g.When("the resource is a Secret", func() {
+				g.BeforeEach(func() {
+					tc.data.resources = []string{"secrets"}
+				})
+				g.It("should allow read access", func() {
+					tc.data.verbs = []string{"get", "list", "watch"}
+					tc.run(context.TODO(), cs, true)
+					gomega.Expect(tc.output.passed).To(gomega.BeTrue(), tc.output.String())
```
linki, line 517: Let's also add a test for write access?

test/e2e/authorization.go
```diff
@@ -0,0 +1,572 @@
...
+				})
+				g.It("should allow write access", func() {
+					tc.data.verbs = []string{"create", "update", "delete", "patch"}
+					tc.run(context.TODO(), cs, true)
+					gomega.Expect(tc.output.passed).To(gomega.BeTrue(), tc.output.String())
+				})
```
linki, line 534: Since we expand the `verbs` anyway, we could also combine this into one test with `"should allow full access"`. Not sure if that makes it clearer but just a style fyi.

cluster/config-defaults.yaml
demonCoder95: Hmm I could change it to EKS only?

My plan is to ultimately keep the `role-sync-controller` enabled and disable the authz webhook at some point, for all clusters. This allows us:
1. single authz setup with no parity between eks/non-eks
2. test the setup by rolling it out on channels and fix issues if we find any

We can hold off on that while we still have some work pending on the admission-controller, I guess? What do you suggest?

test/e2e/authorization.go
demonCoder95: That's actually a good suggestions. I'll do that. Thanks.

test/e2e/authorization.go
demonCoder95: Sounds good. I'll do it.

test/e2e/authorization.go
demonCoder95: I agree. I'll add.

test/e2e/authorization.go
demonCoder95: No 🙁  Do you know what's the right permission here? The current setup uses `daemonset-controller` SA. I'm not sure if we keep it that way or change it to `controller-manager`.

test/e2e/authorization.go
demonCoder95: Not sure why we added this check 😀

test/e2e/authorization.go
demonCoder95: I'll check if we can combine. Maybe define the "list stateful sets" as the container block 🤔

test/e2e/authorization.go
demonCoder95: I can do that.

test/e2e/authorization.go
demonCoder95: Yeah 🤞

test/e2e/authorization.go
demonCoder95: I'll add. Thanks.

test/e2e/authorization.go
demonCoder95: Actually that sounds a lot better. I can do that.

cluster/config-defaults.yaml
demonCoder95: I updated to only enable this for EKS.

demonCoder95:
The remaining comments will be addressed in a follow-up PR, as agreed.

demonCoder95:
👍

cluster/config-defaults.yaml
linki: I'm fine with that plan.

But with this, it would create additional bindings in the namespaces that we don't need because auth-webhook is still running. I'm just concerned about confusion (it should be fine from a permissions perspective).

test/e2e/authorization.go
linki: I don't know for sure. Maybe you can `git blame` it to get more context.

test/e2e/authorization.go
linki: Or just one test with this?

```go
tc.data.users = []string{"system:serviceaccount:default:default", "system:serviceaccount:non-default:default"}
```

linki:
👍

test/e2e/authorization.go
demonCoder95: Added in this commit: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8536/commits/1b13defef78ff9c53dbdc6c43a9e14659f868991

Looks much better 🙂

test/e2e/authorization.go
demonCoder95: Done in this commit: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8536/commits/6f7f8c96c5ef8339c41832b8ffdc86a8a9b46b97

I found another write protection test that failed so I skipped that as well, and marked it with a comment in the code.

test/e2e/authorization.go
demonCoder95: Added in the commit: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8536/commits/6f7f8c96c5ef8339c41832b8ffdc86a8a9b46b97

test/e2e/authorization.go
demonCoder95: Ah yes. you're right. They're identical 😅 I was focussing too much on copying the tests 1 to 1. I merged them into one test case specifying both users.

Combined them in one test in: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8536/commits/4eba59b8cb63f1d56b7ef05fad86022bfe76ebae

test/e2e/authorization.go
demonCoder95: This was taken care of when these two test cases were combined in: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8536/commits/4eba59b8cb63f1d56b7ef05fad86022bfe76ebae

test/e2e/authorization.go
demonCoder95: I actually tried specifying `foo-configmap` but it's still undecided. Not sure why this is happening. Should I still add the name?

test/e2e/authorization.go
demonCoder95: > Actually RBAC also gives credentials provider read/write to secret. Maybe it makes sense to add a test for it as well?

Looking at[ the `ClusterRoleBinding`](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/roles/credentials-provider-rbac.yaml#L20-L27) attached to this user, it's allowed all operation on `Secrets` except `delete`. 

I created two test cases, one for deletion, another for "all other" operations: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8536/commits/39e6e18d57eb00c4da9ef0a7d4a1fa4c044c91ee

test/e2e/authorization.go
demonCoder95: Added it here: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8536/commits/04827d63f832859c141450ad58b07440dd42b616

test/e2e/authorization.go
demonCoder95: Added here: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8536/commits/1646c0671183e2aa08fea97d3fb23933872c86d5

PR #8533, opened 2024-11-21
[e2e] Test refactored RBAC on EKS

demonCoder95:
Building on the change in #8532 , we test the new RBAC test suite on the EKS.

We also drop permissions to read `secrets` from the `poweruser` `ClusterRole`. This is now handled by the `role-sync-controller` on a per-namespace basis.

This PR is only to test the change and shouldn't be merged.


cluster/manifests/roles/poweruser-role.yaml
```diff
@@ -58,19 +58,6 @@ rules:
...
-  - secrets
-  verbs:
-  - create
-  - delete
-  - deletecollection
-  - get
-  - list
-  - patch
-  - update
-  - watch
```
demonCoder95, line 73: Write permissions are already specified for `secrets` in the rule above, here they are redundant, so they're dropped alongside read permissions.

demonCoder95:
The change has now been merged to `dev` and this is covered in `dev-to-eks` PR #8535 . Closing this PR now.

PR #8536, opened 2024-11-22
[e2e] Improvements on the RBAC e2es

demonCoder95:
This addresses the review comments on the RBAC e2e PR #8532.

Also adds a missing test case for testing `poweruser` read access to `secrets` in non-system namespaces.

### DIFF
--- a/test/e2e/authorization.go
+++ b/test/e2e/authorization.go
@@ -207,10 +207,18 @@ var _ = g.Describe("Authorization [RBAC] [Zalando]", func() {
 				tc.data.resources = namespacedResources
 				tc.data.verbs = writeOperations
 			})
-			// These should be covered by the admission-controller tests.
-			// They're written here for completeness.
-			g.It("should deny write access in kube-system and visibility namespaces", func() {})
-			g.It("should allow write access in namespaces other than kube-system and visibility", func() {})
+			// These should be covered by the admission-controller tests. They will
+			// be skipped here. Later when we cover everything with RBAC, we can run them again.
+			g.It("should deny write access in kube-system and visibility namespaces", func() {
+				tc.data.namespaces = []string{"kube-system", "visibility"}
+				tc.run(context.TODO(), cs, false)
+				gomega.Expect(tc.output.passed).To(gomega.BeTrue(), tc.output.String())
+			})
+			g.It("should allow write access in namespaces other than kube-system and visibility", func() {
+				tc.data.namespaces = []string{"", "teapot"}
+				tc.run(context.TODO(), cs, true)
+				gomega.Expect(tc.output.passed).To(gomega.BeTrue(), tc.output.String())
+			})
 		})
 		g.When("the resource is a global resource", func() {
 			g.BeforeEach(func() {
@@ -279,6 +287,8 @@ var _ = g.Describe("Authorization [RBAC] [Zalando]", func() {
 			tc.run(context.TODO(), cs, true)
 			gomega.Expect(tc.output.passed).To(gomega.BeTrue(), tc.output.String())
 		})
+		// This should be covered by the admission-controller tests. It will
+		// be skipped here. Later when we cover everything with RBAC, we can run it again.
 		g.It("should deny deletion of kube-system or visibility namespaces", func() {
 			tc.data.resources = []string{"namespaces"}
 			tc.data.names = []string{"kube-system", "visibility"}
@@ -292,9 +302,13 @@ var _ = g.Describe("Authorization [RBAC] [Zalando]", func() {
 				tc.data.resources = namespacedResources
 				tc.data.verbs = writeOperations
 			})
-			// This should be covered by the admission-controller tests.
-			// It's written here for completeness.
-			g.It("should deny write access in kube-system namespace", func() {})
+			// This should be covered by the admission-controller tests. It will
+			// be skipped here. Later when we cover everything with RBAC, we can run it again.
+			g.It("should deny write access in kube-system namespace", func() {
+				tc.data.namespaces = []string{"kube-system"}
+				tc.run(context.TODO(), cs, false)
+				gomega.Expect(tc.output.passed).To(gomega.BeTrue(), tc.output.String())
+			})
 			g.It("should allow write access in namespaces other than kube-system", func() {
 				tc.data.namespaces = []string{"", "teapot"}
 				tc.run(context.TODO(), cs, true)
--- a/test/e2e/authorization.go
+++ b/test/e2e/authorization.go
@@ -503,6 +503,11 @@ var _ = g.Describe("Authorization [RBAC] [Zalando]", func() {
 					tc.run(context.TODO(), cs, true)
 					gomega.Expect(tc.output.passed).To(gomega.BeTrue(), tc.output.String())
 				})
+				g.It("should allow write access", func() {
+					tc.data.verbs = writeOperations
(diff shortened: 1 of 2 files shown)


demonCoder95:
👍

test/e2e/authorization.go
```diff
@@ -23,19 +23,53 @@ var (
...
+
+	// "secrets" are not included as they have their own set of test cases.
+	namespacedResources = []string{
+		"pods",
+		"apps/deployments",
+		"apps/statefulsets",
+		"apps/deployments/scale",
+		"apps/statefulsets/scale",
+		"services",
+		"persistentvolumes",
```
linki: This is a global resource IIRC

test/e2e/authorization.go
```diff
@@ -198,39 +215,33 @@ var _ = g.Describe("Authorization [RBAC] [Zalando]", func() {
...
+			})
+			// These should be covered by the admission-controller tests. They will
+			// be skipped here. Later when we cover everything with RBAC, we can run them again.
+			g.It("should deny write access in kube-system and visibility namespaces", func() {
+				tc.data.namespaces = []string{"kube-system", "visibility"}
+				tc.run(context.TODO(), cs, false)
+				gomega.Expect(tc.output.passed).To(gomega.BeTrue(), tc.output.String())
+			})
+			g.It("should allow write access in namespaces other than kube-system and visibility", func() {
+				tc.data.namespaces = []string{"", "teapot"}
```
linki: Given that the test is about namespaced resources, how does it play together with `""` (global namespace) here? Does the test do the right thing?

test/e2e/authorization.go
demonCoder95: Hmm, good point. I'll remove it from the list.

test/e2e/authorization.go
demonCoder95: Looking at the authv1 library definition, it says that `""` for a namespaced resources means "all namespaces". This now has me concerned. I think maybe it's best to replace `""` with `"default"`. What do you think?

Good catch!

demonCoder95: I fixed this `re/""/"default"/` in the next commit.

test/e2e/authorization.go
demonCoder95: I moved `persistentvolumes` to `globalresources` list in the next commit.

demonCoder95: The tests have started failing now. Do we actually not allow write access to `persistentvolumes` to collaborators?

mikkeloscar: We do allow write access here: https://github.com/zalando-incubator/kubernetes-on-aws/blob/0b3171b3f858587dc09eb37e33e451bcb69ef68a/cluster/manifests/roles/poweruser-role.yaml#L27-L45

mikkeloscar: The way the collaborator role is defined is that it's assumed you also have the PowerUser/engineer role for all the poweruser permissions it's not a superset of poweruser by itself.

We only give the collaborator role these specific permissions: https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/roles/collaborator-roles.yaml

demonCoder95: ah, that makes sense. I'll remove the `persistentvolumes` from the global resources then.

demonCoder95:
👍

demonCoder95:
Merging this will also fix the `dev-to-eks` e2e that currently fails.

test/e2e/run_e2e.sh
```diff
@@ -185,6 +185,7 @@ if [ "$e2e" = true ]; then
     ginkgo -procs=25 -flake-attempts=2 \
         -focus="(\[Conformance\]|\[StatefulSetBasic\]|\[Feature:StatefulSet\]\s\[Slow\].*mysql|\[Zalando\])" \
         -skip="(\[Serial\]|validates.that.there.is.no.conflict.between.pods.with.same.hostPort.but.different.hostIP.and.protocol|Should.create.gradual.traffic.routes)" \
+        -skip="should.deny.write.access.in.kube-system.and.visibility.namespaces|should.allow.write.access.in.namespaces.other.than.kube-system.and.visibility|should.deny.write.access.in.kube-system.namespace|should.deny.deletion.of.kube-system.or.visibility.namespaces" \
```
mikkeloscar: Does this work correctly for skipping? 🤔 I would think it not valid regex. Maybe the flag being defined twice doesn't work?

Why do we want to skip these tests?

demonCoder95: Yes, it works. I tested the regex by using it in a `-focus` argument like so:

``` bash
$ ginkgo --dry-run -focus "should.deny.write.access.in.kube-system.and.visibility.namespaces|should.allow.write.access.in.namespaces.other.than.kube-system.and.visibility|should.deny.write.access.in.kube-system.namespace|should.deny.deletion.of.kube-system.or.visibility.namespaces" -v

...

Ran 4 of 6479 Specs in 0.121 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 6475 Skipped
PASS

Ginkgo ran 1 suite in 18.862650333s
Test Suite Passed
```

The flag defined twice also works. We should have a total of `472` specs, but the pipeline ran here only executed `468`, so the skip for these 4 is working fine.

The reason we're skipping these is as Martin suggested earlier, we want to keep the implementation here, for when we move completely to RBAC, and just skip them for now so they don't fail. As they're being tested in admission-controller, not in SubjectAccessReview.

mikkeloscar: > for when we move completely to RBAC

But this seems unrealistic on the short term or even long term as we have discussed what can and can't be implemented  purely in RBAC.

demonCoder95: @linki what do you think then? 

We can go back to removing the implementation and removing the skip as well, how it currently is. Leave the empty spec there for completeness, so we can keep track of the tests that we need to add in admission-control.

mikkeloscar: For me it just seems strange to add tests that we skip right away and may never unskip in that form.

I understand the purpose of having them for completeness, but maybe there is a better way e.g. record them in an issue that can then be picked up at some point.

linki: I see. I agree it's weird to skip them like this. I thought about something like this:

```go
g.It("should deny deletion of kube-system or visibility namespaces", func() { g.Skip("handled by admission-controller") }
```

but we can also remove them completely. Then we should make sure to clearly define the boundaries between what is done by RBAC and what is done by admission-controller (like a high-level description so other people can understand, currently it's a bit confusing what needs to go where)

demonCoder95: I created an issue and recorded these tests there. I'll remove the tests from here completely, as agreed.

demonCoder95: > g.It("should deny deletion of kube-system or visibility namespaces", func() { g.Skip("handled by admission-controller") }

I like this suggestion, I think we can keep this to make it easier to read the RBAC tests and not feel like something is missing.

demonCoder95: I think it looks better now: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8536/commits/acd2bdf1e2c87310e34869a6bd0da7c20ffbdfe9

demonCoder95:
👍

mikkeloscar:
:+1:

PR #8537, opened 2024-11-24
[RBAC] drop secret read permission from poweruser ClusterRole

demonCoder95:
This drops the permissions to read `Secrets` from the RBAC `ClusterRole` of poweruser. The write permissions dropped are redundant here, as another rule above this one allows for them. We also extend the `collaborator` Role to include permissions to read Secrets in the `visibility` Namespace, as the e2e test otherwise fails for it.

This is done because:
1. In the current Authz setup which relies on a combination of RBAC and Authz-Webhook, even though we allow these permission in the RBAC, they are blocked by the webhook, so having them here is of no affect.
2. In the future Authz setup, which relies on a combination of per-namespace RBAC and admission-controller, these are still not needed. This is since the permission to read Secrets is provided through a per-namespace `RoleBinding` for different users; through the `role-sync-controller`. Therefore, we can simply drop these permission from this definition here.

This patch is also needed to fully rollout the Authz system we need for EKS.

### DIFF
--- a/cluster/manifests/roles/collaborator-roles.yaml
+++ b/cluster/manifests/roles/collaborator-roles.yaml
@@ -14,6 +14,14 @@ rules:
   - update
   - patch
   - delete
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
 ---
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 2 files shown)


demonCoder95:
This is not to be merged until the end of the CW deployment-policy.

demonCoder95:
Wrapped the RBAC changes behind the flag that enables the `role-sync-controller` so we can easily roll back.

demonCoder95:
We can merge this do `dev`, so change can also propagate to `eks`.

demonCoder95:
👍

mikkeloscar:
:+1:

PR #8539, opened 2024-11-26
[e2e] Skip authz test for access handled by admission-controller

demonCoder95:
Skipping another test that was previously overlooked and is also covered by admission-controller. It's for  the access for `PowerUser`, `Manual` and `Emergency` groups to delete the `kube-system` or `visibility` namespace.

I updated the internal issue that tracks these, and added this test that has to be implemented in e2e for admission-controller.


demonCoder95:
👍

mikkeloscar:
:+1:

PR #8541, opened 2024-11-27
Update custom metrics image to work with ipv6

mikkeloscar:
Updating the image used for autoscaling e2e to work in ipv6 clusters.

The previous image was a simple python app listening on `0.0.0.0` so it wasn't reachable via ipv6. The new image is an equivalent app in Go which listens on any interface and thereby work both on ipv4 and ipv6.

It's only relevant for EKS, but adding it to dev to keep e2e tests consistent.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #8542, opened 2024-11-27
Exclude eks-pod-identity-agent from admission-controller to avoid 🐔🥚

mikkeloscar:
Filter the `eks-pod-identity-agent` daemonset out of the admission-controller as the admission-controller depend on it to run.


linki:
👍

mikkeloscar:
:+1:

PR #8548, opened 2024-12-02
adding a special instance-types string 'not-specified'

myaser:
render karpenter node pools without  instance-types requirements if the strategy is `not-specified`

related: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/833

### DIFF
--- a/cluster/node-pools/worker-karpenter/provisioners.yaml
+++ b/cluster/node-pools/worker-karpenter/provisioners.yaml
@@ -124,7 +124,7 @@ spec:
 #  {{ end}}
 #{{ end}}
 
-#{{ if and (eq (len .NodePool.InstanceTypes) 1) (eq (index .NodePool.InstanceTypes 0) "default-for-karpenter") }}
+#{{ if (eq .NodePool.KarpenterInstanceTypeStrategy "default-for-karpenter" )  }}
         - key: "karpenter.k8s.aws/instance-family"
           operator: In
           values:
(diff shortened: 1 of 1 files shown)


demonCoder95:
I don't see we're using the strategy `not-specified` anywhere? We're using `custom` instead?

myaser:
> I don't see we're using the strategy `not-specified` anywhere? We're using `custom` instead?

since we have only 3 cases

line number 168
```
#{{ if not (eq .NodePool.KarpenterInstanceTypeStrategy "custom") }}
```
will apply the safety guards in case the strategy were `default-for-karpenter` or `not-specified`, we could make it more explicit but the if statement will be slightly more complex (or statement)

while number 127 `#{{ if not (eq .NodePool.KarpenterInstanceTypeStrategy "custom") }}` will apply the default instance types and line 158 `#{{ else if (eq .NodePool.KarpenterInstanceTypeStrategy "custom" ) }}` will apply the specified instances in the requirements.

this means, when the strategy is non-specified, it only applies the safety guards and leave the instance-type requirements blank

myaser:
👍🏻

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -165,7 +165,7 @@ spec:
 #{{ end }}
 
 # safety guards to prevent the use of unwanted instance types in case the user has not specified any specific instance types
-#{{ if or (eq .NodePool.KarpenterInstanceTypeStrategy "default-for-karpenter") (eq .NodePool.KarpenterInstanceTypeStrategy "not-specified") }}
+#{{ if not (eq .NodePool.KarpenterInstanceTypeStrategy "custom") }}
```
mikkeloscar: `not (eq)` can be simplified to `ne`

myaser:
👍🏻

mikkeloscar:
:+1:

PR #8549, opened 2024-12-02
Drop kube-node-ready

mikkeloscar:
Drop configuration related to kube-node-ready daemonset.

It was already disabled in all clusters via config-cluster-manager.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #8550, opened 2024-12-03
[EKS] Add an admitter that validates that unprivileged users cannot exec into postgres pods

linki:
This uses a new admitter and endpoint (`pod/exec`) that is designed to capture exec requests. Upon receiving such a request the admitter will reject the request if it's for a postgres (`application=spilo`) pod, otherwise it allows it.

This is configured to only run for non-privileged namespaces (userland) [1]. It's also _not_ run for admins as well as postgres-admins. This has the effect that admins and postgres-admins can exec into any userland pod. Normal users will trigger the admitter which then allows the request for any pod (in userland) that isn't a postgres pod, hence only admins and postgres-admins can exec into postgres pods in userland. Note, that postgres-admins _cannot_ exec into postgres pods in privileged namespaces. Also note, that anyone who _only_ has the postgres-admin role can exec into postgres pods as well as any other pod in a non-privileged names. In other words, the exec permission for postgres-admins is not limited to just postgres pods.

[1] exec requests in privileged namespaces are handled by the `namespaced-deny-admitter.teapot.zalan.do` admitter and only allows admins and internal components to exec.

### DIFF
--- a/cluster/manifests/02-admission-control/deployment.yaml
+++ b/cluster/manifests/02-admission-control/deployment.yaml
@@ -33,7 +33,7 @@ spec:
       priorityClassName: system-cluster-critical
       containers:
       - name: admission-controller
-        image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/admission-controller:master-226
+        image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/admission-controller:master-227
         lifecycle:
           preStop:
             exec:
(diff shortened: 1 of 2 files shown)


mikkeloscar:
Would make sense to have this change against `dev` (behind the feature flag) such that we can align EKS and non-eks at some point if they can share implementation.

The e2e changes for auth was already going in this direction. What @demonCoder95 was working on.

linki:
Obsolete. Merged via dev-to-eks: https://github.com/zalando-incubator/kubernetes-on-aws/commit/78bf84adcd2489bb672b0cbfe0d25cd59306b899

PR #8575, opened 2024-12-05
Add ingress test for opaAuthorizeRequest filter

wisinghe:
## Changes
* Skipper rbac configuration is ordered to be applied before deployment to make sure the service account is created before the deployment.
* teapot-e2e cluster configuration is changed to support opa
* opaAuthorizeRequest is initialized and verified within ingress.go
* `[Opa]` tag is added to the opaAuthorizeRequest related `It` case so that it will be executed in e2e test step


test/e2e/cluster_config.sh
```diff
@@ -47,6 +47,11 @@ clusters:
     karpenter_pools_enabled: "true"
     okta_auth_client_id: "kubernetes.cluster.teapot-e2e"
     teapot_admission_controller_validate_pod_images_soft_fail_namespaces: "^kube-system$"
+    skipper_open_policy_agent_enabled: "true"
```
mjungsbluth: We should discuss if we need a negative test, i.e. verify that clusters without OPA can still be deployed. We might otherwise not catch config errors where without this flag the yaml is invalid.

wisinghe: @mikkeloscar is this something that can be done here?

AFAIU within these e2e testing, it pre-configures one cluster and do validations on that. Changing the base cluster configurations during the test might not be possible. May be there is a different location where we can test different config flags.

mikkeloscar: We currently don't have the capability to test against different permutations of configuration.

test/e2e/ingress.go
```diff
@@ -315,7 +315,133 @@ var __ = describe("Ingress tests simple", func() {
 	})
 })
 
-var ___ = describe("Ingress tests paths", func() {
+var ___ = describe("Ingress tests for OPA filters", func() {
```
MustafaSaber: minor: move the test down instead of renaming all other testcases

delivery.yaml
```diff
@@ -318,6 +340,230 @@ pipeline:
               cpu: 500m
               memory: 1Gi
 
+- id: create-opa-enabled-cluster
```
MustafaSaber: TBH I don't like that we create another cluster in the pipeline; It will add another 30m to each PR, but that's for teapot to decide.

test/e2e/run_e2e.sh
```diff
@@ -45,8 +46,15 @@ CDP_TARGET_COMMIT_ID="${CDP_TARGET_COMMIT_ID:-"dev"}"
 CDP_HEAD_COMMIT_ID="${CDP_HEAD_COMMIT_ID:-"$(git describe --tags --always)"}"
 RESULT_BUCKET="${RESULT_BUCKET:-""}"
 
-export CLUSTER_ALIAS="${CLUSTER_ALIAS:-"e2e-${CDP_BUILD_VERSION}"}"
-export LOCAL_ID="${LOCAL_ID:-"e2e-${CDP_BUILD_VERSION}"}"
+if [ "$OPA_ENABLED" == "true" ]; then
```
MustafaSaber: this is a matter of taste, so you can ignore

instead of using `OPA_ENABLED`

add 2 environment variables one for FOCUS which defaults to the current value and another for cluster suffix, then you can 

```sh
FOCUS="${FOCUS:-(\[Conformance\]|\[StatefulSetBasic\]|\[Feature:StatefulSet\]\s\[Slow\].*mysql|\[Zalando\])}"
export CLUSTER_ALIAS="${CLUSTER_ALIAS:-"e2e-${CDP_BUILD_VERSION}${CLUSTER_SUFFIX}"}"
export LOCAL_ID="${LOCAL_ID:-"e2e-${CDP_BUILD_VERSION}${CLUSTER_SUFFIX}"}"
```

test/e2e/ingress.go
```diff
@@ -1027,3 +1027,129 @@ var _______ = describe("Ingress tests simple NLB", func() {
...
+
+		_, err = cs.NetworkingV1().Ingresses(ns).Get(context.TODO(), ing.Name, metav1.GetOptions{ResourceVersion: "0"})
+		framework.ExpectNoError(err)
+
+		// skipper http -> https redirect
+		By("Waiting for skipper route to default redirect from http to https, to see that our ingress-controller and skipper works")
+		err = waitForResponse(addr, "http", waitTime, isRedirect, true)
+		framework.ExpectNoError(err)
+
+		// ALB ready
```
MustafaSaber: I know in all (or most) other test cases it's ALB but let's just rename to LB, and we clean other tests later.

test/e2e/ingress.go
```diff
@@ -1027,3 +1027,129 @@ var _______ = describe("Ingress tests simple NLB", func() {
...
+	)
+
+	It("Should activate OPA filter without issue [Ingress] [Opa]", func() {
+		jig = ingress.NewIngressTestJig(f.ClientSet)
+		cs = f.ClientSet
+		serviceName := "skipper-ingress-test"
+		ns := f.Namespace.Name
+		hostName := fmt.Sprintf("%s-%d.%s", serviceName, time.Now().UTC().Unix(), E2EHostedZone())
+		labels := map[string]string{
+			"app": serviceName,
```
MustafaSaber: I remember when testing OPA that the application name should be equal policy name without `-policiy`, is this true?

Maybe it's better to name resources for this as `styra-smoketest` and use it elsewhere.

delivery.yaml
MustafaSaber: this PR time

```
Total steps runtime: 1 h, 34 m, 26 s
```

other PR times (went through 5 not all all of them)

```
Total steps runtime: 1 h, 8 m, 47 s
```

wisinghe: Hmm.. I understand that it would be a concern. Will wait for Teapot's feedback for that. However we do not hope to keep these steps forever. It's only until we are confident enough to enable this feature for all the clusters. 

Also the overall completion time for a build is not affected as these 3 steps run parallel to other create-cluster, e2e steps.

test/e2e/ingress.go
wisinghe: `the application name should be equal policy name without -policiy, is this true?` this is more like a convention. When writing policies for a certain application, we name the policy repo with the template `<application-name>-policies`

delivery.yaml
demonCoder95: > we do not hope to keep these steps forever. It's only until we are confident enough to enable this feature for all the clusters.

The e2e environment is the same for all the infrastructure, it's not dependent on channels and having something we're unsure about is fine to have as a work in progress. Additionally, we can also iterate over the e2e tests safely, so we can merge this version and add improvements later on. Just have to keep backwards compatibility in mind, so as long as the existing test suite is not "disturbed", it's fine to merge this, I'd say.

 So, I'd recommend having `SKIPPER_OPA_ENABLED: true` for the existing cluster and integrating the OPA test into the exiting tests. 

We can evaluate and see how to manage the longer build time. Two clusters add too much complexity to keep the setup sustainable.

delivery.yaml
```diff
@@ -80,6 +80,8 @@ pipeline:
             value: "teapot-kubernetes-e2e-results"
           - name: ETCD_ENDPOINTS
             value: "https://etcd-server.etcd.teapot-e2e.zalan.do:2479"
+          - name: SKIPPER_OPA_ENABLED
+            value: "false"
```
demonCoder95: Let's change this to `true` and drop all the new steps we added, as I explained below.

test/e2e/ingress.go
```diff
@@ -1027,3 +1027,129 @@ var _______ = describe("Ingress tests simple NLB", func() {
...
+			quit <- struct{}{}
+		}()
+		url := "https://" + hostName + "/"
+		req, err := http.NewRequest("GET", url, nil)
+		framework.ExpectNoError(err)
+		resp, err := rt.RoundTrip(req)
+		framework.ExpectNoError(err)
+		s, err := getBody(resp)
+		framework.ExpectNoError(err)
+		Expect(s).To(Equal(backendContent))
```
demonCoder95: This appears to be identical to the testcase above. It would be best to "combine" both test cases so that we only do the setup once and use it for both tests. 

This is, I believe, also adding a lot to the test running time because this process would take a few minutes to complete.

test/e2e/ingress.go
```diff
@@ -1027,3 +1027,129 @@ var _______ = describe("Ingress tests simple NLB", func() {
...
+			map[string]string{
+				"zalando.org/skipper-filter": fmt.Sprintf(`opaAuthorizeRequest("%s")`, opaPolicyName),
+			},
+			port,
+		)
+		ingressUpdate, err := cs.NetworkingV1().Ingresses(ingressCreate.ObjectMeta.Namespace).Update(context.TODO(), updatedIng, metav1.UpdateOptions{})
+		framework.ExpectNoError(err)
+
+		By(fmt.Sprintf("Waiting for ingress %s/%s we wait to get a 403 with opaAuthorizeRequest %s policy", ingressUpdate.Namespace, ingressUpdate.Name, opaPolicyName))
+		time.Sleep(20 * time.Second) // wait for routing change propagation
```
demonCoder95: This might turn out to make the test flakey, are we sure `20s` would be enough for the route propagation?

test/e2e/ingress.go
```diff
@@ -1027,3 +1027,129 @@ var _______ = describe("Ingress tests simple NLB", func() {
...
+		By(fmt.Sprintf("Waiting for ingress %s/%s we wait to get a 200 with opaAuthorizeRequest %s policy", ingressUpdate.Namespace, ingressUpdate.Name, opaPolicyName))
+		req.Header.Set("Authorization", "Basic valid_token") //Authorized request
+		resp, err = getAndWaitResponse(rt, req, 10*time.Second, http.StatusOK)
+		framework.ExpectNoError(err)
+		Expect(resp.StatusCode).To(Equal(http.StatusOK))
+		s, err = getBody(resp)
+		framework.ExpectNoError(err)
+		Expect(s).To(Equal(backendContent))
+	})
+})
```
demonCoder95: Let's split these into two different test cases with `It` closure. Makes the test cases more readable.

test/e2e/run_e2e.sh
```diff
@@ -44,9 +44,11 @@ CDP_TARGET_REPOSITORY="${CDP_TARGET_REPOSITORY:-"github.com/zalando-incubator/ku
 CDP_TARGET_COMMIT_ID="${CDP_TARGET_COMMIT_ID:-"dev"}"
 CDP_HEAD_COMMIT_ID="${CDP_HEAD_COMMIT_ID:-"$(git describe --tags --always)"}"
 RESULT_BUCKET="${RESULT_BUCKET:-""}"
+CLUSTER_SUFFIX="${CLUSTER_SUFFIX:-""}"
```
demonCoder95: Why are we adding this `CLUSTER_SUFFIX`? If it's the part of the multi-cluster setup suggested, I'd suggest we drop it.

delivery.yaml
demonCoder95: I already found one potential step that will definitely make a big impact on the running time of the OPA tests, please see the comment below.

test/e2e/ingress.go
demonCoder95: You can check https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/test/e2e/authorization.go for inspiration on how to design the test suite and extract common setup.

test/e2e/ingress.go
wisinghe: From random checks, OPA starts with bundles within 15 seconds. So far this did not cause a problem in the pipeline as well.

test/e2e/run_e2e.sh
wisinghe: `CLUSTER_SUFFIX` was used to make the LOCAL_ID and ALIAS unique for the 2nd cluster. Will cleanup

test/e2e/ingress.go
wisinghe: The test above `It("Should create simple NLB ingress [Ingress]"`, (or any other Ingress related test for that matter) are not executed in the pipeline e2e testing AFAIU. I came to this conclusion looking at the FOCUS here https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/test/e2e/run_e2e.sh#L186
Also the junit report confirms this. If that is the case, it does not add time to the pipeline time anyway.
```
<testcase name="[It] [zalando] Ingress tests simple NLB Should create simple NLB ingress [Ingress]" classname="Kubernetes e2e suite" status="skipped" time="0">
              <skipped message="skipped"></skipped>
```

We would actually like to have the OPA related tests run during the e2e step in the pipeline. So I feel like it's good to keep it separated from others.

delivery.yaml
mikkeloscar: It's not so trivial to double the amount of clusters we create in e2e. We have limited resources in the account so if each PR is creating two clusters we're likely to hit those limits with the current setup which will just render the e2e setup not working.

For this reason I would be reluctant to add two separate clusters for e2e. I understand the motivation of testing skipper in the different conditions, but our e2e setup is simply not able to handle multiple permutations of configurations as it is today, so it's a trade-off we have to make or we have to invest more in the e2e setup up front.

test/e2e/ingress.go
wisinghe: Actually when doing this, it might need refactoring the complete file. I will skip it as I would prefer not to change the complete suite for the ingress tests

MustafaSaber:
👍

MustafaSaber:
👍

MustafaSaber:
👍

test/e2e/ingress.go
```diff
@@ -1027,3 +1027,129 @@ var _______ = describe("Ingress tests simple NLB", func() {
...
+	f.NamespacePodSecurityEnforceLevel = admissionapi.LevelBaseline
+	var (
+		cs  kubernetes.Interface
+		jig *ingress.TestJig
+	)
+
+	It("Should activate OPA filter without issue [Ingress] [Zalando]", func() {
+		jig = ingress.NewIngressTestJig(f.ClientSet)
+		cs = f.ClientSet
+		serviceName := "styra-smoketest"
```
torwunder: opa-filter-smoketest? styra is not really involved

wisinghe: It's actually a registered application https://github.bus.zalan.do/corporate-iam/styra-smoketest-policies/blob/main/authorization.yaml

It was initially created for Styra smoke tests and now used for this one as well. I would prefer not to register another application for this one.

test/e2e/ingress.go
```diff
@@ -1027,3 +1027,129 @@ var _______ = describe("Ingress tests simple NLB", func() {
...
+		allow if {
+		  input.attributes.request.http.method == "GET"
+		  auth_header_val := input.attributes.request.http.headers.authorization
+		  startswith(auth_header_val, "Basic ")
+		  token := substring(auth_header_val, count("Basic "), -1)
+		  token == "valid_token"
+		}
+		*/
+		path := "/"
+		opaPolicyName := "styra-smoketest"
```
torwunder: opa-filter-smoketest?

wisinghe: will rename to `aaas-infrastructure-test` so that we can use one common application to test different parts of the solution.

test/e2e/ingress.go
```diff
@@ -1027,3 +1027,129 @@ var _______ = describe("Ingress tests simple NLB", func() {
...
+			serviceName,
+			path,
+			netv1.PathTypeImplementationSpecific,
+			ingressCreate.ObjectMeta.Labels,
+			map[string]string{
+				"zalando.org/skipper-filter": fmt.Sprintf(`opaAuthorizeRequest("%s")`, opaPolicyName),
+			},
+			port,
+		)
+		ingressUpdate, err := cs.NetworkingV1().Ingresses(ingressCreate.ObjectMeta.Namespace).Update(context.TODO(), updatedIng, metav1.UpdateOptions{})
```
mikkeloscar: It seems like for the test you only need an ingress. So you could drop the whole setup of deployment and service which is copied from the other tests. What @demonCoder95 was also commenting on a bit above.

wisinghe: Updated the tests to only create ingress. Routes are configured to shunt after the policy decision is taken.

test/e2e/aaas.go
```diff
@@ -0,0 +1,158 @@
...
+	framework.ExpectNoError(err)
+
+	// Ensure ingress exists
+	_, err = cs.NetworkingV1().Ingresses(ns).Get(context.TODO(), ing.Name, metav1.GetOptions{ResourceVersion: "0"})
+	framework.ExpectNoError(err)
+
+	// Verify skipper routing
+	By("Waiting for skipper route to default redirect from http to https")
+	err = waitForResponse(addr, "http", waitTime, isRedirect, true)
+	framework.ExpectNoError(err)
```
mikkeloscar: This is used for testing http to https redirect. We already do this in other tests, so maybe this can simply be dropped here.

test/e2e/aaas.go
```diff
@@ -0,0 +1,158 @@
...
+
+	By("Waiting for ALB to create endpoint and skipper route")
```
mikkeloscar: This is not ALB. Maybe change to more generic message like: `Waiting for LB and skipper routes to be ready`

mikkeloscar:
:+1:

test/e2e/aaas.go
```diff
@@ -0,0 +1,153 @@
...
+		netv1.PathTypeImplementationSpecific,
+		ingressCreate.ObjectMeta.Labels,
+		map[string]string{
+			"zalando.org/skipper-routes": ingressRoute,
+		},
+		port,
+	)
+	ingressUpdate, err := cs.NetworkingV1().Ingresses(ingressCreate.ObjectMeta.Namespace).Update(context.TODO(), updatedIng, metav1.UpdateOptions{})
+	framework.ExpectNoError(err)
+	time.Sleep(2 * time.Minute) // wait for routing change propagation
```
MustafaSaber, line 150: Let's see if this will flake in the next pipelines.

MustafaSaber:
👍

PR #8577, opened 2024-12-05
Exclude kubenurse from being considered for a node to get ready

mikkeloscar:
Don't treat `kubenurse` as a critical daemonset that must be ready before a node is ready. This way if kubenurse it failing for whatever reason, nodes will still be ready for user workloads.

depends on #8576


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #8571, opened 2024-12-04
update role-sync-controller

demonCoder95:
This update includes the refactoring in the controller to not make any bindings for the `poweruser` Role, instead use per-namespace `secrets-reader` Role and RoleBinding.

PR #8588, opened 2024-12-06
Enable previously disabled e2e tests for authz

mikkeloscar:
Enable previously disabled authz related e2e tests now that it should be working via #8571

### DIFF
--- a/test/e2e/run_e2e.sh
+++ b/test/e2e/run_e2e.sh
@@ -204,6 +204,7 @@ if [ "$e2e" = true ]; then
         SKIPPED_TESTS+=(
             "Mirror pods should be created for the main Kubernetes components \[Zalando\]"
             "Should audit API calls to create, update, patch, delete pods. \[Audit\] \[Zalando\]"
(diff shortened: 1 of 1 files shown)


demonCoder95:
For EKS, these need to remain skipped, since the tests have been rewritten to use client-go, specifically for an EKS environment.

For the `dev` branch, we can unskip them and the pipeline should succeed, as both new and old tests should be successful for non-EKS setup. 

However, we can't then merge this change in EKS for the reasons mentioned above, the pipeline would fail, as the existing RBAC test suite is not compatible with the EKS apiserver.

demonCoder95:
I notice the secrets reading RBAC from the new suite is also failing still. I've identified the issue with the RBAC permissions of the role-sync-controller. I'll use this PR to test the change.

demonCoder95:
Cherry-picked the fix in #8608 to test the fix.

demonCoder95:
This should now be fixed. 

I updated the comment on the TODO because we cannot unskip those tests as they break fundamentally. Once we finish rewriting the RBAC tests and remove the existing tests from the setup, we can remove this line.

demonCoder95:
👍

demonCoder95:
👍

mikkeloscar:
:+1:

PR #8639, opened 2024-12-11
Update to Kubernetes v1.31.4

mikkeloscar:
This updates the AMI to Kubernetes v1.31.4: https://github.com/kubernetes/kubernetes/releases/tag/v1.31.4

Additionally it makes the following changes:

* Switch to upstream version of kube-controller-manager. This is possible because the fix for https://github.com/kubernetes/kubernetes/issues/127792 is part of Kubernetes v1.31.4
* Add the CNI config to the standard path `/etc/cni/net.d/` this is following a change in the AMI to expect the config at that path. This is done to be compatible with AWS CNI plugin which is expecting the standard path. We need later a follow up to remove the legacy path from the userdata. Added a TODO comment to cover this.

### DIFF
--- a/cluster/node-pools/master-default/userdata.yaml
+++ b/cluster/node-pools/master-default/userdata.yaml
@@ -928,8 +928,26 @@ write_files:
       {{- end}}
 {{- end}}
 
+  # TODO: Remove this once all nodes are running an AMI compatible with /etc/cni/net.d/10-flannel.conflist
   - owner: root:root
     path: /etc/kubernetes/cni/net.d/10-flannel.conflist
+    content: |
+      {
+        "name": "podnet",
+        "cniVersion": "0.3.1",
+        "plugins": [
+          {
+            "type": "flannel",
+            "delegate": {
+              "isDefaultGateway": true,
+              "hairpinMode": true
+            }
+          }
+        ]
+      }
+
+  - owner: root:root
+    path: /etc/cni/net.d/10-flannel.conflist
     content: |
       {
         "name": "podnet",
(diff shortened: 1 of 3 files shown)


cluster/node-pools/master-default/userdata.yaml
```diff
@@ -928,6 +928,7 @@ write_files:
       {{- end}}
 {{- end}}
 
+  # TODO: Remove this once all nodes are running an AMI compatible with /etc/cni/net.d/10-flannel.conflist
```
demonCoder95, line 931: Do we cover this in some internal issue? Would be better recorded there. I'm guessing we have an issue for the VPC CNI work?

mikkeloscar: I created an internal issue: 3756

demonCoder95: Thanks 🙏

demonCoder95:
I think the image wasn't promoted yet to the registry when the pipeline ran, so it failed

```
time="2024-12-11T14:13:22Z" level=fatal msg="Fail to provision: unable to read configuration defaults: template: main/cluster/config-defaults.yaml:743:38: executing \"main/cluster/config-defaults.yaml\" at <amiID \"zalando-ubuntu-jammy-22.04-kubernetes-production-v1.31.4-amd64-master-359\" \"861068367966\">: error calling amiID: no image found with name: zalando-ubuntu-jammy-22.04-kubernetes-production-v1.31.4-amd64-master-359 and owner: 861068367966"
```

The name does appear correct comparing to the AMI pipeline, so I just retriggered.

demonCoder95:
Seems like that was it, e2e passes now.

demonCoder95:
👍

linki:
👍

PR #8651, opened 2024-12-12
skipper: Fix formatting of flags

mikkeloscar:
The formatting of the skipper flags causes an issue in EKS clusters. Simplify by using `'` around the whole flag value and only use `"` for the eskip notation.

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -182,7 +182,7 @@ spec:
           - "-wait-for-healthcheck-interval={{ .Cluster.ConfigItems.skipper_wait_for_healthcheck_interval }}"
 {{ if ne .Cluster.ConfigItems.skipper_routesrv_enabled "exec" }}
           - "-kubernetes-east-west-range-domains=ingress.cluster.local"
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
:+1:

cluster/manifests/skipper/deployment.yaml
```diff
@@ -316,7 +316,7 @@ spec:
 {{ end }}
 {{ if or (eq .Cluster.ConfigItems.nlb_switch "pre") (eq .Cluster.ConfigItems.nlb_switch "exec") }}
           - "-forwarded-headers=X-Forwarded-For,X-Forwarded-Proto=https,X-Forwarded-Port=443"
-          - "-forwarded-headers-exclude-cidrs={{ .cluster_internal_cidrs | join "," }}"
+          - '-forwarded-headers-exclude-cidrs={{ .cluster_internal_cidrs | join "," }}'
```
AlexanderYastrebov, line 319: This change is not necessary as CIDRs do not contain any quotes

mikkeloscar: Do you prefer not to have it?

AlexanderYastrebov: There is no difference at the end so we can keep it.

PR #8659, opened 2024-12-12
Add test suite that covers auth-related admission checks

linki:
Adds test cases to cover auth-related admission-controller checks to circumvent the limitations of RBAC (e.g. deny functionality).

It sets up EKS Access Entries and subsequently assumes them to access the Kubernetes API with a particular `Group`, e.g. `zalando:administrator` for privileged and `zalando:readonly` for unprivileged access. The access entries are setup with full RBAC permissions (`AmazonEKSClusterAdminPolicy`) to simulate the cases where requests need to be caught by admission-controller instead of RBAC (see below).

The most common use case are ClusterRoles that should give permission in _all_ namespaces except `kube-system` (and other system namespaces). We currently implement these things by using ClusterRoles giving access in every namespace. We then selectively reject certain privileged requests via `admission-controller` after the request passed the RBAC check.

Another case is handling requests that require more context. Execing into Postgres pods should only be allowed to Postgres Admins and Administrators but not for normal users. To implement that, all mentioned user groups have RBAC permissions that allow execing into _all Pods_ in all user namespaces. However, once the request reaches the `admission-controller` (and was allowed by RBAC), it checks whether the Pod is a Postgres pod (identified by a label) and rejects the request if the requestor isn't a Postgres admin. This cannot be expressed in pure RBAC.

The tests for the above behaviour are in this suite.

Tasks:
- [x] protect all resources in the system namespaces `kube-system`, `visibility` and `kubenurse`
- [x] protect global resources that have the label `admission.zalando.org/infrastructure-component: true`
- [x] protect exec into postgres/spilo pods by users other than admins or postgres-admins


cluster/manifests/02-admission-control/teapot.yaml
```diff
@@ -519,8 +519,6 @@ webhooks:
     matchConditions:
       - name: 'exclude-privileged-groups'
         expression: 'request.userInfo.groups.all(g, !(g in ["okta:common/administrator", "zalando:administrator"]))'
-      - name: 'exclude-postgres-admins'
-        expression: 'request.userInfo.groups.all(g, !(g in ["okta:common/postgres-admin"]))'
```
linki, line 523: This is now handled in-code in admission-controller version `master-229`: https://github.bus.zalan.do/teapot/admission-controller/pull/266

The main reason was the additional allow-rule for `postgres-operator` which just made the rules more complicated. The exclusion rules for `okta:common/administrator`, `system:masters` etc. and are mainly used to resolve chicken-and-egg problems. We don't have those for postgres exec handling and therefore we can handle them inside admission-controller's code.

linki:
👍

zaklawrencea:
:+1:

PR #8679, opened 2024-12-19
feat: add wiz manifests

vinaythupili:
This PR will add the manifests for WIZ Connector and WIZ Sensor components for active monitoring of Kubernetes clusters.
The following Resources will be added as part of the PR .
**WIZ Connector**
- Deployment (WIZ broker)
- Jobs  (WIZ connector)
- Service Accounts 
- Role 
- Role binding
- Cluster Role Binding
- Secrets 

**WIZ Sensor** 
- Daemon set (Sensor)
- Secret
- Service Accounts
- Cluster Role 
- Cluster Role Binding


cluster/manifests/wiz/sensor/clusterrole.yaml
```diff
@@ -14,14 +14,15 @@ metadata:
     app.kubernetes.io/managed-by: Helm
 rules:
   - apiGroups: [""]
-    resources: ["pods"]
+    resources: ["pods", "namespaces", "nodes", "replicationcontrollers", "serviceaccounts",]
```
zaklawrencea:
```diff
- resources: ["pods", "namespaces", "nodes", "replicationcontrollers", "serviceaccounts",]
+ resources: ["pods", "namespaces", "nodes", "replicationcontrollers", "serviceaccounts"]
```

cluster/manifests/wiz/sensor/clusterrole.yaml
```diff
@@ -14,14 +14,15 @@ metadata:
...
     verbs: ["get", "list", "watch"]
 
-  - apiGroups: ["", "apps", "batch"]
-    resources: [
-      "namespaces", "nodes", "daemonsets", "replicasets", "deployments",
-      "jobs", "cronjobs", "statefulsets", "replicationcontrollers", "serviceaccounts",
-    ]
+  - apiGroups: ["apps"]
+    resources: ["daemonsets", "replicasets", "deployments", "statefulsets",]
```
zaklawrencea:
```diff
- resources: ["daemonsets", "replicasets", "deployments", "statefulsets",]
+ resources: ["daemonsets", "replicasets", "deployments", "statefulsets"]
```

cluster/manifests/wiz/sensor/clusterrole.yaml
```diff
@@ -14,14 +14,15 @@ metadata:
...
+    verbs: ["get", "list", "watch"]
+
+  - apiGroups: ["batch"]
+    resources: ["cronjobs",]
```
zaklawrencea:
```diff
- resources: ["cronjobs",]
+ resources: ["cronjobs"]
```

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,220 @@
...
+          mountPath: /wiz-sensor-store/
+        - name: tmp-store
+          mountPath: /tmp/
+        - name: api-client-secret
+          mountPath: /api-client/
+        - name: api-endpoint-name-secret
+          mountPath: /api-endpoint-name/
+        resources:
+          limits:
+            cpu: 300m
```
zaklawrencea: Lets set this to be the same as the `requests` to be consistent, even though the cpu limits are disabled.

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,220 @@
...
+            memory: 300Mi
```
zaklawrencea: Memory limits should equal requests.

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,220 @@
...
+          mountPath: /wiz-host-cache/
+        - name: sensor-store
+          mountPath: /wiz-sensor-store/
+        - name: tmp-store
+          mountPath: /tmp/
+        - name: api-client-secret
+          mountPath: /api-client/
+        - name: api-endpoint-name-secret
+          mountPath: /api-endpoint-name/
+        resources:
```
zaklawrencea: Perhaps we should add config items to control the cpu/mem resources, this would also make it easier to adjust in the future per cluster or as an emergency response.

See an example here:
https://github.com/zalando-incubator/kubernetes-on-aws/blob/3d593d0065e198c05ee5c26ac720e644bd52e853/cluster/manifests/audittrail-adapter/daemonset.yaml#L71-L77

vinaythupili:
👍

cluster/manifests/wiz/sensor/dameonset.yaml
nctram: +1 to this suggestion. We have seen time and again the usefulness of setting values as config items.

vinaythupili: as we are updating the resources at globally here https://github.bus.zalan.do/teapot/cluster-config-manager/blob/cbdb347320c35c0b6e6860479f243a4c36aba064/clusters/secops-test.yaml#L5 , we can keep the resources constant across all clusters

cluster/manifests/wiz/connector/clusterrole.yaml
```diff
@@ -0,0 +1,39 @@
...
+    app.kubernetes.io/managed-by: Helm
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name:  readonly # readonly role created by default in out kubernetes environment
+subjects:
+- kind: ServiceAccount
+  name: wiz-cluster-reader
+  namespace: "wiz"
+{{end}}}
```
zaklawrencea:
```suggestion
{{end}}
```

This additional curly bracket is present in most of the files.

cluster/manifests/wiz/sensor/dameonset.yaml
zaklawrencea: @vinaythupili the reserved value is to indicate how much space is free for user pods on a node and to protect us from disrupting their workloads, however it doesn't control the actual resources used by a specific Daemonset.

linki:
We should also populate the [deletions.yaml](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/deletions.yaml). Otherwise, the ConfigItem cannot be used to turn it off again.

cluster/manifests/wiz/connector/deployment.yaml
```diff
@@ -0,0 +1,79 @@
...
+kind: Deployment
+metadata:
+  name: wiz-connector-agent
+  namespace: "wiz"
+  labels:
+    helm.sh/chart: wiz-broker-2.1.0
+    app.kubernetes.io/name: wiz-broker
+    app.kubernetes.io/instance: wiz-connector
+    app.kubernetes.io/version: "2.5"
+    app.kubernetes.io/managed-by: Helm
```
linki: In our environment we need an `application` (and potentially `component`) label.

cluster/manifests/wiz/connector/deployment.yaml
```diff
@@ -0,0 +1,79 @@
...
+            items:
+              - key: connectorData
+                path: data
+      containers:
+        - name: wiz-broker
+          securityContext:
+            allowPrivilegeEscalation: false
+            runAsNonRoot: true
+            runAsUser: 1000
+          image: "wiziopublic.azurecr.io/wiz-app/wiz-broker:2.5"
```
linki: In our environment all images need to come from `container-registry.zalando.net`.

cluster/manifests/wiz/connector/deployment.yaml
```diff
@@ -0,0 +1,79 @@
...
+              secretKeyRef:
+                name: wiz-api-token
+                key: clientToken
+          - name: TARGET_IP
+            value: kubernetes.default.svc.cluster.local
+          - name: TARGET_PORT
+            value: "443"
+          resources:
+            null
+{{end}}}
```
linki:
```suggestion
{{end}}
```

cluster/manifests/wiz/connector/job.yaml
```diff
@@ -0,0 +1,167 @@
...
+    app.kubernetes.io/name: wiz-kubernetes-connector
+    app.kubernetes.io/instance: wiz-connector
+    app.kubernetes.io/version: "2.5"
+    app.kubernetes.io/managed-by: Helm
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded 
+    rollme.wizApiTokenHash: ce8124bc1b0fbc0cb5cd47338ca0c7d5f5446d79936e443a201d96b192a7bd65
+    rollme.proxyHash: 9aa53d69075371b3fa23ebeea2fd2416ea81fb533499d071ca2d576f17c7c886
+    rollme.brokerHash: 115ba85431eeaf8db3ff2173aee02d16e67df1555d5e1ef74cfa7ac0d812cab2
```
linki: These values look like they are computed based on the configuration value in order to trigger an update when they change. Having these values static here doesn't provide any value.

We do have [similar functionality](https://github.com/zalando-incubator/kubernetes-on-aws/blob/3d593d0065e198c05ee5c26ac720e644bd52e853/cluster/manifests/deployment-service/status-service-deployment.yaml#L28) which can be used if it's really needed.

cluster/manifests/wiz/connector/job.yaml
```diff
@@ -0,0 +1,167 @@
...
+          imagePullPolicy: Always
+          command:
+            - "wiz-broker"
+          args:
+            
+            - create-kubernetes-connector
+            - --api-server-endpoint
+            - "https://kubernetes.default.svc.cluster.local"
+            - --secrets-namespace
+            - "default"
```
linki: Should this be `wiz`?

cluster/manifests/wiz/connector/secrets.yaml
```diff
@@ -0,0 +1,58 @@
...
+    helm.sh/chart: wiz-sensor-1.0.4760
+    image/tag: v1
+    app.kubernetes.io/name: wiz-sensor
+    app.kubernetes.io/instance: wiz-sensor
+    app.kubernetes.io/version: "1.0.4708"
+    app.kubernetes.io/managed-by: Helm
+type: Opaque
+stringData:
+  clientId: "{{ .Cluster.ConfigItems.wiz_api_client_id | base64 }}"
+  clientToken: "{{ .Cluster.ConfigItems.wiz_api_client_token | base64 }}"
```
linki: When using `stringData`, then `base64` encoding might not be needed, just fyi if you see issues. https://kubernetes.io/docs/concepts/configuration/secret/#restriction-names-data

cluster/manifests/wiz/connector/job.yaml
zaklawrencea: Nice spot, updated!

cluster/manifests/wiz/connector/secrets.yaml
zaklawrencea: @vinaythupili now that I think about it, the values are already base64 encoded before the encryption is done. I think we meant to do `base64Decode` here originally, but since the secret can be base64 anyway, we probably don't need this encoding and can just use the `data` field instead of `stringData`.

cluster/manifests/wiz/sensor/dameonset.yaml
vinaythupili: updated

cluster/manifests/wiz/connector/deployment.yaml
vinaythupili: updated

vinaythupili:
> We should also populate the [deletions.yaml](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/deletions.yaml). Otherwise, the ConfigItem cannot be used to turn it off again.

updated 👍

cluster/manifests/wiz/connector/secrets.yaml
vinaythupili: updated to data 👍

cluster/config-defaults.yaml
```diff
@@ -1194,3 +1194,8 @@ role_sync_controller_enabled: "true"
 {{ else }}
 role_sync_controller_enabled: "false"
 {{ end }}
+
+#Wiz Configs
+wiz_enable_runtime_monitoring_daemonset: "false"
+wiz_adapter_cpu: "300m "
+wiz_adapter_memory: "300Mi""
```
katyanna:
```suggestion
wiz_adapter_memory: "300Mi"
```

cluster/manifests/deletions.yaml
```diff
@@ -339,3 +339,59 @@ post_apply:
 - name: kube-janitor
   kind: ClusterRoleBinding
 {{- end }}
+{{- if eq .Cluster.ConfigItems.wiz_enable_runtime_monitoring_daemonset "true"}
```
katyanna:
```suggestion
{{- if eq .Cluster.ConfigItems.wiz_enable_runtime_monitoring_daemonset "true"}}
```

cluster/manifests/deletions.yaml
```diff
@@ -339,3 +339,59 @@ post_apply:
 - name: kube-janitor
   kind: ClusterRoleBinding
 {{- end }}
+{{- if eq .Cluster.ConfigItems.wiz_enable_runtime_monitoring_daemonset "true"}}
```
zaklawrencea:
```suggestion
{{- if ne .Cluster.ConfigItems.wiz_enable_runtime_monitoring_daemonset "true" }}
```

cluster/manifests/deletions.yaml
```diff
@@ -339,3 +339,59 @@ post_apply:
...
+- name: wiz-kubernetes-connector-delete-connector
+  kind : Job
+  namespace: wiz
+- name: wiz-connector-agent
+  kind : Deployment
+  namespace: wiz
+- name: wiz-cluster-reader
+  kind : ClusterRoleBinding
+  namespace: wiz
+{{- end }}
```
demonCoder95: nit: Missing empty line at the end 😅

cluster/manifests/wiz/connector/clusterrole.yaml
```diff
@@ -0,0 +1,39 @@
...
+#   labels:
+#     helm.sh/chart: wiz-kubernetes-connector-3.1.1
+#     app.kubernetes.io/name: wiz-kubernetes-connector
+#     app.kubernetes.io/instance: wiz-connector
+#     app.kubernetes.io/version: "2.5"
+#     app.kubernetes.io/managed-by: Helm
+# rules:
+#   - apiGroups: ["*"]
+#     resources: ["*"]
+#     verbs: ["get", "list", "watch"]
```
demonCoder95: Do we want to keep this manifest here? I think the comment is sufficient to communicate that we don't need a global reader role 🤔

cluster/manifests/wiz/connector/clusterrole.yaml
```diff
@@ -0,0 +1,39 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: wiz-cluster-reader
+  labels:
+    helm.sh/chart: wiz-kubernetes-connector-3.1.1
+    app.kubernetes.io/name: wiz-kubernetes-connector
+    app.kubernetes.io/instance: wiz-connector
+    app.kubernetes.io/version: "2.5"
+    app.kubernetes.io/managed-by: Helm
```
demonCoder95: Are these labels important for the Wiz deployment? I think we might need to remove the ones related to `helm`, also we need to provide the Zalando environment labels like `application: foo` and `component: bar`.

demonCoder95: I see a similar concern was raised by Martin in another manifest, so let's fix the labels for all manifests 🙂 
1. Drop the helm related labels.
2. Add `application` and `component` labels.

cluster/manifests/wiz/connector/deployment.yaml
demonCoder95: You probably need to create an internal repository that "republishes" this public image to our internal ECR registry and then use the link as Martin suggests here. I communicated in our internal chat thread.

cluster/manifests/wiz/connector/deployment.yaml
```diff
@@ -0,0 +1,81 @@
...
+              - key: connectorData
+                path: data
+      containers:
+        - name: wiz-broker
+          securityContext:
+            allowPrivilegeEscalation: false
+            runAsNonRoot: true
+            runAsUser: 1000
+          image: "wiziopublic.azurecr.io/wiz-app/wiz-broker:2.5"
+          imagePullPolicy: Always
```
demonCoder95: `imagePullPolicy` should be `IfNotPresent`. 

Images pulled are always cached by the runtime and reusing the cached image also results in faster pod startup times. This is only set to `Always` in cases where you expect frequent image updates, which is not the case for our infrastructure.

cluster/manifests/wiz/connector/deployment.yaml
```diff
@@ -0,0 +1,81 @@
...
+            mountPath: /etc/connectorData
+            readOnly: true
+          args: [
+            /etc/connectorData/data
+          ]
+          env:
+          - name: LOG_LEVEL
+            value: info
+          - name: WIZ_ENV
+            value:
```
demonCoder95: Do we want to create an empty environment variable?

cluster/manifests/wiz/connector/deployment.yaml
```diff
@@ -0,0 +1,81 @@
...
+            runAsUser: 1000
+          image: "wiziopublic.azurecr.io/wiz-app/wiz-broker:2.5"
+          imagePullPolicy: Always
+          volumeMounts:
+          - name: connector-data
+            mountPath: /etc/connectorData
+            readOnly: true
+          args: [
+            /etc/connectorData/data
+          ]
```
demonCoder95: It's a lot cleaner to specify a list with `-` instead of square brackets like this, like

```yaml
args:
  - /etc/connectorData/data
```

We use the same format in all of the repository, e-g see: https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/role-sync-controller/cronjob.yaml#L31

cluster/manifests/wiz/connector/deployment.yaml
```diff
@@ -0,0 +1,81 @@
...
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: wiz-broker
+      app.kubernetes.io/instance: wiz-connector
+  template:
+    metadata:
+      annotations:
+        rollme: "Cd4Gg"
```
demonCoder95: What's the purpose of this annotation? Is it to control rolling updates? We already have the CLM perform updates upon manifest changes made in the repository.

cluster/manifests/wiz/connector/job.yaml
demonCoder95: I commented the same on the `rollme` annotation on the connector deployment manifest above. It appears this is used to keep track of changes in the deployment manifests, not in the configuration. So, we probably don't need this whole setup anyway. Maybe it's a helm thing? CLM does this job for us.

cluster/manifests/wiz/connector/job.yaml
```diff
@@ -0,0 +1,171 @@
...
+    rollme.proxyHash: 9aa53d69075371b3fa23ebeea2fd2416ea81fb533499d071ca2d576f17c7c886
+    rollme.brokerHash: 115ba85431eeaf8db3ff2173aee02d16e67df1555d5e1ef74cfa7ac0d812cab2   
+
+spec:
+  ttlSecondsAfterFinished: 60
+  manualSelector: true
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: wiz-kubernetes-connector
+      app.kubernetes.io/instance: wiz-connector
```
demonCoder95: I think we probably need to change these labels to `application: wiz` and `component: connector` everywhere. This is how we manage applications and components in our infrastructure.

cluster/manifests/wiz/connector/job.yaml
```diff
@@ -0,0 +1,171 @@
...
+        runAsNonRoot: true
+        runAsUser: 1000
+      containers:
+        - name: wiz-connector-creator
+          securityContext:
+            allowPrivilegeEscalation: false
+            runAsNonRoot: true
+            runAsUser: 1000
+          image: "wiziopublic.azurecr.io/wiz-app/wiz-broker:2.5"
+          imagePullPolicy: Always
```
demonCoder95: Same comments here to update the image URL and `imagePullPolicy`.

cluster/manifests/wiz/connector/job.yaml
```diff
@@ -0,0 +1,171 @@
...
+          command:
+            - "wiz-broker"
+          args:
+
```
demonCoder95: nit: remove empty line

cluster/manifests/wiz/connector/job.yaml
```diff
@@ -0,0 +1,171 @@
...
+  manualSelector: true
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: wiz-kubernetes-connector
+      app.kubernetes.io/instance: wiz-connector
+  backoffLimit: 1
+  template:
+    metadata:
+      labels:
+
```
demonCoder95: nit: remove empty line.

cluster/manifests/wiz/connector/job.yaml
```diff
@@ -0,0 +1,171 @@
...
+        runAsNonRoot: true
+        runAsUser: 1000
+      containers:
+        - name: wiz-connector-delete
+          securityContext:
+            allowPrivilegeEscalation: false
+            runAsNonRoot: true
+            runAsUser: 1000
+          image: "wiziopublic.azurecr.io/wiz-app/wiz-broker:2.5"
+          imagePullPolicy: Always
```
demonCoder95: same suggestion here to update this

cluster/manifests/wiz/connector/job.yaml
```diff
@@ -0,0 +1,171 @@
...
+                key: clientId
+                optional: false
+          - name: WIZ_CLIENT_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: wiz-api-token
+                key: clientToken
+                optional: false
+          - name: WIZ_ENV
+            value: ""
```
demonCoder95: do we need an empty env var?

cluster/manifests/wiz/connector/role.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - apiGroups: [""]
+    resources: ["secrets"]
+    resourceNames: ["wiz-connector-connector"]
+    verbs: ["update", "get"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    resourceNames: [
+      "wiz-api-token",
+      "wiz-cluster-reader-token",
+    ]
```
demonCoder95: As suggested above, let's use the `-` format when listing values on multiple lines:
```yaml
resourceNames:
  - wiz-api-token
  - wiz-cluster-reader-token
```

cluster/manifests/wiz/connector/job.yaml
```diff
@@ -0,0 +1,171 @@
...
+            runAsNonRoot: true
+            runAsUser: 1000
+          image: "wiziopublic.azurecr.io/wiz-app/wiz-broker:2.5"
+          imagePullPolicy: Always
+          command: ["/bin/sh", "-c"]
+          args:
+            - >
+              wiz-broker delete-kubernetes-connector 
+              --input-secrets-namespace 
+              "default"
```
demonCoder95: I think this namespaces should be `wiz` and not `default`. 

I'm looking at the `secrets.yaml` file below and it's deploying the secret in the `wiz` namespace.

cluster/manifests/wiz/connector/secrets.yaml
demonCoder95: Do we still need the `base64` template function, now that we're using the `data` field that's already base64 encoded when the secret object will be stored? (That's what I understand from Zak's comment above)

cluster/manifests/wiz/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: wiz
```
demonCoder95: I believe you need to rename the file to `001-namespace.yaml` to ensure that CDP deploys it before any other manifest. Otherwise, you will run into issues. Similarly, you want to ensure that the RBAC resources are deployed before any deployment/job objects are deployed.

You can see how we do it for `karpenter` for inspiration on how to manage precedence of manifest application: https://github.com/zalando-incubator/kubernetes-on-aws/tree/dev/cluster/manifests/z-karpenter

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,223 @@
+{{ if eq .Cluster.ConfigItems.wiz_enable_runtime_monitoring_daemonset "true"}}
+---
+# Source: wiz-sensor/templates/daemonset.yaml
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: wiz-sensor
+  labels:
+
```
demonCoder95: nit: empty line should be removed

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,223 @@
...
+            requiredDuringSchedulingIgnoredDuringExecution:
+              nodeSelectorTerms:
+              - matchExpressions:
+                - key: master
+                  operator: DoesNotExist
+              - matchExpressions:
+                - key: dedicated
+                  operator: NotIn
+                  values:
+                  - skipper-ingress
```
demonCoder95: We also have `skipper` and `skipper-ingress-redis` labelled nodes that we will have to skip if we're skipping `skipper-ingress`.

We also have other "system" dedicated nodes like `zmon-worker-arm64`, `zmon-redis` and `zmon-appliance` labelled nodes for ZMON.

demonCoder95: I'm assuming we are scoping the Wiz system only to "application worker nodes", please correct me if this isn't necessarily the case.

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,223 @@
...
+        - key: nvidia.com/gpu
+          value: present
+          effect: NoSchedule
+
+      restartPolicy: Always
+
+      containers:
+      - name: wiz-sensor
+        image: wizio.azurecr.io/sensor:v1
+        imagePullPolicy: Always
```
demonCoder95: todo: this needs updating, as suggested above

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,223 @@
...
+                  values:
+                  - skipper-ingress
+
+      hostPID: true
+      hostIPC: true
+      tolerations:
+        - key: karpenter.sh/disruption
+          operator: Exists
+        - key: node.kubernetes.io/not-ready
+          operator: Exists
```
demonCoder95: I see for `zalando.org/node-not-ready` taint below we added `effect: NoSchedule`. We probably need the same behavior for the `node.kubernetes.io/not-ready` taint as well?

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,223 @@
...
+            user: "system_u"
+            role: "system_r"
+            level: "s0"
+            type: "spc_t"
+
+        env:
+        - name: MIN_SENSOR_COMPAT_BUILD_VER
+          value: "4142"
+        - name: HELM_CHART_COMPAT_VER
+          value: "1"
```
demonCoder95: Probably need to drop this ENVVAR.

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,223 @@
...
+        - name: WIZ_TMP_STORE
+          value: "/wiz-sensor-store/"
+        - name: WIZ_HOST_STORE
+          value: "/wiz-host-cache/"
+        - name: WIZ_RAMFS_STORE
+          value: "/tmp/"
+        - name: LOG_FILE
+          value: "/wiz-sensor-store/sensor.log"
```
demonCoder95: I see below the `tmp-store` `emptyDir` type volume that is backing the `/wiz-sensor-store` directory is only `100Mi` in size. This might not be suitable for logs as I see the log level set to `info`. We probably need a bigger volume here.

I see you specified `"Memory"` backend. From the [Kubernetes docs](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir):

> While tmpfs is very fast be aware that, unlike disks, files you write count against the memory limit of the container that wrote them

I suggest you can just drop the `"Memory"` backend and use a disk backed `emptyDir` volume that's not "too large", to rely on the root FS, at the same time not exhaust the root disk of the node where it runs.

We might want to make a note of monitoring the log volume Wiz generates once we roll it out.

demonCoder95: Also, I see the `logging-destination` annotation missing on the pod templates in the application manifests, please add that. Example: https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/z-karpenter/deployment.yaml#L28

This will ensure that the Wiz logs are sent to Scalyr as well. This is a built-in feature provided in the infrastructure and the necessary information required to setup the connectivity with the Scalyr account is already injected to the pod when it's admitted, so you just need to provide this annotation and logs should show up in Scalyr.

That should make it safe to have a somewhat smaller volume for logging locally to the disk, since logs will be pushed to Scalyr already.

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,223 @@
...
+        - name: CRI_SOCKET_CUSTOM_PATH
+          value: 
+        - name: HTTP_PROXY_URL
+          value: 
+        - name: HTTP_PROXY_USERNAME
+          value: 
+        - name: HTTP_PROXY_PASSWORD
+          value: 
+        - name: HTTP_PROXY_CERT
+          value:
```
demonCoder95: Maybe we can just drop these empty ENVVARs?

cluster/manifests/wiz/sensor/secrets.yaml
```diff
@@ -0,0 +1,39 @@
...
+    helm.sh/chart: wiz-sensor-1.0.4760
+    image/tag: v1
+    app.kubernetes.io/name: wiz-sensor
+    app.kubernetes.io/instance: wiz-sensor
+    app.kubernetes.io/version: "1.0.4708"
+    app.kubernetes.io/managed-by: Helm
+type: Opaque
+data:
+  clientId: "{{ .Cluster.ConfigItems.wiz_api_client_id | base64 }}"
+  clientToken: "{{ .Cluster.ConfigItems.wiz_api_client_token | base64 }}"
```
demonCoder95: I suppose the same applies here, do we need to drop the `base64` template function?

cluster/manifests/wiz/sensor/secrets.yaml
```diff
@@ -0,0 +1,39 @@
...
+    
+    helm.sh/chart: wiz-sensor-1.0.4760
+    image/tag: v1
+    app.kubernetes.io/name: wiz-sensor
+    app.kubernetes.io/instance: wiz-sensor
+    app.kubernetes.io/version: "1.0.4708"
+    app.kubernetes.io/managed-by: Helm
+  namespace: wiz
+data:
+  .dockerconfigjson: "{{ .Cluster.ConfigItems.wiz_sensor_dockerconfigjson }}"
```
demonCoder95: This looks strange 😅  I suggest we do the following:
1. Instead of a Secret, make this manifest a `ConfigMap` that keeps the docker config JSON.
2. We expose only the token itself as a config-item, instead of `wiz_sensor_dockerconfigjson`. (or username and token both can be config-items)
3. We populate the token in the Docker config JSON using that config-item in the ConfigMap instead.

cluster/manifests/wiz/connector/job.yaml
vinaythupili: Yeah these annotations are added while unpacking the Helm chart of wiz , we can remove helm annotations `helm.sh` , i'm not sure of the `rollme.hashes`

cluster/manifests/wiz/connector/secrets.yaml
vinaythupili: @zaklawrencea has encoded this with base64, so i assume this is needed.

cluster/manifests/wiz/connector/deployment.yaml
vinaythupili: Not sure why WIZ has provided this empty ENV vars here

cluster/manifests/wiz/sensor/dameonset.yaml
vinaythupili: The Target is to Deploy it on all worker nodes.

cluster/manifests/wiz/sensor/dameonset.yaml
vinaythupili: Took the tolerations similar to core-dns https://github.com/zalando-incubator/kubernetes-on-aws/blob/891cbaa76e062210f1bbe74446ca0cc0a2191684/cluster/manifests/01-coredns-local/daemonset-coredns.yaml#L237-L260

cluster/manifests/wiz/connector/deployment.yaml
demonCoder95: We decided to keep these for now. We can clean up later.

cluster/manifests/wiz/connector/job.yaml
demonCoder95: We decided to keep the empty vars for now.

cluster/manifests/wiz/connector/secrets.yaml
demonCoder95: Okay, let's keep it then.

cluster/manifests/wiz/sensor/dameonset.yaml
demonCoder95: Alright. Thanks for sharing 🙏 I'll take another look later, we can keep the comment open for now, but it shouldn't be a blocker to move forward.

cluster/manifests/wiz/sensor/dameonset.yaml
demonCoder95: This will be evaluated later. It's a TODO for now. Not a blocker.

cluster/manifests/wiz/sensor/dameonset.yaml
demonCoder95: We decided to keep the empty vars for now.

cluster/manifests/wiz/sensor/secrets.yaml
demonCoder95: As discussed above, let's keep it.

cluster/manifests/wiz/connector/job.yaml
demonCoder95: We decided also to remove the `rollme` related annotations as they're used by Helm to provide rolling updates. We have CLM managing that for us.

cluster/manifests/wiz/connector/deployment.yaml
demonCoder95: We decided to remove these.

cluster/manifests/wiz/namespace.yaml
demonCoder95: This has been done now.

cluster/manifests/wiz/sensor/dameonset.yaml
demonCoder95: This was done because `skipper-ingress` nodes weren't large enough in some cases to accommodate the wiz daemonset pods. As wiz needs to run on all worker nodes, we need to work with GW&Proxy to address this properly. It's a TODO for now, not a blocker.

cluster/manifests/wiz/sensor/secrets.yaml
demonCoder95: This is a TODO for now, will be done later. Shouldn't be a blocker, but we should definitely do this in a cleaner way, instead of passing an encrypted Docker config JSON file.

cluster/manifests/wiz/connector/deployment.yaml
```diff
@@ -0,0 +1,72 @@
...
+            items:
+              - key: connectorData
+                path: data
+      containers:
+        - name: wiz-broker
+          securityContext:
+            allowPrivilegeEscalation: false
+            runAsNonRoot: true
+            runAsUser: 1000
+          image: "container-registry-test.zalando.net/secops-systems/wiz-broker:latest"
```
demonCoder95: `latest` tag will not work in production. So, we definitely need to fix this to the latest build version before merging. It's fine for testing.

cluster/manifests/wiz/connector/job.yaml
```diff
@@ -0,0 +1,145 @@
...
+      component: "connector"
+  backoffLimit: 1
+  template:
+    metadata:
+      labels: 
+        helm.sh/chart: wiz-kubernetes-connector-3.1.1
+        app.kubernetes.io/name: wiz-kubernetes-connector
+        app.kubernetes.io/instance: wiz-connector
+        app.kubernetes.io/version: "2.5"
+        app.kubernetes.io/managed-by: Helm
```
demonCoder95: Need to update these as well.

cluster/manifests/wiz/connector/job.yaml
```diff
@@ -0,0 +1,145 @@
...
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1000
+      containers:
+        - name: wiz-connector-creator
+          securityContext:
+            allowPrivilegeEscalation: false
+            runAsNonRoot: true
+            runAsUser: 1000
+          image: "container-registry-test.zalando.net/secops-systems/wiz-broker:latest"
```
demonCoder95: need to use the latest build instead of `latest` before merging.

cluster/manifests/wiz/connector/job.yaml
```diff
@@ -0,0 +1,145 @@
...
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1000
+      containers:
+        - name: wiz-connector-delete
+          securityContext:
+            allowPrivilegeEscalation: false
+            runAsNonRoot: true
+            runAsUser: 1000
+          image: "container-registry-test.zalando.net/secops-systems/wiz-broker:latest"
```
demonCoder95: need to use latest build before merging instead of `latest`.

cluster/manifests/wiz/connector/serviceaccount.yaml
```diff
@@ -0,0 +1,42 @@
...
+kind: ServiceAccount
+metadata:
+  name: wiz-broker
+  namespace: "wiz"
+  labels:
+    helm.sh/chart: wiz-broker-2.1.0
+    app.kubernetes.io/name: wiz-broker
+    app.kubernetes.io/instance: wiz-connector
+    app.kubernetes.io/version: "2.5"
+    app.kubernetes.io/managed-by: Helm
```
demonCoder95: need to update these

cluster/manifests/wiz/connector/serviceaccount.yaml
```diff
@@ -0,0 +1,42 @@
...
+  namespace: "wiz"
+  labels:
+    helm.sh/chart: wiz-kubernetes-connector-3.1.1
+    app.kubernetes.io/name: wiz-kubernetes-connector
+    app.kubernetes.io/instance: wiz-connector
+    app.kubernetes.io/version: "2.5"
+    app.kubernetes.io/managed-by: Helm
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
```
demonCoder95: need to fix these. Remove the annotations and fix the labels as done in other manifests.

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,214 @@
...
+          effect: NoSchedule
+        - key: nvidia.com/gpu
+          value: present
+          effect: NoSchedule
+
+      restartPolicy: Always
+
+      containers:
+      - name: wiz-sensor
+        image: container-registry-test.zalando.net/secops-systems/wiz-sensor:latest
```
demonCoder95: fix image tag before merging

cluster/manifests/wiz/sensor/secrets.yaml
```diff
@@ -0,0 +1,32 @@
...
+  clientToken: "{{ .Cluster.ConfigItems.wiz_api_client_token | base64 }}"
+---
+# Source: wiz-sensor/templates/imagepullsecret.yaml
+apiVersion: v1
+kind: Secret
+type: kubernetes.io/dockerconfigjson
+metadata:
+  name: wiz-sensor-imagepullkey
+  labels:
+
```
demonCoder95: remove empty line

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,214 @@
...
+              containerName: wiz-sensor
+              resource: limits.memory
+              divisor: "1Mi"
+        - name: POD_CPU_LIMITS
+          valueFrom:
+            resourceFieldRef:
+              containerName: wiz-sensor
+              resource: limits.cpu
+              divisor: "1m"
+
```
demonCoder95: remove empty line

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,214 @@
...
+          runAsGroup: 2202
+          readOnlyRootFilesystem: true
+          seccompProfile:
+            type: Unconfined
+          seLinuxOptions:
+            user: "system_u"
+            role: "system_r"
+            level: "s0"
+            type: "spc_t"
+
```
demonCoder95: remove empty line

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,214 @@
...
+          value: present
+          effect: NoSchedule
+
+      restartPolicy: Always
+
+      containers:
+      - name: wiz-sensor
+        image: container-registry-test.zalando.net/secops-systems/wiz-sensor:latest
+        imagePullPolicy: IfNotPresent
+
```
demonCoder95: remove empty line

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,214 @@
...
+          operator: Exists
+        - key: zalando.org/node-not-ready
+          operator: Exists
+          effect: NoSchedule
+        - key: nvidia.com/gpu
+          value: present
+          effect: NoSchedule
+
+      restartPolicy: Always
+
```
demonCoder95: remove empty line

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,214 @@
...
+          operator: Exists
+        - key: aws.amazon.com/spot
+          operator: Exists
+        - key: zalando.org/node-not-ready
+          operator: Exists
+          effect: NoSchedule
+        - key: nvidia.com/gpu
+          value: present
+          effect: NoSchedule
+
```
demonCoder95: remove empty line

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,214 @@
...
+              nodeSelectorTerms:
+              - matchExpressions:
+                - key: master
+                  operator: DoesNotExist
+              - matchExpressions:
+                - key: dedicated
+                  operator: NotIn
+                  values:
+                  - skipper-ingress # ToDo : Target is to deploy on all worker nodes
+
```
demonCoder95: remove empty line

cluster/manifests/wiz/sensor/dameonset.yaml
```diff
@@ -0,0 +1,214 @@
...
+    metadata:
+      labels:
+        helm.sh/chart: wiz-sensor-1.0.4760
+        application: "wiz"
+        component: "sensor"
+      annotations:
+        container.apparmor.security.beta.kubernetes.io/wiz-sensor: unconfined
+        cluster-autoscaler.kubernetes.io/enable-ds-eviction: "true"
+        node-ready.cluster.zalando.org/exclude: "true"
+
```
demonCoder95: remove empty line

cluster/manifests/wiz/sensor-dameonset.yaml
zaklawrencea, line 1: Typo in file name:

```diff
-.../sensor-dameonset.yaml
+.../sensor-daemonset.yaml
```

cluster/manifests/wiz/sensor-daemonset.yaml
```diff
@@ -0,0 +1,196 @@
...
+        - key: nvidia.com/gpu
+          value: present
+          effect: NoSchedule
+        - key: dedicated
+          value: skipper-ingress
+          effect: NoSchedule
+      restartPolicy: Always
+      containers:
+      - name: wiz-sensor
+        image: container-registry-test.zalando.net/secops-systems/wiz-sensor:v1-pr-1-4
```
mikkeloscar: The images should be updated to use a production image (There are more places were a test image is used)

vinaythupili: updated 👍

mikkeloscar: Not only the tag needed to be updated. Also the url:

`container-registry-test.zalando.net` -> `container-registry.zalando.net` (without `-test`)

cluster/config-defaults.yaml
```diff
@@ -1202,3 +1202,8 @@ role_sync_controller_enabled: "true"
 {{ else }}
 role_sync_controller_enabled: "false"
 {{ end }}
+
+#Wiz Configs
+wiz_enable_runtime_monitoring_daemonset: "false"
```
mjungsbluth: I would strongly suggest to split this in two:
- wiz_enable_runtime_sensor
- wiz_enable_k8s_connector

The name is a bit misleading and splitting this not only makes rollout more granular, it also will allow graceful degradation in the future. You can then for example disable the sensor selectively without also going blind due to the missing connector. 

The resources seem to already to be split.

cluster/config-defaults.yaml
```diff
@@ -1202,3 +1202,8 @@ role_sync_controller_enabled: "true"
...
+wiz_adapter_cpu: "300m"
```
mjungsbluth: This same setting is applied for very different resources, are we sure this will work?

cluster/manifests/wiz/005-connector-job.yaml
```diff
@@ -0,0 +1,155 @@
...
+          securityContext:
+            allowPrivilegeEscalation: false
+            runAsNonRoot: true
+            runAsUser: 1000
+          image: "container-registry.zalando.net/secops-systems/wiz-broker:2.5-main-2"
+          imagePullPolicy: IfNotPresent
+          command:
+            - "wiz-broker"
+          args:
+            - create-kubernetes-connector
```
mjungsbluth, line 45: What does this Job actually do? I looks like it is supposed to create K8s resources via the Kubernetes API. The attached service account only has permissions to update the connector secret though...

cluster/manifests/wiz/005-connector-job.yaml
```diff
@@ -0,0 +1,155 @@
...
+            - --api-server-endpoint
+            - "https://kubernetes.default.svc.cluster.local"
+            - --secrets-namespace
+            - "wiz"
+            - --service-account-token-secret-name
+            - "wiz-cluster-reader-token"
+            - --output-secret-name
+            - "wiz-connector-connector"
+            - --is-on-prem=true
```
mjungsbluth, line 54: This sounds wrong, not sure what this is supposed to do.

cluster/manifests/wiz/connector-deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+      application: "wiz"
+      component: "connector"
+  template:
+    metadata:
+      labels:
+        helm.sh/chart: wiz-broker-2.1.0
+        application: "wiz"
+        component: "connector"
+    spec:
+      serviceAccountName: wiz-broker
```
mjungsbluth, line 27: This service account seems to have no role binding, is that really enough for its operations?

cluster/manifests/wiz/sensor-daemonset.yaml
```diff
@@ -0,0 +1,196 @@
...
+            - SYS_RAWIO # file hashing
+            - DAC_OVERRIDE # file hashing
+            - DAC_READ_SEARCH # file hashing
+            - NET_ADMIN # network events
+            - NET_RAW # network events
+            - IPC_LOCK # eBPF
+            - FOWNER # file hashing
+            - SYS_PTRACE # eBPF
+            - SYSLOG # kernel symbol resolve
+            - KILL # response
```
mjungsbluth, line 77: Do we want to enable this in the first iteration?

cluster/config-defaults.yaml
vinaythupili: That's recommended from WIZ , as sensor is a light weight agent. Also for connector at least till now now issues.

cluster/manifests/wiz/005-connector-job.yaml
vinaythupili: This job creates a required configuration in the wiz backend. so wiz connector will be able to contact the Kubernetes API using wiz broker .

cluster/manifests/wiz/005-connector-job.yaml
vinaythupili: As we are using Kubernetes API default SVC kubernetes.default.svc.cluster.local  , wiz treats as on-prem or as a private cluster

cluster/manifests/wiz/connector-deployment.yaml
vinaythupili: Yeah i think it should be fine as the SA have the default permissions

mikkeloscar: by default a service account has NO permissions.

mikkeloscar: Actually, we may still have our legacy setup left-over where a service account has read-only access by default.

But this you should not rely on

See this for context on this legacy setup: https://github.com/zalando-incubator/kubernetes-on-aws/pull/6586

cluster/manifests/wiz/sensor-daemonset.yaml
vinaythupili: This is used as part of WIZ incident response , When as a part of response action a remediation step like Terminate process is needed.

cluster/config-defaults.yaml
mjungsbluth: When I look at the Wiz resources, it is actually different for sensor and connector:
* https://github.com/wiz-sec/charts/blob/a1dc571bffdad4178b02f56fda8ad968e9cd0a7c/wiz-kubernetes-connector/values.yaml#L154
* https://github.com/wiz-sec/charts/blob/a1dc571bffdad4178b02f56fda8ad968e9cd0a7c/wiz-sensor/values.yaml#L269

cluster/manifests/wiz/005-connector-job.yaml
mjungsbluth: > This job creates a required configuration in the wiz backend

Sounds like the Wiz token has more privileges than I expected. Let us discuss this internally...

cluster/config-defaults.yaml
vinaythupili: updated 👍

cluster/manifests/deletions.yaml
```diff
@@ -339,6 +339,59 @@ post_apply:
 - name: kube-janitor
   kind: ClusterRoleBinding
 {{- end }}
+{{- if ne .Cluster.ConfigItems.wiz_enable_runtime_monitoring_daemonset "true" }}
```
zaklawrencea: Let's also update the deletions with the new config items.

cluster/config-defaults.yaml
vinaythupili: as we discussed separated resources config items for sensor and connector components

cluster/manifests/wiz/sensor-daemonset.yaml
vinaythupili: removed the KILL capability as discussed.

cluster/manifests/wiz/connector-deployment.yaml
vinaythupili: So the SA here is intended just to avoid the default service account to the POD , when there is non assigned.  more as best practise and work load identity

mikkeloscar: Not only that, if your service needs to read from the Kubernetes API then it would loose access to read once we disable the legacy default service account feature.

It should use a dedicated service account and specify the needed permissions via RBAC.

cluster/config-defaults.yaml
```diff
@@ -1202,3 +1202,12 @@ role_sync_controller_enabled: "true"
 {{ else }}
 role_sync_controller_enabled: "false"
 {{ end }}
+
+#Wiz Configs
+wiz_enable_runtime_sensor: "false"
```
mikkeloscar, line 1210: please describe with 1-2 lines in a comment what these configs do. See how we do it in the rest of this file.

memory/cpu config-items don't need much description, those are obvious, but the others e.g. like `wiz_node_feature_rollout` does need a bit of description how to use it.

cluster/manifests/deletions.yaml
```diff
@@ -339,6 +339,61 @@ post_apply:
...
+- name: wiz-sensor-apikey
+  kind: Secret
+  namespace: wiz
+- name: wiz-sensor
+  kind : ClusterRole
+  namespace: wiz
+- name: wiz-sensor
+  kind : ClusterRoleBinding
+  namespace: wiz
+{{- end }}
```
mikkeloscar, line 396: If both `runtime_sensor` and `runtime_connector` are false, then it should also delete the `wiz` namespace as a last step.

cluster/manifests/wiz/sensor-daemonset.yaml
```diff
@@ -0,0 +1,203 @@
...
+  labels:
+    helm.sh/chart: wiz-sensor-1.0.4760
+    application: "wiz"
+    component: "sensor"
+  namespace: wiz
+spec:
+  selector:
+    matchLabels:
+      application: "wiz"
+      component: "sensor"
```
mikkeloscar: please use `daemonset: wiz-sensor` as the only label selector here. This is how we do it for all other daemonsets/deployments in this repo and it allows easier change of `application/component` labels if ever needed in the future.

cluster/manifests/wiz/connector-deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+  labels:
+    helm.sh/chart: wiz-broker-2.1.0
+    application: "wiz"
+    component: "connector"
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      application: "wiz"
+      component: "connector"
```
mikkeloscar: please use `deployment: wiz-connector-agent` as the only label selector here. This is how we do it for all other daemonsets/deployments in this repo and it allows easier change of `application/component` labels if ever needed in the future.

cluster/manifests/wiz/sensor-daemonset.yaml
```diff
@@ -0,0 +1,203 @@
...
+        cluster-autoscaler.kubernetes.io/enable-ds-eviction: "true"
+        node-ready.cluster.zalando.org/exclude: "true"
+    spec:
+      serviceAccountName: wiz-sensor
+      nodeSelector:
+        {{ if eq .Cluster.ConfigItems.wiz_node_feature_rollout "true" }}
+        node-feature.zalando.org/wiz: enabled
+        {{ else }}
+        node.kubernetes.io/role: worker
+        {{ end }}
```
mikkeloscar, line 36: This should not be `node-feature.zalando.org/wiz: enabled` or `node.kubernetes.io/role: worker`. It should always select nodes with `node.kubernetes.io/role: worker` and optionally it should limit this futher to nodes with: `node-feature.zalando.org/wiz: enabled`

E.g. you want this:

```yaml
nodeSelector:
        {{ if eq .Cluster.ConfigItems.wiz_node_feature_rollout "true" }}
        node-feature.zalando.org/wiz: enabled
        {{ end }}
        node.kubernetes.io/role: worker
```

cluster/manifests/wiz/sensor-daemonset.yaml
```diff
@@ -0,0 +1,203 @@
...
+      labels:
+        helm.sh/chart: wiz-sensor-1.0.4760
+        application: "wiz"
+        component: "sensor"
+      annotations:
+        container.apparmor.security.beta.kubernetes.io/wiz-sensor: unconfined
+        cluster-autoscaler.kubernetes.io/enable-ds-eviction: "true"
+        node-ready.cluster.zalando.org/exclude: "true"
+    spec:
+      serviceAccountName: wiz-sensor
```
mikkeloscar, line 32: This service account should be added to the privileged list here: https://github.com/zalando-incubator/kubernetes-on-aws/blob/5f4fdd525d1fbabd0fbd717e387367c8843528ca/cluster/manifests/01-admission-control/config.yaml#L95-L113

it should ofc. only be added if `eq .Cluster.ConfigItems.wiz_enable_runtime_sensor "true"`

This was currently set per cluster via: `teapot_admission_controller_pod_security_policy_privileged_service_accounts` but it makes more sense to make it a condition of the `wiz_enable_runtime_sensor: true|false` config-item.

cluster/manifests/wiz/sensor-daemonset.yaml
```diff
@@ -0,0 +1,203 @@
...
+            - SYS_RESOURCE # eBPF
+            - SYS_RAWIO # file hashing
+            - DAC_OVERRIDE # file hashing
+            - DAC_READ_SEARCH # file hashing
+            - NET_ADMIN # network events
+            - NET_RAW # network events
+            - IPC_LOCK # eBPF
+            - FOWNER # file hashing
+            - SYS_PTRACE # eBPF
+            - SYSLOG # kernel symbol resolve
```
mikkeloscar, line 75: These capabilities should be enabled conditionally here: https://github.com/zalando-incubator/kubernetes-on-aws/blob/5f4fdd525d1fbabd0fbd717e387367c8843528ca/cluster/manifests/01-admission-control/config.yaml#L119-L136

(Those that are not already mentioned in the list)

cluster/manifests/wiz/sensor-daemonset.yaml
```diff
@@ -0,0 +1,203 @@
...
+          items:
+            - key: clientEndpoint
+              path: clientEndpoint
+      - name: sensor-store
+        emptyDir:
+          sizeLimit: "1Gi"
+      - name: tmp-store
+        emptyDir:
+          sizeLimit: "100Mi"
+          medium: "Memory"
```
mikkeloscar, line 202: Note that this memory volume will contribute to the memory limit of the pod. Would be good to understand with wiz if this must be memory backed or could be backed by disk and thereby allow to potentially lower the memory request/limit of daemonset pod.

vinaythupili: Sure @mikkeloscar 👍 , will check with the team if we can adapt the disk instead of memory.

cluster/manifests/wiz/sensor-daemonset.yaml
vinaythupili: updated 🙌

cluster/manifests/wiz/sensor-daemonset.yaml
vinaythupili: updated 🙌

cluster/manifests/wiz/sensor-daemonset.yaml
vinaythupili: updated 🙌

cluster/manifests/wiz/connector-deployment.yaml
vinaythupili: updated 🙌

cluster/manifests/wiz/sensor-daemonset.yaml
vinaythupili: updated 🙌

cluster/manifests/deletions.yaml
vinaythupili: updated 🙌

cluster/config-defaults.yaml
vinaythupili: updated 🙌

cluster/manifests/wiz/005-connector-job.yaml
```diff
@@ -10,20 +10,21 @@ metadata:
     helm.sh/chart: wiz-broker-2.1.0
     application: "wiz"
     component: "connector"
+    deployment: "wiz-connector-agent"
```
mikkeloscar: Actually, this should be `job: wiz-connector-agent` as we use the kind to avoid that job and deployment resources with the same name could overlap.

Sorry if I suggested `deployment`, then I didn't realize this was a Job manifest.

cluster/manifests/01-admission-control/config.yaml
```diff
@@ -139,9 +142,20 @@ data:
...
-  pod.pod-security-policy.allowed-unsafe-sysctls.{{ $sysctl }}: ""
+  pod.pod-security-policy.allowed-restricted-capabilities.DAC_READ_SEARCH: ""
+  pod.pod-security-policy.allowed-restricted-capabilities.IPC_LOCK: ""
+  pod.pod-security-policy.allowed-restricted-capabilities.NET_ADMIN: ""
+  pod.pod-security-policy.allowed-restricted-capabilities.SYS_ADMIN: ""
+  pod.pod-security-policy.allowed-restricted-capabilities.SYS_MODULE: ""
+  pod.pod-security-policy.allowed-restricted-capabilities.SYS_PTRACE: ""
+  pod.pod-security-policy.allowed-restricted-capabilities.SYS_RESOURCE: ""
+  pod.pod-security-policy.allowed-restricted-capabilities.SYS_RAWIO: ""
+  pod.pod-security-policy.allowed-restricted-capabilities.SYSLOG: ""
```
mikkeloscar: This change looks wrong. It removes the `allowed-unsafe-sysctls.{{ $sysctl }}: ""` setting and adds the capbilties based on the `allowed_unsafe_sysctls` config-item.

The capabilties should be behind the config-item: `{{- if eq. .Cluster.ConfigItems.wiz_enable_runtime_sensor "true" }}` and they should apply to the `allowed-privileged-capabilities` not `allowed-restricted-capabilities`. We only want to enable these for privileged pods.

vinaythupili: updated the capabilities for privileged pods [74b22dc](https://github.com/zalando-incubator/kubernetes-on-aws/pull/8679/commits/74b22dc5c25a208fcc387f5de8b3cc1fd2289cc9)

cluster/manifests/01-admission-control/config.yaml
```diff
@@ -139,6 +142,24 @@ data:
...
+  pod.pod-security-policy.allowed-privileged-capabilities.SYS_RESOURCE: ""
+  pod.pod-security-policy.allowed-privileged-capabilities.SYS_RAWIO: ""
+  pod.pod-security-policy.allowed-privileged-capabilities.SYSLOG: ""
+  pod.pod-security-policy.allowed-privileged-capabilities.NET_RAW: ""
+  pod.pod-security-policy.allowed-privileged-capabilities.FOWNER: ""
+  pod.pod-security-policy.allowed-privileged-capabilities.KILL: ""
+  pod.pod-security-policy.allowed-privileged-capabilities.DAC_OVERRIDE: ""
+  pod.pod-security-policy.allowed-privileged-capabilities.SYS_CHROOT: ""
+  pod.pod-security-policy.allowed-privileged-capabilities.MKNOD: ""
+  pod.pod-security-policy.allowed-privileged-capabilities.SETFCAP: ""
```
mikkeloscar: Please only add the capabilities here, which are not already defined above under `allowed-restricted-capabilities` as the privileged capabilities will inherit what is allowed for restricted. 

These should not be defined here:

```
NET_RAW
FOWNER
KILL
DAC_OVERRIDE
SYS_CHROOT
MKNOD
SETFCAP
```

vinaythupili: updated the changes via [48df05c](https://github.com/zalando-incubator/kubernetes-on-aws/pull/8679/commits/48df05c439eccc4f85c5cf67b53f5c4d7ee31160)

mikkeloscar:
:+1:

mikkeloscar:
:+1:

vinaythupili:
👍

PR #8685, opened 2024-12-23
[e2e] use p4 instead of old p3 instances

demonCoder95:
We got notified in the e2e account to migrate away from `p3` instance type as AWS is decommissioning it in the upcoming months. This updates it to use `p4` instead, as suggested by AWS.

> Please migrate to newer generation EC2 G or P instances by December 20, 2025

### DIFF
--- a/test/e2e/cluster_config.sh
+++ b/test/e2e/cluster_config.sh
@@ -131,7 +131,6 @@ clusters:
       taints: dedicated=node-tests:NoSchedule
   - discount_strategy: spot
     instance_types:
-    - "p4d.24xlarge"
     - "g3s.xlarge"
     - "g3.4xlarge"
     - "g4dn.xlarge"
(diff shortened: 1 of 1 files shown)


demonCoder95:
👍

demonCoder95:
The smallest instance available is `p4d.24xlarge`, we probably need to decide if we want to add this in our e2e setup, or if we can drop the `p3` type altogether instead of replacing it. https://aws.amazon.com/ec2/instance-types/p4/

Another suggestion from AWS is to move to `g6` type, not sure what the availability in `eu-central-1` is for that type.

demonCoder95:
Keeping it `do-not-merge` until we decide upon a replacement.

mikkeloscar:
The e2e test depending on this is "just" testing if we can spin up a node with GPU and interact with the GPU it doesn't test performance etc. The goal here should simply be to have a pool of GPU types to increase the chance of getting one for the e2e test. the p4d.24xlarge seems like a big type or this. If there are smaller g6 available, let's try to add them instead. Otherwise just removing the p3 will also be enough as we have others in the pool already and likely are getting those if the p3 is being phased out by AWS.

demonCoder95:
We have `g6` available in smaller sizes; `xlarge`, `2xlarge`, and `4xlarge` so I added them instead.

mikkeloscar:
:+1:

demonCoder95:
👍

PR #8690, opened 2024-12-27
use upgraded instance types in default-for-karpenter

demonCoder95:
We decided to perform instance type upgrades:

| Instance Type | Replacement |
|---------------|-------------|
| c5            | c6i         |
| m5            | m6i         |
| r5            | r6i         |
| ~c6g~           | ~c7g~         |

For the `default-for-karpenter` type, we can simply drop the older types as we already have the new ones in the list. `c6g` is not in the list so no replacement is needed.

### DIFF
--- a/cluster/node-pools/worker-karpenter/provisioners.yaml
+++ b/cluster/node-pools/worker-karpenter/provisioners.yaml
@@ -128,9 +128,6 @@ spec:
         - key: "karpenter.k8s.aws/instance-family"
           operator: In
           values:
-            - "c5"
-            - "m5"
-            - "r5"
             - "c5d"
             - "m5d"
             - "r5d"


demonCoder95:
👍

katyanna:
| Instance Type | Replacement |
|---------------|-------------|
| c5            | c6i         |
| m5            | m6i         |
| c6g           | c7g         |

quick review of the table in the description, is that correct?

katyanna:
| Instance Type | Replacement |
|---------------|-------------|
| c5            | c6i         |
| m5            | m6i         |
| c6g           | c7g         |

quick review of the table in the description; similar to my comment on the e2e PR, I think there was a typo on the m5 replacement and there's no c6 changes in this PR. Is that correct?

demonCoder95:
I fixed the typo, thanks for highlighting it 🙂 Yes, that's correct. I copied the table from the internal issue we have for the replacement strategy, I copied the typo over 🙈 I fixed it now.

mikkeloscar:
The PR looks ok, but the description also mentions c6g/c7g which should not (and is not) included.

Those should not be included as there would be the risk of spinning up an ARM node for pods with a container image only working on amd64.

I crossed those out in the description.

mikkeloscar:
:+1:

PR #8692, opened 2024-12-27
add aws-efa-device-plugin manifests

demonCoder95:
This was requested to be enabled by the data team. The manifest source is: https://github.com/aws/eks-charts/tree/master/stable/aws-efa-k8s-device-plugin

Dry-run deployed the manifest to get a rendered DaemonSet YAML, and edited it accordingly, taking inspiration from the nvidia-gpu-device-plugin manifests at: https://github.com/zalando-incubator/kubernetes-on-aws/tree/dev/cluster/manifests/nvidia

```bash
$ helm repo add eks https://aws.github.io/eks-charts
$ helm install test-efa eks/aws-efa-k8s-device-plugin --dry-run --debug
```

TO-DOs:
- [x] Put the plugin behind a feature flag.
- [x] Create an image republish internal pipeline and use that image.
- [x] Adjust the instance types list to accommodate the Zalando specific GPU instance types only.

### DIFF
--- a/cluster/manifests/aws-efa-device-plugin/rbac.yaml
+++ b/cluster/manifests/aws-efa-device-plugin/rbac.yaml
@@ -1,0 +1,10 @@
+{{ if eq .Cluster.ConfigItems.aws_efa_device_plugin_enabled "true" }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: aws-efa-k8s-device-plugin
+  namespace: kube-system
+  labels:
+    application: kubernetes
+    component: aws-efa-k8s-device-plugin
+{{ end }}
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -427,6 +427,11 @@ nvidia_dcgm_exporter_enabled: "false"
 nvidia_dcgm_exporter_cpu: "10m"
 nvidia_dcgm_exporter_memory: "200Mi"
 
+# AWS EFA device plugin
+aws_efa_device_plugin_enabled: "false"
+aws_efa_device_plugin_cpu: "10m"
+aws_efa_device_plugin_memory: "20Mi"
+
 # static egress controller settings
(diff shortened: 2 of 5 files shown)


demonCoder95:
I removed `-metal` instances from the `nodeSelector` since we don't support them in our infrastructure. We still need to check the list and remove any other instance types we don't support.

cluster/manifests/aws-efa-device-plugin/rbac.yaml
```diff
@@ -0,0 +1,27 @@
...
+metadata:
+  name: aws-efa-k8s-device-plugin
+  namespace: kube-system
+  labels:
+    application: kubernetes
+    component: aws-efa-k8s-device-plugin
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: privileged-psp
```
mikkeloscar: This doesn't work anymore. We need to define the service account in this list: https://github.com/zalando-incubator/kubernetes-on-aws/blob/fae0f905acdd619b18bca11cd9d25cc5b67a3bca/cluster/manifests/01-admission-control/config.yaml#L95-L114

(And remove this role binding which is not needed)

mikkeloscar:
> I removed `-metal` instances from the `nodeSelector` since we don't support them in our infrastructure. We still need to check the list and remove any other instance types we don't support.

It should be fine to list instance types we don't support to avoid complexity of maintaining an overlapping list. As long as the list defined here include all we support it should work it's just to ensure it can only schedule on nodes of the right type, but we still control which node types are created in the first place via node pool configs.

cluster/manifests/aws-efa-device-plugin/rbac.yaml
mikkeloscar: Changed!

mikkeloscar:
:+1:

desinurch:
👍

PR #8730, opened 2025-01-07
Use default kube-controller-manager image

mikkeloscar:
This reverts #8395 to use the latest version of the kube-controller-manager image which includes the fix that was the original motivation for pinning the version.

This is possible after Kubernetes v1.31.4 which includes the fix: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8639


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #8729, opened 2025-01-07
[WIP] add aws-efa-device-plugin manifests

mikkeloscar:
Same as #8692 but on top of #8623 for testing.

PR #8733, opened 2025-01-07
Allow configuring associatePublicIp on instance launch

mikkeloscar:
By default we configure AssociatePublicIPAddress on the launch templates of the node pools such that an IP address is added at instance launch.

For some special usage cases e.g. what is being tested in #8729 we need to launch instances with multiple network interfaces. This is not supported with AssociatePublicIPAddress in EC2 as you get the error:

```
 InvalidParameterCombination: The associatePublicIPAddress parameter cannot be specified when launching with multiple network interfaces
```

See also this for more details on the problem: https://github.com/aws/karpenter-provider-aws/pull/5666

The workaround for this problem is to set MapPublicIPOnLaunch on the subnets and not set associatePublicIPAddress on the launch template. This PR puts the config behind a config-item so it can be disabled per cluster.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -44,6 +44,9 @@ karpenter_max_pods_per_node: "32"
 # legacy => 0.36.2-main-25.patched
 karpenter_version: "current"
 
+# Configure whether to associate public ip when launching instances.
+associate_public_ip_on_launch: "true"
+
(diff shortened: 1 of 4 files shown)


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #8738, opened 2025-01-08
Set prefix delegation false by default

mikkeloscar:
Set Prefix delegation false by default.

Not using prefix delegation gives a better utilization of ipv4 address space without sacrificing pod density on nodes too much in our environment.

We calculated that only 35 nodes would have less density than current usage across all production clusters.


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #8742, opened 2025-01-09
Configure associate Public IP at node Pool level

mikkeloscar:
Follow up to #8733 

Allow configuring the associatePublicIP setting at node pool level. It can still be configured cluster wide as before.


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #8758, opened 2025-01-13
Add optional support for internel node subnets

mikkeloscar:
Add optional support for launching nodes in internal subnets.

The purpose of this is to be able to discover and launch a subnet of node pools in internal subnets. This is useful for special instances running with NAT gateways or simply for running nodes in internal subnets.

A concrete use case is getting p5.48xlarge instances working with EFA. Those can't automatically get a public IP so we need another way to give them internet access e.g. via NAT.

This can be configured at node pool level.

depends on https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/840


mikkeloscar:
:+1:

zaklawrencea:
:+1:

mikkeloscar:
:+1:

PR #8775, opened 2025-01-15
fabric: update crd

veronikavlktna:
CRD validation for the Fabric Gateway name: must not exceed 63 chars

for https://github.com/zalando-incubator/kubernetes-on-aws/pull/8759


AlexanderYastrebov:
:+1:

demonCoder95:
👍

demonCoder95:
We should probably put this validation behind a feature flag to be able to control it if we run into issues once we start rolling out to production clusters.

AlexanderYastrebov:
> We should probably put this validation behind a feature flag to be able to control it if we run into issues once we start rolling out to production clusters.

We've already checked cluster resources across the board.

AlexanderYastrebov:
We agreed to re-create this from the proper user account.

PR #8780, opened 2025-01-17
Make internal subnet control associate-public-ip

mikkeloscar:
When `internal_node_subnets_enabled: true` setting `associate_public_ip_on_launch: true` doesn't make any sense. So ignore `associate_public_ip_on_launch` when `internal_node_subnets_enabled: true` to avoid needing to set both config items always.

Follow up to #8742 and #8758


demonCoder95:
👍

mikkeloscar:
:+1:

PR #8786, opened 2025-01-17
Fix forwarded headers exclude cirds for IPv6

mikkeloscar:
In EKS IPv6 clusters the ingress is both IPv6 and IPv4 enabled. Therefore we need to make skipper-ingress aware of the IPv4 VPC CIDR range for the `--forwarded-headers-exclude-cirds` flag.

* [x] depends on: https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/841


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #8805, opened 2025-01-22
e2e: remove EoL g3 instances from e2e pool

demonCoder95:
As per the email received from AWS for the e2e account:

> You are expected to migrate to newer generation G instances by October 31, 2025. Newer generation instances typically offer better price-performance than older generation instances.

As we already use newer generation instances in our e2e pool, will simply drop the older ones.


demonCoder95:
👍

mikkeloscar:
:+1:

PR #8806, opened 2025-01-22
Fix *ingress.cluster.local. on IPv6

mikkeloscar:
The DNS config for the "magic" `*.ingress.cluster.local.` address was wrongly configured such that it responded with `AAAA` on an `A` record query in IPv6 clusters.

This made it not work in Go applications as they would get an error like:

```
lookup deployment-status-service.ingress.cluster.local. on [2a05:d014:b94:3404::c4c3]:53: no such host
```

Somehow `curl` was able to handle this misconfiguration and use the IPv6 address responded on `A` query, this is why the issue was not noticed in the initial implementation :shrug:

This fixes the issue by replying with AAAA record if the query is for AAAA in IPv6 mode and no reply with anything on A query in IPv6 mode and vice versa.

### DIFF
--- a/cluster/manifests/01-coredns-local/configmap-local.yaml
+++ b/cluster/manifests/01-coredns-local/configmap-local.yaml
@@ -70,16 +70,26 @@ data:
         log
 {{ end }}
         template IN A {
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

demonCoder95:
👍

PR #8826, opened 2025-01-24
skipper-ingress: inline healthcheck routes

AlexanderYastrebov:
Skipper provides two flags to control kubernetes healthcheck routes: `-kubernetes-healthcheck` true by default and `-reverse-source-predicate` to select source predicate type.

This change disables automatic healthcheck routes and instead explicitly defines them to allow further modification.

In configuration with routesrv enabled skipper will not add default
filters to healthcheck routes but will do it with routesrv disabled.

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -161,6 +161,7 @@ spec:
 {{ else }}
           - "-kubernetes"
           - "-kubernetes-in-cluster"
+          - "-kubernetes-healthcheck=false" # see -inline-routes
           - "-kubernetes-path-mode=path-prefix"
           - "-kubernetes-backend-traffic-algorithm={{ .Cluster.ConfigItems.skipper_ingress_backend_traffic_algorithm }}"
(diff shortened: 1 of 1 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -308,9 +308,20 @@ spec:
...
+            kube__healthz_down: Path("/kube-system/healthz")
+              && Shutdown()
+              && SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128")
+              -> status(503)
+              -> <shunt>;
+            kube__healthz_up: Path("/kube-system/healthz")
+              && SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128")
+              -> disableAccessLog(200)
+              -> status(200)
+              -> <shunt>;
```
AlexanderYastrebov, line 322: https://github.com/zalando/skipper/blob/c2dbda66b929187234795d6e38ee007ac566808c/dataclients/kubernetes/kube.go#L474-L478

PR #8827, opened 2025-01-27
Assume RAID0 for instances with local storage

mikkeloscar:
Configure Karpenter to assume RAID0 setup for instances with local storage.

This maps to how we configure such instances at boot and is needed to make Karpenter know what ephemeral storage size to expect when choosing instances for a pod requesting ephemeral storage.

This is how it's used when calculating ephemeral storage for an instance type: https://github.com/aws/karpenter-provider-aws/blob/5f55b1d20d26302e9d3a0fc5943416dcfca1688c/pkg/providers/instancetype/types.go#L228-L235

This setting is added behind a config-item, on by default, to make it quick to turn off should it cause unexpected issues. (The change was tested/validated in a pet cluster)

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -44,6 +44,11 @@ karpenter_max_pods_per_node: "32"
 # legacy => 0.36.2-main-25.patched
 karpenter_version: "current"
 
+# configure whether karpenter should assume instances with local storage use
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #8841, opened 2025-01-27
Correctly exclude wiz from node readiness check

mikkeloscar:
Follow up to #8679 

Moves the annotation: `node-ready.cluster.zalando.org/exclude: "true"` to the right place (daemonset level) to have the intended effect of not marking nodes unready if wiz fails to run.


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #8854, opened 2025-01-30
Give more time for the routing change

wisinghe:
Give more time for the routing change

### DIFF
--- a/test/e2e/aaas.go
+++ b/test/e2e/aaas.go
@@ -147,7 +147,7 @@ func updateIngressAndWait(serviceName, hostName, path, ingressRoute string, port
 	)
 	ingressUpdate, err := cs.NetworkingV1().Ingresses(ingressCreate.ObjectMeta.Namespace).Update(context.TODO(), updatedIng, metav1.UpdateOptions{})
 	framework.ExpectNoError(err)
-	time.Sleep(2 * time.Minute) // wait for routing change propagation
+	time.Sleep(4 * time.Minute) // wait for routing change propagation
 
 	return ingressUpdate
 }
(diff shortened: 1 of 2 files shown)


test/e2e/aaas.go
```diff
@@ -147,7 +147,7 @@ func updateIngressAndWait(serviceName, hostName, path, ingressRoute string, port
 	)
 	ingressUpdate, err := cs.NetworkingV1().Ingresses(ingressCreate.ObjectMeta.Namespace).Update(context.TODO(), updatedIng, metav1.UpdateOptions{})
 	framework.ExpectNoError(err)
-	time.Sleep(2 * time.Minute) // wait for routing change propagation
+	time.Sleep(4 * time.Minute) // wait for routing change propagation
```
mikkeloscar: Sleeping here seems quite hacky, especially when we have to bump it like this.

It seems better to retry the expected response until a certain threshold here: https://github.com/zalando-incubator/kubernetes-on-aws/blob/55ae102966b0262778b98d8be4f4e24613195b0c/test/e2e/aaas.go#L62 e.g. increase the passed timeout from 10 seconds to minutes (and maybe update the logic in the `getAndWaitResponse` to try a bit less often than every 1 sec, maybe every 10 sec.).

Do we understand why it could take minutes to update the routes? skipper should do it in seconds usually.

wisinghe: The reason could be that OPA should download a bundle from S3 for the filter to be ready.

`getAndWaitResponse` is a common function from util.go, so I am not sure about changing the checking frequency there. But will increase the threshold for these calls.

wisinghe: Introduced a new util function to control the calling frequency and timeout

mikkeloscar:
:+1:

wisinghe:
👍

test/e2e/aaas.go
MustafaSaber: This can be one of the bottlenecks that we need to address, because it will slow all routing table updates.

PR #8857, opened 2025-01-31
Allow powerusers to use the subjectaccessreviews API

mikkeloscar:
Allow power users to use the `subjectaccessreviews` API in order to check what different identities has access to do.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #8871, opened 2025-02-04
Handle secret read clusterroles

mikkeloscar:
Improves handling of secret read cluster roles in EKS clusters

The implementation is described as follows:

1. Extend admission-controller with a clusterrole admitter which detects clusterroles that defined secret read permissions `({ApiGroups: [""], Resources:["secrets"],Verbs:["get|list|watch"]}`). If detected, it extracts the permissions from the role and moves them to an annotation: `clusterroles.admission-controller.zalando.org/secret-read-verbs: "get,list,watch"`. This achieves two things 1) Users can't deploy clusterroles with secret read permissions because those permissions will be removed during admission. 2) The role gets annotated such that it can be detected if the users intended to have read secret permissions. This is implemented in https://github.com/zalando-incubator/kubernetes-on-aws/pull/8987
2. Extend the role-sync-controller to look for clusterroles withs the annotation: `clusterroles.admission-controller.zalando.org/secret-read-verbs: "get,list,watch"` and resolve the clusterrolebindings binding to those roles. With the bindings we get a list of subjects that want read secret permissions at cluster level and those subjects can be added to the per namespace secrets role bindings. This is implemented in https://github.com/zalando-incubator/kubernetes-on-aws/pull/8986
3. Allow read secret permissions at the cluster level for poweruser/CDP. This effectively allows users to apply a clusterrole with read secret permissions, but because of 1 and 2, the clusterrole is not applied in that way but the subject(s) binding to the clusterrole will instead bind to the per namespace secret reader roles and the users get effectively the same permissions as in non-EKS, without changing their manifests.
4. Lastly, to prevent powerusers from reading from kube-system we remove secret read permissions from the cluster-role and only attach those permissions to the cdp/deployment-service service accounts. This means a poweruser, user, e.g. Manual access, can't create a cluster-role with read secret permissions. This is however still possible via CDP.

## TODO

* [ ] ~Add e2e covering this feature.~

### DIFF
--- a/cluster/manifests/02-admission-control/teapot.yaml
+++ b/cluster/manifests/02-admission-control/teapot.yaml
@@ -583,4 +583,33 @@ webhooks:
         expression: '!(request.userInfo.username in ["system:kube-controller-manager", "system:kube-scheduler", "zalando-iam:zalando:service:k8sapi_credentials-provider"])'
       - name: 'exclude-eks-components'
         expression: '!request.userInfo.username.startsWith("eks:")'
+  {{- if eq .Cluster.ConfigItems.role_sync_controller_enabled "true" }}
+  - name: clusterrole-admitter.teapot.zalan.do
+    clientConfig:
+      {{- if eq .Cluster.Provider "zalando-eks"}}
+      service:
+        name: "admission-controller"
+        namespace: "kube-system"
+        path: "/node"
+      {{- else }}
+      url: "https://localhost:8085/clusterrole"
+      {{- end }}
+      caBundle: "{{ .Cluster.ConfigItems.ca_cert_decompressed }}"
+    admissionReviewVersions: ["v1beta1"]
+    failurePolicy: Fail
+    matchPolicy: Equivalent
+    sideEffects: "None"
+    rules:
+      - operations: [ "CREATE", "UPDATE" ]
+        apiGroups: ["rbac.authorization.k8s.io"]
+        apiVersions: ["v1"]
+        resources: ["clusterroles"]
+    matchConditions:
+      - name: 'exclude-privileged-groups'
+        expression: 'request.userInfo.groups.all(g, !(g in ["system:masters", "system:nodes", "system:serviceaccounts:kube-system", "okta:common/administrator", "zalando:administrator"]))'
+      {{- if eq .Cluster.Provider "zalando-eks"}}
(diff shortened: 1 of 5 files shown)


cluster/manifests/roles/poweruser-role.yaml
```diff
@@ -58,21 +58,14 @@ rules:
...
   - get
-{{ if ne .Cluster.ConfigItems.role_sync_controller_enabled "true" }}
 - apiGroups:
   - ''
   resources:
   - secrets
   verbs:
-  - create
-  - delete
-  - deletecollection
```
mikkeloscar, line 69: These permissions are already defined for `secrets` above.

demonCoder95:
> Add e2e covering this feature

Don't the admission-controller unit tests cover this already? What kind of coverage are you thinking about for e2e tests? 🤔

mikkeloscar:
> > Add e2e covering this feature
> 
> Don't the admission-controller unit tests cover this already? What kind of coverage are you thinking about for e2e tests? 🤔

Nvm, the role-sync-controller was not working correctly making some e2e fail. This was fixed in https://github.com/zalando-incubator/kubernetes-on-aws/pull/9005

There is one thing which is not covered by e2e, will track it in an internal issue:

With this change the cdp and deployment-service _can_ read secrets from kube-system because they need this access to allow users to deploy read secret cluster roles. This can not be defined as a generic test as it would fail on non-eks where it's not allowed.

mikkeloscar:
:+1:

cluster/manifests/02-admission-control/teapot.yaml
```diff
@@ -583,4 +583,33 @@ webhooks:
...
       - name: 'exclude-eks-components'
         expression: '!request.userInfo.username.startsWith("eks:")'
+  {{- if eq .Cluster.ConfigItems.role_sync_controller_enabled "true" }}
+  - name: clusterrole-admitter.teapot.zalan.do
+    clientConfig:
+      {{- if eq .Cluster.Provider "zalando-eks"}}
+      service:
+        name: "admission-controller"
+        namespace: "kube-system"
+        path: "/node"
```
linki:
```suggestion
        path: "/clusterrole"
```

linki:
👍

mikkeloscar:
:+1:

PR #8879, opened 2025-02-05
feat: add new config item wiz_enable_runtime_connector_broker

vinaythupili:
- Currently both the manifests for connector job and broker deployment being implemented, Some times when the broker deployment started before the connector job finishes.
- The broker pod goes in to crash loop back and we have to run again the connector job and restart the broker pod.
- With this new config item `wiz_enable_runtime_connector_broker`. it will gives flexibility to deploy the Job first and broker deployment.


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #8880, opened 2025-02-05
Update to VPA 1.3.0

mikkeloscar:
https://github.com/kubernetes/autoscaler/releases/tag/vertical-pod-autoscaler-1.3.0

The biggest change is updating to only support `autoscaling.k8s.io/v1`. We need to first migrate all manifests using `autoscaling.k8s.io/v1beta1`


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #8881, opened 2025-02-05
efa-plugin: use label selector: vpc.amazonaws.com/efa: enabled

mikkeloscar:
This adds a nodeSelector to the efa plugin:

```
vpc.amazonaws.com/efa: enabled
```

This means node-pools that intend to run with EFA enabled instances should set this label. The reason for this is that even if an instance has the capability to attach EFA, it might not do it. If it doesn't then the efa-plugion daemonset pod just exits and the pod and node never gets ready.

This is a mitigation to that problem. Ideally the plugin would behave more graceful.


mikkeloscar:
:+1:

desinurch:
👍

PR #8885, opened 2025-02-06
Fix trustpolicy for skipper-ingress AWS role

mikkeloscar:
Follow up to #8875

Fix the trustpolicy on the skipper-ingress AWS IAM role in EKS mode.


mikkeloscar:
:+1:

wisinghe:
👍

PR #8886, opened 2025-02-06
feat: update wiz deletions to wiz_enable_runtime_connector_broker

vinaythupili:
-- In the Previous [PR](https://github.com/zalando-incubator/kubernetes-on-aws/pull/8879/files) the change has not been adapted to the deletions.


cluster/manifests/deletions.yaml
```diff
@@ -339,7 +339,7 @@ post_apply:
 - name: kube-janitor
   kind: ClusterRoleBinding
 {{- end }}
-{{- if ne .Cluster.ConfigItems.wiz_enable_runtime_connector "true" }}
+{{- if ne .Cluster.ConfigItems.wiz_enable_runtime_connector_broker "true" }}
```
mikkeloscar: Shouldn't this only be around the wiz-connector-agent?

Otherwise every wiz component is deleted if the broker config-item is not set true..

vinaythupili: the connector-agent(broker) is using the config item `wiz_enable_runtime_connector_broker` and the connector job is using `wiz_enable_runtime_connector`

mikkeloscar: So if you only enable `wiz_enable_runtime_connector` the job will be deleted right when it's created if `wiz_enable_runtime_connector_broker` is still false. This seems wrong.

vinaythupili: Yeah ! the job will be deleted right after the completion.

mikkeloscar: This sounds very fragile, you can't ensure the job will complete before being deleted..

IMO all the resources related to the connector job should be wrapped in `{{- if ne .Cluster.ConfigItems.wiz_enable_runtime_connector "true" }}` and the rest in `{{- if ne .Cluster.ConfigItems.wiz_enable_runtime_connector_broker "true" }}`

Then it's more clear what resources are deployed by what config-item. It will ofc. mean you need to do the following:

1. `wiz_enable_runtime_connector: true`
2. `wiz_enable_runtime_connector_broker: true` and `wiz_enable_runtime_connector: false`

vinaythupili: updated the PR

cluster/manifests/deletions.yaml
```diff
@@ -339,13 +339,24 @@ post_apply:
 - name: kube-janitor
   kind: ClusterRoleBinding
 {{- end }}
-{{- if ne .Cluster.ConfigItems.wiz_enable_runtime_connector "true" }}
+{{ - if ne .Cluster.ConfigItems.wiz_enable_runtime_connector "true" }}
```
mikkeloscar: this seems wrong :)

vinaythupili:
👍

zaklawrencea:
:+1:

PR #8890, opened 2025-02-07
feat: Update WIZ Broker and Connector Job resources 

vinaythupili:
As certain resources like the WIZ Service Account and Secrets are having dependency for both the Connector Job and Broker deployment. When we initially roll out the connector job and set the `wiz_enable_runtime_connector` config item to false after the job completion, there is a chance that these resources will be deleted. So setting them OR so if  `wiz_enable_runtime_connector_broker` or `wiz_enable_runtime_connector` is set they will not be deleted.


vinaythupili:
👍

zaklawrencea:
:+1:

PR #8891, opened 2025-02-07
Add kube-node-decommissioner CronJob

mikkeloscar:
This adds a simple cronJob which checks all nodes in a cluster and sees if the node should be replaced for some reason.

Currently it only implements checking for nodes in an unresponsive state where all pods are in terminating state.

This is intended as a simple mitigation to that problem. E.g. auto detect the condition and automatically mark the node for decommissioning. This is both faster than having an on-call person get paged and do it manually, and it's additionally less toil for 24x7.

It's enabled by default but can be turned off by setting `kube_node_decommissioner_enabled: false`


mikkeloscar:
:+1:

demonCoder95:
👍

PR #8894, opened 2025-02-10
skipper-ingress: explicit healthcheck routes

AlexanderYastrebov:
Skipper provides two flags to control kubernetes healthcheck routes: `-kubernetes-healthcheck` (true by default) that enables automatic healthcheck routes and `-reverse-source-predicate` to select source predicate type.

Currently when routesrv is enabled (default state) it adds automatic healthcheck routes and appends default filters to all routes including healthcheck routes.

This change disables automatic healthcheck routes and instead explicitly defines them in skipper pods to avoid adding default filters to them and enable further modification.

When routesrv is disabled though skipper will add default filters to explicitly-defined healthcheck routes like it happens currently. In this case default filters should be configured to not hinder healthchecks.

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -171,6 +171,7 @@ spec:
 {{ else }}
           - "-kubernetes"
           - "-kubernetes-in-cluster"
+          - "-kubernetes-healthcheck=false" # see -inline-routes
           - "-kubernetes-path-mode=path-prefix"
           - "-kubernetes-backend-traffic-algorithm={{ .Cluster.ConfigItems.skipper_ingress_backend_traffic_algorithm }}"
           - "-kubernetes-default-lb-algorithm={{ .Cluster.ConfigItems.skipper_ingress_default_lb_algorithm }}"
@@ -187,7 +188,6 @@ spec:
           - '-kubernetes-annotation-filters-append={{ .Cluster.ConfigItems.skipper_kubernetes_annotation_filters_append }}'
           - '-kubernetes-east-west-range-annotation-predicates={{ .Cluster.ConfigItems.skipper_kubernetes_east_west_range_annotation_predicates }}'
           - '-kubernetes-east-west-range-annotation-filters-append={{ .Cluster.ConfigItems.skipper_kubernetes_east_west_range_annotation_filters_append }}'
-          - "-reverse-source-predicate"
 {{ end }}
           - "-proxy-preserve-host"
           - "-compress-encodings={{ .Cluster.ConfigItems.skipper_compress_encodings }}"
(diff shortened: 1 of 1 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -310,9 +310,20 @@ spec:
...
+            kube__healthz_down: Path("/kube-system/healthz")
+              && Shutdown()
+              && SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128")
+              -> status(503)
+              -> <shunt>;
+            kube__healthz_up: Path("/kube-system/healthz")
+              && SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128")
+              -> disableAccessLog(200)
+              -> status(200)
+              -> <shunt>;
```
AlexanderYastrebov, line 324: This is the same as [here](https://github.com/zalando/skipper/blob/128b43767fe91d691551873aa6fc07e6fba6cb1f/dataclients/kubernetes/kube.go#L476-L477)

cluster/manifests/skipper/deployment.yaml
```diff
@@ -310,9 +310,21 @@ spec:
...
+              && Shutdown()
+              && SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128")
+              -> disableAccessLog()
+              -> status(503)
+              -> <shunt>;
+            kube__healthz_up: Path("/kube-system/healthz")
+              && SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128")
+              -> disableAccessLog()
+              -> status(200)
+              -> <shunt>;
```
AlexanderYastrebov, line 335: This is the same as [here](https://github.com/zalando/skipper/blob/128b43767fe91d691551873aa6fc07e6fba6cb1f/dataclients/kubernetes/kube.go#L476-L477) but disables access log completely.

AlexanderYastrebov:
Pipeline consistently fails

AlexanderYastrebov:
> Pipeline consistently fails

Could be due to https://github.com/zalando-incubator/kubernetes-on-aws/pull/9203

cluster/manifests/skipper/deployment.yaml
```diff
@@ -320,9 +320,21 @@ spec:
...
           - '-forwarded-headers-exclude-cidrs={{ .cluster_internal_cidrs | join "," }}'
 {{ end }}
-{{ if .Cluster.ConfigItems.skipper_ingress_inline_routes }}
-          - "-inline-routes={{ .Cluster.ConfigItems.skipper_ingress_inline_routes }}"
-{{ end }}
+          - "-inline-routes"
+          - |
+            kube__healthz_down: Path("/kube-system/healthz")
+              && Shutdown()
+              && SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128")
```
MustafaSaber, line 327: I think for EKS clusters we are missing WhitelistedIPs
https://github.com/zalando-incubator/kubernetes-on-aws/blob/18cb6e0afd46c6244d06ea79fd98fd81a0396b3f/cluster/manifests/skipper/deployment.yaml#L556

in skipper we append those IPs to a global value 

https://github.com/zalando/skipper/blob/15385fae75b3e384d6bed671266d4a8906e055fa/dataclients/kubernetes/kube.go#L309

Maybe that's why EKS clusters are failing

MustafaSaber: What I mean we need to append this IP in the route if it's EKS & IP family is IPv6

AlexanderYastrebov:
> Pipeline consistently fails

The problem was due to simultaneous change to both skipper and routesrv - new routesrv configuration did not serve healthcheck routes which resulted in unhealthy old skipper pods that picked up routes update without healthchecks.
The solution is to deploy configuration change in two steps.

Closing in favour of #9206 + #9241

cluster/manifests/skipper/deployment.yaml
MustafaSaber: This was done in https://github.com/zalando-incubator/kubernetes-on-aws/pull/9206

MustafaSaber:
> > Pipeline consistently fails
> 
> 
> 
> The problem was due to simultaneous change to both skipper and routesrv - new routesrv configuration did not serve healthcheck routes which resulted in unhealthy old skipper pods that picked up routes update without healthchecks.
> 
> The solution is to deploy configuration change in two steps.


And for EKS cluster creation failure was due to not whitelisting the IPs for IPv6.

PR #8914, opened 2025-02-17
kube-node-decommissioner: fix image

mikkeloscar:
Follow up to #8891 

Fixes two issues:

* Built image was wrong so the cronjob didn't succeed in pulling and running
* Drop `--debug` flag which is not implemented (was copied from another cronjob).


mikkeloscar:
:+1:

demonCoder95:
👍

PR #8915, opened 2025-02-17
Move registerWithTaint to userdata.yaml

mikkeloscar:
This PR updates the AMI with the following changes:

* Moves the node taint settings from being passed to kubelet via a flag `--register-with-taints` to be set in the kubelet config file. This makes it simpler to template the taints without having to modify the AMI.
* Drops the legacy flannel config (legacy/unused path)
* Removes dependency on registry.k8s.io for `pause` image
* Speeds up boot of instances with local storage

The changes are implemented behind a config-item `kuberuntu_ami_version: new|old` which controls which AMI and which userdata is used. The config-item is set to `new` by default which ensures the changes are tested during e2e.

The roll out strategy is to set: `kuberuntu_ami_version: old` for all clusters and then enable it controlled e.g. per channel. This allows running the change in each channel for a longer time without blocking channel merges of other changes. Additionally it avoid complex temporary channel flow as we usually do with Kubernetes version upgrades.

I have tested in a pet cluster that with `kuberuntu_ami_version: old` there is no change to how the userdata is rendered compared to current `dev`.

### TODO

* [x] Update AMI to a version that doesn't set the flag.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -756,12 +756,14 @@ tracing_coredns_local_zone_traces_endpoint: ""
 # AMI id given the image name and the Image AWS account owner.
 #
 # [0]: https://github.com/zalando-incubator/cluster-lifecycle-manager/blob/8a9bd1cb2d094038a9e23e646421f8146b48886a/provisioner/template.go#L116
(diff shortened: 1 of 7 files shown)


demonCoder95:
👍

mikkeloscar:
:+1:

PR #8929, opened 2025-02-18
Fix permissions for kube-node-decommissioner

linki:
Fixes permission issues in `kube-node-decommissioner`.

```
2025/02/18 10:22:01 failed to retire nodes: failed to map nodes to pods: pods is forbidden: User "system:serviceaccount:kube-system:kube-node-decommissioner" cannot list resource "pods" in API group "" at  the cluster scope: access undecided system:serviceaccount:kube-system:kube-node-decommissioner/[system:serviceaccounts system:serviceaccounts:kube-system system:authenticated]
```

[edit] also changed nodes `patch` to `update` to match the function call in the code.

### DIFF
--- a/cluster/manifests/kube-node-decommissioner/01-rbac.yaml
+++ b/cluster/manifests/kube-node-decommissioner/01-rbac.yaml
@@ -17,8 +17,11 @@ metadata:
     component: kube-node-decommissioner
 rules:
 - apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["list"]
+- apiGroups: [""]
   resources: ["nodes"]
-  verbs: ["list", "patch"]
+  verbs: ["list", "update"]
 ---
 # This role binding allows service-account "kube-node-decommissioner" to
 # list and patch nodes.


linki:
👍

cluster/manifests/kube-node-decommissioner/01-rbac.yaml
```diff
@@ -16,6 +16,9 @@ metadata:
     application: kubernetes
     component: kube-node-decommissioner
 rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["list"]
 - apiGroups: [""]
   resources: ["nodes"]
   verbs: ["list", "patch"]
```
mikkeloscar: We likely need `update` instead of `patch` here :facepalm:  can you also add `update` just to be on the safe side.

linki: Changed!

mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #8930, opened 2025-02-18
Enable postgresql owning-application annotation check in prod

hughcapet:
Block production CREATEs and UPDATEs of a postgresql resource without `zalando.org/owning-application `annotation specified in the manifest or when its value is a non-existent/inactive application.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -605,6 +605,7 @@ teapot_admission_controller_validate_pod_template_resources: "true"
 teapot_admission_controller_preemption_enabled: "true"
 teapot_admission_controller_postgresql_delete_protection_enabled: "true"
 teapot_admission_controller_namespace_delete_protection_enabled: "true"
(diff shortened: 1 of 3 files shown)


cluster/config-defaults.yaml
```diff
@@ -616,6 +617,7 @@ teapot_admission_controller_validate_pod_template_resources: "false"
 teapot_admission_controller_preemption_enabled: "true"
 teapot_admission_controller_postgresql_delete_protection_enabled: "false"
 teapot_admission_controller_namespace_delete_protection_enabled: "false"
+teapot_admission_controller_postgresql_owning_application_check_enabled: "false"
```
mikkeloscar: We don't run postgresql in e2e so we don't need to explicitly disable it here.

mikkeloscar:
The base branch was deleted, can you open again pointing to dev?

hughcapet:
uh, I thought it will automatically change target:(

hughcapet:
https://github.com/zalando-incubator/kubernetes-on-aws/pull/8932

PR #8934, opened 2025-02-18
Enable postgresql owning-application annotation check in prod

hughcapet:
And yet
Block production CREATEs and UPDATEs of a postgresql resource without zalando.org/owning-application annotation specified in the manifest or when its value is a non-existent/inactive application.


mikkeloscar:
:+1:

hughcapet:
👍

mikkeloscar:
:+1:

hughcapet:
👍

PR #8939, opened 2025-02-19
Increase resources for kube-node-decommissioner and make configurable

mikkeloscar:
Increase resources for `kube-node-decommissioner` cronjob and make them configurable.

The cronjob is crashing in big clusters hence the need to increase the resources.


linki:
👍

mikkeloscar:
:+1:

PR #8943, opened 2025-02-20
Optional support for AWS Fault Injection Service with EKS

mikkeloscar:
This implements the foundation for integrating AWS Fault Injection Service (FIS) with EKS

It basically creates an IAM role to be used in FIS which has access to create a FIS executor pod in the target namespace.

image

This is the very first version of adding support, there are several complexities not yet solved, however having this in the EKS branch makes it simpler to enable and test in clusters.

https://docs.aws.amazon.com/fis/latest/userguide/eks-pod-actions.html

https://aws.amazon.com/blogs/containers/aws-fault-injection-simulator-supports-chaos-engineering-experiments-on-amazon-eks-pods/

### DIFF
--- a/cluster/manifests/eks-fis/01-rbac.yaml
+++ b/cluster/manifests/eks-fis/01-rbac.yaml
@@ -1,0 +1,48 @@
+{{- if eq .Cluster.ConfigItems.eks_fis_support_enabled "true" }}
+# {{ range $namespace := split .Cluster.ConfigItems.eks_fis_namespaces "," }}
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  namespace: "{{ $namespace }}"
+  name: fis-experiment-executor
+---
+# {{ end }}
+kind: ClusterRole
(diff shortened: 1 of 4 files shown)


cluster/manifests/eks-fis/01-rbac.yaml
```diff
@@ -0,0 +1,48 @@
...
+  name: fis-experiment-executor
+subjects:
+# {{ range $namespace := split .Cluster.ConfigItems.eks_fis_namespaces "," }}
+- kind: ServiceAccount
+  name: fis-experiment-executor
+  namespace: "{{ $namespace }}"
+# {{ end }}
+- apiGroup: rbac.authorization.k8s.io
+  kind: User
+  name: fis-experiment-executor
```
linki, line 43: Is this the `User` that's referenced in the AccessEntry?

mikkeloscar: Yes

mikkeloscar:
:+1:

linki:
👍

PR #8901, opened 2025-02-12
Pre karpenter v1.1.1 changes [2/x]

mikkeloscar:
* [x] This depends on full roll out of #8900
  * [x] dev
  * [x] alpha
  * [x] beta
  * [x] stable

This prepares for updating to Karpenter v1.1.x by migrating the CRDs from v1beta1 to v1.

The biggest change is the move of `kubelet` config field from `NodePool` to `EC2NodeClass`.

This essentially implements the upgrade requirements outlined in https://karpenter.sh/docs/upgrading/upgrade-guide/#upgrading-to-110

PR #8954, opened 2025-02-21
Fix karpenter CRDs for v1.1.1 upgrade

mikkeloscar:
Follow up to #8901

Fixes issues with creating the ec2nodeclass in new clusters (issue didn't happen when migration from existing v1beta1 version of the resource).

* Set `consolidateAfter` to 0s which is the same behavior as with v1beta1, but allow configuring this via `consolidate_after`.
* Add `group` and `kind` to nodeClassRef


mikkeloscar:
:+1:

linki:
👍

PR #8974, opened 2025-02-25
Eks register with taints kubelet config

mikkeloscar:
Same as #8915 but for EKS branch


mikkeloscar:
Closing in favor of making #8915 config-item controlled.

PR #8981, opened 2025-02-25
use alias-independent variable for the hosted zone

mikkeloscar:
This ports a commit from the `eks` branch to dev so it can be prepared before the `eks` branch is merged.

The motivation is to add a hostname for the cluster-health probe which is unique when there are multiple clusters in a single account.


mikkeloscar:
:+1:

linki:
👍

PR #9006, opened 2025-02-28
Slow down coredns shutdown

mikkeloscar:
Follow up to #8963

Slow down coreDNS daemonset shutdown to 40 seconds. The motivation for this is that daemonsets have a default terminationGracePeriod of 30 seconds and e.g. `logging-agent` ships logs before it shuts down. If coredns is stopping before then last log shipping can fail.


mikkeloscar:
:+1:

mikkeloscar:
:+1:

katyanna:
👍

PR #9033, opened 2025-03-05
Avoid problem of anchor-merging lists by using multiple envFroms

linki:
We cannot anchor-merge two lists of `env`s which led to some duplication. By switching to multiple `envFrom`s and letting Kubernetes take care of the merging, we can remove the duplication again.


mikkeloscar:
:+1:

linki:
👍

PR #9037, opened 2025-03-05
Update kube-node-decommissioner CronJob concurrencyPolicy to Replace

zaklawrencea:
Update the `kube-node-decommissioner` CronJob to `concurrencyPolicy: Replace`

> Replace: If it is time for a new Job run and the previous Job run hasn't finished yet, the CronJob replaces the currently running Job run with a new Job run

https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#concurrency-policy

We have seen the pod landing on a node and becoming stuck in `Terminating` state itself. With this change in policy, a new pod should run and allow the `kube-node-decommissioner` to continue to work.


zaklawrencea:
:+1:

demonCoder95:
👍

PR #9049, opened 2025-03-06
external-dns: Drop --debug flag

mikkeloscar:
Was introduced with #7808 but is not really needed.

Also drop TODO that is not valid anymore


mikkeloscar:
:+1:

linki:
👍

PR #9052, opened 2025-03-06
Test the upgrade path of EKS clusters in e2e tests

linki:
Similar to legacy clusters, test the upgrade path as well, now that it's possible.


linki:
👍

mikkeloscar:
👍

PR #9053, opened 2025-03-07
add startupProbe to metrics-server Pod to protect against race condition

demonCoder95:
We have a race condition when a new metrics-server Pod comes up, between the time that the server updates the node informer cache and it scrapes the nodes in the cache, Karpenter can terminate many nodes at the same time. This leads to the scraper stuck in a loop attempting to connect to the kubelet on terminated/terminating nodes and not progressing from NotReady state for up-to 30mins and we get paged for a kube-system Pod not running. There is no way to configure the `max-retries` to a kubelet, so there is no straight-forward way to configure this in the metrics-server itself.

```go
E0303 15:51:02.423322       1 scraper.go:149] "Failed to scrape node" err="Get \"https://ip-10-151-1-173.eu-central-1.compute.internal:10250/metrics/resource\": dial tcp 10.151.1.173:10250: connect: connection refused" node="ip-10-151-1-173.eu-central-1.compute.internal"
```

This introduces a `startupProbe` on the container where we can have the kubelet kill and restart the container after `2mins` if it's stuck like this, and the restarted Pod should have an updated node informer cache, and proceed to `Ready` state.

### DIFF
--- a/cluster/manifests/metrics-server/deployment.yaml
+++ b/cluster/manifests/metrics-server/deployment.yaml
@@ -52,6 +52,17 @@ spec:
             port: https
             scheme: HTTPS
           periodSeconds: 10
+        # This is needed to fix the issue where when a new metrics-server Pod comes up at
+        # the same time when many nodes are Terminated, it will get stuck and not be marked
+        # Ready for a long time. This provides a failsafe where the kubelet can kill the Pod
+        # and it can be restarted.
+        startupProbe:
+          periodSeconds: 10
+          failureThreshold: 12 #120s = 2min startup timeout
+          httpGet:
+            path: /healthz
+            port: https
+            scheme: HTTPS
         securityContext:
           readOnlyRootFilesystem: true
           runAsNonRoot: true


mikkeloscar:
Can you point to an upstream issue or the relevant code in metrics-server for this issue? It sounds strange that it can only happen at startup and not in general 🤔

demonCoder95:
Okay, after looking at the code in detail, I think this doesn't solve the problem.

What's apparently happening is:

1. The metrics-server starts, it sets up informers and start the scraping loop.
2. The loop runs once every `scrape-interval` which is `15s`.
3. In each scrape loop, nodes are listed. For each node, a goroutine is started that fetches metrics from the `<node>:10250/metrics` endpoint with a timeout of `10s`.
4. Once all goroutines are started, another loop reads the response channel to collect all the metrics reported by the scraping goroutines, and stores them.

I realised that restarting the Pod actually won't fix anything, because it wasn't an outdated informer cache, but Terminating nodes that are the problem. Looking carefully at the karpenter logs, it took Karpenter ~4mins to go from tainting a node it wishes to terminate, and the node completely being removed from the cluster. In these 4mins, we see the metrics server fail the readinessProbe (3*10 = 30s). This causes the Pod to be marked `NotReady` until the node is completely gone, so probably `3.5mins` the Pod will be marked `NotReady`. Once the node is fully gone, the informer cache will remove it, the scraper won't scrape it anymore and we don't have the issue.

What apparently happens, is when many nodes terminate in a short period of time this `NotReady` period keeps increasing, this leads to the Pod staying `NotReady` long enough for our alerting to pick it up and alert us. So, if we chain enough node terminations in a cluster, we end up prolonging this `NotReady` status by a really long time, up to an hour say, or more.

I think the best way to fix this would be for the scraper to somehow "ignore" nodes that are Terminating? Or implement a `max-retires` config, that in combination with the `scrape-timeout` determines when to "give up" on a node, so we can avoid this runaway kind of a situation.

I don't see any issues in the upstream repo about this, probably because this issue is only most apparent when you have:
1. A lot of nodes running, from them multiple nodes can be terminated in a short period of time.
2. Have a readinessProbe setup on the metrics-server.
3. Have alerting set to trigger if the Pod is not ready for more than `8mins`.

demonCoder95:
As discussed, I'll close this PR and move this to an internal issue.

PR #9059, opened 2025-03-07
Allow collaborators write access in collaborator namespaces

linki:
Collaborators should have access in a subset of the (currently marked as) privileged namespaces. This includes visibility and kubenurse based on the old configuration.

### DIFF
--- a/cluster/manifests/02-admission-control/teapot.yaml
+++ b/cluster/manifests/02-admission-control/teapot.yaml
@@ -538,7 +538,7 @@ webhooks:
       matchExpressions:
         - key: kubernetes.io/metadata.name
           operator: In
-          values: [ "kube-system", "visibility", "kubenurse" ]
+          values: [ "kube-system", "kubenurse" ]
     rules:
       - operations: [ "*" ]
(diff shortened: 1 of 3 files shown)


cluster/manifests/02-admission-control/teapot.yaml
```diff
@@ -552,6 +552,39 @@ webhooks:
...
+      caBundle: "{{ .Cluster.ConfigItems.ca_cert_decompressed }}"
+    admissionReviewVersions: ["v1beta1"]
+    failurePolicy: Fail
+    sideEffects: "NoneOnDryRun"
+    matchPolicy: Equivalent
+    namespaceSelector:
+      matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: In
+          values: [ "kubenurse", "visibility" ]
```
mikkeloscar: Not sure if we define `kubenurse` as collaborator namespace in other places. But there is no need for it to be one. It should be equivalent to kube-system in terms of who has access.

mikkeloscar:
:+1:

linki:
👍

PR #9060, opened 2025-03-07
Enable network policies by default for EKS

linki:
Enable the NetworkPolicy controller in EKS by default. This will also enable it for all e2e test runs.


linki:
👍

katyanna:
Let's talk about safe guards before enabling it by default.

linki:
Deprioritized for now.

PR #9071, opened 2025-03-11
Disable resource protection via admission-controller in legacy clusters

linki:
Enable the new resource protection via admission-controller only in EKS clusters where it's actually needed.

This is to unblock the channel propagation here: https://github.com/zalando-incubator/kubernetes-on-aws/pull/9056

Especially this rule has me concerned: https://github.com/zalando-incubator/kubernetes-on-aws/blob/097abfad7096d9345ece1933507813c8920026ef/cluster/manifests/02-admission-control/teapot.yaml#L493-L521

It would block exec calls in user namespaces in production clusters, which is fine. But it would also block in test and cannot be overridden by manual access as far as I can tell. Let's unblock the rollout while figuring this out.


linki:
👍

zaklawrencea:
:+1:

PR #9078, opened 2025-03-12
Update Karpenter to v1.3.0 (this should fail)

linki:
Starting to work on https://github.bus.zalan.do/teapot/issues/issues/3791

This attempts to reproduce the issue we saw when updating the karpenter version 1.3.0. First with some green tests and then by introducing the broken 1.3.0 version to see if we can catch the error in e2e. 1.3.1 should then result in green tests again.


linki:
For EKS I can reproduce the issue because it doesn't go the update path (at the moment) and gets the broken karpenter version right from the start.

For legacy clusters, karpenter gets updated, prints the same AWS API error but still schedules the pods 🤷

PR #9085, opened 2025-03-13
Extend test case to ensure that node pools can scale-out after cluster update

linki:
For https://github.bus.zalan.do/teapot/issues/issues/3791

Adds a pod anti-affinity to the `reserve-pools` and scales them out as part of the e2e suite. This keeps the original behaviour and ensures that all node pools can be scheduled on (pods with replicas 1). After cluster update, during the e2e run it scales out all reserve pools to 2 which requires a new node for each node pool. This then tests that scaling-out also works in the updated cluster.

### DIFF
--- a/test/e2e/infra.go
+++ b/test/e2e/infra.go
@@ -11,6 +11,7 @@ import (
 	admissionapi "k8s.io/pod-security-admission/api"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	applyconfigurationsautoscalingv1 "k8s.io/client-go/applyconfigurations/autoscaling/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
 	"k8s.io/kubernetes/test/e2e/framework/deployment"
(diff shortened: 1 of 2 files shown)


linki:
👍

linki:
👍

test/e2e/infra.go
```diff
@@ -35,7 +36,34 @@ var _ = describe("Infrastructure tests", func() {
...
+			err = deployment.WaitForDeploymentComplete(cs, deploy)
+			framework.ExpectNoError(err)
+
+			// Scale out deployment to one more replica. In combination with Pod-Anti-Affinity, this should require one more node.
+			_, err = cs.AppsV1().Deployments("default").ApplyScale(
+				context.Background(),
+				fmt.Sprintf("pool-reserve-%s", pool),
+				applyconfigurationsautoscalingv1.Scale().WithSpec(applyconfigurationsautoscalingv1.ScaleSpec().WithReplicas(2)),
+				metav1.ApplyOptions{FieldManager: "e2e.test", Force: true},
+			)
```
mikkeloscar, line 62: Does this correctly wait for the pod to be scheduled to verify it got a node? :thinking:

linki: Sort of. It applies the seven Deployments sequentially and then waits for all seven in the next loop. By the time it enters the next loop, the first deployment in the list already has a pending pod.

But I can also add a small time delay to be sure. (Would skip running the tests though to speed up the merge)

linki: To clarify: This call doesn't wait. The wait happens in the next loop below.

mikkeloscar: I guess I was/am unsure if `err = deployment.WaitForDeploymentComplete(cs, deploy)` waits for deployment to have 2 ready pods.

If it does that, then all good.

linki: Yes, I can confirm that it does that.

mikkeloscar:
👍

PR #9092, opened 2025-03-14
eks: Break circular dependency on bootstrap

mikkeloscar:
When bootstrapping a production EKS cluster we have the following circular dependency

image

Which prevents the majority of pods from starting because the check for production ready images in admission-controller can not execute.

This PR breaks the circular dependency in the following way

1. Exclude components needed for setting up the `<cluster-alias>.<account>.zalan.do` endpoint from admission-controller
   * `zalando-iam-aws-proxy`
   * `external-dns`
   * `kube-ingress-aws-controller`
2. Use the resolved ECR image url for those components as they can't be transformed from the vanity URL when not going through admission-controller
3. Allow those components to run on the cluster-seed nodes such that they can run e.g. before karpenter is running.
4. Exclude `kube-system` from skipper validating webhook ingress/RouteGroup validation.

### DIFF
--- a/cluster/manifests/02-admission-control/teapot.yaml
+++ b/cluster/manifests/02-admission-control/teapot.yaml
@@ -13,13 +13,21 @@ webhooks:
       matchExpressions:
         - key: component
           operator: NotIn
-          values: ["admission-controller", "coredns", "aws-node"]
+          values:
+            - "admission-controller"
+            - "coredns"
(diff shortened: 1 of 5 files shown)


linki:
👍

linki (approved):

mikkeloscar:
:+1:

PR #9093, opened 2025-03-15
eks: Fix disabling control plane logging

mikkeloscar:
This fixes turning off `eks_control_plane_logging` after it has been enabled.

Without this the CF stack update fails because it's not allowed to remove the whole `Logging` section after it has been defined in the template once.


mikkeloscar:
:+1:

mikkeloscar:
:+1:

demonCoder95:
👍

PR #9094, opened 2025-03-17
eks: Make unbound work in ipv4 and ipv6 clusters

mikkeloscar:
This adds the needed changes to make unbound listen on IPv6 and work in IPv6 EKS clusters.

Note, the communication between unbound and coreDNS is via IPv4. SInce this works fine and is transparent to the clients there is no reason to change this (This works because the coredns pod uses hostnetwork).


mikkeloscar:
:+1:

demonCoder95:
👍

PR #9095, opened 2025-03-17
Switch to unbound_exporter for metrics

mikkeloscar:
* depends on #9094

This introduces `unbound_exporter`: https://github.com/letsencrypt/unbound_exporter as metric exporter for unbound instead of unbound-telemetry which has been deprecated: https://github.com/svartalf/unbound-telemetry

`unbound_exporter` is one to one compatible with for the metrics names, at least those we use in the existing metric check.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #9101, opened 2025-03-17
Drop karpenter taint injection via admission-controller

mikkeloscar:
Drops karpenter taint injeciton via admission-controller as this is now done via kubelet: #8915


mikkeloscar:
:+1:

katyanna:
👍

PR #9111, opened 2025-03-18
fix: connector job and broker deployment

vinaythupili:
- This will enable wiz-connector-connector to be created when wiz_enable_runtime_connector set to true and will avoid re applying via CLM 

Fixes: 
- Crashloopback of broker pod when wiz_enable_runtime_connector set to false 
- Duplicate broker connections by re applying the connector job every time CLM runs as wiz_enable_runtime_connector set to true


linki:
👍

zaklawrencea:
:+1:

PR #9118, opened 2025-03-19
Apply RouteGroup CRD after admission-control

mikkeloscar:
Apply RouteGroup CRD after admission-control to ensure the crd admitter is installed to create poweruser and readonly permissions based on the CRD.

Without this powerusers/CDP will not get permissions to RouteGroups.


mikkeloscar:
:+1:

linki:
👍

PR #9119, opened 2025-03-20
Mainly use karpenter pools for e2e

mikkeloscar:
Reduce number of ASG based node pools as we constantly hit rate limits on the ASG API in the e2e account

* Replace `default-worker-splitaz` by a `default-karpenter` pool. `worker-splitaz` profile pool is tested by skipper-ingress indirectly.
* Drop dedicated `worker-instance-storage` pool and just rely on karpenter spinning up a right node to satisfy a pod requesting 500Gi ephemeralStorage.
* Drop dedicated `worker-combined` pool. Master pool and EKS Seed pool are both `worker-combined` and we only use this profile in that context, so should be enough to test as part of e2e.
* Switch GPU pools to be karpenter based
* Drop dedicated `worker-karpenter` pool and rely on `default-karpenter` for testing basic karpenter functionality
* Replaced dedicated `worker-arm64` pool with a karpenter-arm pool that can spin up any arm instance based on `kubernetes.io/arch: arm64` nodeSelector.
* Drop the custom setting: `ebs_root_volume_size: "550"` as karpenter can spin up a node with enough ephemeral storage.

### DIFF
--- a/cluster/manifests/e2e-resources/pool-reserve.yaml
+++ b/cluster/manifests/e2e-resources/pool-reserve.yaml
@@ -1,5 +1,5 @@
 {{ if eq .Cluster.Environment "e2e" }}
-{{ range $pool := split "default-worker-splitaz,worker-limit-az,worker-combined,worker-instance-storage,worker-node-tests,worker-karpenter,worker-arm64" "," }}
+{{ range $pool := split "worker-limit-az,worker-instance-storage,worker-node-tests,worker-arm64" "," }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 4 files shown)


cluster/manifests/e2e-resources/pool-reserve.yaml
```diff
@@ -1,5 +1,5 @@
 {{ if eq .Cluster.Environment "e2e" }}
-{{ range $pool := split "default-worker-splitaz,worker-limit-az,worker-combined,worker-instance-storage,worker-node-tests,worker-karpenter,worker-arm64" "," }}
+{{ range $pool := split "worker-limit-az,worker-instance-storage,worker-node-tests,worker-arm64" "," }}
```
linki, line 2: I think we need to shorten this list as well: https://github.com/zalando-incubator/kubernetes-on-aws/pull/9120/files#diff-b5504898e28d320aa7ff060453dc48970aa8c4995a1350651150bc65cdabf4e0R39-R46

Because the Deployments that are checked there won't exist.

linki: Fixed

linki:
👍

mikkeloscar:
:+1:

PR #9123, opened 2025-03-20
Disable egress controller in EKS by default

mikkeloscar:
The egress controller is not designed to run multiple times per AWS region as it only makes sense to have a single NAT setup per VPC.

Disable the controller by default in EKS clusters such that it doesn't run in the second cluster during migration phase.

This is a simple way to reduce risk, we need to think about how to make the migration work in respect to egress-controller.


linki:
👍

mikkeloscar:
:+1:

PR #9127, opened 2025-03-20
drop karpenter_pools_enabled config-item

demonCoder95:
As Karpenter has been running in production for a while and we don't require the ability to rollback, this cleans up the config-item that enables Karpenter.

PR #9131, opened 2025-03-21
set karpenter_pools_enabled to true by default

demonCoder95:
This is required as a pre-condition to dropping the config-item altogether in #9127 . We have this enabled everywhere already, so we can set it `true` now.


demonCoder95:
👍

linki:
👍

PR #9137, opened 2025-03-24
Fix indentation to make template valid

mikkeloscar:
Fix indentation to make template valid.

Without this fix we get the error:

```
error remarshaling manifest main/cluster/manifests/deployment-service/controller-rbac.yaml: yaml: line 208: did not find expected key
```

When `deployment_service_executor_cdp_permissions: "true"`


mikkeloscar:
:+1:

linki:
👍

PR #9143, opened 2025-03-25
Config Item to require instance types with in-transit encryption support

mikkeloscar:
This adds a config-item: `karpenter_in_transit_support_required` which, when set set a requirement on Karpenter node pools to restrict to instances with in-transit encryption support

The idea is that this can be set in clusters where this is a requirement.

ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/data-protection.html#encryption-transit

### DIFF
--- a/cluster/node-pools/worker-karpenter/provisioners.yaml
+++ b/cluster/node-pools/worker-karpenter/provisioners.yaml
@@ -158,7 +158,7 @@ spec:
             - nitro
 #{{ end }}
 #{{ end }}
-#{{ if eq .NodePool.ConfigItems.karpenter_in_transit_support_required "true") }}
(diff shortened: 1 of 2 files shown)


cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -150,6 +150,10 @@ spec:
           values:
             - nitro
 #{{ end }}
+#{{ if eq .NodePool.ConfigItems.karpenter_in_transit_support_required "true") }}
```
linki:
```suggestion
#{{ if eq .NodePool.ConfigItems.karpenter_in_transit_support_required "true" }}
```

mikkeloscar:
Tested in pet cluster!

mikkeloscar:
:+1:

linki:
👍

PR #9146, opened 2025-03-25
add nitro requirement for Karpenter pools

demonCoder95:
For Karpenter node-pools that use the instance type strategy of `not-specified`, we want to restrict the usage of Nitro hypervisor as this will be required for all nodes when running EKS with IPv6.

This adds this behind a toggle so it's easier to rollback in case we see issues when rolling this out.  Once the rollout is complete, the toggle will be dropped.

This is only relevant for clusters that have the scheduling controls enabled, since only in those clusters we have Karpenter based node-pools with `instance-types: not-specified`.

### DIFF
--- a/cluster/node-pools/worker-karpenter/provisioners.yaml
+++ b/cluster/node-pools/worker-karpenter/provisioners.yaml
@@ -150,6 +150,14 @@ spec:
           values:
             - nitro
 #{{ end }}
+#{{ if eq .Cluster.ConfigItems.karpenter_nitro_support_required "true" }}
+#{{ if and (eq .Cluster.Provider "zalando-aws") (eq .NodePool.KarpenterInstanceTypeStrategy "not-specified") }}
+        - key: karpenter.k8s.aws/instance-hypervisor
+          operator: in
+          values:
+            - nitro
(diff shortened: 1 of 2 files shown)


cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -150,6 +150,14 @@ spec:
           values:
             - nitro
 #{{ end }}
+#{{ if eq .Cluster.ConfigItems.karpenter_nitro_support_required "true" }}
+#{{ if and (eq .Cluster.Provider "zalando-aws") (eq .NodePool.KarpenterInstanceTypeStrategy "not-specified") }}
```
linki, line 154: > when running EKS with IPv6

This line excludes EKS. Why?

demonCoder95: For `eks` we already have the lines right above this one, where we have enabled this already. So, my thinking is, once we move to EKS, we can simply drop this section altogether, and keep the one above it.

linki:
👍

demonCoder95:
👍

PR #9106, opened 2025-03-18
Give api-monitoring-controller access to kube-system configmap

mikkeloscar:
The api-monitoring-controller needs access to write to a specific configmap in `kube-system`. Exclude it from namespaced-deny admission-controller to allow write access.

The errors from the controller illustrate the issue:

```
%!q(*model.ErrorSummary=&{FATAL failed to update config map k8s gateway [resource=replace configmaps, status-code=400, message={"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"admission webhook \"namespaced-deny-admitter.teapot.zalan.do\" denied the request: write operations are forbidden","code":400}] 1 []})] interval="30s" status="FAILED"

```

PR #6497, opened 2023-09-20
roles: allow ReadOnly group to read routesrv routes

AlexanderYastrebov:
* commented template actions to make file a valid yaml
* reformatted yaml

See previous https://github.com/zalando-incubator/kubernetes-on-aws/pull/4584

PR #9150, opened 2025-03-26
admission-control: allow routesrv proxy

AlexanderYastrebov:
Allow engineers to read routes from routesrv service proxy.
The access level should be the same as for listing Ingresses and RouteGroups (because one can technically generate the same by running routesrv locally against api server)

See also
* #6497
* #9106

### DIFF
--- a/cluster/manifests/02-admission-control/teapot.yaml
+++ b/cluster/manifests/02-admission-control/teapot.yaml
@@ -575,6 +575,16 @@ webhooks:
             object.kind == "ConfigMap" &&
             object.metadata.name == "skipper-default-filters"
           )
+      - name: 'allow-routesrv-routes-access'
+        expression: |
+          !(
+            "okta:common/engineer" in request.userInfo.groups &&
+            request.name == "skipper-ingress-routesrv" &&
+            request.resource.resource == "services" &&
+            request.subResource == "proxy" &&
+            request.operation == "CONNECT"
+          )
+
   - name: collaborator-deny-admitter.teapot.zalan.do
     clientConfig:
       {{- if eq .Cluster.Provider "zalando-eks"}}


cluster/manifests/02-admission-control/teapot.yaml
```diff
@@ -575,6 +575,16 @@ webhooks:
...
           )
+      - name: 'allow-routesrv-routes-access'
+        expression: |
+          !(
+            "okta:common/engineer" in request.userInfo.groups &&
+            request.name == "skipper-ingress-routesrv" &&
+            request.resource.resource == "services" &&
+            request.subResource == "proxy" &&
+            request.operation == "CONNECT"
+          )
```
AlexanderYastrebov, line 586: This was tested to allow getting routes from routesrv but not connect to other services.

cluster/manifests/02-admission-control/teapot.yaml
```diff
@@ -575,6 +575,16 @@ webhooks:
             object.kind == "ConfigMap" &&
             object.metadata.name == "skipper-default-filters"
           )
+      - name: 'allow-routesrv-routes-access'
+        expression: |
+          !(
+            "okta:common/engineer" in request.userInfo.groups &&
```
AlexanderYastrebov, line 581: Should it be `okta:common/read-only` instead?

linki: Let's ask @mikkeloscar about that 👆 

For now, let's keep it as engineer. This way it doesn't allow any new permission because I think you could just do a `zkubectl tunnel` and curl the `routesrv` service to do the same.

mikkeloscar: I agree with engineer to keep it like it is today.

linki:
👍

mikkeloscar:
👍

AlexanderYastrebov:
:+1:

PR #9156, opened 2025-03-28
Switch deployment targets from `stups-test` to `stups-euc1-test`

linki:
Runs e2e tests on `stups-euc1-test` instead of `stups-test`.


linki:
👍

demonCoder95:
👍

PR #9168, opened 2025-04-02
fix requirement for encryption in-transit

demonCoder95:
This was added in https://github.com/zalando-incubator/kubernetes-on-aws/pull/9143. The label exists on all Karpenter nodes, so we need to check for a value of `true` instead of checking if it exists.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #9170, opened 2025-04-02
Configure response header timeouts for styra calls

wisinghe:
Configure response header timeouts for styra API calls

This configuration will make sure any outgoing calls to styra will timeout after 2 seconds. The default timeout is 10 seconds. Our Styra endpoints has p999 below 1s and it makes sense the calls do not have to wait longer in case there's a service degradation.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -378,6 +378,9 @@ skipper_ingress_routegroup_crd_require_hosts: "true"
 skipper_open_policy_agent_enabled: "false"
 skipper_open_policy_agent_styra_token: ""
 
+# Default timeout value in seconds for outgoing http calls from Open Policy Agent in a skipper filter
+skipper_open_policy_agent_styra_response_header_timeout: 2
+
 #
 # FabricGateway controller config
 #
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -379,7 +379,7 @@ skipper_open_policy_agent_enabled: "false"
 skipper_open_policy_agent_styra_token: ""
 
 # Default timeout value in seconds for outgoing http calls from Open Policy Agent in a skipper filter
-skipper_open_policy_agent_styra_response_header_timeout: 2
+skipper_open_policy_agent_styra_response_header_timeout: "2"
 
 #
 # FabricGateway controller config
--- a/cluster/manifests/skipper/configmap-open-policy-agent.yaml
+++ b/cluster/manifests/skipper/configmap-open-policy-agent.yaml
@@ -23,7 +23,7 @@ data:
           token: "{{ `{{ .Env.STYRA_TOKEN }}` }}"
       name: styra
       url: "{{ .Cluster.ConfigItems.skipper_open_policy_agent_observability_url }}"
-      response_header_timeout_seconds: 2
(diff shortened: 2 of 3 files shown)


cluster/manifests/skipper/configmap-open-policy-agent.yaml
```diff
@@ -23,6 +23,7 @@ data:
           token: "{{ `{{ .Env.STYRA_TOKEN }}` }}"
       name: styra
       url: "{{ .Cluster.ConfigItems.skipper_open_policy_agent_observability_url }}"
+      response_header_timeout_seconds: 2
```
MustafaSaber: why not 1 then?

mikkeloscar: Let's make it configurable via a config-item so it's possible to adjust per cluster if needed.

wisinghe: There can be situations that Styra is a bit slow but still able to handle requests. And Opa initialization will not be impacted by 2 seconds waiting time. It felt like 2 seconds is good value which will give a bit more chance for Opa's to report when Styra is slightly slow, but not wait too long.

wisinghe: Raised a PR here https://github.bus.zalan.do/teapot/kubernetes-on-aws-internal/pull/301

MustafaSaber: I think it can be done here in this PR

mikkeloscar: I agree with @MustafaSaber this is a config we can just add to this repo. Doesn't have to be in the internal one.

cluster/config-defaults.yaml
```diff
@@ -378,6 +378,9 @@ skipper_ingress_routegroup_crd_require_hosts: "true"
 skipper_open_policy_agent_enabled: "false"
 skipper_open_policy_agent_styra_token: ""
 
+# Default timeout value in seconds for outgoing http calls from Open Policy Agent in a skipper filter
+skipper_open_policy_agent_styra_response_header_timeout: 2
```
mikkeloscar: All values should be string, please define as `"2"`.

test/e2e/cluster_config.sh
```diff
@@ -52,6 +52,7 @@ clusters:
     skipper_open_policy_agent_bucket_arn: "${SKIPPER_OPA_BUCKET_ARN}"
     skipper_open_policy_agent_observability_url: "${SKIPPER_OPA_OBSERVABILITY_URL}"
     skipper_open_policy_agent_bundles_url: "${SKIPPER_OPA_BUNDLES_URL}"
+    skipper_open_policy_agent_styra_response_header_timeout: 2
```
mikkeloscar: This is not needed as it's defined at config-defaults level.

cluster/config-defaults.yaml
wisinghe: Will try that. However I saw some integers in the file. Eg https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/config-defaults.yaml#L173

mikkeloscar:
:+1:

wisinghe:
👍

PR #9174, opened 2025-04-03
Exclude base image validation in kube-system

mikkeloscar:
To avoid a circular dependency as explained in #9092 exclude the base image validation in kube-system,

This check is disabled in e2e making it different from production. It's tricky to enable in e2e as we have many e2e images that would not pass.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -606,6 +606,7 @@ kubelet_image_gc_low_threshold: 40
 {{if eq .Cluster.Environment "production"}}
 teapot_admission_controller_validate_application_label: "true"
 teapot_admission_controller_validate_base_images: "true"
+teapot_admission_controller_validate_base_images_namespaces: "^kube-system$"
 
 # Check container image compliance in production clusters. Be careful when thinking about changing this: Setting it to
 # false will allow any container image to run in production clusters.


mikkeloscar:
:+1:

cluster/config-defaults.yaml
```diff
@@ -606,6 +606,7 @@ kubelet_image_gc_low_threshold: 40
 {{if eq .Cluster.Environment "production"}}
 teapot_admission_controller_validate_application_label: "true"
 teapot_admission_controller_validate_base_images: "true"
+teapot_admission_controller_validate_base_images_namespaces: "^kube-system$"
```
linki, line 609: Doesn't this enable the base image check _only_ in `kube-system` this way?

image

https://github.bus.zalan.do/teapot/admission-controller/blob/f3b5344ce3ad2efdf7fee04c484a8311f9263e57/pkg/config/config.go#L88-L89

linki: I don't think we did base image checks in `kube-system` before. Looks like it's excluded here (as a protected namespace): https://github.bus.zalan.do/teapot/admission-controller/blob/6de653566e2674a442397f983a895829274a2711/pkg/podfactory/resource_validator.go#L194

mikkeloscar: Yeah, I realized this is at least wrong.

Will try to replicate the prod issue in a pet cluster to better think about a solution.

mikkeloscar:
Closing as this didn't work

PR #9176, opened 2025-04-04
karpenter: Only allow nitro instances

mikkeloscar:
Only allow nitro instances in general and drop the complexity of a config-item and eks ipv6 conditions.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #9181, opened 2025-04-06
kube2iam: Fix for EKS ipv6

mikkeloscar:
Make kube2iam work on EKS with IPv6

This is done in the following way:

* Enable IPv6 EC2 metadata endpoint `http://[fd00:ec2::254]` (`HttpProtocolIpv6: enabled`) (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html)
* Update AMI to a version which create an `ip6tables` rule for routing traffic targeting `http://[fd00:ec2::254]` to kube2iam.
* Update kube2iam to fix handling IPv6 addresses (https://github.com/jtblin/kube2iam/pull/389)
* Inject `AWS_EC2_METADATA_SERVICE_ENDPOINT=http://[fd00:ec2::254]` and `AWS_EC2_METADATA_SERVICE_ENDPOINT_MODE=IPv6` envars into pods to ensure the AWS SDKs uses the IPv6 endpoint (https://docs.aws.amazon.com/sdkref/latest/guide/feature-imds-credentials.html).

PR #9191, opened 2025-04-08
Avoid userdata change

mikkeloscar:
In #9181 we changed the value of `kuberuntu_ami_version` from `old|new` to `aws|eks`. This had the side effect of changing the userdata which was not intended.

This fixes it by removing the use of `kuberuntu_ami_version` inside the userdata and ensuring that the userdata renders exactly the same as before to avoid rolling nodes even just for a whitespace in the userdata.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #9198, opened 2025-04-09
Only attach skipper nodes to NLB in e2e

mikkeloscar:
We run with a dedicated skipper-ingress pool in e2e as in all other clusters, so only attach instances of that pool to the NLB(s) in e2e.


MustafaSaber:
👍

mikkeloscar:
:+1:

PR #9203, opened 2025-04-09
Add custom PDB for skipper-ingress covering main and canary deployment

mikkeloscar:
Since introducing the canary skipper-ingress deployment there is no automatic PodDisruptionBudget created by `pdb-controller` in test and e2e clusters. The reason is that it only creates a PDB for deployments with 2 or more replicas. Since there can be a canary with 1 replica and main deployment with minReplicas: 1 it doesn't create a PDB.
This means that both pods could be unavailable at the same time, making skipper-ingress 100% unavailable without any protection.

This fixes the issue by adding a custom PDB that matches pods across both deployments.

I noticed this when debugging the flakyness of the load tests done as part of e2e and I believe this is causing skipper to be unavailable once in a while during e2e, making the load test fail.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -113,11 +113,7 @@ skipper_ingress_binpack: "false"
 
 # skipper node-pool
 enable_dedicate_nodepool_skipper: "true"
-{{if eq .Cluster.Environment "e2e"}}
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #9206, opened 2025-04-10
skipper-ingress: explicit healthcheck routes on the pod level

ponimas:
Skipper provides two flags to control kubernetes healthcheck routes:
`-kubernetes-healthcheck` (true by default) that enables automatic healthcheck routes
and `-reverse-source-predicate` to select source predicate type.

Currently when routesrv is enabled (default state) it adds
automatic healthcheck routes and appends default filters to
all routes including healthcheck routes.

This change defines healthcheck routes in skipper pods to avoid adding default
filters to them and enable further modifications. 

The automated healthcheck routes are still enabled on the routesrv level, to 
make it possible to deploy the change and avoid the situation when the skipper
pod without inlined routes consumes data from the routerrv with healthchecks
 disabled. These routes are disabled in the [following PR](https://github.com/zalando-incubator/kubernetes-on-aws/pull/9241). 

When routesrv is disabled though skipper will add default filters
to explicitly-defined healthcheck routes like it happens currently.
In this case default filters should be configured to not hinder
healthchecks.

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -323,20 +323,22 @@ spec:
           - "-forwarded-headers=X-Forwarded-For,X-Forwarded-Proto=https,X-Forwarded-Port=443"
           - '-forwarded-headers-exclude-cidrs={{ .cluster_internal_cidrs | join "," }}'
 {{ end }}
-          - "-inline-routes"
-          - |
-            kube__healthz_down: Path("/kube-system/healthz")
-              && Shutdown()
-              && SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128")
-              -> disableAccessLog()
-              -> status(503)
-              -> <shunt>;
-            kube__healthz_up: Path("/kube-system/healthz")
-              && SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128")
-              -> disableAccessLog()
-              -> status(200)
-              -> <shunt>;
-            {{ .Cluster.ConfigItems.skipper_ingress_inline_routes }};
+          - >-
+            -inline-routes=
+            kube__healthz_down:
+            Path("/kube-system/healthz") &&
+            Shutdown() &&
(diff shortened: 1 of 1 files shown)


ayruslore (approved):

shyamz-22:
👍🏽

cluster/manifests/skipper/deployment.yaml
```diff
@@ -323,9 +323,23 @@ spec:
...
 {{ end }}
-{{ if .Cluster.ConfigItems.skipper_ingress_inline_routes }}
-          - "-inline-routes={{ .Cluster.ConfigItems.skipper_ingress_inline_routes }}"
-{{ end }}
+          - >-
+            -inline-routes=
+            kube__healthz_down:
+            Path("/kube-system/healthz") &&
+            Shutdown() &&
+            SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128"{{- if and (eq .Cluster.Provider "zalando-eks") (eq .Cluster.ConfigItems.eks_ip_family "ipv6")}}{{ range $ip := split .Values.subnet_ipv6_cidrs "," }},"{{ $ip }}"{{ end }}{{- end }})
```
mikkeloscar: I'm not sure it makes sense to have the ipv6 ranges AND the specific ranges for IPv6 when zalando-eks is configured.

Basically the ranges: `"fd00::/8", "::1/128"` has no meaning in our setup from my understanding.

ponimas: fixed

ponimas:
👍

ayruslore:
👍

cluster/manifests/skipper/deployment.yaml
szuecs: Hmm ::1 should be useful for debugging purposes if you execute into the pod. Don't you think so?

ponimas: But from inside the pods ipv4 loopback should be available?

szuecs: It depends if we have IPv4 enabled or disabled. If we have dualstack enabled we also have IPv4 but it simplifies the network if we would disable IPv4 and maybe it fixes a teapot issue with the svc type load balancer issue if we would disable IPv4.

ponimas: Makes sense, i'll enable the looopback back.

ponimas:
👍

MustafaSaber:
This PR & the following supersedes https://github.com/zalando-incubator/kubernetes-on-aws/pull/8894 and fix pipeline issues there.

MustafaSaber:
👍

PR #9213, opened 2025-04-10
Disable emergency-access-service by default

mikkeloscar:
Disable Emergency-access-Service by default

This service is replaced by Okta/Savyint so there is no need to run it anymore. A handful of users are still connecting to in in the last month, will send communication to them and inform about using Okta instead.


mikkeloscar:
Will disable it via cluster-config-manager first so it's easier to control.

mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #9217, opened 2025-04-10
[HOTFIX] beta to stable e2e fix

mikkeloscar:
This introduces #9119 to stable as a separate PR.

The reason is that this change is not compatible with the previous version so e2e will fail. Therefore don't combine it with other changes.


mikkeloscar:
:+1:

linki:
👍

PR #9218, opened 2025-04-11
skipper: whitelist IPv6 CIDR

MustafaSaber:
Currently we only whitelist IPv6 with RouteSRV running in front of skipper, see https://github.com/zalando-incubator/kubernetes-on-aws/blob/18cb6e0afd46c6244d06ea79fd98fd81a0396b3f/cluster/manifests/skipper/deployment.yaml#L556
If we disable RouteSRV healthchecks will fail.


mikkeloscar:
:+1:

MustafaSaber:
👍

PR #9221, opened 2025-04-11
Add taint to arm64 pool

mikkeloscar:
Add taint to ARM e2e pool to avoid e2e test containers that are only amd64 compatible gets scheduled on an ARM node with empty space.


mikkeloscar:
:+1:

linki:
👍

PR #9226, opened 2025-04-11
Allow Manual role Admin access in e2e clusters

mikkeloscar:
Hack to give `Manual` role Admin access in e2e clusters

Right now, we don't have a simple way to grant admin access to a single cluster, so simple use Manual in e2e clusters as a special work-around.


demonCoder95:
👍

katyanna:
👍

mikkeloscar:
:+1:

PR #9234, opened 2025-04-14
exclude a1 instance family from Karpenter pools

demonCoder95:
The `a1` instance family has an ARM architecture that is too old and flannel-tc container crashes on the node:

```bash
Fatal glibc error: This version of Amazon Linux requires a newer ARM64 processor
                   compliant with at least ARM architecture 8.2-a with Cryptographic
                   extensions. On EC2 this is Graviton 2 or later.
```

We have to exclude this from Karpenter pools so these instances aren't used.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #9272, opened 2025-04-22
Only update flannel OnDelete

mikkeloscar:
Change the updateStrategy to OnDelete to only update flannel pods when nodes rotate.

The motivation is that shutdown of a flannel pod can lead to temporary unavailability of pod network on the node because flannel [cleans up iptables during shutdown](https://github.com/flannel-io/flannel/blob/986fc24fc75619160bb0ad9a104790e585462ed0/pkg/trafficmngr/iptables/iptables.go#L59-L62).

This avoids issues if we accidentally update flannel until we have a proper fix.

### DIFF
--- a/test/e2e/daemonset-updated/main.go
+++ b/test/e2e/daemonset-updated/main.go
@@ -192,11 +192,19 @@ func onDeleteDaemonsets(ctx context.Context, client kubernetes.Interface) (map[d
 
 	for _, ds := range daemonsets.Items {
 		if ds.Spec.UpdateStrategy.Type == appsv1.OnDeleteDaemonSetStrategyType {
+			var templateGeneration int64
+			if genStr, ok := ds.Annotations["deprecated.daemonset.template.generation"]; ok {
+				templateGeneration, err = strconv.ParseInt(genStr, 10, 64)
+				if err != nil {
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

linki:
👍

test/e2e/daemonset-updated/main.go
```diff
@@ -192,11 +192,19 @@ func onDeleteDaemonsets(ctx context.Context, client kubernetes.Interface) (map[d
...
+					templateGeneration = 0
+				}
+
+			}
 			onDeleteDaemonsets[dsID{
 				Name:      ds.Name,
 				Namespace: ds.Namespace,
 				UID:       ds.UID,
-			}] = ds.Generation
+			}] = templateGeneration
```
mikkeloscar, line 207: This fix makes the script correctly check if the pods need an update by comparing template generation and not DS generation as before which is different.

Interestingly enough. The annotation `deprecated.daemonset.template.generation` is deprecated since forever, but there is no alternative and it's still available in Kubernetes v1.32 :shrug:

mikkeloscar:
:+1:

linki:
👍

PR #9276, opened 2025-04-22
e2e: only consolidate empty nodes

mikkeloscar:
Make e2e tests more reliable by configuring karpenter to only consolidate when nodes are empty to avoid kicking out e2e pods to consolidate.

Also move `consolidate_after` to the top level instead of defining on every node pool.


mikkeloscar:
👍

demonCoder95:
👍

PR #9279, opened 2025-04-23
Add the `aws-load-balancer-controller` to support ServiceType LoadBalancer on EKS

linki:
This adds the https://github.com/kubernetes-sigs/aws-load-balancer-controller (ALBC) to EKS-based clusters.

With it it's possible to use Services with `Type: LoadBalancer` on EKS with IPv6. It's implemented as an NLB. The controller also takes care of opening and closing the ports on the nodes via updating the security group.

A working service could look like this:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: nginx
  annotations:
    # dualstack enables IPv6, there's no IPv6-only option
    # injected: this will be injected by admission-controller on EKS if not defined
    service.beta.kubernetes.io/aws-load-balancer-ip-address-type: dualstack
    # configures IPv6 Pod-addresses in the Target Group, bypassing the need to configure nodes (which doesn't work currently)
    # optional: this is the default and can be omitted or changed to `instance`
    service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
    # makes this NLB available via the Internet
    # optional: this is the default and can be omitted or changed to `internal`
    service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
    # exposes this port from the below port-list via SSL
    # optional: if not defined, all ports would be used
    service.beta.kubernetes.io/aws-load-balancer-ssl-ports: "443"
    # configure a certificate for SSL termination
    # optional: if not defined, no SSL endpoint will be available
    service.beta.kubernetes.io/aws-load-balancer-ssl-cert: arn:aws:acm:eu-central-1:1234567890:certificate/some-id
    # configure DNS to point this name to the created NLB
    # optional: if not defined, no custom DNS name will be available. Needed when using SSL.
    external-dns.alpha.kubernetes.io/hostname: nginx.playground.zalan.do
spec:
  type: LoadBalancer
  # marks this service for being processed by ALBC rather than the in-tree Service controller
  # injected: this will be injected by admission-controller on EKS if not defined
  loadBalancerClass: service.k8s.aws/nlb
  selector:
    application: nginx
  ports:
    - name: http
      protocol: TCP
      port: 80
      targetPort: 80
    - name: https
      protocol: TCP
      port: 443
      targetPort: 80
```

The part that triggers the new controller is:

```yaml
spec:
  type: LoadBalancer
  loadBalancerClass: service.k8s.aws/nlb
```

which marks the Service as a "load balancer" and gives it a class that's handled by ALBC. This prevents the in-tree, or cloud-provider (formerly in-tree) controllers to process it. `loadBalancerClass` is optional and will be injected by `admission-controller` unless it's already defined.

Another required part is the annotation: `service.beta.kubernetes.io/aws-load-balancer-ip-address-type: dualstack`. Since our EKS clusters run on IPv6 this is the only value that works. This annotation is optional and will be injected by `admission-controller` unless it's already defined.

All other annotations are optional but using them either overrides a default or triggers new functionality:
* `service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip`: backend can be registered to load balancers in two modes, IP-mode attaches IPv6 pod IPs, whereas instance-mode attaches instances. Currently, only ip-mode works but it could be changed here.
* `service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing`: by default load balancers are internet facing to match the old behaviour. This annotations can be used to make private load balancers.
* `service.beta.kubernetes.io/aws-load-balancer-ssl-cert: arn:aws:acm:eu-central-1:1234567890:certificate/some-id`: optionally enable SSL termination by specifying a certificate. If not set, traffic will be unencrypted.
* `service.beta.kubernetes.io/aws-load-balancer-ssl-ports`: when using SSL, all ports are exposed encrypted. If not desired (such as when doing a typical 80/443 setup) a subset can be selected
* `external-dns.alpha.kubernetes.io/hostname: nginx.playground.zalan.do`: assign a custom DNS name via ExternalDNS. Likely needed when using SSL, otherwise clients will present warnings when the NLB endpoint is used directly.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -3222,7 +3222,7 @@ Resources:
           PolicyName: root
       RoleName: "{{.Cluster.LocalID}}-audittrail-adapter"
     Type: 'AWS::IAM::Role'
-# {{- if eq .Cluster.Provider "zalando-eks"}}
+# {{- if and (eq .Cluster.Provider "zalando-eks") (eq .Cluster.ConfigItems.eks_ip_family "ipv6") }}
   AWSLoadBalancerControllerIAMRole:
     Properties:
       AssumeRolePolicyDocument: !Sub
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -3222,6 +3222,221 @@ Resources:
           PolicyName: root
       RoleName: "{{.Cluster.LocalID}}-audittrail-adapter"
     Type: 'AWS::IAM::Role'
+# {{- if eq .Cluster.Provider "zalando-eks"}}
+  AWSLoadBalancerControllerIAMRole:
+    Properties:
+      AssumeRolePolicyDocument: !Sub
+        - |
+          {
+            "Version": "2012-10-17",
+            "Statement": [
+              {
+                "Effect": "Allow",
+                "Principal": {
+                  "Federated": [
+                    "arn:aws:iam::${AWS::AccountId}:oidc-provider/${OIDC}"
+                  ]
+                },
+                "Action": [
+                  "sts:AssumeRoleWithWebIdentity"
+                ],
+                "Condition": {
+                  "StringEquals": {
+                    "${OIDC}:sub": "system:serviceaccount:kube-system:aws-load-balancer-controller"
+                  }
+                }
+              }
+            ]
+          }
+{{- if eq .Cluster.Provider "zalando-eks" }}
+        - OIDC: !Select [1, !Split ["//", !GetAtt EKSCluster.OpenIdConnectIssuerUrl]]
+{{- else }}
+        - OIDC: "{{.Cluster.LocalID}}.{{.Values.hosted_zone}}"
+{{- end }}
+      Path: /
+      Policies:
+        - PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+            - Effect: Allow
+              Action:
+              - iam:CreateServiceLinkedRole
+              Resource: "*"
+              Condition:
+                StringEquals:
+                  iam:AWSServiceName: elasticloadbalancing.amazonaws.com
+            - Effect: Allow
+              Action:
+              - ec2:DescribeAccountAttributes
+              - ec2:DescribeAddresses
+              - ec2:DescribeAvailabilityZones
+              - ec2:DescribeInternetGateways
+              - ec2:DescribeVpcs
+              - ec2:DescribeVpcPeeringConnections
+              - ec2:DescribeSubnets
+              - ec2:DescribeSecurityGroups
+              - ec2:DescribeInstances
+              - ec2:DescribeNetworkInterfaces
+              - ec2:DescribeTags
+              - ec2:GetCoipPoolUsage
+              - ec2:DescribeCoipPools
+              - ec2:GetSecurityGroupsForVpc
+              - ec2:DescribeIpamPools
+              - elasticloadbalancing:DescribeLoadBalancers
+              - elasticloadbalancing:DescribeLoadBalancerAttributes
+              - elasticloadbalancing:DescribeListeners
+              - elasticloadbalancing:DescribeListenerCertificates
+              - elasticloadbalancing:DescribeSSLPolicies
+              - elasticloadbalancing:DescribeRules
+              - elasticloadbalancing:DescribeTargetGroups
+              - elasticloadbalancing:DescribeTargetGroupAttributes
+              - elasticloadbalancing:DescribeTargetHealth
+              - elasticloadbalancing:DescribeTags
+              - elasticloadbalancing:DescribeTrustStores
+              - elasticloadbalancing:DescribeListenerAttributes
+              - elasticloadbalancing:DescribeCapacityReservation
+              Resource: "*"
+            - Effect: Allow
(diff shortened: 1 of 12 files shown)


linki:
👍

cluster/cluster.yaml
```diff
@@ -3222,6 +3222,221 @@ Resources:
...
+                  }
+                }
+              }
+            ]
+          }
+{{- if eq .Cluster.Provider "zalando-eks" }}
+        - OIDC: !Select [1, !Split ["//", !GetAtt EKSCluster.OpenIdConnectIssuerUrl]]
+{{- else }}
+        - OIDC: "{{.Cluster.LocalID}}.{{.Values.hosted_zone}}"
+{{- end }}
```
demonCoder95: The outer block starting on line 3225 has the condition `.Cluster.Provider == "zalando-eks"` already, won't that make it impossible for this `else` block to ever be executed?

demonCoder95: Maybe we need to terminate the outer block on line 3252?

cluster/manifests/aws-load-balancer-controller/deployment.yaml
```diff
@@ -0,0 +1,78 @@
...
+          initialDelaySeconds: 30
+          timeoutSeconds: 10
+        name: controller
+        ports:
+        - containerPort: 9443
+          name: webhook-server
+          protocol: TCP
+        resources:
+          limits:
+            cpu: 100m
```
demonCoder95: Would it make sense to use a config-item for CPU as well?

cluster/manifests/aws-load-balancer-controller/vpa.yaml
```diff
@@ -0,0 +1,22 @@
...
+    apiVersion: apps/v1
+    kind: Deployment
+    name: aws-load-balancer-controller
+  updatePolicy:
+    updateMode: Auto
+  resourcePolicy:
+    containerPolicies:
+    - containerName: controller
+      maxAllowed:
+        memory: {{ .Cluster.ConfigItems.aws_load_balancer_controller_mem }}
```
demonCoder95: This should be `minAllowed`, I think. Having this `maxAllowed` defeats the purpose of having a VPA as the memory requests will never be increased. Since we've already set the memory requests to the same value.

We should also change the config-item's name to append `_min` to clarify this.

cluster/cluster.yaml
linki: I left it just in case we removed the outer if (to run this controller on legacy clusters) in which case it would just work. Otherwise, it's likely to be missed.

However, I dropped a similar if in another place, so I'm going to remove it here, too.

cluster/manifests/aws-load-balancer-controller/deployment.yaml
linki: Added

cluster/cluster.yaml
```diff
@@ -3222,6 +3222,217 @@ Resources:
           PolicyName: root
       RoleName: "{{.Cluster.LocalID}}-audittrail-adapter"
     Type: 'AWS::IAM::Role'
+# {{- if eq .Cluster.Provider "zalando-eks"}}
```
mikkeloscar: Should we also condition all this to `eks_ip_family: "ipv6"`? Since on IPv4 the normal "in-tree" controller works.

cluster/manifests/aws-load-balancer-controller/vpa.yaml
linki: This was indeed wrong but I left it at `maxAllowed`. I took other controller definitions as a guide, such as this one: https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/pdb-controller/vpa.yaml and others.

It initially starts it with a lot of extra memory and then lets VPA take care of adjusting it _down_ over time.

cluster/cluster.yaml
linki: I didn't think about that case. We could do it, but it should also work like it is now. `admission-controller` shifts _all_ Services with LoadBalancer type to the new controller on EKS, regardless of IP version.

So, it depends on which controller we prefer. I think there's benefit in using a single controller for EKS rather than two based on IP version.

Do we intend to run IPv4 EKS clusters going forward?

cluster/manifests/aws-load-balancer-controller/vpa.yaml
demonCoder95: I see, you've increased the memory a bit, so I suppose that's fine this way as well :)

demonCoder95:
👍

cluster/manifests/aws-load-balancer-controller/vpa.yaml
linki: `4Gi` seems to be what other components use. It's more like a placeholder than a consciously chosen value.

cluster/cluster.yaml
linki: Given that we have existing EKS/IPv4 clusters it's probably better to add the check.

mikkeloscar:
:+1:

linki:
👍

PR #9284, opened 2025-04-24
allow excluding t type instances from Karpenter node-pools

demonCoder95:
The `t` type instances have burstable CPU which makes them unpredicable to use in production environments. This allows to toggle the usage of these instances. The default behavior will be to exlude them, but we will allow them in test environment.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #9302, opened 2025-04-25
Don't display cluster-scoped CRDs in `kubectl get all`

linki:
`ClusterScalingSchedule` is the only global CRD that is displayed as part of `kubectl get all`. This is fairly confusing when you want to see "all" resources in a namespace (knowing that all is not really everything) but you constantly _also_ see the `ClusterScalingSchedule` everytime.

This was recently introduced by a bug fix but turns out the old (bugged) behaviour was actually better.


zaklawrencea:
:+1:

linki:
👍

PR #9308, opened 2025-04-28
skipper: use internal CIDRs variable for healthcheck routes

AlexanderYastrebov:
* reuse existing `cluster_internal_cidrs` variable that accounts for `zalando-eks` provider configuration instead of ad-hoc formatting
* remove `-whitelisted-healthcheck-cidr` flag added by #9218 which is not used since #9206 configured explicit healthcheck routes
* indent and preserve newlines in healthcheck routes

Follow up on #9206

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -177,9 +177,6 @@ spec:
           - "-kubernetes-default-lb-algorithm={{ .Cluster.ConfigItems.skipper_ingress_default_lb_algorithm }}"
           - "-kubernetes-disable-catchall-routes={{ .Cluster.ConfigItems.skipper_ingress_disable_catchall_routes }}"
           - "-enable-kubernetes-endpointslices={{ .Cluster.ConfigItems.skipper_endpointslices_enabled }}"
-{{ if and (eq .Cluster.Provider "zalando-eks") (eq .Cluster.ConfigItems.eks_ip_family "ipv6")}}
-          - "-whitelisted-healthcheck-cidr={{ .Values.subnet_ipv6_cidrs }}"
-{{ end }}
 {{ end }}
           - "-address=:9999"
           - "-wait-first-route-load"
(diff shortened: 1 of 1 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -323,21 +323,18 @@ spec:
           - "-forwarded-headers=X-Forwarded-For,X-Forwarded-Proto=https,X-Forwarded-Port=443"
           - '-forwarded-headers-exclude-cidrs={{ .cluster_internal_cidrs | join "," }}'
 {{ end }}
-          - >-
```
AlexanderYastrebov, line 326: Config is multiline so I propose we keep newlines to make actual value close to the configured.

cluster/manifests/skipper/deployment.yaml
```diff
@@ -323,21 +323,18 @@ spec:
...
-            kube__healthz_up:
-            Path("/kube-system/healthz") &&
-            SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "::1/128"{{- if and (eq .Cluster.Provider "zalando-eks") (eq .Cluster.ConfigItems.eks_ip_family "ipv6")}}{{ range $ip := split .Values.subnet_ipv6_cidrs "," }},"{{ $ip }}"{{ end }}{{- else }}, "fd00::/8"{{- end }})
-            -> disableAccessLog()
-            -> status(200)
-            -> <shunt>;
+          - "-inline-routes"
+          - |
+            kube__healthz_down: Path("/kube-system/healthz") && Shutdown()
+              && SourceFromLast("{{ $cluster_internal_cidrs | join `","` }}", "10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128")
```
AlexanderYastrebov: I kept hardcoded `"10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128"` based on this discussion https://github.com/zalando-incubator/kubernetes-on-aws/pull/9206#discussion_r2058074066

Some values likely could be removed but it also does not hurt as all of them are private ranges.

ponimas:
👍

MustafaSaber:
In the description we say we remove the whitelist added but I don't see that it's removed. Also are we sure we only use it in healthchecks routes?

AlexanderYastrebov:
> In the description we say we remove the whitelist added but I don't see that it's removed. 

My bad :facepalm: - fixed, thanks!

> Also are we sure we only use it in healthchecks routes?

Yes, checked Skipper code to confirm.

ponimas:
👍

AlexanderYastrebov:
:+1:

ponimas:
👍

AlexanderYastrebov:
:+1:

ponimas:
👍

PR #9314, opened 2025-04-29
security: update Ingress default ssl policy `ELBSecurityPolicy-TLS13-1-2-2021-06`

ayruslore:
updates the default ssl policy from ELBSecurityPolicy-TLS-1-2-2017-01 to ELBSecurityPolicy-TLS13-1-2-2021-06 for ingress

Outdated ssl-policy can lead to low level attacks like MitM to break TLS connections. Also TLS ranking tools show us not to be best in class, so we should take care to make it best in class.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -681,7 +681,7 @@ Resources:
     Properties:
       AlpnPolicy:
         - {{ if eq .Cluster.ConfigItems.experimental_nlb_alpn_h2_enabled "true" }}HTTP2Preferred{{else}}None{{end}}
-      SslPolicy: "ELBSecurityPolicy-TLS13-1-2-2021-06"
+      SslPolicy: "ELBSecurityPolicy-TLS-1-2-2017-01"
       Certificates:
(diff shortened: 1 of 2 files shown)


cluster/cluster.yaml
```diff
@@ -614,7 +614,7 @@ Resources:
     Properties:
       AlpnPolicy:
         - {{ if eq .Cluster.ConfigItems.experimental_nlb_alpn_h2_enabled "true" }}HTTP2Preferred{{else}}None{{end}}
-      SslPolicy: "ELBSecurityPolicy-TLS-1-2-2017-01"
```
AlexanderYastrebov: Do we want to change control plane policy and ingress policy in one PR?

ayruslore: The change is minor considering the ssl policy update, but if we think it would cause a problem recreating LBs for both control plane and master in the same rollout, then I can split the update into two PRs.

ayruslore: removed the control plane update. Will create a separate PR for that.

AlexanderYastrebov:
:+1:

ayruslore:
👍

PR #9315, opened 2025-04-29
Add internal path for loadtest

mikkeloscar:
This extends the existing load test to test internal path in the cluster e.g. to test the pod network via the path illustrated below:

image

The purpose of this is to test the pod network during a cluster upgrade e.g. during flannel rotation which can cause a short drop of network availability (https://github.com/zalando-incubator/kubernetes-on-aws/pull/9272).

To make the test work it further improves the load test

* Fix checking if daemonset pods are up to date in the `wait-for-update.py` script
* Prevent scale-down of the vegeta/load test client to ensure it always running also when flannel is rotated.
* Use `kubectl apply` instead of `create` to make it simpler to run `start-load-test.sh` locally.

### TODO

* [x] Drop flannel change which is currently there to validate the test would catch it.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -689,7 +689,7 @@ spot_node_rescheduler_cpu: "50m"
 
 # Enable and configure prevent scale down annotation
 teapot_admission_controller_prevent_scale_down_enabled: "true"
-{{if eq .Cluster.Environment "production"}}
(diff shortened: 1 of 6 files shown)


MustafaSaber:
👍

mikkeloscar:
:+1:

demonCoder95:
👍

PR #9321, opened 2025-04-30
Use newer and recommended TLS 1.3 security policy for NLBs

linki:
Let's use the latest and recommended TLS security policy for new Services on EKS clusters. This is the same policy as proposed for Ingresses: https://github.com/zalando-incubator/kubernetes-on-aws/pull/9314

image


mikkeloscar:
👍

linki:
👍

PR #9335, opened 2025-05-02
skipper: enable backend and response metrics

ayruslore:
This change adds new flags for enabling backend and response metrics for skipper pods.

The defaults are false and this change enables setting these values as true for different clusters and collect metrics when required.


AlexanderYastrebov:
:+1:

ayruslore:
👍

PR #9347, opened 2025-05-05
security: update Control Plane default ssl policy to `ELBSecurityPolicy-TLS13-1-2-2021-06`

ayruslore:
updates the default ssl policy from ELBSecurityPolicy-TLS-1-2-2017-01 to ELBSecurityPolicy-TLS13-1-2-2021-06 for the control plane

Outdated ssl-policy can lead to low level attacks like MitM to break TLS connections. Also TLS ranking tools show us not to be best in class, so we should take care to make it best in class.

Ingress update was done here : https://github.com/zalando-incubator/kubernetes-on-aws/pull/9314


linki:
👍

ayruslore:
👍

PR #9351, opened 2025-05-05
Add e2e admission-controller role access tests

katyanna:
Extend test roles Admission Controller coverage for EKS.

| **User Role \\ Namespace** | **user namespace** | **visibility**     | **kube-system** |
|----------------------------|---------------------|---------------------|------------------|
| **admin**                  | ✅ Write Access     | ✅ Write Access     | ✅ Write Access  |
| **collaborator**           | ✅ Write Access     | ✅ Write Access     |  ❌ No Access    |
| **engineer**               | ✅ Write Access     | ❌ No Access        | ❌ No Access     |

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -285,7 +285,7 @@ Resources:
             AWS: !Sub "arn:aws:iam::${AWS::AccountId}:root"
         Version: 2012-10-17
       Path: /
-      RoleName: "{{.Cluster.LocalID}/-e2e-eks-iam-test-read-only-role"
+      RoleName: "{{.Cluster.LocalID}}-e2e-eks-iam-test-read-only-role"
     Type: 'AWS::IAM::Role'
   E2EEKSIAMTestAccessEntryReadOnly:
     Type: "AWS::EKS::AccessEntry"
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -273,7 +273,7 @@ Resources:
       AddonName: eks-pod-identity-agent
       ClusterName: !Ref EKSCluster
   {{ if eq .Cluster.Environment "e2e" }}
-  E2EEKSIAMTestRoleUnprivileged:
+  E2EEKSIAMTestRoleReadOnly:
     Properties:
       AssumeRolePolicyDocument:
         Statement:
@@ -285,9 +285,9 @@ Resources:
             AWS: !Sub "arn:aws:iam::${AWS::AccountId}:root"
         Version: 2012-10-17
       Path: /
-      RoleName: "{{.Cluster.LocalID}}-e2e-eks-iam-test-unprivileged-role"
+      RoleName: "{{.Cluster.LocalID}/-e2e-eks-iam-test-read-only-role"
     Type: 'AWS::IAM::Role'
-  E2EEKSIAMTestAccessEntryUnprivileged:
+  E2EEKSIAMTestAccessEntryReadOnly:
     Type: "AWS::EKS::AccessEntry"
     Properties:
       AccessPolicies:
@@ -295,16 +295,16 @@ Resources:
           Type: "cluster"
         PolicyArn: "arn:aws:eks::aws:cluster-access-policy/AmazonEKSClusterAdminPolicy"
       ClusterName: !Ref EKSCluster
-      PrincipalArn: !GetAtt E2EEKSIAMTestRoleUnprivileged.Arn
+      PrincipalArn: !GetAtt E2EEKSIAMTestRoleReadOnly.Arn
       Username: !Join
           - ''
           - - !Sub 'arn:aws:sts::${AWS::AccountId}:assumed-role/'
-            - !Ref E2EEKSIAMTestRoleUnprivileged
+            - !Ref E2EEKSIAMTestRoleReadOnly
             - '/{{`{{SessionName}}`}}'
       KubernetesGroups:
       - zalando:readonly
       Type: "STANDARD"
-  E2EEKSIAMTestRolePrivileged:
+  E2EEKSIAMTestRoleAdministrator:
     Properties:
       AssumeRolePolicyDocument:
         Statement:
@@ -316,9 +316,9 @@ Resources:
             AWS: !Sub "arn:aws:iam::${AWS::AccountId}:root"
         Version: 2012-10-17
       Path: /
-      RoleName: "{{.Cluster.LocalID}}-e2e-eks-iam-test-privileged-role"
+      RoleName: "{{.Cluster.LocalID}}-e2e-eks-iam-test-administrator-role"
     Type: 'AWS::IAM::Role'
-  E2EEKSIAMTestAccessEntryPrivileged:
+  E2EEKSIAMTestAccessEntryAdministrator:
     Type: "AWS::EKS::AccessEntry"
     Properties:
       AccessPolicies:
@@ -326,11 +326,11 @@ Resources:
           Type: "cluster"
         PolicyArn: "arn:aws:eks::aws:cluster-access-policy/AmazonEKSClusterAdminPolicy"
       ClusterName: !Ref EKSCluster
-      PrincipalArn: !GetAtt E2EEKSIAMTestRolePrivileged.Arn
+      PrincipalArn: !GetAtt E2EEKSIAMTestRoleAdministrator.Arn
       Username: !Join
           - ''
           - - !Sub 'arn:aws:sts::${AWS::AccountId}:assumed-role/'
-            - !Ref E2EEKSIAMTestRolePrivileged
+            - !Ref E2EEKSIAMTestRoleAdministrator
             - '/{{`{{SessionName}}`}}'
       KubernetesGroups:
       - zalando:administrator
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -366,6 +366,37 @@ Resources:
       KubernetesGroups:
       - zalando:collaborator
       Type: "STANDARD"
+  E2EEKSIAMTestRoleEngineer:
+    Properties:
+      AssumeRolePolicyDocument:
+        Statement:
+        - Action:
+          - 'sts:AssumeRole'
+          - 'sts:SetSourceIdentity'
+          Effect: Allow
+          Principal:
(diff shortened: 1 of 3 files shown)


cluster/cluster.yaml
```diff
@@ -366,6 +366,37 @@ Resources:
...
+      ClusterName: !Ref EKSCluster
+      PrincipalArn: !GetAtt E2EEKSIAMTestRoleEngineer.Arn
+      Username: !Join
+          - ''
+          - - !Sub 'arn:aws:sts::${AWS::AccountId}:assumed-role/'
+            - !Ref E2EEKSIAMTestRoleEngineer
+            - '/{{`{{SessionName}}`}}'
+      KubernetesGroups:
+      - zalando:engineer
+      Type: "STANDARD"
```
linki, line 399: This is just to get a token with `zalando:engineer` group.

test/e2e/authorization.go
```diff
@@ -663,6 +708,7 @@ var _ = g.Describe("Authorization via admission-controller [RBAC] [Zalando]", fu
 				framework.ExpectNoError(err)
 			})
 
+			// TODO: is that correct?? We use "readonly" as the role in the test case, why should this be able to delete something?
```
linki: Ok, it's a bit annoying.

The tests are designed in this way: The user in the the test case is always an "admin" in terms of RBAC. This is to ensure that any "prevention" must come from admission-controller, rather than from RBAC. Unfortunately, that makes the expectations here sort of wrong (maybe).

linki: So, the tests here are only for the admission-controller's response of the request. They don't represent the real response which is a combination of RBAC+admission-controller.

test/e2e/authorization.go
```diff
@@ -642,6 +643,19 @@ var _ = g.Describe("Authorization via admission-controller [RBAC] [Zalando]", fu
 				framework.ExpectNoError(err, "failed to create pod: %s in namespace: %s", nonSystemResource.Name, nonSystemResource.Namespace)
 			})
 
+			// Not needed actually
```
linki: I commented this out because the missing test cases were about the namespaces itself rather than resources within the namespace.

But we could also add this test case. We would just need to create a resource first, otherwise the delete call doesn't work correctly.

test/e2e/authorization.go
```diff
@@ -651,6 +665,41 @@ var _ = g.Describe("Authorization via admission-controller [RBAC] [Zalando]", fu
 				_, err := client.CoreV1().Pods(systemResource.Namespace).Create(context.Background(), systemResource, metav1.CreateOptions{DryRun: []string{"All"}})
 				gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("write operations are forbidden")))
 			})
+
+			// Not needed actually
```
linki: See similar comment above.

test/e2e/authorization.go
```diff
@@ -717,6 +767,72 @@ var _ = g.Describe("Authorization via admission-controller [RBAC] [Zalando]", fu
...
+			g.It("should deny write access for system resources", func() {
+				err := client.RbacV1().ClusterRoles().Delete(context.Background(), systemResource.Name, metav1.DeleteOptions{DryRun: []string{"All"}})
+				gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("write operations are forbidden")))
+			})
+
+			// test specific namespaces
+
+			g.It("should deny deletion of kube-system namespace", func() {
+				err := client.CoreV1().Namespaces().Delete(context.Background(), "kube-system", metav1.DeleteOptions{DryRun: []string{"All"}})
+				gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("this namespace may not be deleted")))
```
linki, line 755: This error message doesn't come from `admission-controller` but from Kubernetes itself: https://github.com/kubernetes/kubernetes/blob/v1.32.4/staging/src/k8s.io/apiserver/pkg/admission/plugin/namespace/lifecycle/admission.go#L80 (immortal namespace). That's why the error message is different than the one for `visibility`.

cluster/manifests/02-visibility/01-namespace.yaml
```diff
@@ -2,3 +2,5 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: visibility
+  labels:
+    admission.zalando.org/infrastructure-component: "true"
```
linki: This is what protects the `visibility` namespace. We could do the same for `kube-system` (to maybe get the same error message) but we don't have a YAML for that at the moment.

linki:
If we can we should also add a test case for this - currently untested - behaviour: https://github.com/zalando-incubator/kubernetes-on-aws/blob/6e651edb911e143051d5b0de6ac655a85b649015/cluster/manifests/02-admission-control/teapot.yaml#L578-L586

test/e2e/authorization.go
```diff
@@ -717,6 +767,72 @@ var _ = g.Describe("Authorization via admission-controller [RBAC] [Zalando]", fu
...
+			})
+
+			g.It("should deny deletion of visibility namespace", func() {
```
katyanna, line 780: `collaborator` role should not be able to delete the `collaborator` namespace?

test/e2e/authorization.go
```diff
@@ -651,6 +665,41 @@ var _ = g.Describe("Authorization via admission-controller [RBAC] [Zalando]", fu
...
+			var client *kubernetes.Clientset
+
+			g.BeforeEach(func() {
+				var err error
+
+				client, err = getEngineerClient(eksCluster, awsAccountID)
+				framework.ExpectNoError(err)
+			})
+
+			// these tests are similar to the ones for unprivileged (let's think about that) (also see remarks of the first test in the block below)
```
katyanna: Let's keep all access tested and not depend on tests for unprivileged even if this role is built upon it.

test/e2e/authorization.go
linki: Yes.

But even if not there's no point in deleting this namespace.

test/e2e/authorization.go
```diff
@@ -849,8 +969,13 @@ func getCollaboratorClient(cluster *types.Cluster, awsAccountID string) (*kubern
...
-// getUnprivilegedClient returns a client with the `zalando:readonly` group.
-func getUnprivilegedClient(cluster *types.Cluster, awsAccountID string) (*kubernetes.Clientset, error) {
+// getEngineerClient returns a client with the `zalando:engineer` group.
+func getEngineerClient(cluster *types.Cluster, awsAccountID string) (*kubernetes.Clientset, error) {
+	return newClientWithRole(cluster, fmt.Sprintf("arn:aws:iam::%s:role/%s-e2e-eks-iam-test-engineer-role", awsAccountID, aws.ToString(cluster.Name)))
+}
+
+// getReadOnlyClient returns a client with the `zalando:readonly` group.
+func getReadOnlyClient(cluster *types.Cluster, awsAccountID string) (*kubernetes.Clientset, error) {
 	return newClientWithRole(cluster, fmt.Sprintf("arn:aws:iam::%s:role/%s-e2e-eks-iam-test-unprivileged-role", awsAccountID, aws.ToString(cluster.Name)))
```
linki: Let's also rename the role itself from unprivileged to read-only.

test/e2e/authorization.go
```diff
@@ -839,8 +959,8 @@ var _ = g.Describe("Authorization via admission-controller [RBAC] [Zalando]", fu
 	})
 })
 
-// getPrivilegedClient returns a client with the `zalando:administrator` group.
-func getPrivilegedClient(cluster *types.Cluster, awsAccountID string) (*kubernetes.Clientset, error) {
+// getAdminClient returns a client with the `zalando:administrator` group.
+func getAdminClient(cluster *types.Cluster, awsAccountID string) (*kubernetes.Clientset, error) {
 	return newClientWithRole(cluster, fmt.Sprintf("arn:aws:iam::%s:role/%s-e2e-eks-iam-test-privileged-role", awsAccountID, aws.ToString(cluster.Name)))
```
linki: Let's also rename the role itself from privileged to admin/administrator.

test/e2e/authorization.go
```diff
@@ -711,19 +752,98 @@ var _ = g.Describe("Authorization via admission-controller [RBAC] [Zalando]", fu
...
+
+			g.It("should allow write access for non-system resources", func() {
+				err := client.RbacV1().ClusterRoles().Delete(context.Background(), nonSystemResource.Name, metav1.DeleteOptions{DryRun: []string{"All"}})
+				framework.ExpectNoError(err, "failed to delete cluster role: %s", nonSystemResource.Name)
+			})
+
+			g.It("should deny write access for system resources", func() {
+				err := client.RbacV1().ClusterRoles().Delete(context.Background(), systemResource.Name, metav1.DeleteOptions{DryRun: []string{"All"}})
+				gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("write operations are forbidden")))
+			})
```
linki, line 777: Let's add the third test case from `collaborator` here as well to test that it's different, e.g.

```
g.It("should deny write access for collaborator resources", func() {
```

linki: Unless that's not what we want. I'm not sure actually.

test/e2e/authorization.go
```diff
@@ -711,19 +752,98 @@ var _ = g.Describe("Authorization via admission-controller [RBAC] [Zalando]", fu
 				framework.ExpectNoError(err, "failed to delete cluster role: %s", nonSystemResource.Name)
 			})
 
+			g.It("should allow write access for collaborator resources", func() {
+				err := client.RbacV1().ClusterRoles().Delete(context.Background(), collaboratorResource.Name, metav1.DeleteOptions{DryRun: []string{"All"}})
```
linki: `collaboratorResource` is not initialized in the `BeforeEach`. But given that the outer context is "for global resources", it doesn't make sense to have collaborator resources here because we don't have this concept yet. When referring to the collaborator namespace itself, we can just use the string "visibility".

katyanna:
`read-only` role is not scoped, since it's entirely handled by RBAC.

demonCoder95:
👍

katyanna:
:+1:

PR #9392, opened 2025-05-12
Updating Wiz to Helm 0.2.91

cantwel-c:
Updating Wiz Sensor and Broker to a more recent version of Helm 0.2.91

### DIFF
--- a/cluster/manifests/wiz/connector-deployment.yaml
+++ b/cluster/manifests/wiz/connector-deployment.yaml
@@ -31,7 +31,7 @@ spec:
       volumes:
         - name: connector-data
           secret:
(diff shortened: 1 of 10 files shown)


linki:
@cantwel-c This only changes labels at the moment. Is that intended?

cluster/manifests/wiz/connector-deployment.yaml
```diff
@@ -31,7 +31,7 @@ spec:
       volumes:
         - name: connector-data
           secret:
-            secretName: wiz-connector-connector
```
zaklawrencea, line 34: This change doesn't look related to the title/description. We would also need to update the name of  the `Secret` resource.

PR #9413, opened 2025-05-13
Drop dnsmasq as dns cache option

mikkeloscar:
Drops `dnsmasq` as option for `dns_cache` and drops the `dns_cache` config-item, making `unbound` the only DNS cache option.

`unbound` is already the default in all clusters so this is just a cleanup to get rid of dnsmasq.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #9426, opened 2025-05-15
admission: use builtin preStop sleep

AlexanderYastrebov:
We already use it in other manifests and these two are the last ones using shell sleep.

See https://github.com/kubernetes/enhancements/issues/3960

### DIFF
--- a/cluster/manifests/02-admission-control/deployment.yaml
+++ b/cluster/manifests/02-admission-control/deployment.yaml
@@ -36,8 +36,8 @@ spec:
         image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/admission-controller:master-257
         lifecycle:
           preStop:
-            exec:
-              command: ["/bin/sh", "-c",  "sleep 60"]
+            sleep:
+              seconds: 60
         readinessProbe:
           httpGet:
             scheme: HTTPS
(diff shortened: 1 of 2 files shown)


cluster/manifests/02-admission-control/deployment.yaml
```diff
@@ -36,8 +36,8 @@ spec:
         image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/admission-controller:master-257
         lifecycle:
           preStop:
-            exec:
-              command: ["/bin/sh", "-c",  "sleep 60"]
+            sleep:
+              seconds: 60
```
AlexanderYastrebov: Uncovered a bug
```
time="2025-05-15T15:00:23Z" level=error msg="The Deployment \"teapot-admission-controller\" is invalid: spec.template.spec.containers[0].lifecycle.preStop.sleep: Invalid value: 60: must be greater than 0 and less than terminationGracePeriodSeconds (30). Enable AllowPodLifecycleSleepActionZeroValue feature gate for zero sleep." module=02-admission-control
```

mikkeloscar: The feature gate is actually called: `PodLifecycleSleepActionAllowZero` and will be default (beta) in 1.33

https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

AlexanderYastrebov:
:+1:

cluster/manifests/02-admission-control/deployment.yaml
AlexanderYastrebov: I've reduced sleep to 20 such that its below default terminationGracePeriodSeconds which is 30.

linki:
👍

PR #9427, opened 2025-05-15
Provide default OIDC trust relationship template as a deployment variable

linki:
Provide default OIDC trust relationship template as a deployment variable, so users don't need to specify it and we can update it.


linki:
👍

demonCoder95:
👍

PR #9428, opened 2025-05-15
fg: only use one snapshot

MustafaSaber:
We already use 1 snapshot everywhere this is a clean up, then will clean the override.


MustafaSaber:
👍

ayruslore (approved):

PR #9429, opened 2025-05-16
kube-apiserver: enable MutatingAdmissionPolicy

AlexanderYastrebov:
MutatingAdmissionPolicies allow simple modifications to resources during admission.

It is an Alpha feature since 1.30 and Beta release target is 1.34.

See
* https://kubernetes.io/docs/reference/access-authn-authz/mutating-admission-policy/
* https://github.com/kubernetes/enhancements/issues/3962


AlexanderYastrebov:
Hold on to discuss whether we want alpha feature for the use case we have.

AlexanderYastrebov:
The usecase was implemented in admission-controller https://github.com/zalando-incubator/kubernetes-on-aws/pull/9530

PR #9464, opened 2025-05-22
Updating Helm to a more recent verison 

cantwel-c:
Updating Helm labels to a more recent verison as well as moving to the updated secret managment


demonCoder95:
@cantwel-c seems like we have conflicts in the change. If it's still relevant, please resolve them so we can merge. If not, we should close the PR, it's been dormant for a long time.

PR #9472, opened 2025-05-26
FlinkDeployment: add flink-operator

fymo:
Adding Flink Operator controller to support FlinkDeployment


cluster/manifests/flinkdeployment/deployment.yaml
```diff
@@ -0,0 +1,202 @@
+{{ if eq .Cluster.ConfigItems.flink_operator_enabled "true" }}
```
linki, line 1: Please add this condition to all other resources, so that it becomes a "no-op" when disabled.

linki: Please also add all the guarded resources to https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/deletions.yaml. This way it's possible to turn it off after it was turned on (in case that's necessary).

cluster/manifests/flinkdeployment/deployment.yaml
```diff
@@ -0,0 +1,202 @@
...
+---
+# Source: flink-kubernetes-operator/templates/flink-operator.yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: flink-operator-config
+  namespace: kube-system
```
linki: I wonder if it should rather be in its own namespace.

cluster/manifests/flinkdeployment/rbac.yaml
```diff
@@ -0,0 +1,184 @@
...
+    application: kubernetes
+    component: flink-operator
+roleRef:
+  kind: Role
+  name: flink
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+  - kind: ServiceAccount
+    name: flink
+    namespace: kube-system
```
linki, line 183: namespace should be removed for non-cluster-wide role bindings.

fymo: If I understood correctly, you mean removing namespace from ClusterRole and ClusterRoleBinding, right?

mikkeloscar:
Closing as we discussed internally to roll out via CDP for now

PR #9480, opened 2025-05-30
Enable  skipper-ingress aws role for reading opa bundles by default

adetunjiakintundeakinde:
Enable  skipper-ingress aws role for reading opa bundles by default


szuecs:
:+1:

MustafaSaber:
👍

mikkeloscar:
This breaks our pet clusters as we now need to define a valid bucket arn.

Why is the role enabled if OPA is not?

PR #9494, opened 2025-06-03
Enable and allow to configure feature gate for image volumes

linki:
Image Volumes is a feature that allows to mount container images as volumes: https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/


linki:
👍

katyanna:
👍

PR #9508, opened 2025-06-04
Set increased WIZ Priority as default 

zaklawrencea:
The initial rollout of WIZ is now complete and there are no longer any Pending WIZ Pods in our environment. This PR changes the default WIZ priority to `system-node-critical`, which is inline with the rest of our core DaemonSet pods.


zaklawrencea:
:+1:

katyanna:
👍

PR #9521, opened 2025-06-05
Remove legacy cadvisor and related config-items

mikkeloscar:
This removes the legacy cadvisor container from the node-monitor daemonset and related resources. Also removes the cadvisor config-items.

This is done as no cluster depend on it anymore and instead rely on the metrics from kubelet.


mikkeloscar:
:+1:

katyanna:
👍

PR #9527, opened 2025-06-06
Add config item for controlling spot in Karpenter capacity-type

zaklawrencea:
This PR adds a config item: `karpenter_enable_spot` (true by default) which allows controlling whether Karpenter nodepools contain `spot` as a `capacity-type`. This config allows us to easily toggle whether Karpenter can provision spot instances alongside on-demand.

### DIFF
--- a/cluster/node-pools/worker-karpenter/provisioners.yaml
+++ b/cluster/node-pools/worker-karpenter/provisioners.yaml
@@ -249,7 +249,9 @@ spec:
         - key: "karpenter.sh/capacity-type"
           operator: In
           values:
+#{{ if eq .NodePool.ConfigItems.karpenter_enable_spot "true" }}
             - "spot"
+#{{ end }}
             - "on-demand"
         - key: "kubernetes.io/arch"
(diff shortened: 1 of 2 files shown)


cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -249,7 +249,9 @@ spec:
         - key: "karpenter.sh/capacity-type"
           operator: In
           values:
+#{{ if eq .Cluster.ConfigItems.karpenter_enable_spot "true" }}
```
mikkeloscar: Let's do this at node pool level e.g. `.NodePool.ConfigItems.karpenter_enable_spot`. That will allow us to set this at cluster level _or_ node pool level.

zaklawrencea: Good point, updated.

mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #9547, opened 2025-06-11
Make deployment-service's S3 bucket name configurable

linki:
Adds a config item for deployment-service's S3 bucket name. The default is the current value.

In order to shorten it in the future, we can configure the current value for all existing clusters in cluster-registry and then change the default for all clusters going forward. We could also do the opposite and only configure the value if the default exceeds the length limit.


linki:
👍

mikkeloscar:
:+1:

PR #9574, opened 2025-06-13
Add kube2iam to combined trust relationship

linki:
Also adds the kube2iam trust relationships to`CDP_OIDC_TRUST_RELATIONSHIP_TEMPLATE` variable.

Doesn't quite work yet. Need a way to know EKS cluster local ID (or cluster alias) from the legacy cluster context.


cluster/config-defaults.yaml
```diff
@@ -1273,6 +1273,7 @@ eks_zalando_iam_aws_proxy_hpa_cpu_target: "80"
 eks_zalando_iam_aws_proxy_hpa_memory_target: "80"
 eks_okta_identity_provider: "true"
 eks_legacy_cluster_local_id: "kube-1"
+eks_eks_cluster_local_id: ""
```
linki, line 1276: This needs to be set for legacy clusters to the local ID of the EKS cluster (i.e. cluster alias). But we could generate the default that works for most clusters. Something like:

```
{{ if eq .Cluster.Environment "test" }}
eks_eks_cluster_local_id: "{{ .Cluster.InfrastructureAccount | getAWSAccountGroup }}-{{ .Cluster.Region | shortVersion }}-test"
{{ else }}
eks_eks_cluster_local_id: "{{ .Cluster.InfrastructureAccount | getAWSAccountGroup }}-{{ .Cluster.Region | shortVersion }}"
{{ end }}
```

linki:
Replaced in favour of https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/866

PR #9523, opened 2025-06-06
Update to Kubernetes v1.33

mikkeloscar:
Upgrade to Kubernetes v1.33

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.33.md#changelog-since-v1320

https://kubernetes.io/blog/2025/04/23/kubernetes-v1-33-release/

PR #9580, opened 2025-06-18
Remove TopologyAwareHints Feature Flag

mikkeloscar:
This removes the custom config for `TopologyAwareHints` featureflag and default to true which is the only option since Kubernetes v1.33. It's no longer possible to set the Feature flag.

This change must be rolled out _before_ #9523 otherwise `kube-proxy` will not get ready on the new master node during roll out


szuecs:
:+1:

mikkeloscar:
:+1:

PR #9582, opened 2025-06-19
expose-opa-ast-optimization-configuration-per-cluster

amak3:
Exposing OPA optimization configuration.

New configuration should be disabled by default and it should be possible to enable it per cluster. Enabling should only be possible if `skipper_open_policy_agent_enabled` is enabled.

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -317,6 +317,7 @@ spec:
           - "-enable-open-policy-agent"
           - "-open-policy-agent-config-template=/etc/skipper/open-policy-agent/opaconfig.yaml"
           - "-open-policy-agent-envoy-metadata=/etc/skipper/open-policy-agent/envoymetadata.json"
+          - "-enable-open-policy-agent-data-preprocessing-optimization={{ .Cluster.ConfigItems.skipper_open_policy_agent_data_preprocessing_optimization_enabled }}"
 {{ end }}
 {{ if or (eq .Cluster.ConfigItems.nlb_switch "pre") (eq .Cluster.ConfigItems.nlb_switch "exec") }}
           - "-forwarded-headers=X-Forwarded-For,X-Forwarded-Proto=https,X-Forwarded-Port=443"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -372,6 +372,7 @@ skipper_ingress_routegroup_crd_require_hosts: "true"
 # Set defaults values that would enable Open Policy Agent in a skipper filter
 skipper_open_policy_agent_enabled: "false"
 skipper_open_policy_agent_styra_token: ""
+skipper_open_policy_agent_data_preprocessing_optimization_enabled: "false"
 
 # Default timeout value in seconds for outgoing http calls from Open Policy Agent in a skipper filter
 skipper_open_policy_agent_styra_response_header_timeout: "2"


cluster/manifests/skipper/deployment.yaml
```diff
@@ -317,6 +317,10 @@ spec:
           - "-enable-open-policy-agent"
           - "-open-policy-agent-config-template=/etc/skipper/open-policy-agent/opaconfig.yaml"
           - "-open-policy-agent-envoy-metadata=/etc/skipper/open-policy-agent/envoymetadata.json"
+          - "-enable-open-policy-agent-data-preprocessing-optimization"
```
mefarazath: This should not be here i guess (since you this guarded by the flag `skipper_open_policy_agent_data_preprocessing_optimization_enabled` again in line 321)

amak3: ah, good catch, I missed it, thanks!

cluster/manifests/skipper/deployment.yaml
```diff
@@ -317,6 +317,9 @@ spec:
           - "-enable-open-policy-agent"
           - "-open-policy-agent-config-template=/etc/skipper/open-policy-agent/opaconfig.yaml"
           - "-open-policy-agent-envoy-metadata=/etc/skipper/open-policy-agent/envoymetadata.json"
+  {{ if eq .Cluster.ConfigItems.skipper_open_policy_agent_data_preprocessing_optimization_enabled "true" }}
+          - "-enable-open-policy-agent-data-preprocessing-optimization"
+  {{ end }}
```
AlexanderYastrebov: I think we can use it directly like

```
- "-enable-open-policy-agent-data-preprocessing-optimization={{ .Cluster.ConfigItems.skipper_open_policy_agent_data_preprocessing_optimization_enabled }}"
```

We can only use this flag when both canary and main use the version >= v0.22.51 that [introduced this](https://github.com/zalando/skipper/commit/91c0bae09f7e6c0cdc72ba639699384ccc73a0be).

AlexanderYastrebov: Please also squash commits and add `skipper: ...` prefix to the commit title

amak3: I think this would change the logic:

current logic: only if both OPA (`skipper_open_policy_agent_enabled`)  and `skipper_open_policy_agent_data_preprocessing_optimization_enabled` are enabled, the optimization flag is added as a separate argument

after a suggestion: it's possible to set 
`skipper_open_policy_agent_data_preprocessing_optimization_enabled` to "true", even if OPA is disabled, which is not expected behaviour. 

Enabling of `skipper_open_policy_agent_data_preprocessing_optimization_enabled` should only be possible if `skipper_open_policy_agent_enabled` is enabled.

AlexanderYastrebov: The flag is inside of `{{ if eq .Cluster.ConfigItems.skipper_open_policy_agent_enabled "true" }}` block in both cases, I suggest to remove if in favour of using the boolean value directly

amak3: I made changes, thank you 🙏

AlexanderYastrebov:
> We can only use this flag when both canary and main use the version >= v0.22.51 that introduced this.

Update https://github.com/zalando-incubator/kubernetes-on-aws/pull/9597 to v0.22.52  was rolled out to stable.

mefarazath:
👍🏻

amak3:
👍🏻

PR #9573, opened 2025-06-13
increase subdomain_max_length for stups-test

demonCoder95:
After #9479 , e2e tests in the `stackset-controller` broke as the hostnames there exceed this default limit of `57` for subdomain length. 

```go
time="2025-06-13T12:40:22Z" level=error msg="Unable to reconcile stack resources: admission webhook \"routegroup-admitter.teapot.zalan.do\" denied the request: subdomain 'd-9owpufu1wutzdgxfdrpytgzw1-stackset-update-add-routegroup' is too long, keep each subdomain at a max of 57 characters" controller=stackset namespace=d-9owpufu1wutzdgxfdrpytgzw1 stack=stackset-update-add-routegroup-v2 stackset=stackset-update-add-routegroup
```

It's tricky to change the domain names and since it's just an e2e test environment, it's easier to enable this limit to a higher value, to make the e2e tests work like they worked before enforcing this limit in the admission-controller.

PR #9585, opened 2025-06-20
drop subdomain_max_length config for stups-test

demonCoder95:
We decided to shorten the subdomain names instead. So, this change introduced in #9573 can be dropped. 
If it's really needed in the future, we can set it per cluster instead, instead of doing it here.


linki:
👍

demonCoder95:
👍

PR #9593, opened 2025-06-23
Wait twice as long for reserve pools

linki:
We currently wait the default of 5 min. for the deployment to be ready after scaling up and sometimes it takes longer to schedule those pods requiring ARM. Let's increase it by 2x.


linki:
👍

demonCoder95:
👍

PR #9601, opened 2025-06-24
Leverage new IAM properties on CLM's cluster object

linki:
Use new member variables on cluster objects provided by CLM to improve combined trust relationship. The old trust relationship is left in place until all applications have migrated away from it.


linki:
👍

mikkeloscar:
:+1:

PR #9620, opened 2025-06-30
feat: Add sandbox controller configuration, disabled by default.

ponimas:
Added the deployment manifest for the sandbox-controller component.

A new cluster configuration item, sandbox_controller_enabled, has been introduced to control its deployment. The controller is disabled by default, and setting this config item to "true" is required to enable its deployment in the cluster.

A conditional deletion rule was added to ensure the deployment is removed when the configuration item is not set to "true".


cluster/manifests/sandbox-controller/deployment.yaml
```diff
@@ -0,0 +1,42 @@
+# {{ $image := "container-registry.zalando.net/gwproxy/sandbox-controller:main-8" }}
+# {{ $version := index (split $image ":") 1 }}
+
+apiVersion: apps/v1
```
linki: This should be guarded by:

```
{{ if eq .Cluster.ConfigItems.sandbox_controller_enabled "true" }}
```

cluster/manifests/sandbox-controller/deployment.yaml
```diff
@@ -0,0 +1,42 @@
...
+        kubernetes-log-watcher/scalyr-parser: |
+          [{"container": "controller", "parser": "keyValue"}]
+        logging/destination: "{{ .Cluster.ConfigItems.log_destination_both }}"
+        # no metrics exposed so far
+        # prometheus.io/path: /metrics
+        # prometheus.io/port: "7979"
+        # prometheus.io/scrape: "true"
+    spec:
+      priorityClassName: "{{ .Cluster.ConfigItems.system_priority_class }}"
+      serviceAccountName: sandbox-controller
```
linki: This ServiceAccount also needs to be created.

linki:
What's the purpose of this controller?

cluster/manifests/sandbox-controller/deployment.yaml
ponimas: ~Could you point me to the docs on how to do that?~

ponimas:
> What's the purpose of this controller?

It will create routegroups/ingresses to allow testing in production.
[Solution design document](https://docs.google.com/document/d/1ytLltsASWhKRfPucAmw_8IWl5NdhRjHidxZ4A1V03nk/edit?tab=t.0#heading=h.cd1shwdl7itu)

cluster/manifests/sandbox-controller/sandbox_crd.yaml
```diff
@@ -0,0 +1,38 @@
+{{ if eq .Cluster.ConfigItems.sandbox_controller_enabled "true" }}
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
szuecs: I think we need to create this first before we create the rbac

szuecs:
:+1:

szuecs:
error is unrelated

ponimas:
👍

cluster/manifests/sandbox-controller/30-deployment.yaml
```diff
@@ -0,0 +1,44 @@
+# {{ $image := "container-registry.zalando.net/gwproxy/sandbox-controller:main-8" }}
+# {{ $version := index (split $image ":") 1 }}
```
AlexanderYastrebov, line 2: You may yaml-comment other template instructions like this elsewhere to make manifests a valid yaml and enable e.g. autoformatting.

PR #9629, opened 2025-07-01
Use cluster's member function instead of template functions

linki:
Replaces the various ways we use to get to the account ID by a single method.


mikkeloscar:
:+1:

linki:
👍

PR #9632, opened 2025-07-02
Make coredns tolerate the karpenter dedicated taint

mikkeloscar:
This fixes a bug for scheduling annotation dedicated nodes where coredns would not be scheduled because it didn't tolerate the new taint set on those nodes.

Scheduling of pods worked fine, but DNS doesn't work as there is no CoreDNS.


mikkeloscar:
:+1:

linki:
👍

PR #9633, opened 2025-07-03
Make deployment-service roles configurable

linki:
Add config items for several deployment-service roles. The intention is to override them with a shorter version in the short-term. A long-term term solution should change defaults.


linki:
👍

mikkeloscar:
:+1:

PR #9637, opened 2025-07-03
test skipper-ingress with eks karpenter ipv6 with hostNetwork and hostPort

ayruslore:
Test PR

Testing skipper ingress on eks with karpenter ipv6

With AWSCNI enabled and hostNetwork and hostPort enabled


ayruslore:
We need the load balancer ip address type to be dual stack so that we can add ipv6 target groups.

See error message from cf stack

```
This AWS::ElasticLoadBalancingV2::Listener resource is in a UPDATE_FAILED state.

Resource handler returned message: "You cannot associate an IPv6 target group with a load balancer that has an 'ipv4' IP 
address type (Service: ElasticLoadBalancingV2, Status Code: 400, Request ID: 1c8e8ce0-eefa-4ae4-b61a-95e72bca5812) 
(SDK Attempt Count: 1)" (RequestToken: ced9b7e0-78da-123f-803d-f48404feb18c, HandlerErrorCode: InvalidRequest)
```

PR #9476, opened 2025-05-27
optimize: skipper-ingress-redis to use karpenter node pools

szuecs:
optimize: skipper-ingress-redis to use karpenter node pools with instance type passed weights

PR #9642, opened 2025-07-03
Skipper ingress Redis scheduling annotations

mikkeloscar:
This implements scheduling annotations for skipper-ingress-redis such that a dedicated node pool is not needed, but it will land on dedicated nodes via dedicated scheduling annotation feature.

(Since kube-system currently is not covered by scheduling annotations, the annotation is not defined, but the nodeSelector/toleration which is normally injected by admission-controller, is defined manually)

The changes includes a few more optimizations:

* Set: `karpenter.sh/do-not-disrupt: "true"` which is the equivalent of `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` for karpenter
* Set minDomains in topologyspread which is just a best practice.
* Include the instance type priority from #9476

### DIFF
--- a/test/e2e/cluster_config.sh
+++ b/test/e2e/cluster_config.sh
@@ -166,7 +166,7 @@ EOFF
     - "g6.xlarge"
     - "g6.2xlarge"
     - "g6.4xlarge"
-    name: karpenter-gpu
+    name: karpenter-gpu-tesla
     profile: worker-karpenter
     min_size: 0
     max_size: 0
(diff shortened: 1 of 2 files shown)


test/e2e/cluster_config.sh
```diff
@@ -166,7 +166,7 @@ EOFF
     - "g6.xlarge"
     - "g6.2xlarge"
     - "g6.4xlarge"
-    name: karpenter-gpu
+    name: karpenter-gpu-tesla
```
szuecs, line 169: @mikkeloscar is this part of the PR?

mikkeloscar: This is part of the PR because it's the first time we run the "node pool less" config in e2e which is needed for `zalando.org/dedicated: skipper-ingress-redis`

We run this setup in all clusters already though.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #9647, opened 2025-07-04
pod-metric: adds availability zone label to pod metrics

madumalt:
## Description

This change will result in otel-collectors to get the `topology.kubernetes.io/zone:<availability-zone>` pod annotation and push it along with the pod metrics into Lightstep as an attribute of those metrics.


szuecs:
:+1:

madumalt:
👍

MustafaSaber:
Was this tested on playground or teapot?

PR #9655, opened 2025-07-08
Test enabling rolling of flannel

mikkeloscar:
This validates that updating the flannel image as done in #9371 will not fail the e2e introduced in #9315 validating that the updated flannel image has a fix for dropping traffic when pods are being rotated.


mikkeloscar:
Closing after validating that e2e passes with this version of flannel.

Once the new flannel version has been _fully_ rolled out then the daemonset can be changed to `RollingUpdate`

PR #9673, opened 2025-07-10
fabric-gateway: test rolling back

MustafaSaber:
testing fabric rolling back from versioned to unversioned


MustafaSaber:
This was done by @greeshma1196

PR #9677, opened 2025-07-11
Only configure main domain as cluster-domain for legacy clusters

linki:
Internal discussion revealed that the `cluster-domain` argument is merely used to detect/filter Ingress/RouteGroup domains on per-stack resources. The main hostname of the stackset isn't affected by this setting.

Since we intend to stop supporting the main hostname variants for individual stacks on EKS clusters, we can just drop it entirely from the list to achieve that.

Users can still use the cluster-internal DNS domain to access individual stacks.


test/e2e/apply/config.yaml
```diff
@@ -26,6 +25,7 @@ metadata:
   name: "{{{APPLICATION}}}-{{{COMPONENT}}}-config-provider-aws"
 data:
   CLUSTER_PROVIDER: "zalando-aws"
+  CLUSTER_DOMAIN_INTERNAL: "teapot-e2e.zalan.do"
```
linki, line 28: `CLUSTER_DOMAIN_INTERNAL` is just the name of the variable to give a second cluster domain to `stackset-controller`. It doesn't actually need to be internal.

cluster/manifests/stackset-controller/deployment.yaml
```diff
@@ -55,8 +55,10 @@ spec:
         - "--sync-ingress-annotation=zalando.org/aws-waf-web-acl-id"
         - "--sync-ingress-annotation=kubernetes.io/ingress.class"
 {{ end}}
-        - "--cluster-domain={{ .Values.hosted_zone }}"
         - "--cluster-domain=ingress.cluster.local"
+{{- if eq .Cluster.Provider "zalando-aws" }}
+        - "--cluster-domain={{ .Values.hosted_zone }}"
+{{- end }}
```
linki, line 61: I moved this down so that the common domain between the providers is first.

linki:
👍

myaser:
👍🏻

PR #9706, opened 2025-07-24
ingress-controller: comment template actions

AlexanderYastrebov:
Comment template actions to make file a valid YAML.

Follow up on #9092


ponimas:
👍

ayruslore (approved):

PR #9718, opened 2025-07-29
Change buffer type default event limit number

adetunjiakintundeakinde:
# Description
Change buffer type default event limit number from 1000 to 10000

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -380,7 +380,7 @@ skipper_open_policy_agent_styra_response_header_timeout: "2"
 # Decision logging use event buffer type
 skipper_open_policy_agent_decision_logs_buffer_type_event_enable: "false"
 # Decision logging sets the maximum number of decision log events that can be buffered before being dropped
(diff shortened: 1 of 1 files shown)


cluster/config-defaults.yaml
```diff
@@ -380,7 +380,7 @@ skipper_open_policy_agent_styra_response_header_timeout: "2"
 # Decision logging use event buffer type
 skipper_open_policy_agent_decision_logs_buffer_type_event_enable: "false"
 # Decision logging sets the maximum number of decision log events that can be buffered before being dropped
-skipper_open_policy_agent_decision_logs_buffer_type_event_limit: "1000"
+skipper_open_policy_agent_decision_logs_buffer_type_event_limit: "10000"
```
MustafaSaber, line 383: Do we have any metric that this won't break anything?

adetunjiakintundeakinde: The default buffer size is 10000 for styra and this new buffer type is not in use so nothing should break

demonCoder95:
👍

MustafaSaber:
👍

PR #9662, opened 2025-07-09
Make WIZ Connector default config items conditional to cluster provider

hagenmathis:
makes the default config items conditional to prevent deployment of the broker and its associated resources to eks clusters. Neither broker nor connector should be needed with those, as access to Wiz is granted using [access entries](https://docs.aws.amazon.com/eks/latest/userguide/access-entries.html)

PR #9724, opened 2025-07-29
Only enable wiz broker components on non eks clusters

hagenmathis:
Reopening #9662 to debug the error in the e2e tests. From the original PR:
makes the default config items conditional to only deploy the broker and its associated resources to non eks clusters. Neither broker nor connector should be needed with eks, as access to Wiz is granted using [access entries](https://docs.aws.amazon.com/eks/latest/userguide/access-entries.html)


hagenmathis:
:+1:

zaklawrencea (approved):

PR #9725, opened 2025-07-29
add EFS SG definition and rule for EKS cluster

demonCoder95:
This adds a SecurityGroup definition for EFS traffic and then adds an ingress rule from the EKSWorkerSecurityGroup. This is based on how we do it for legacy non-EKS clusters. This is required to enable EFS backed PersistentVolumes in EKS.

I created a somewhat "duplicate" `EKSEFSWorkerSecurityGroup` so that it's easier to drop support for legacy clusters in the future and we don't have overlapping resource between the two clusters.


demonCoder95:
👍

ni9hty:
:+1:

PR #9732, opened 2025-07-30
skipper/hostname-credentials-controller: disable retry

AlexanderYastrebov:
Do not retry cronjob on failure once (backoffLimit) but wait one minute until next scheduled time.
E.g. it makes no sense to retry in case of `429 Too Many Requests` from API server.


greeshma1196:
👍

AlexanderYastrebov:
:+1:

PR #9735, opened 2025-07-31
Add e2e tests for cdp and deployment-service secret read permissions

katyanna:
This PR adds e2e tests to verify that CDP and deployment-service can read secrets from any namespace (including kube-system) and that this permission is later revoked by the Admission Controller.

- RBAC: Allows CDP and deployment-service to read secrets across namespaces
- Admission Controller: Rewrites these permissions to deny further access to read secrets

This test coverage helps validate the secret access workflow end-to-end on the redesign of RBAC for EKS.

Ref.: https://github.com/zalando-incubator/kubernetes-on-aws/pull/8871


test/e2e/authorization.go
```diff
@@ -885,6 +931,69 @@ var _ = g.Describe("Authorization via admission-controller [RBAC] [Zalando]", fu
...
+
+			g.It("should deny secret read access to user namespace", func() {
+				_, err := client.CoreV1().Secrets(testSecret.Namespace).Get(context.Background(), testSecret.Name, metav1.GetOptions{})
+				gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("read operations are forbidden")))
+            })
+
+			g.It("should deny secret read access to kube-system namespace", func() {
+				_, err := client.CoreV1().Secrets(systemSecret.Namespace).Get(context.Background(), systemSecret.Name, metav1.GetOptions{})
+				gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("read operations are forbidden")))
+			})
```
linki, line 1001: Hmm, that doesn't add up. First of all, why do we "deny" in both cases? Isn't the point of all this stuff the ability to give permissions on all but one namespace (or some variation of this)? If we want to deny access in all namespaces we could just use RBAC.

Secondly, admission-controller isn't involved in GET requests, so these won't go through admission-controller anyways.

Can we define again what we actually wanted to achieve?

PR #9737, opened 2025-07-31
[ALPHA] [HOTFIX] Update Cluster Autoscaler version

zaklawrencea:
dev-to-alpha e2es do not pass as the cluster creation step initially tries to use the old CA image version which does not work.

This PR introduces the CA version update separately.


zaklawrencea:
:+1:

myaser:
👍🏻

PR #9741, opened 2025-07-31
exclude g6f instance family from Karpenter nodepools

demonCoder95:
The `g6f` instance family has issues when scheduling workloads, we need to disable it by default and test it properly before we can make it generally available. For now, adding the config-item so we can block provisioning these instances.


zaklawrencea:
:+1:

demonCoder95:
👍

PR #9742, opened 2025-07-31
exclude g6f instance family from Karpenter nodepools

demonCoder95:
This brings the patch in https://github.com/zalando-incubator/kubernetes-on-aws/pull/9741 as hotfix to `stable` so we can fix an upcoming issue in production


demonCoder95:
This fix has been merged all the way to `stable` now, so the hot-fix branch can be dropped.

PR #9750, opened 2025-08-05
[DO-NOT-MERGE] try unskipping mysql test

demonCoder95:
Need to check if this e2e test still fails or if it works.


demonCoder95:
Seems like the test still fails. We've decided to keep it skipped, as functionality provided by other e2e tests covers our use-cases sufficiently.

demonCoder95:
Looking closer at the test, it only runs the image to check it gets a cluster ready. So, we can skip the test. Will close the PR.

PR #9768, opened 2025-08-15
Add string docs for RBAC read-only testing

katyanna:
Adds a docstring to the RBAC read-only role tests explaining that non-mutating requests bypass the admission controller and are enforced solely by RBAC. This clarifies why admission controller testing is unnecessary for this role.


katyanna:
:+1:

demonCoder95:
👍

PR #9770, opened 2025-08-18
update flannel DaemonSet strategy to RollingUpdate

demonCoder95:
This is needed to affirm rolling out the fix for the network disruption during node rotation.

### DIFF
--- a/cluster/manifests/flannel/daemonset.yaml
+++ b/cluster/manifests/flannel/daemonset.yaml
@@ -10,7 +10,7 @@ metadata:
     component: flannel
 spec:
   updateStrategy:
-    type: OnDelete
+    type: RollingUpdate
   selector:
     matchLabels:
       daemonset: kube-flannel


katyanna:
:+1:

linki:
For some reason we postponed this until now. We wanted to test it properly I guess? What's the trigger to do it now?

mikkeloscar:
> For some reason we postponed this until now. We wanted to test it properly I guess? What's the trigger to do it now?

This was postponed because all nodes need to run the new (current) version of flannel _before_ RollingUpdate can be enabled. If not, then it would start to roll the old pods which has the bug and causes the issue. Since flannel is in OnDelete at the moment pods will only be updated when nodes rotate, so we essentially need to ensure all nodes have been rotated once before. Not sure if there was a node update in the last month?

(This is not caught by e2e because the e2e env already runs the fixed version, so even if it rolls during e2e it doesn't trigger the bug).

zaklawrencea:
> Not sure if there was a node update in the last month

No, we are still running Nodes with the previous flannel version. In which case it would be easier to wait until after v1.33 has forced a full rotation before moving this forward?

linki:
Then let's just wait for v1.33.

szuecs:
@linki we have now 1.33 :)

tcondeixa:
now all cluster are running kubernetes 1.33, so we can merge this one @linki

tcondeixa:
👍

PR #9777, opened 2025-08-21
Add Log Group with expiring retention period for EKS Control Plane logging

zaklawrencea:
Add a new Log Group `ControlPlaneLogGroup` to the main CF stack which allows us to configure the retention period of the EKS Control Plane logs. The name of the Log Group is fixed by EKS and we cannot point the logs to a new Log Group.

For existing EKS clusters, as the Log Group will first need to be manually imported into the Stack, this PR also introduces a helper config item `eks_control_plane_logging_migration` to control when the resource will be applied by CLM.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -158,6 +158,9 @@ Resources:
     Type: 'AWS::EC2::SecurityGroup'
   EKSCluster:
     Type: AWS::EKS::Cluster
+{{- if eq .Cluster.ConfigItems.eks_control_plane_logging "true" }}
+    DependsOn: ControlPlaneLogGroup
+{{- end }}
     Properties:
       Name: "{{.Cluster.Name}}"
       Version: "1.32"
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -159,7 +159,9 @@ Resources:
   EKSCluster:
     Type: AWS::EKS::Cluster
 {{- if eq .Cluster.ConfigItems.eks_control_plane_logging "true" }}
+{{- if eq .Cluster.ConfigItems.eks_control_plane_logging_migration "true" }}
     DependsOn: ControlPlaneLogGroup
+{{- end }}
 {{- end }}
     Properties:
       Name: "{{.Cluster.Name}}"
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -3023,6 +3023,16 @@ Resources:
                 - BucketArn: !GetAtt AuditTrailBucket.Arn
 {{- end }}
 
+{{- if eq .Cluster.ConfigItems.eks_control_plane_logging "true" }}
+{{- if eq .Cluster.ConfigItems.eks_control_plane_logging_migration "true" }}
+  ControlPlaneLogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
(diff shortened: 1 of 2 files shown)


tcondeixa:
I think you need to add a forced dependency `DependsOn` in the EKS cluster resource to be sure the log group is created before. I don't think we have any implicit dependency at the moment with any attribute from the log group in the cluster.
https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-dependson.html

Adding the `DependsOn` here also with the templated if conditions
https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/cluster.yaml#L160

cluster/cluster.yaml
```diff
@@ -158,6 +158,9 @@ Resources:
     Type: 'AWS::EC2::SecurityGroup'
   EKSCluster:
     Type: AWS::EKS::Cluster
+{{- if eq .Cluster.ConfigItems.eks_control_plane_logging "true" }}
+    DependsOn: ControlPlaneLogGroup
```
tcondeixa, line 163: It's better you also add the other condition here temporary while doing the import. If it was not migrated it will not be there, and this dependency might fail the CF stack.
`{{- if eq .Cluster.ConfigItems.eks_control_plane_logging_migration "true" }}`

zaklawrencea:
> I think you need to add a forced dependency DependsOn in the EKS cluster resource to be sure the log group is created before

I think you're right. The EKS cluster takes a while to provision and the Log Group has "won the race" in the tests, but a hard dependency is a good idea.

tcondeixa:
👍

zaklawrencea:
:+1:

cluster/cluster.yaml
```diff
@@ -3023,6 +3028,16 @@ Resources:
                 - BucketArn: !GetAtt AuditTrailBucket.Arn
 {{- end }}
 
+{{- if eq .Cluster.ConfigItems.eks_control_plane_logging "true" }}
+{{- if eq .Cluster.ConfigItems.eks_control_plane_logging_migration "true" }}
+  ControlPlaneLogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: "/aws/eks/{{.Cluster.LocalID}}/cluster"
```
mikkeloscar, line 3036: Maybe we can use `.Cluster.Name` here so it matches the value here: https://github.com/zalando-incubator/kubernetes-on-aws/blob/a70f7d7e2298ea32288ff4c28605798b70aa8272/cluster/cluster.yaml#L167

tcondeixa: @zaklawrencea mentioned that EKS enforce that log group name, so we cannot define a different name

mikkeloscar: I understand, but those two variables should render to the same value in EKS, it's just more logical to have .Cluster.Name in both places.

zaklawrencea: Fair enough, I didn't have `Cluster.Name` in mind when I made the PR, but if they render to the same value then when we make the follow up to remove the `eks_control_plane_logging_migration` config item I can also update this.

zaklawrencea: Var was updated in https://github.com/zalando-incubator/kubernetes-on-aws/pull/9805.

PR #9780, opened 2025-08-22
move  credprov rollout date config  to the default configs

tcondeixa:
This config item does not need to be change and it can be in the default with current date


tcondeixa:
closing since this it's not enough to have it for templating, we need the value in registry

PR #9783, opened 2025-08-26
deploy: polarsignals to skipper-ingress nodes

szuecs:
deploy: polarsignals to skipper-ingress nodes
deploy is feature toggled

first deploy internal apikey

### DIFF
--- a/cluster/manifests/polarsignals/02-rbac.yaml
+++ b/cluster/manifests/polarsignals/02-rbac.yaml
@@ -1,0 +1,55 @@
+{{ if eq .Cluster.ConfigItems.polarsignals_enabled "true" }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    app.kubernetes.io/component: continuous-profiler
+    app.kubernetes.io/instance: polarsignals-agent
+    app.kubernetes.io/name: polarsignals-agent
+  name: polarsignals-agent
+  namespace: polarsignals
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/component: continuous-profiler
+    app.kubernetes.io/instance: polarsignals-agent
+    app.kubernetes.io/name: polarsignals-agent
+  name: polarsignals-agent
+  namespace: polarsignals
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+    verbs:
+      - list
+      - watch
+  - apiGroups:
+      - ""
+    resources:
+      - nodes
+    verbs:
+      - get
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  labels:
+    app.kubernetes.io/component: continuous-profiler
+    app.kubernetes.io/instance: polarsignals-agent
+    app.kubernetes.io/name: polarsignals-agent
+  name: polarsignals-agent
+  namespace: polarsignals
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: polarsignals-agent
+subjects:
+  - kind: ServiceAccount
+    name: polarsignals-agent
+    namespace: polarsignals
+---
+{{ end }}
--- a/cluster/manifests/polarsignals/02-rbac.yaml
+++ b/cluster/manifests/polarsignals/02-rbac.yaml
@@ -4,9 +4,8 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   labels:
-    app.kubernetes.io/component: continuous-profiler
-    app.kubernetes.io/instance: polarsignals-agent
-    app.kubernetes.io/name: polarsignals-agent
+    component: agent
+    application: polarsignals
   name: polarsignals-agent
   namespace: polarsignals
 ---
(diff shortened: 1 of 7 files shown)


cluster/manifests/polarsignals/daemonset.yaml
```diff
@@ -0,0 +1,119 @@
+{{ if eq .Cluster.ConfigItems.polarsignals_enabled "true" }}
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    app.kubernetes.io/component: continuous-profiler
+    app.kubernetes.io/instance: polarsignals-agent
+    app.kubernetes.io/name: polarsignals-agent
```
mikkeloscar: let's add `application` and `component` labels here and at pod spec level.

cluster/manifests/polarsignals/daemonset.yaml
```diff
@@ -0,0 +1,119 @@
+{{ if eq .Cluster.ConfigItems.polarsignals_enabled "true" }}
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
```
mikkeloscar, line 5: I would recommend to run with this annotation:

https://github.com/zalando-incubator/kubernetes-on-aws/blob/a70f7d7e2298ea32288ff4c28605798b70aa8272/cluster/manifests/kubenurse/daemonset.yaml#L10-L11

This means that this daemonset won't be checked for readiness to mark the node ready, this ensures that if this tool breaks for whatever reason it won't block scale up of nodes. I would consider this not critical to run at node startup.

cluster/manifests/polarsignals/02-rbac.yaml
```diff
@@ -0,0 +1,55 @@
...
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    app.kubernetes.io/component: continuous-profiler
+    app.kubernetes.io/instance: polarsignals-agent
+    app.kubernetes.io/name: polarsignals-agent
+  name: polarsignals-agent
+  namespace: polarsignals
```
mikkeloscar, line 10: This service account should be defined in admission-controller config, to allow it to run privileged: Similar like we do here: https://github.com/zalando-incubator/kubernetes-on-aws/blob/a70f7d7e2298ea32288ff4c28605798b70aa8272/cluster/manifests/02-admission-control/config.yaml#L115-L117

szuecs: thanks, done

szuecs:
:+1:

cluster/manifests/polarsignals/daemonset.yaml
```diff
@@ -0,0 +1,117 @@
...
+            - --remote-store-bearer-token-file=/var/polarsignals-agent/token
+            - --debuginfo-strip
+            - --debuginfo-temp-dir=/tmp
+            - --debuginfo-upload-cache-duration=5m
+          env:
+            - name: NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+          image: ghcr.io/parca-dev/parca-agent:v0.39.1 # TODO(sszuecs) use local build
```
mikkeloscar: I guess this is missing to be updated!

szuecs: fixed

szuecs:
:+1:

I think we can ignore the error as I see routegroup is there and created NLB/ALB + DNS record and it shows that up to skipper-ingress it works. And we have e2e test completed for eks and the PR does not really change the behavior for e2e tests at all.

szuecs:
:+1:

cluster/manifests/polarsignals/daemonset.yaml
```diff
@@ -0,0 +1,117 @@
...
+              name: bpffs
+            - mountPath: /var/run/dbus/system_bus_socket
+              name: dbus-system
+            - mountPath: /var/polarsignals-agent
+              name: token
+      hostPID: true
+      serviceAccountName: polarsignals-agent
+      nodeSelector:
+        kubernetes.io/os: linux
+        dedicated: skipper-ingress
```
MustafaSaber, line 83: why this needs to run on dedicated nodepools?

cluster/manifests/polarsignals/daemonset.yaml
```diff
@@ -0,0 +1,117 @@
...
+        - hostPath:
+            path: /sys/kernel/debug
+          name: debugfs
+        - hostPath:
+            path: /var/run/dbus/system_bus_socket
+          name: dbus-system
+        - secret:
+            secretName: polarsignals-agent
+          name: token
+{{ end }}
```
MustafaSaber, line 117: missing newline?

cluster/manifests/polarsignals/daemonset.yaml
szuecs: This needs to run on skipper-ingress nodes and not any other nodes.

cluster/manifests/polarsignals/daemonset.yaml
szuecs: It doesn't matter

cluster/manifests/polarsignals/daemonset.yaml
mikkeloscar: In the pilot phase it will be limited to only the skipper-ingress use case.

mikkeloscar:
:+1:

PR #9784, opened 2025-08-26
Adding playground poweruser policy to deployment services

Anupama-P:
Adding playground poweruser policy to deployment services for CI/CD role to have the same permissions as users.


tcondeixa:
👍

Anupama-P:
👍

PR #9791, opened 2025-08-28
Polarsignals ds fixes

mikkeloscar:
This adds some fixes for the polarsignals daemonset instroduced in #9783 

1. Fix image name (was missing `main-1` suffix in version)
2. Drop NoExecute toleration which our admission-controller doesn't allow as it would mean the DS _could_ run on master nodes. The DS is deployed in a non-system namespace `polarsignals` and therefore not allowed. Alternative solution is to make polarsignals namespace a system namespace. This requires a change in admission-controller.
3. Drop the `/bin/parca-agent` from `args` it's already defined as command in the dockerfile
4. Remove Readiness Probe which fails and is also not defined upstream: https://www.parca.dev/docs/kubernetes/#setting-up-parca-agent

### DIFF
--- a/cluster/manifests/polarsignals/daemonset.yaml
+++ b/cluster/manifests/polarsignals/daemonset.yaml
@@ -37,7 +37,7 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
-          image: container-registry.zalando.net/gwproxy/parca-agent:v0.41.0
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

szuecs:
Interesting , I think I copied from their install docs.
@brancz fyi

mikkeloscar:
:+1:

mikkeloscar:
(merged since this is anyway not covered by e2e, can be tested in teapot, I also tested manually)

PR #9801, opened 2025-08-28
Allow downscaling statefulsets

mikkeloscar:
Optionally allow downscaling `statefulsets` via `kube-downscaler`.

This was previously disabled in #3179

The intention is to re-visit this and see if there is an opportunity for easy cost savings. It's still disabled by default but can be enabled per cluster.


mikkeloscar:
:+1:

tcondeixa (approved):

PR #9804, opened 2025-08-29
fix: parca-agent restart oom

szuecs:
fix: parca-agent restart oom
fix: parca-agent was not scraped by prometheus

### DIFF
--- a/cluster/manifests/polarsignals/daemonset.yaml
+++ b/cluster/manifests/polarsignals/daemonset.yaml
@@ -17,6 +17,10 @@ spec:
       application: polarsignals
   template:
     metadata:
+      annotations:
+        prometheus.io/path: /metrics
+        prometheus.io/port: 7071
+        prometheus.io/scrape: "true"
       labels:
         component: agent
         application: polarsignals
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

cluster/manifests/polarsignals/daemonset.yaml
```diff
@@ -17,6 +17,10 @@ spec:
       application: polarsignals
   template:
     metadata:
+      annotations:
+        prometheus.io/path: /metrics
```
mikkeloscar, line 21: I don't think this will work because prometheus is configured to only detect pods in `kube-system` iirc.

mikkeloscar:
:+1:

cluster/manifests/polarsignals/daemonset.yaml
szuecs: good point will add `polarsignals` to the namespace to be scraped, then.

szuecs: I validated and it should work out of the box: https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/prometheus/configmap.yaml#L274-L283

PR #9805, opened 2025-08-29
Drop additional EKS control plane log config item after migration

zaklawrencea:
Follow up to https://github.com/zalando-incubator/kubernetes-on-aws/pull/9777

Now that the Log Groups have been imported into the Cluster Stacks, we can drop the helper config item: `eks_control_plane_logging_migration`. Also update the template variable for the LogGroupName to `Cluster.Name`.

This change also temporarily stops the control plane logging in e2e clusters to solve the issue of the Cluster Stack attempting to create a conflicting Resource. Currently, on the first parse of the create-cluster-eks step, the EKS Cluster is created as well as the Control Plane Log Group, however the Log Group is not part of the Stack. Upon applying the PR changes, the Log Group resource fails to create as a Log Group with the same name (created on the first parse) already exists but was never imported. A follow up PR will be created to re-enable the logging in e2e. This issue will then be avoided as the Log Group will not exist in new e2e clusters on the first parse and will only be created upon the PR changes being applied. From then on, the Log Group will already be part of the Cluster Stack.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -159,9 +159,7 @@ Resources:
   EKSCluster:
     Type: AWS::EKS::Cluster
 {{- if eq .Cluster.ConfigItems.eks_control_plane_logging "true" }}
-{{- if eq .Cluster.ConfigItems.eks_control_plane_logging_migration "true" }}
     DependsOn: ControlPlaneLogGroup
-{{- end }}
 {{- end }}
     Properties:
       Name: "{{.Cluster.Name}}"
@@ -2337,8 +2335,8 @@ Resources:
             Version: 2012-10-17
           PolicyName: root
       RoleName: "{{.Cluster.LocalID}}-etcd-backup"
-{{- end }}
     Type: 'AWS::IAM::Role'
+{{- end }}
   StaticEgressControllerIAMRole:
     Properties:
       AssumeRolePolicyDocument: !Sub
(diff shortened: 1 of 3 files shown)


cluster/cluster.yaml
```diff
@@ -2337,8 +2335,8 @@ Resources:
             Version: 2012-10-17
           PolicyName: root
       RoleName: "{{.Cluster.LocalID}}-etcd-backup"
-{{- end }}
     Type: 'AWS::IAM::Role'
+{{- end }}
```
zaklawrencea, line 2346: @tcondeixa this misplaced conditional end was responsible for the duplicated `Type: 'AWS::IAM::Role'` in the template :)

cluster/cluster.yaml
```diff
@@ -3035,14 +3033,12 @@ Resources:
 {{- end }}
 
 {{- if eq .Cluster.ConfigItems.eks_control_plane_logging "true" }}
-{{- if eq .Cluster.ConfigItems.eks_control_plane_logging_migration "true" }}
   ControlPlaneLogGroup:
     Type: AWS::Logs::LogGroup
     Properties:
-      LogGroupName: "/aws/eks/{{.Cluster.LocalID}}/cluster"
+      LogGroupName: "/aws/eks/{{.Cluster.Name}}/cluster"
       RetentionInDays: 545
```
tcondeixa: @mikkeloscar would not be enough to have 30 days retention as we have with non EKS clusters? this might save us costs on long run the more EKS clusters are adopted.

mikkeloscar: I was thinking about this. For apiserver, scheduler, controller-manager etc. yes, 30 days should be enough. For audit logs we should have longer, at least until/if we have another place where they will be stored.

I don't know if it can be further segmented or if they are all in the same "group"? 🤔

tcondeixa: let's keep this value for now and we can adjust later

tcondeixa:
👍

tcondeixa:
👍

mikkeloscar:
:+1:

PR #9817, opened 2025-09-02
Add eks fis v2 support

mikkeloscar:
This sets up a new implementation of EKS+FIS which is simpler to manage than #8943

## TODO

* [x] Dynamically provision service account + role per namespace.

### DIFF
--- a/cluster/manifests/eks-fis/01-rbac.yaml
+++ b/cluster/manifests/eks-fis/01-rbac.yaml
@@ -1,12 +1,4 @@
 {{- if eq .Cluster.ConfigItems.eks_fis_support_enabled "true" }}
-# {{ range $namespace := split .Cluster.ConfigItems.eks_fis_namespaces "," }}
-kind: ServiceAccount
-apiVersion: v1
-metadata:
-  namespace: "{{ $namespace }}"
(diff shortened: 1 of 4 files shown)


mikkeloscar:
:+1:

mikkeloscar:
:+1:

cluster/manifests/eks-fis/01-rbac.yaml
```diff
@@ -1,12 +1,4 @@
 {{- if eq .Cluster.ConfigItems.eks_fis_support_enabled "true" }}
-# {{ range $namespace := split .Cluster.ConfigItems.eks_fis_namespaces "," }}
```
demonCoder95, line 2: Is the config-item `eks_fis_namespaces` used anywhere else? I see it's being removed from 2 places now.

 If not, we should also clean it up, if not in this PR then a follow-up.

demonCoder95:
👍

cluster/manifests/eks-fis/01-rbac.yaml
mikkeloscar: Yeah, let me do it in a follow up!

PR #9824, opened 2025-09-03
fg: make `ValidatingAdmissionPolicy` customizable

MustafaSaber:
This will allow us to control which clusters to enable. And disable on e2e clusters.

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -30,6 +30,14 @@ post_apply:
   namespace: cron
   kind: LimitRange
 {{ end }}
+{{ if ne .Cluster.ConfigItems.fabric_gateway_custom_validation_enabled "true" }}
+- name: "fabricgateway-policy.zalando.org"
+  namespace: default
+  kind: ValidatingAdmissionPolicy
+- name: "fabricgateway-policy.zalando.org"
+  namespace: default
+  kind: ValidatingAdmissionPolicyBinding
+{{ end }}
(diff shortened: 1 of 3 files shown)


greeshma1196:
👍

cluster/manifests/deletions.yaml
```diff
@@ -30,6 +30,14 @@ post_apply:
...
   kind: LimitRange
 {{ end }}
+{{ if ne .Cluster.ConfigItems.fabric_gateway_custom_validation_enabled "true" }}
+- name: "fabricgateway-policy.zalando.org"
+  namespace: default
+  kind: ValidatingAdmissionPolicy
+- name: "fabricgateway-policy.zalando.org"
+  namespace: default
+  kind: ValidatingAdmissionPolicyBinding
+{{ end }}
```
madumalt, line 40: I do not understand this part. May I ask why do we want this part here?

MustafaSaber: let's say you enable this resource on `cluster-x` then make the config-item false, without this deletions file the resource won't be deleted from the cluster, it's like you do `kubectl apply -f ./deploy`  then you remove the yaml file from `deply` folder and do `kubectl apply -f ./deploy` it won't be applied again but also it won't be deleted

madumalt: got it, thanks for the explanation!

madumalt (approved):

szuecs:
:+1:

PR #9827, opened 2025-09-04
Make the IAM proxy's endpoint available via the alias

linki:
If `Alias != LocalID`, let's use the Alias for zalando-iam-proxy's endpoint.


demonCoder95:
👍

linki:
Nevermind, let's follow the old approach here as well.

PR #9840, opened 2025-09-09
[EKS] Manage kube-proxy as an addon

mikkeloscar:
Manage `kube-proxy` as an EKS addon.

The motivation for this is that currently `kube-proxy` is not managed on EKS meaning it's set up when the cluster is created by EKS, but afterwards it's not updated. We currently run v1.31 in some v1.33 clusters.

By adding it as an addon we can update the version over time.

The change in the admission controller config is needed to allow EKS to delete the addon during cluster deletion.


tcondeixa (approved):

mikkeloscar:
:+1:

PR #9841, opened 2025-09-09
eks-pod-identity-agent: Update to latest version

mikkeloscar:
Update to latest version of the `eks-pod-identity-agent` addon. Without this the version stays at the version which was available when the cluster was created.


mikkeloscar:
:+1:

tcondeixa (approved):

PR #9848, opened 2025-09-09
Disable gr6f instance-types

mikkeloscar:
Disable `gr6f` instance types because:

1. Has the same karpenter scheduling issue as `g6f`
2. Doesn't work on our AMI because of no available tesla driver (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/install-nvidia-driver.html#nvidia-driver-types)
3. Documented use case is video encoding, not the common use case for GPUs at Zalando (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/install-nvidia-driver.html#nvidia-driver-types)


mikkeloscar:
:+1:

tcondeixa (approved):

PR #9860, opened 2025-09-10
fabric-gateway: remove namespace for validation

MustafaSaber:
The resource shouldn't be namespaced.


mikkeloscar:
:+1:

MustafaSaber:
👍

MustafaSaber:
I think we can merge since it's only deletion and we don't need the e2e for this.

PR #9887, opened 2025-09-12
kube-proxy: verbose level 1

mikkeloscar:
In Kubernetes v1.33, kube-proxy has new verbose 2 level logging introduced: https://github.com/kubernetes/kubernetes/blob/74cdb4273add43f53ddcad2de8ea9fd93c810dc4/pkg/proxy/topology.go#L192-L195

This leads to high (top 10) log volume after rolling out to test clusters.

Set the default to verbose: 1 to reduce the log volume.


mikkeloscar:
:+1:

tcondeixa (approved):

PR #9919, opened 2025-09-18
Exclude kube-proxy in admission config

mikkeloscar:
Follow up to #9840 

We need to also exclude `kube-proxy` service account from the admission-controller, otherwise the kube-proxy addon fails to delete during cluster deletion and thus the CF Stack deletion fails.


mikkeloscar:
:+1:

ayruslore (approved):

PR #9922, opened 2025-09-18
Do not allow Karpenter to disrupt nodes due to Drift

zaklawrencea:
We rely on internal components to implement the same logic as Drift disruption by Karpenter. Additionally, Karpenter does not respect our internal features (e.g. to block node disruption during cluster updates) and SLAs for forced node rotations.

This PR configures the Karpenter disruption budget to disable node disruption due to Drift.

### DIFF
--- a/cluster/node-pools/worker-karpenter/provisioners.yaml
+++ b/cluster/node-pools/worker-karpenter/provisioners.yaml
@@ -278,6 +278,9 @@ spec:
     # limit the maximum number of nodes that can be removed from the NodePool at once to 10 nodes for clusters with gte 200 nodes
     budgets:
       - nodes: "5%"
+        reasons: 
+        - "Empty"
+        - "Underutilized"
       - nodes: "10"
         reasons: 
         - "Empty"
--- a/cluster/node-pools/worker-karpenter/provisioners.yaml
+++ b/cluster/node-pools/worker-karpenter/provisioners.yaml
@@ -279,6 +279,13 @@ spec:
     budgets:
       - nodes: "5%"
       - nodes: "10"
+        reasons: 
+        - "Empty"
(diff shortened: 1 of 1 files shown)


cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -279,6 +279,13 @@ spec:
     budgets:
       - nodes: "5%"
```
tcondeixa, line 280: don't you also need to define the `reasons` for the `5%`?

zaklawrencea: ~I believe this is stacked with the `"10"` below, with the more restrictive one taking priority (either 5% or 10 nodes) and applied for both of the reasons specified underneath.~

zaklawrencea: Actually I may need to update this.

Now I read it as:
- 5% is applied to all disruptions
- 10 nodes applies to those two specified reasons
- 0 nodes applies to the `Drifted` reason.

zaklawrencea: I originally had only:

```yaml
budgets:
- nodes: "5%"
- nodes: "10"
- nodes: "0"
  reasons: 
  - "Drifted"
```

but was trying to be explicit with the additional reasons to avoid surprises in the future.

tcondeixa: I didn't check the spec, but I would read it as you mentioned.
It's safer to also update the `5%`

tcondeixa (approved):

mikkeloscar:
:+1:

PR #9950, opened 2025-09-29
docs: Add field descriptions to sandbox CRD spec

ponimas:
Enhanced the sandbox Custom Resource Definition by adding descriptions for the testContext, sourceHosts, and target fields. The descriptions clarify how testContext is used for request routing with X-Zalando-Client-Id headers, explain that sourceHosts defines which traffic gets rerouted when testContext matches, and specify that target represents the destination host for sandboxed traffic.


ponimas:
👍

MustafaSaber:
👍

PR #9970, opened 2025-10-01
Exclude kube-proxy from admission-controller

mikkeloscar:
Similar to #9919 extending to one more admission endpoint which can fail during update:

image


mikkeloscar:
:+1:

tcondeixa (approved):

PR #9975, opened 2025-10-01
Allow enabling HPAConfigurableTolerance alpha feature

mikkeloscar:
Kubernetes v1.33 has alpha support for the feature gate `HPAConfigurableTolerance` which allows defining HPA scaling tolerance _per_ HPA. Something that was normally only possible at cluster level: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#tolerance

This PR adds a config-item to optionally enable the feature gate, disabled by default.


mikkeloscar:
:+1:

tcondeixa:
👍

PR #9992, opened 2025-10-07
Drop eks_fis_namespaces config-item

mikkeloscar:
Follow up to #9817

Drop the unused config-item `eks_fis_namespaces` which is no longer needed with the EKS FIS v2 implementation.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #10012, opened 2025-10-09
Update to Karpenter v1.8.0

mikkeloscar:
https://github.com/kubernetes-sigs/karpenter/releases/tag/v1.8.0

## TODO

* [x] Update karpenter image from test image.


myaser:
the replicas new feature could be useful for situations like this
where they needed always hot gpu instances to speed up bootstrapping. the instances were reserved so we were paying for it regardless of being running or not

the replicas feature would improve this implementation as it allows for node rotation to do operational tasks like updating AMI or so
https://github.com/zalando-build/zooport/issues/5059#issuecomment-2856566804

katyanna:
:+1:

mikkeloscar:
:+1:

PR #10019, opened 2025-10-10
feature: base eks migration stackset with routegroup

szuecs:
feature: base eks migration stackset with routegroup

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -122,6 +122,7 @@ skipper_ingress_binpack: "false"
 
 # skipper EKS migration support
 skipper_eks_migration_enabled: "false"
+skipper_eks_migration_svc_enabled: "false"
 worker_sg_legacy_cluster: ""
 
 # skipper node-pool
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -120,6 +120,9 @@ skipper_ingress_binpack: "true"
 skipper_ingress_binpack: "false"
 {{end}}
 
+# skipper EKS migration support
+skipper_eks_migration_enabled: "false"
+
 # skipper node-pool
 enable_dedicate_nodepool_skipper: "true"
 skipper_attach_only_to_skipper_node_pool: "true"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -122,6 +122,7 @@ skipper_ingress_binpack: "false"
 
 # skipper EKS migration support
 skipper_eks_migration_enabled: "false"
+worker_sg_legacy_cluster: ""
 
 # skipper node-pool
 enable_dedicate_nodepool_skipper: "true"
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -558,7 +558,7 @@ spec:
         args:
           - "routesrv"
 {{ if eq .Cluster.ConfigItems.skipper_eks_migration_enabled "true" }}
-          - "-forward-backend-url=http://skipper-ingress-eks.{{ .Cluster.Alias }}.zalan.do"
+          - "-forward-backend-url=http://skipper-ingress-eks.{{ .Values.hosted_zone }}"
 {{ end }}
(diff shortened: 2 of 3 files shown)


szuecs:
:+1:

cluster/manifests/skipper/deployment.yaml
```diff
@@ -557,6 +557,9 @@ spec:
           protocol: TCP
         args:
           - "routesrv"
+{{ if eq .Cluster.ConfigItems.skipper_eks_migration_enabled "true" }}
+          - "-forward-backend-url=http://skipper-ingress-eks.{{ .Cluster.Alias }}.zalan.do"
```
mikkeloscar:
```suggestion
          - "-forward-backend-url=http://skipper-ingress-eks.{{ .Values.hosted_zone }}"
```

Using the `hosted_zone` variable provides the right account level hosted zone. `Alias` would only work for the non-EKS cluster.

cluster/manifests/skipper/service-eks-internal.yaml
```diff
@@ -0,0 +1,35 @@
+# {{ if eq .Cluster.ConfigItems.skipper_eks_migration_enabled "true" }}
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    external-dns.alpha.kubernetes.io/hostname: skipper-ingress-eks.{{ .Cluster.Alias }}.zalan.do
```
mikkeloscar:
```suggestion
    external-dns.alpha.kubernetes.io/hostname: skipper-ingress-eks.{{ .Values.hosted_zone }}
```

cluster/manifests/skipper/service-eks-internal.yaml
```diff
@@ -0,0 +1,35 @@
+# {{ if eq .Cluster.ConfigItems.skipper_eks_migration_enabled "true" }}
```
mikkeloscar: Maybe this could just use `worker_sg_legacy_cluster` not empty as condition, so not two config-items are needed. Alternatively it needs both here to not fail if `worker_sg_legacy_cluster` is not set.

cluster/config-defaults.yaml
```diff
@@ -120,6 +120,9 @@ skipper_ingress_binpack: "true"
 skipper_ingress_binpack: "false"
 {{end}}
 
+# skipper EKS migration support
+skipper_eks_migration_enabled: "false"
+
```
mikkeloscar, line 127: Please define: `worker_sg_legacy_cluster: ""` so we have a default value defined/documented here.

cluster/manifests/skipper/service-eks-internal.yaml
szuecs: I put `worker_sg_legacy_cluster` to all old clusters first because I hope on your cluster creations you just copy the config items and override the ones that you have to.

cluster/config-defaults.yaml
szuecs: But all clusters should have this value set or you create a bad cluster!
Will fix it anyways

szuecs: done

cluster/manifests/skipper/service-eks-internal.yaml
mikkeloscar: But this is _only_ needed to be configured on the EKS clusters from my understanding, so for all non-EKS clusters we don't need it set and we don't deploy this service.

But we need to configure the forward rule on skipper-ingress in the non-EKS clusters.

szuecs: yes that's true, maybe we should add here another condition or a different condition for the 2 different cases.

cluster/manifests/skipper/service-eks-internal.yaml
```diff
@@ -0,0 +1,35 @@
+# {{ if eq .Cluster.ConfigItems.skipper_eks_migration_svc_enabled "true" }}
```
szuecs, line 1: @mikkeloscar what do you think like this?

mikkeloscar:
:+1:

szuecs:
:+1:

PR #10034, opened 2025-10-15
Skipper open policy agent data preprocessing optimization enabled by default

amak3:
Configuration of Skipper open policy agent data preprocessing optimization should be enabled by default for all the clusters that are having OPA enabled.  The decision was made after successful experimental phase when it was enabled for some clusters.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -394,7 +394,7 @@ skipper_ingress_routegroup_crd_require_hosts: "true"
 # Set defaults values that would enable Open Policy Agent in a skipper filter
 skipper_open_policy_agent_enabled: "false"
 skipper_open_policy_agent_styra_token: ""
-skipper_open_policy_agent_data_preprocessing_optimization_enabled: "false"
+skipper_open_policy_agent_data_preprocessing_optimization_enabled: "true"
 
 # Default timeout value in seconds for outgoing http calls from Open Policy Agent in a skipper filter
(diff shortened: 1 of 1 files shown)


demonCoder95:
I see `skipper_open_policy_agent` disabled by default. So, enabling a dependency by default will be incorrect. You also mention in the description that 

> should be enabled by default for all the clusters that are having OPA enabled

But I don't see a check in the configuration template for this, so please add that check, and make this conditional for clusters which have OPA enabled.

amak3:
> I see `skipper_open_policy_agent` disabled by default. So, enabling a dependency by default will be incorrect. You also mention in the description that
> 
> > should be enabled by default for all the clusters that are having OPA enabled
> 
> But I don't see a check in the configuration template for this, so please add that check, and make this conditional for clusters which have OPA enabled.

the check is being done here:
https://github.com/zalando-incubator/kubernetes-on-aws/blob/d93c551766e8f591c5c9ef88df02b2cabcbfcf4b/cluster/manifests/skipper/deployment.yaml#L320

the behaviour should be like this:

if skipper_open_policy_agent_enabled is true
then enable-open-policy-agent-data-preprocessing-optimization flag should be set to skipper_open_policy_agent_data_preprocessing_optimization_enabled

and now skipper_open_policy_agent_data_preprocessing_optimization_enabled should be by default set to true

szuecs:
:+1:

linki:
👍

PR #10045, opened 2025-10-17
Enable deletion protection on EKS cluster resource

linki:
New feature that allows to enable deletion protection on EKS cluster resources. Trying it out for now.

**Rational**: For most clusters, especially the important ones, there's rarely a reason to be able to delete it via an API call or a click of a button. If we ever need to really delete it, then it's OK to go through one extra step. Whether that has to be done manually in the UI or via setting an accompanying config item (similar to deletion protection for Postgres clusters) is to be decided.

Before merging there are a couple of things that need to be clarified:
1. how to keep deleting e2e clusters
2. how to work seamlessly with pet clusters
3. how to configure playground for its weekly decommissioning
4. how to handle intended cluster deletions on protected clusters

Proposals:
1. disable for e2e clusters
2. disable for pet clusters
3. disable for playground or let cluster-cleaner lift the protection (as a reference on how it can work)
4. change it manually in the UI or set a config item with the date for CLM to disable it (it's also possible to switch the config item introduced here, but the date approach used by postgres-operator is safer)

Once this is merged update documentation on how to delete clusters if there are any additional steps required.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -206,6 +206,7 @@ Resources:
 {{- else }}
           EnabledTypes: []
 {{- end }}
+      DeletionProtection: {{ .Cluster.ConfigItems.eks_deletion_protection_enabled }}
       # Tags:
       # - Key: "application"
       #   Value: "kubernetes"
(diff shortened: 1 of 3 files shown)


linki:
Putting this on hold, because:
* it blocks CF stack deletion and the AWS SDK v1 of CLM doesn't have the API call to disable it ([need to update to v2 first](https://github.com/zalando-incubator/cluster-lifecycle-manager/pull/808))
* if the CF stack _is_ terminated it doesn't stop deleting half the cluster (e.g. OIDC provider, node pools) and make it dysfunctional
* in the UI there's a confirmation dialog even without deletion protection (there isn't one on the CLI I guess)

PR #10061, opened 2025-10-21
Add EKSAccessEntrySecurityAuth resource to cluster.yaml

ni9hty:
- start implement native AWS EKS IAM integration as replacement for Okta IDP

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -297,7 +297,7 @@ Resources:
       PrincipalArn: !Sub "arn:aws:iam::${AWS::AccountId}:role/SecAudit
       Type: "STANDARD"
       KubernetesGroups:
-        - zalando:security
+        - ReadOnly
   EKSAddonPodIdentityAgent:
     Type: AWS::EKS::Addon
     Properties:
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -298,7 +298,7 @@ Resources:
     Type: "AWS::EKS::AccessEntry"
     Properties:
       ClusterName: !Ref EKSCluster
-      PrincipalArn: !Sub "arn:aws:iam::${AWS::AccountId}:role/SecAudit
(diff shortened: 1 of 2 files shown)


cluster/cluster.yaml
```diff
@@ -290,6 +290,18 @@ Resources:
...
+    Properties:
+      ClusterName: !Ref EKSCluster
+      PrincipalArn: !Sub "arn:aws:iam::${AWS::AccountId}:role/SecAudit
+      Type: "STANDARD"
+      AccessPolicies:
+        - AccessScope:
+            Type: "cluster"
+          PolicyArn: "arn:aws:eks::aws:cluster-access-policy/AmazonEKSViewPolicy"
+      KubernetesGroups:
+        - zalando:security
```
tcondeixa: I don't think we have any cluster role nor cluster role binding for security?
I don't think we need to have both `AccessPolicies` and `KubernetesGroups` if we have the binings, but I see the existing EKS Access Entry relying on EKS policy.

tcondeixa: you can add the `zalando:read-only` here https://github.com/zalando-incubator/kubernetes-on-aws/blob/c00db6d3fa25f83bc4daa47e1e33b9bc595d42dc/cluster/manifests/roles/readonly-binding.yaml#L1-L44
instead of adding the EKSViewPolicy

tcondeixa: as we discussed we don't need to create the `zalando:security` group, we can just use `zalando:read-only`

tcondeixa: you still need to add the `zalando:read-only` in the file above that I mentioned @ni9hty 
you can search for `okta:common/read-only` in this repo, and the `zalando:read-only` should be added in the same places, but I think the file I mentioned is the only place we need.

tcondeixa (approved):

tcondeixa:
👍

ni9hty:
:+1:

PR #10084, opened 2025-10-23
enable egress-controller on EKS when it's the only cluster in the account

linki:
Instead of always off, let's enable egress-controller when we're in a EKS-only account.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -490,7 +490,8 @@ aws_efa_device_plugin_cpu: "10m"
 aws_efa_device_plugin_memory: "20Mi"
 
 # static egress controller settings
-{{- if eq .Cluster.Provider "zalando-eks" }}
+# by default static egress controller is enabled on legacy clusters and disabled on zalando-eks clusters with multiple clusters
+{{- if and (eq .Cluster.Provider "zalando-eks") (gt (len .Cluster.AccountClusters) 1) }}
 static_egress_controller_enabled: "false"
 {{- else }}
 static_egress_controller_enabled: "true"


cluster/config-defaults.yaml
```diff
@@ -490,7 +490,8 @@ aws_efa_device_plugin_cpu: "10m"
 aws_efa_device_plugin_memory: "20Mi"
 
 # static egress controller settings
-{{- if eq .Cluster.Provider "zalando-eks" }}
+# by default static egress controller is enabled on legacy clusters and disabled on zalando-eks clusters with multiple clusters
+{{- if and (eq .Cluster.Provider "zalando-eks") (gt (len .Cluster.AccountClusters) 1) }}
```
tcondeixa, line 494: when does this len of Cluster.AccountClusters will be just 1? Do we need to decommission the cluster for it not be counted?
We might have a period when we don't have the static egress controller active in any cluster, but maybe it's fine.

szuecs:
Likely it's simpler to copy the configmaps to the eks cluster and migrate the controller to eks and stop it in the zalando-aws cluster.
Simple in the sense of safety.

cluster/config-defaults.yaml
tcondeixa: this is just to understand a bit more, but I think this is anyway better than what we have at the moment.

linki: It was just a quick workaround to not have to enable egress on stand-alone EKS cluster manually (it will be enabled by this).

`.AccountClusters` are the clusters in this account (I need to double check how the region plays into this).

The idea is that when there's just one cluster (legacy account or new EKS account) then the egress-controller can run. If there are two clusters, then it's likely a legacy+EKS combination in which case the egress-controller should only run on the legacy cluster (it ran there before so it just keeps running and no new one is running on EKS).

This is just for the default case which currently is `legacy: on, EKS: off`.

linki: Best would be if both egress-controllers can just run (like most other components do).

The problem is that the NAT gateways are shared since the nodes run in the same subnets. If we could just have two separate group of NAT gateways this would also be fine (apart from cost) but I don't know if it's possible and then how extensive the change in the controller would be (if needed).

tcondeixa:
👍

linki:
> Likely it's simpler to copy the configmaps to the eks cluster and migrate the controller to eks and stop it in the zalando-aws cluster. Simple in the sense of safety.

That's basically a migration. But it would also mean users in the legacy cluster cannot create new egress rules.

cluster/config-defaults.yaml
tcondeixa: as I said this is anyway better than what we have at the moment, so we should proceed IMO

tcondeixa: The main issue I see that is not directly related with this PR, is that users might delete the static egress from the legacy cluster after they migrate, so this will cause the routes to be removed and an incident.

linki: That's right 🤦  But this would also happen without this PR.

linki:
👍

PR #10093, opened 2025-10-24
use resolved image name for CoreDNS

mikkeloscar:
Always use the long image name for CoreDNS

When we introduced EKS we needed to use the resolved image name to let CoreDNS run _before_ admission-controller. We conditioned this to only EKS because on non-EKS admission-controller run before CoreDNS and thus can resolve the images during pod creation.

However, this setup means we always need multiple PRs to update an image and it adds no value to split it like this. So simply use the resolved name in all environments.


mikkeloscar:
:+1:

linki:
👍

PR #10097, opened 2025-10-24
Allow defining scheduled scaling for Prometheus

mikkeloscar:
The Scheduled Scaling VPA defines a Vertical Pod Autoscaler (zVPA) resource which is an enriched version of the [upstream VPA](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler#vertical-pod-autoscaler) that adds the ability to change scaling based on [Scaling Schedules](https://github.com/zalando-incubator/kube-metrics-adapter?tab=readme-ov-file#scalingschedule-collectors).

An example zVPA looks like this:

```yaml
apiVersion: zalando.org/v1
kind: VerticalPodAutoscaler
metadata:
  name: scheduled-scaling-vpa-test
spec:
  resourcePolicy:
    containerPolicies:
    - containerName: main
      minAllowed:
        cpu: 0
        memory: 500Mi
  targetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: scheduled-scaling-vpa-test
  updatePolicy:
    updateMode: Auto
  # above this line looks like a normal VPA
  # below this line are references to scaling schedules and the resourcePolicy
  # overwrite.
  schedules:
  - name: "scheduling-event"
    preStartMinutes: 10
    resourcePolicy:
      containerPolicies:
      - containerName: main
        minAllowed:
          cpu: 0
          memory: 1Gi
```

The resource is combined with a simple controller that can read zVPA resources and generate/update VPA resources. If a referenced schedule is active it will use the resourcePolicy of the respective schedule, otherwise it will use the default resourcePolicy.

---

This PR enables also enables Prometheus to use this feature for scaling based on central events.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -553,6 +553,13 @@ prometheus_remote_max_backoff: "10s"
 # Comma-separated list of user ids allowed to access Prometheus UI
 prometheus_ui_users: ""
 
+# Scheduled scaling events for Prometheus VPA (in the format
+# <name>:<pre-start-min>:<cpu>:<memory>, multiple events separated by commas)
+prometheus_scheduled_scaling_events: ""
+
+# Enable scheduled scaling for VPA
+scheduled_scaling_vpa_enabled: "false"
+
 metrics_service_cpu: "100m"
 metrics_service_mem_max: "4Gi"
 metrics_server_metric_resolution: "15s"
--- a/cluster/manifests/02-scheduled-scaling-vpa/rbac.yaml
+++ b/cluster/manifests/02-scheduled-scaling-vpa/rbac.yaml
@@ -1,0 +1,44 @@
+{{ if eq .Cluster.ConfigItems.scheduled_scaling_vpa_enabled "true" }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: scheduled-scaling-vpa
+  namespace: kube-system
+  labels:
+    application: kubernetes
+    component: scheduled-scaling-vpa
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: scheduled-scaling-vpa
+  labels:
+    application: kubernetes
+    component: scheduled-scaling-vpa
+rules:
+- apiGroups: ["autoscaling.k8s.io"]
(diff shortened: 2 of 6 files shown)


cluster/config-defaults.yaml
```diff
@@ -553,6 +553,13 @@ prometheus_remote_max_backoff: "10s"
 # Comma-separated list of user ids allowed to access Prometheus UI
 prometheus_ui_users: ""
 
+# Scheduled scaling events for Prometheus VPA (in the format
+# <name>:<pre-start-min>:<cpu>:<memory>, multiple events separated by commas)
+prometheus_scheduled_scaling_events: ""
```
tcondeixa, line 558: do we need this? I don't see it used to define the cpu nor memory

mikkeloscar: This would be defined per cluster in cluster-config-manager. There is an open PR internally showing how it would be used.

tcondeixa: ok, we want to define it here just to be on the safe side with a default in case it's enable but without defining the events

mikkeloscar: Yeah, it's mostly to document what this config item is since it's referenced in a manifest here.

In the manifest we check for not empty string, we could ofc. also check if the config-item is defined or not, but I think this pattern is more clear with some description.

cluster/manifests/02-scheduled-scaling-vpa/rbac.yaml
```diff
@@ -0,0 +1,44 @@
...
+kind: ClusterRole
+metadata:
+  name: scheduled-scaling-vpa
+  labels:
+    application: kubernetes
+    component: scheduled-scaling-vpa
+rules:
+- apiGroups: ["autoscaling.k8s.io"]
+  resources: ["verticalpodautoscalers"]
+  verbs: ["get", "list", "watch", "create", "update", "patch"]
```
demonCoder95: Do we need the `watch` permission here? I don't recall the controller needing that 🤔

mikkeloscar: True, this was just general "read permissions". But I can drop `watch`.

demonCoder95:
👍

tcondeixa:
👍

PR #10104, opened 2025-10-27
Enable zone-based topology-spread for deployment-service's status component

linki:
Let's make sure the three pods for deployment-service's status component are distributed across zones (and therefore nodes).

### DIFF
--- a/cluster/manifests/deployment-service/status-service-deployment.yaml
+++ b/cluster/manifests/deployment-service/status-service-deployment.yaml
@@ -26,6 +26,7 @@ spec:
         prometheus.io/port: "9090"
         prometheus.io/scrape: "true"
         config/hash: '{{"01-config.yaml" | manifestHash}}'
+        zalando.org/topology-spread: 'true'
     spec:
       serviceAccountName: "deployment-service-status-service"
       containers:


linki:
👍

cluster/manifests/deployment-service/status-service-deployment.yaml
```diff
@@ -26,6 +26,7 @@ spec:
         prometheus.io/port: "9090"
         prometheus.io/scrape: "true"
         config/hash: '{{"01-config.yaml" | manifestHash}}'
+        zalando.org/topology-spread: 'true'
```
szuecs, line 29: don't you want to add also PreferClose to the svc?

szuecs:
:+1:

cluster/manifests/deployment-service/status-service-deployment.yaml
linki: Haven't used it yet. I'll look into it via a separate PR. Is it something we might want to set for every service?

mikkeloscar: Setting the annotation doesn't work in `kube-system` because it's excluded from admission-controller actions.

We either need to set it manually like we do for skipper-ingress or we need to allow _some_ mutations in admission-controller in `kube-system`.

PR #10113, opened 2025-10-28
Update aws-node-decommissioner pipeline

mikkeloscar:
Updates the aws-node-decommissioner image to the new internal pipeline. This change includes

* Update the build and use static base image to reduce possible CVEs
* Update to aws-sdk-go v2
* Pass AWS_REGION as needed by aws-sdk-go v2 and make it work in regions other than `eu-central-1`.

### DIFF
--- a/cluster/manifests/aws-node-decommissioner/cronjob.yaml
+++ b/cluster/manifests/aws-node-decommissioner/cronjob.yaml
@@ -28,7 +28,10 @@ spec:
           restartPolicy: Never
           containers:
           - name: aws-node-decommissioner
-            image: container-registry.zalando.net/teapot/aws-node-decommissioner:master-690846a9-master-19
+            image: container-registry.zalando.net/cloud-platform/aws-node-decommissioner:main-4
+            env:
+            - name: AWS_REGION
+              value: "{{.Cluster.Region}}"
             args:
               - --debug
             resources:


cluster/manifests/aws-node-decommissioner/cronjob.yaml
```diff
@@ -28,7 +28,10 @@ spec:
           restartPolicy: Never
           containers:
           - name: aws-node-decommissioner
-            image: container-registry.zalando.net/teapot/aws-node-decommissioner:master-690846a9-master-19
+            image: container-registry.zalando.net/cloud-platform/aws-node-decommissioner:main-4
+            env:
+            - name: AWS_REGION
```
tcondeixa, line 33: I think for go aws sdk it needs to be `AWS_DEFAULT_REGION`, at least this is what we use in other projects for aws sdk v2

tcondeixa: From the docs it seems fine to use `AWS_REGION`
https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#specifying-the-region

tcondeixa: we probably should change that env in other projects

mikkeloscar:
:+1:

tcondeixa:
👍

PR #10120, opened 2025-10-28
Remove spot-node-rescheduler

mikkeloscar:
The `spot-node-rescheduler` was a simple cronjob aiming to continously introduce spot instances for workloads that allowed running on spot and managed via Cluster Autoscaler.

Now we run Karpenter for all workloads (except skipper-ingress) and Karpenter does a much better job on managing spot.

Therefore remove the `spot-node-rescheduler` completely. It was only enabled in 3 clusters and it has been disabled in all already.


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #10122, opened 2025-10-30
Revert "enable egress-controller on EKS when it's the only cluster in the account"

linki:
Reverts zalando-incubator/kubernetes-on-aws#10084

This logic doesn't work because `.AccountClusters` _does_ include decommissioned clusters (I thought it didn't - for trust relationship these are filtered out in CLM).

This means that "single" clusters are not detected as such when there was an old cluster that is now decommissioned. This leads to egress-controller not running when it should.

Let's go back to the "simple" manual approach and find a better way based on that.


linki:
👍

mikkeloscar:
:+1:

PR #10129, opened 2025-10-31
coredns: Fix overlapping port names

mikkeloscar:
The port name `dns-tcp` was defined both for the `unbound` container and for `coredns` container.

This produces a warning.

Change it for `unbound` as the service already selects on the name for coredns


linki:
👍

mikkeloscar:
:+1:

PR #10135, opened 2025-10-31
Use small seed node

mikkeloscar:
Follow up to #10117 

Use a small seed node for e2e EKS clusters to validate that this works continously.

This depends on #10117 being fully rolled out first

* [x] #10117
* [x] alpha
* [x] beta
* [x] stable


mikkeloscar:
:+1:

tcondeixa (approved):

PR #10144, opened 2025-11-03
Put skipper-validation webhook back onto the seed pool

linki:
skipper-validation-webhook's route-group admitter needs karpenter and node pools that are defined later, leading to a chicken egg problem. Let's keep running it on the seed pool which doesn't require karpenter.


linki:
👍

linki:
Closed in favour of https://github.com/zalando-incubator/kubernetes-on-aws/pull/10145

PR #10161, opened 2025-11-06
feat: upgrade wiz images

hagenmathis:
- drop unnecessary labels
- upgrade the images
- update connector job command accordingly
- drop capabilities for connector job


hagenmathis:
will be reopened after cyberweek

PR #10163, opened 2025-11-06
Add initial configuration for EKS Control Plane logging SIEM integration

zaklawrencea:
This PR adds the configuration for the SIEM logging integration of EKS control plane logs.

With EKS Control Plane logging enabled, the logs are being routed to a `Log Group` (per account). These changes introduce CloudFormation configuration to additionally enable sending these logs to our SIEM solution. This is done by attaching a `Subscription Filter` to the existing EKS Control Plane `Log Groups` which utilise `Firehose` to send the logs to a centralised S3 Bucket.

The configuration is behind an additional feature flag: `eks_siem_logging`.


test/e2e/cluster_config.sh
```diff
@@ -48,6 +48,10 @@ clusters:
     skipper_open_policy_agent_observability_url: "${SKIPPER_OPA_OBSERVABILITY_URL}"
     skipper_open_policy_agent_bundles_url: "${SKIPPER_OPA_BUNDLES_URL}"
     eks_ip_family: "ipv6"
+    eks_siem_logging: "true"
```
mikkeloscar, line 51: Is this writing to some test endpoint of SIEM or the same as production? Maybe we don't want to actually write to prod from every e2e test.

zaklawrencea: This is using the same endpoint as production. The current recommendation and preference from Security is to use this endpoint for all cluster environments. The alternative is to use a dev SIEM instance, however they are heavily restricted on licenses there.

mikkeloscar: I guess this is ok for test clusters in general. But e2e is special because they are short lived and we create _a lot_ of those over time.

I worry that we ship a lot of useless data into SIEM from e2e clusters and I think the value of having this in e2e is minimal as we would enable it in dev to validate that it works in general.

zaklawrencea: I see your point. Having the e2e configuration has been useful for verifying the config during development, but once we are happy to move forward we can also disable this and rely on the dev clusters.

cluster/cluster.yaml
```diff
@@ -166,6 +166,88 @@ Resources:
...
+      DestinationArn: !GetAtt EKSControlPlaneLogsDataFirehose.Arn
+  EKSControlPlaneLogsDataFirehose:
+    Type: AWS::KinesisFirehose::DeliveryStream
+    Properties:
+      DeliveryStreamName: "EKSCtrlPlaneDeliveryStream-{{.Cluster.Name}}"
+      DeliveryStreamType: DirectPut
+      ExtendedS3DestinationConfiguration:
+        BucketARN: "{{.Cluster.ConfigItems.eks_siem_bucket}}"
+        RoleARN: !GetAtt EKSControlPlaneFirehoseS3Role.Arn
+        FileExtension: ".json.gz"
```
demonCoder95, line 186: I'm curious how we determined that the log files will be gzip compressed? 🤔

tcondeixa:
👍

cluster/cluster.yaml
zaklawrencea: base64 encoding and gzip compression is the default when using CloudWatch subscription filters to send to other services: https://docs.aws.amazon.com/eks/latest/userguide/control-plane-logs.html. In the case of CW -> Firehose -> S3, the base64 encoding and decoding is seemingly handled transparently by AWS and we are left with just the gzip compressed data. We also inspected sample events to confirm the format. The file extension is added just to make it clear what the format is for humans (and other software).

zaklawrencea:
:+1:

cluster/cluster.yaml
demonCoder95: Ah, right. That's exactly what I was looking for. Thanks! 🙂

PR #10170, opened 2025-11-07
add the option to set tags on karpenter EC2 instances per node pool

tcondeixa:
add the option to set tags on karpenter EC2 instances per node pool


tcondeixa:
Can you confirm this will not force the node rotation and only new nodes will have the new tags?

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -71,6 +71,12 @@ spec:
     node.kubernetes.io/node-pool: "{{ .NodePool.Name }}"
     zalando.org/pod-max-pids: "{{ .NodePool.ConfigItems.pod_max_pids }}"
     'zalando.de/cluster-local-id/{{ .Cluster.LocalID }}': owned
+#{{ if index .NodePool.ConfigItems "tags" }}
+#  {{ range split .NodePool.ConfigItems.tags "," }}
+#    {{ $tag := split . "=" }}
+    '{{index $tag 0}}': "{{index $tag 1}}"
+#  {{ end }}
+#{{ end }}
```
linki, line 79: Here, "tags" is a field under `ConfigItems`.  This doesn't match with your proposed definition in the other repo (it's one level up).

mikkeloscar:
> Can you confirm this will not force the node rotation and only new nodes will have the new tags?

99% sure that it does not trigger a rotation because we don't have drift enabled in Karpenter. We do some check in CLM to determine if rotation is needed but I don't _think_ it checks for tags.

(I would confirm via a pet cluster :))

tcondeixa:
> > Can you confirm this will not force the node rotation and only new nodes will have the new tags?
> 
> 99% sure that it does not trigger a rotation because we don't have drift enabled in Karpenter. We do some check in CLM to determine if rotation is needed but I don't _think_ it checks for tags.
> 
> (I would confirm via a pet cluster :))

 I can also add one tag in one of the dev clusters after merge to see if the nodes are rotated. Or do you recommend to do it in pet cluster?

cluster/node-pools/worker-karpenter/provisioners.yaml
tcondeixa: I will adjust the proposed definition in the other repo

tcondeixa: done

tcondeixa:
I tested this solution and the instances were tagged with the new tag without forcing the rotation of the instances by Karpenter.
This is ready to be merged after CW.

tcondeixa:
👍

mikkeloscar:
:+1:

PR #10165, opened 2025-11-06
skipper-validation-webhook: enabled missing configuration for custom filters

vlktna:
- OAuth2 Tokeninfo Filters
- OAuth2 Grant Flow
- Open Policy Agent
- Rate Limiter
- Compression

PR #10181, opened 2025-11-13
cluster/manifests: disable loading routing table

vlktna:
Set the `source-poll-timeout` flag to the max int64 value to disable loading the routing table for the data client. The routing table loads by default whenever a data-client config is present. In this case, the config is only needed, so the filters get added to the registry for validation


MustafaSaber:
It's not needed here anyway since we don't use k8s dataclient, please add it to https://github.com/zalando-incubator/kubernetes-on-aws/pull/10165

vlktna:
moved to https://github.com/zalando-incubator/kubernetes-on-aws/pull/10165

PR #10182, opened 2025-11-14
Firedrills

demonCoder95:
Let's break the control plane to learn how to fix it!

Part of firedrills session. This will be used as an override on a test cluster. This is based on a custom branch to not disrupt the override already there on the cluster.


demonCoder95:
Exercise completed, so will close this now.

PR #10189, opened 2025-12-01
Enable OAuth2 UI login in e2e environment

linki:
The ~~production~~playground configuration suffers from a chicken-and-egg problem wrt to the encryption key. Let's see if it makes sense / is possible to reproduce it in e2e tests.

[EDIT] Actually, it's fine. It found it eventually. But let's see what e2e says about `skipper_oauth2_ui_login: true` in e2e tests anyways.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -372,12 +372,7 @@ skipper_tokeninfo_cache_ttl: "30s"
 # skipper-ingress ingress OAuth2 grant flow
 #
 # oauth2 UI login - grant flow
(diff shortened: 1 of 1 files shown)


szuecs:
I think it won't work because we would need a valid application and a cronjob has to create the PCS based on this.

linki:
> I think it won't work because we would need a valid application and a cronjob has to create the PCS based on this.

Ok, thanks. Can we somehow "fake" some values so that it at least runs the components? Would allow us to catch some config errors, such as the chicken-and-egg problem.

PR #10195, opened 2025-12-02
update images to be detected as compliant again

szuecs:
update images to be detected as compliant again


szuecs:
problem is a test registry

PR #10197, opened 2025-12-04
Configure egress-controller to use platform credentials

linki:
Roll out v0.3.6 of egress-controller: https://github.com/szuecs/kube-static-egress-controller/tree/v0.3.6 that includes https://github.com/szuecs/kube-static-egress-controller/pull/60. Configure it to use platform credentials instead of service account tokens.

### DIFF
--- a/cluster/manifests/kube-static-egress-controller/01-rbac.yaml
+++ b/cluster/manifests/kube-static-egress-controller/01-rbac.yaml
@@ -24,6 +24,18 @@ roleRef:
   kind: ClusterRole
   name: kube-static-egress-controller
 subjects:
-- kind: ServiceAccount
-  name: kube-static-egress-controller
+- apiGroup: rbac.authorization.k8s.io
+  kind: User
+  name: zalando-iam:zalando:service:stups_kubernetes
+---
+apiVersion: zalando.org/v1
+kind: PlatformCredentialsSet
+metadata:
+  name: kube-static-egress-controller-credentials
   namespace: kube-system
+spec:
+  application: kubernetes
+  token_version: v2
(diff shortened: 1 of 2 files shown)


linki:
👍

cluster/manifests/kube-static-egress-controller/01-rbac.yaml
```diff
@@ -24,6 +24,18 @@ roleRef:
   kind: ClusterRole
   name: kube-static-egress-controller
 subjects:
-- kind: ServiceAccount
-  name: kube-static-egress-controller
+- apiGroup: rbac.authorization.k8s.io
+  kind: User
+  name: zalando-iam:zalando:service:stups_kubernetes
```
demonCoder95, line 29: Why do we need this subject in the binding?

demonCoder95:
Nice branch name 😅 Seeing as we've needed to revert it a few times in the past, should we keep this credentials mechanism change behind a feature flag to be able to revert easily? 🤔

linki:
No, it was reverted because it was merged closely before BF and not needed.

cluster/manifests/kube-static-egress-controller/01-rbac.yaml
linki: That's basically the username you get when using the PCS. So it gives the container that mounts the PCS the specific RBAC permissions by this binding.

demonCoder95: Ah, okay. Makes sense 🙂

demonCoder95:
👍

PR #10198, opened 2025-12-04
fix: reduce redis connections to bare minimum 

szuecs:
fix: reduce redis connections to bare minimum for the webhook that is not using redis


vlktna:
👍

szuecs:
:+1:

PR #10203, opened 2025-12-04
fix: skipper-validation-webhook increase resources to not fail

szuecs:
fix: skipper-validation-webhook increase resources to not fail


tcondeixa (approved):

szuecs:
:+1:

PR #10215, opened 2025-12-08
sandbox crd: testContext rename and URL validation

ponimas:
Rename testContext -> testProject to match the TDD.
Also add URL validation for the target field to ensure it uses a valid URL with http scheme only allowed for cluster.local domains and https for all others.

see:
- https://github.com/zalando-build/sandbox-controller/pull/50
- https://github.com/zalando-build/sandbox-controller/pull/45


ponimas:
👍

ayruslore (approved):

PR #10228, opened 2025-12-09
Configure egress-controller to use external API server endpoint

linki:
In order to actually use PCS against the API server consistently in both environments (`zalando-aws`, `zalando-eks`) let's switch to using the external API server URL.


linki:
👍

mikkeloscar:
:+1:

PR #10234, opened 2025-12-10
Configure egress-controller to manage all clusters in the account

linki:
Configure egress-controller to manage all clusters in the account

### DIFF
--- a/cluster/manifests/kube-static-egress-controller/deployment.yaml
+++ b/cluster/manifests/kube-static-egress-controller/deployment.yaml
@@ -32,10 +32,8 @@ spec:
       - name: controller
         image: container-registry.zalando.net/teapot/kube-static-egress-controller:v0.4.0-master-56
         args:
-        # {{ with $data := . }}
-        # {{ range $cluster := $data.Cluster.AccountClusters }}
-        - "--master=https://{{ $cluster.LocalID }}.{{ $data.Values.hosted_zone }}"
-        # {{ end }}
+        # {{ range $cluster := .Cluster.AccountClusters }}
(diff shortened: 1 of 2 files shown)


linki:
👍

cluster/manifests/kube-static-egress-controller/deployment.yaml
```diff
@@ -32,7 +32,11 @@ spec:
       - name: controller
         image: container-registry.zalando.net/teapot/kube-static-egress-controller:v0.4.0-master-56
         args:
-        - "--master=https://{{ .Cluster.LocalID }}.{{ .Values.hosted_zone }}"
+        # {{ with $data := . }}
+        # {{ range $cluster := $data.Cluster.AccountClusters }}
+        - "--master=https://{{ $cluster.LocalID }}.{{ $data.Values.hosted_zone }}"
```
tcondeixa: just to learn here, why don't you just do `$.Values.hosted_zone`?

linki: I didn't try this format. 

`.Values.hosted_zone` doesn't work so I took this version from [here](https://github.com/zalando-incubator/kubernetes-on-aws/blob/639045cf8f1b50a6b883cad372ae1c4591018121/cluster/manifests/01-aws-node/pod_subnets.yaml#L4).

tcondeixa: it needs to be with `$` because you are in a loop/range scope, it should be `$.Values.hosted_zone`

tcondeixa:
👍

linki:
👍

tcondeixa:
👍

PR #10238, opened 2025-12-10
Use external cluster endpoint that understands platformcredentialssets

linki:
Although these values should be the same (according to the API), the `.APIServerURL` field is actually rewritten on-the-fly when rendering the manifests. So let's construct the value ourselves.


linki:
👍

tcondeixa (approved):

PR #10241, opened 2025-12-10
validation-webhook: move hostname-credentials-controller to validation webhook

vlktna:
Move hostname-credentials-controller to validation webhook since can't be optional because config is required on initialization and filter creation in the registry

Note: the same applied to the open-policy-agent-config


szuecs:
:+1:

szuecs:
We need this as pre-condition to run the webhook and skipper would be applied much later, so we move all dependencies that are used by skipper and the webhook to the webhook folder.

vlktna:
👍

PR #10242, opened 2025-12-11
sandbox-controller: add status field to Sandbox CRD

ponimas:
Introduces a status subresource with a problems array to track the current state of Sandbox resources according to Kubernetes API conventions.


szuecs:
:+1:

ponimas:
👍

PR #10248, opened 2025-12-11
Filter watched clusters for egress-configs to ready clusters

linki:
We need to manually filter the non-ready clusters because the provided list can contain decommissioned or to-be-provisioned clusters.


tcondeixa (approved):

linki:
👍

PR #10261, opened 2025-12-15
Exclude slow c6id.large instance types

mikkeloscar:
We have had at least 3 cases where scheduling on `c6id.large` takes a long time leading to "pods not scheduled in time alerts". It's always the same pattern of 2-3 `c6id.large` being launched and only the last one getting ready in time.

For this reason, exclude `c6id.large` types by default similar to how we exclude `c5d.large` because of slow SSD.


linki (approved):

mikkeloscar:
:+1:

PR #10276, opened 2025-12-16
rename function that detects the main cluster for egress-controller

linki:
Use the more correct function name so we can phase out the other one.


linki:
👍

tcondeixa:
👍

PR #10278, opened 2025-12-16
skipper-validation: fix cluster bootstrapping

MustafaSaber:
We don't need the full setup for oauthGrant to work so I move hostname-credentials-provider back to skipper and we can use some dummy values as a place holder for filter to work

### DIFF
--- a/cluster/manifests/01-coredns-local/daemonset-coredns.yaml
+++ b/cluster/manifests/01-coredns-local/daemonset-coredns.yaml
@@ -111,7 +111,7 @@ spec:
           name: unbound-socket
           readOnly: false
       - name: coredns
-        image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/coredns:1.12.4-master-29
+        image: 926694233939.dkr.ecr.eu-central-1.amazonaws.com/production_namespace/teapot/coredns:1.13.2-master-30
         args: [ "-conf", "/etc/coredns/Corefile" ]
         env:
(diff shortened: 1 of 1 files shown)


cluster/manifests/skipper/hostname-credentials-controller.yaml
```diff
@@ -1,4 +1,4 @@
-# {{ if eq .Cluster.ConfigItems.skipper_oauth2_ui_login "true" }}
+{{ if eq .Cluster.ConfigItems.skipper_oauth2_ui_login "true" }}
```
vlktna: need to uncomment `# {{ end }}`

MustafaSaber: This was by mistake will cleanup later or you can take over

vlktna:
👍

vlktna:
👍

MustafaSaber:
Testing this by creating a pet cluster from this branch.

MustafaSaber:
This will fail

szuecs:
Maybe you can tell why it failed and how?
Error message or logs or something else?

MustafaSaber:
> Maybe you can tell why it failed and how? Error message or logs or something else?

We merged the other PR see https://github.com/zalando-incubator/kubernetes-on-aws/pull/10280, this one I closed for other reasons, it was failing because I was using 

```
"{{ val | base64 }}"
```
in the secret which CLM considered `val` an undefined function

PR #10280, opened 2025-12-16
skipper-validation: fix cluster bootstrapping

MustafaSaber:
We don't need the full setup for oauthGrant to work so I move hostname-credentials-provider back to skipper and we can use some dummy values as a place holder for filter to work


szuecs:
:+1:

MustafaSaber:
👍

PR #10292, opened 2025-12-17
Skipper ingress Karpenter scheduling annotations

ayruslore:
This implements scheduling annotations for skipper-ingress such that a dedicated node pool is not needed, but it will land on dedicated nodes via dedicated scheduling annotation feature.

(Since kube-system currently is not covered by scheduling annotations, the annotation is not defined, but the nodeSelector/toleration which is normally injected by admission-controller, is defined manually)

The changes includes a few more optimizations:

- Include the instance type priority to prioritize 7th gen instances instead of 6th gen

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -85,9 +85,6 @@ spec:
 {{- if eq .Cluster.ConfigItems.skipper_topology_spread_enabled "true" }}
       topologySpreadConstraints:
         - maxSkew: 1
-{{- if eq .Cluster.ConfigItems.skipper_karpenter_node_pools_enabled "true" }}
-          minDomains: 3
-{{- end }}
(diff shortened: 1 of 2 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -85,12 +85,57 @@ spec:
 {{- if eq .Cluster.ConfigItems.skipper_topology_spread_enabled "true" }}
       topologySpreadConstraints:
         - maxSkew: 1
+{{- if eq .Cluster.ConfigItems.skipper_karpenter_node_pools_enabled "true" }}
+          minDomains: 3
```
szuecs: this we can not set like this. For example teapot had to drop one AZ in logistics cluster for cw.

szuecs:
:+1:

mikkeloscar:
:+1:

ayruslore:
👍

PR #10299, opened 2025-12-19
Update stackset-controller to support forward

mikkeloscar:
This updates stackset-controller to the latest version adding optional support for forwarding traffic to another backend instead of the stack as implemented in: https://github.com/zalando-incubator/stackset-controller/pull/716 and https://github.com/zalando-incubator/stackset-controller/pull/721

The new feature is only enabled if: `skipper_eks_migration_enabled` is set to true which also enables the relevant feature in skipper ingress. The idea is we would only enable it in source clusters for the migration. In target clusters it would not be enabled.


mikkeloscar:
:+1:

katyanna:
👍

PR #10269, opened 2025-12-15
Update to Vertical Pod Autoscaler v1.5.1

mikkeloscar:
This updates the VPA components from v1.3.1 to the latest v1.5.1: https://github.com/kubernetes/autoscaler/releases/tag/vertical-pod-autoscaler-1.5.0

It's possible to switch back to the previous version by setting: `vertical_pod_autoscaler_version: legacy`

### TODO

* [x] Update to production images

PR #10301, opened 2025-12-19
VPA Auto mode is deprecated, use Recreate instead

mikkeloscar:
`updateMode: Auto` is deprecated in the latest VPA version (#10269, see: https://github.com/kubernetes/autoscaler/issues/8424)

Therefore change to `Recreate` which is the same behavior.

The alternative is: `InPlaceOrRecreate` which is available with the InPlace pod resource changes feature. Since we haven't validate that our setup works with this yet, let's use `Recreate` for now.


mikkeloscar:
:+1:

tcondeixa:
👍

PR #10302, opened 2025-12-19
Prometheus scheduled VPA: Fix config-item name

mikkeloscar:
Follow up to #10097 fixing the config-item name used such that the schedules get defined.


mikkeloscar:
:+1:

tcondeixa (approved):

PR #10306, opened 2025-12-19
Update Scheduled Scaling VPA CRD

mikkeloscar:
Follow up to #10097 updating the VPA CRD to the latest version which includes the unitResource field used by the prometheus vpa when schedules are enabled.


mikkeloscar:
:+1:

tcondeixa (approved):

PR #10317, opened 2025-12-22
Update ingress controller filters to catch both asg and karpenter nodes

ayruslore:
see subject


ayruslore:
For this we need dynamic tagging. The current labels propagation only works for dedicated node pools.

PR #10330, opened 2026-01-05
Setup EFS CSI driver Add-on

katyanna:
Add the AWS EFS CSI Driver as a managed addon behind the `efs_csi_driver` flag, which is disabled by default.

Ref: https://github.com/zalando-build/cloud-platform/issues/7708


cluster/manifests/04-efs-csi/serviceaccount.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if eq .Cluster.ConfigItems.eks_efs_csi_support_enabled "true" }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: efs-csi-controller-sa
+  namespace: kube-system
+  labels:
+    app.kubernetes.io/name: aws-efs-csi-driver
```
linki: nit: This should use our `application`, `component` format, e.g. like here: https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/efs-provisioner/depl-efs-provisioner.yaml#L8-L9

linki: We can do it separately (or at least no need to wait for e2e again).

linki:
👍

cluster/manifests/04-efs-csi/serviceaccount.yaml
linki: Or, let's leave it like this because the components managed by AWS also use it.

katyanna:
:+1:

linki:
👍

cluster/cluster.yaml
```diff
@@ -568,6 +576,37 @@ Resources:
...
+                    "sts:AssumeRole",
+                    "sts:TagSession"
+                ]
+            }
+        ]
+      }
+      Path: /
+      ManagedPolicyArns:
+      - arn:aws:iam::aws:policy/service-role/AmazonEFSCSIDriverPolicy
+  EKSEFSCSIDriverPodIdentity:
```
mikkeloscar: This can be defined at addon level via this sort of definition: https://github.com/zalando-incubator/kubernetes-on-aws/pull/10368/files#diff-c7cc960a01f7efb45aaa0ca7dee2d7bc3890f542a23693509a32065ec0e91b55R354-R356

cluster/manifests/04-efs-csi/serviceaccount.yaml
mikkeloscar: Isn't this service account created/managed by the addon? Or is there a reason we need to manage it like this?

cluster/cluster.yaml
linki: image

This part?

cluster/manifests/04-efs-csi/serviceaccount.yaml
katyanna: It's a prerequisite for the addon https://docs.aws.amazon.com/eks/latest/userguide/workloads-add-ons-available-eks.html#add-ons-aws-efs-csi-driver

cluster/cluster.yaml
linki: Ok, this part: https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-eks-addon.html#cfn-eks-addon-podidentityassociations

Makes sense to use it, as the definition is probably shorter.

cluster/manifests/04-efs-csi/serviceaccount.yaml
tcondeixa: I think we only need to declare the IAM role and the Addon, the rest is created by EKS. it should create a service account named `efs-csi-controller-sa` and automatically annotate it with the role.

cluster/cluster.yaml
```diff
@@ -319,6 +319,18 @@ Resources:
...
       KubernetesGroups:
         - zalando:readonly
+{{- if eq .Cluster.ConfigItems.eks_efs_csi_support_enabled "true" }}
+  EKSAddonEFSCSIDriver:
+    Type: AWS::EKS::Addon
+    Properties:
+      AddonName: aws-efs-csi-driver
+      AddonVersion: "v2.1.15-eksbuild.1"
+      ClusterName: !Ref EKSCluster
+      PodIdentityAssociations:
```
tcondeixa, line 329: is PodIdentity supported for EFS Addon? and do we need it for this use-case or do we use `ServiceAccountRoleArn`?

cluster/manifests/04-efs-csi/serviceaccount.yaml
linki: Right, I didn't think about that.

@katyanna We should only specify the minimum that we need, starting with the `AWS::EKS::Addon` definition and then only add what AWS doesn't create (like the EFS filesystem itself for example).

You can test this best with an existing pet cluster and just keep iterating on the configuration.

cluster/cluster.yaml
linki: The definition has both fields: https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-eks-addon.html#cfn-eks-addon-podidentityassociations Why do you think it wouldn't work. Have you seen it before?

I assume both are supported (the docs say you can deploy the addon using IAM role _or_ pod association). I would try to use pod associations as it seems to be the EKS way going forward.

@katyanna Again, we can just quickly check that dynamically in a pet cluster.

tcondeixa: I asked because I remember that no all Addons supported Pod Identity, but it might have changed or it might never have been an issue for EFS Addon.

linki: Thanks, that's a good point. Let's keep it in mind in case something doesn't work.

The [manual setup docs](https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html#efs-install-driver) make use of Pod association so it's likely supported 🤞

cluster/cluster.yaml
```diff
@@ -319,6 +319,18 @@ Resources:
...
+        - RoleArn: !GetAtt EKSEFSCSIDriverIAMRole.Arn
+          ServiceAccount: efs-csi-controller-sa
+          Namespace: kube-system
```
linki: @katyanna [According to the spec](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-eks-addon-podidentityassociation.html) there is no `Namespace` field, so this will fail gain.

Likely also means that the ServiceAccount is created automatically.

cluster/cluster.yaml
katyanna: The EFS add-on does support PodIdentity, it was validated on petcluster tests.

cluster/cluster.yaml
linki: It won't fail here because it's behind a disabled feature flag.

cluster/manifests/04-efs-csi/serviceaccount.yaml
katyanna: @tcondeixa thanks for the insight, the docs put the SA as a prerequisite but automating on CF it really isn't needed.

katyanna:
👍

linki:
👍

tcondeixa:
just a general comment that we don't need to change, but it might be useful for other cases. 
I see most IAM roles `AssumeRolePolicyDocument` and other objects are defined as json sometimes, but usually everything (there might exists an exception) in the resources can be defined as yaml for CloudFormation.

linki:
> just a general comment that we don't need to change, but it might be useful for other cases. I see most IAM roles `AssumeRolePolicyDocument` and other objects are defined as json sometimes, but usually everything (there might exists an exception) in the resources can be defined as yaml for CloudFormation.

Good point. Let's change it as a whole in a separate PR.

I would also be interested to separate the huge CloudFormation file into multiple files and then group related things together (you could have a separate file for the stuff related to a specific addon etc.). But since there's lots of references between the resources it only makes sense if that part doesn't get worse.

PR #10333, opened 2026-01-05
Simplify enabling Prometheus scheduled scaling vpa

mikkeloscar:
Make Prometheus use the Scheduled Scaling VPA resource type if the scheduled-scaling-vpa controller is enabled, even if no schedules are defined for Prometheus for the cluster.

The motivation for this is:

1. When switching to the Scheduled Scaling VPA type we need to delete/change ownership of the existing VPA resource such that it can be owned by the Scheduled Scaling VPA resource, this is a manual, one of, migration step. To avoid having this step again if schedules are disabled for Prometheus we simply use the Scheduled Scaling VPA type for all Prometheus in all clusters where the Scheduled Scaling VPA is enabled.
2. While this makes scheduled scaling VPA more critical to the operation of Prometheus, also in clusters where there are not scaling scheduled defined, it has the benefit of validating the Scheduled Scaling VPA controller e.g. in test clusters without schedules and it means all Prometheus runs with the same VPA type.

This change, byt itself, does not effect current clusters. We will still enable (`scheduled_scaling_vpa_enabled: true`) per cluster for a safe roll out.


mikkeloscar:
:+1:

katyanna:
👍

PR #10344, opened 2026-01-07
fix: routesrv can cause fatal of new ingress pods in test 

szuecs:
fix: routesrv can cause fatal of new ingress pods in test 
Alerting happens if component ingress and routesrv are updated in test clusters at the same time


szuecs:
:+1:

madumalt:
👍

PR #10345, opened 2026-01-08
Fix e2e build

mikkeloscar:
After the CDP golang runtime was updated from Ubuntu 20.04 to 24.04 the e2e tests started failing because the `e2e.test` was dynamically linked.

```
/workdir/test/e2e/e2e.test: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /workdir/test/e2e/e2e.test)
```

The problem was that is used the binary `e2e.test` build from the make target `make build` as opposed to the one from `make build.linux.amd64` which is correctly statically build.

This PR fixes various things around the e2e build related to this:

1. Only add the files needed for e2e to the docker image
2. Explicitly add the binaries under `build/linux/${TARGETARCH}` to the image such that we have the right architecture.
3. Update build to use Go 1.25 (from 1.24).

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -11,7 +11,7 @@ pipeline:
   - event: pull_request
   vm_config:
     type: linux
-    image: "cdp-runtime/go-1.24"
+    image: "cdp-runtime/go-1.25"
     size: large  # speed up building kubernetes/kubernetes
   cache:
     paths:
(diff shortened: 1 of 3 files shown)


ponimas:
👍

mikkeloscar:
:+1:

PR #10355, opened 2026-01-09
sandbox-crd: adds paths support to sandbox spec

madumalt:
Adds paths support to Sandbox. Paths should not be duplicated, and startsWith a '/'

### Related Issue(s)

https://github.com/zalando-build/pg9-issues/issues/871


madumalt:
👍

ponimas:
👍

PR #10368, opened 2026-01-12
Manage Amazon VPC CNI as addon

mikkeloscar:
This moves the custom defined manifests for Amazon VPC CNI into EKS Addon configuration to make it simpler to manage long term and simpler to upgrade.

A few quirks:

* We can't set application/component at daemonset level. We can still set it at pod level which is most important for e.g. cost allocation.
* We already have a PodIdentityAssociations for IAM role to SA association. This means we can't move to the addon level definition of `PodIdentityAssociations`. Behind the scenes it's the same thing, but it's not tied to the addon which is maybe a bit confusing. We could likely migrate but we would need to ensure downtime of AWS access is limited.

Depends on #10377 being fully rolled out

* [x] dev
* [x] alpha
* [x] beta
* [x] stable

### TODO

* [x] Test IPv4 clusters
* [x] Test Custom Networking clusters
* [x] Test network policies
* [x] Fix deletion:
   ```
  This AWS::EKS::Addon resource is in a DELETE_FAILED state.
  
  Resource handler returned message: "Addon moved to failed status during Delete operation. Code: AdmissionRequestDenied, Message: Internal error occurred: failed calling webhook "configmap-admitter.teapot.zalan.do": failed to call webhook: Post "https://admission-controller.kube-system.svc:443/configmap?timeout=10s": no endpoints available for service "admission-controller", ResourceIds: [] " (RequestToken: c0011432-e804-3c3f-e483-5cbf76f326a0, HandlerErrorCode: NotStabilized)
    ```


cluster/cluster.yaml
```diff
@@ -319,6 +319,46 @@ Resources:
...
+      AddonName: vpc-cni
+      AddonVersion: "v1.19.5-eksbuild.3"
+      ClusterName: !Ref EKSCluster
+      ConfigurationValues: |
+{{- if eq .Cluster.ConfigItems.aws_vpc_cni_custom_networking "true" }}
+        eniConfig: {
+          create: true,
+          region: "{{.Cluster.Region}}"
+          subnets:
+{{- with $data := . }}
```
tcondeixa: why do we need to create $data from `.` ? is it to make it cleaner? otherwise we can just do `$.Values....`

mikkeloscar: I guess you're right. I just copied from what we had in the non-addon setup.

I didn't test this part of the config yet (see TODO) but we can try as you suggest when testing it.

tcondeixa: IIRC we also need to remove all the `with` to avoid conditional scoping, but we are not taking an advantage of it anyway since we don't use directly `.` inside the scope.

mikkeloscar: Updated

cluster/cluster.yaml
```diff
@@ -319,6 +319,43 @@ Resources:
...
+      AddonVersion: "v1.19.5-eksbuild.1"
+      ClusterName: !Ref EKSCluster
+      ConfigurationValues: !Sub |
+{{- if eq .Cluster.ConfigItems.aws_vpc_cni_custom_networking "true" }}
+        eniConfig:
+          create: true
+          region: "{{.Cluster.Region}}"
+          subnets:
+{{- range $az := .Values.availability_zones }}
+{{- with $azID := azID $az }}
```
tcondeixa: why do we need the $azID here? from what I see we only use $az for templating

mikkeloscar: True, removed!

tcondeixa:
👍

szuecs:
@mikkeloscar we could label it in our admission-webhook

mikkeloscar:
> @mikkeloscar we could label it in our admission-webhook

The problem is we have a chicken-and-egg problem between this and admission-controller. The aws-node (vpc-cni) daemonset must run before admission-controller can run. We currently exclude it from the admission-controller to avoid the problem, but this means we also can't utilize it for label injection.

mikkeloscar:
:+1:

PR #10374, opened 2026-01-13
Revert "Skipper ingress Karpenter scheduling annotations"

ayruslore:
Reverts zalando-incubator/kubernetes-on-aws#10292

We have decided to not use scheduled annotations, but rather use the standard dedicated node pools. With the dedicated node pools we can automatically propagate the label `ingress-enabled: true` to the EC2 nodes. But. with scheduled annotations this is right now not possible and we need to implement dynamic tagging. This will take longer and hence we proceed with dedicated node pool.

See label propagation https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/node-pools/worker-karpenter/provisioners.yaml#L80-L85 for dedicated node pools.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -128,7 +128,6 @@ worker_sg_legacy_cluster: ""
 # skipper node-pool
 enable_dedicate_nodepool_skipper: "true"
 skipper_attach_only_to_skipper_node_pool: "true"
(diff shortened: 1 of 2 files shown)


ayruslore:
👍

ayruslore:
👍

demonCoder95:
👍

PR #10379, opened 2026-01-14
Test e2e with EFS enabled

katyanna:
Temporary PR for checking e2e response to EFS add-on enablement.


katyanna:
Add-on definition validated ✅

PR #10382, opened 2026-01-14
Wiz version upgrade

hagenmathis:
Upgrades the Wiz component to the latest version using manifests from `wiz-kubernetes-integration` with version 0.3.6.
This is a patch for the sensor and a major version upgrade for the connector.

Changes in this PR:
- runs the wiz-sensor in the e2e clusters
- deletes unused secret: `wiz-sensor-apikey`
- adds jobs and nodes/proxy api groups to sensor cluster role
- removes the startup probe from the sensor
- adds readiness and liveness probes to connector
- major broker version is switching to using the same tunnel as the sensor, toggle with `WIZ_USE_HATUNNEL`
- `WIZ_BROKER_HEARTBEAT_DISABLE_CLUSTER_ID` is a temporary bug fix by Wiz

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -1318,7 +1318,7 @@ wiz_enable_runtime_connector: "true"
 wiz_enable_runtime_connector_broker: "true"
 {{ end }}
 {{if eq .Cluster.Environment "e2e"}}
-wiz_enable_runtime_sensor: "false"
+wiz_enable_runtime_sensor: "true"
 wiz_sensor_cpu: "200m"
 wiz_sensor_memory: "300Mi"
 {{ else }}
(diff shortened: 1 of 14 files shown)


cluster/config-defaults.yaml
```diff
@@ -1317,9 +1317,15 @@ wiz_enable_runtime_connector_broker: "false"
 wiz_enable_runtime_connector: "true"
 wiz_enable_runtime_connector_broker: "true"
 {{ end }}
+{{if eq .Cluster.Environment "e2e"}}
+wiz_enable_runtime_sensor: "false"
+wiz_sensor_cpu: "200m"
+wiz_sensor_memory: "300Mi"
+{{ else }}
```
mikkeloscar, line 1324: What's the motivation for this when the values are the same in the if and else? :)

hagenmathis: the motivation is opening a PR past 6 and overlooking it :D 
Thanks for catching it, I was going to check why the sensors did not deploy today, this makes it a lot easier.

hagenmathis:
:+1:

mikkeloscar:
:+1:

PR #10383, opened 2026-01-14
Add /var/lib/kubelet symlink to AMI

mikkeloscar:
This updates the AMI to a version that adds `/var/lib/kubelet` as symlink to `/opt/podruntime/kubelet`

This is done to ensure tools assuming kubelet data dir is at `/var/lib/kubelet` will work out of the box without work-arounds.

Specifically it supports #10330 

Depends on #10389 being fully rolled out

* [x] dev
* [x] alpha
* [x] beta
* [x] stable


mikkeloscar:
:+1:

katyanna:
👍

PR #10388, opened 2026-01-15
Add both kubelet-dir basedirs ahead of migration

mikkeloscar:
This configures old (`/opt/podruntime/kubelet`) and new (`/var/lib/kubelet`) paths on the ebs-csi to make it compatible with the AMI upgrade that switches kubelet to use `/var/lib/kubelet`.


mikkeloscar:
Closing in favor of #10389

PR #10389, opened 2026-01-15
Add both kubelet-dir basedirs ahead of migration

mikkeloscar:
This is done to enable the rollout of an AMI which has `/var/lib/kubelet` as a symlink: #10383 

This PR is targeting `dev` with the intention it should be rolled out fully to stable before rolling out the AMI change to have a smooth migration.


tcondeixa:
👍

mikkeloscar:
:+1:

PR #10423, opened 2026-01-19
Switch default target type of ALBC to instance

linki:
Switch the default target type from `ip` to `instance`.

`instance` matches the old behaviour more closely and avoids `albc` to manage Pod IPs in target groups. This makes this component less critical as rotating pod IPs don't need to updated in the target group by `albc`. Instead, instances stay in the target group for their entire lifetime and pod IPs are mapped by `kube-proxy`.

There are five clusters that already have active load balancers using `ip` mode. The ~default~Config Item has been set to `ip` in these clusters to keep them unchanged. After this rollout, the existing Services can be migrated separately. This seems doable because the list of services suggests there can be a short downtime for each of them.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -1353,6 +1353,7 @@ aws_vpc_cni_network_policy_enforcing_mode: "standard"
 # aws-load-balancer-controller resource settings
 aws_load_balancer_controller_cpu: "100m"
 aws_load_balancer_controller_mem_max: "4Gi"
+aws_load_balancer_controller_default_target_type: "instance"
 
 # configure if sandbox-controller should be deployed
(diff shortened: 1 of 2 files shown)


tcondeixa:
you PR does not seem to match the description, are we going to switch the default to instance? are we using a config item to set this as `ip` in the cluster-config-manager?

linki:
Yes, this switches the default to `instance`. Unless in five clusters that need special care because they already have some load balancers using `ip` mode.

I set the Config Item for those clusters already. Ideally, it doesn't take long to switch and we can avoid defining the Config Item in cluster-config-manager.

tcondeixa (approved):

linki:
👍

PR #10426, opened 2026-01-20
allow kro and ack EKS capabilities on kube-system namespace for experimentation on cloud-euc1-test

tcondeixa:
allow kro and ack EKS capabilities to have access to the kube-system namespace for experimentation on cloud-euc1-test

### DIFF
--- a/cluster/manifests/02-admission-control/teapot.yaml
+++ b/cluster/manifests/02-admission-control/teapot.yaml
@@ -633,7 +633,8 @@ webhooks:
         expression: '!(request.userInfo.username in ["system:kube-controller-manager", "system:kube-scheduler", "zalando-iam:zalando:service:k8sapi_credentials-provider"])'
       - name: 'exclude-eks-components'
         expression: '!request.userInfo.username.startsWith("eks:")'
-      {{- if eq .Cluster.Provider "zalando-eks"}}
+      # Restricted to cloud-euc1-test until we decide to make it production ready
+      {{- if eq .Cluster.Alias "cloud-euc1-test"}}
       - name: 'exclude-eks-kro-capability'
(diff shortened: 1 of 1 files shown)


cluster/manifests/02-admission-control/teapot.yaml
```diff
@@ -633,6 +633,13 @@ webhooks:
         expression: '!(request.userInfo.username in ["system:kube-controller-manager", "system:kube-scheduler", "zalando-iam:zalando:service:k8sapi_credentials-provider"])'
       - name: 'exclude-eks-components'
         expression: '!request.userInfo.username.startsWith("eks:")'
+      # Restricted to cloud-euc1-test until we decide to make it production ready
+      {{- if eq .Cluster.Alias "cloud-euc1-test"}}
```
linki, line 637: I think it's better to use a config item like `eks_capacility_kro_support_enabled` instead of the cluster alias. You can just set the value to `true` on your cluster with zregistry (even before rollout) to achieve the same thing. But you will be much more flexible once you want to test it somewhere else or roll it out.

tcondeixa: I understand your comment as a general best practice, and for most cases I would do it.
But the Capability (and the IAM Role) was deployed in a separate CloudFormation Stack, and we had to manually add the EKSClusterAdmin policy to the EKS Access Entry created by EKS.
I would rather not have that flexibility `on purpose` until we do things properly after the validation phase; we might even decide not to use KRO or ACK at all.

linki: Ok, got it.

linki:
👍

tcondeixa:
👍

PR #10442, opened 2026-01-20
Create default EFS and storage class

linki:
This creates a default EFS and its storage class. This matches the old behaviour with the additional benefit that the EFS doesn't need to be [created manually](https://cloud-platform.docs.zalando.net/howtos/enable-efs-storage-class/).

Users don't need to know the EFS ID. They only need to specify storage claims and mount them to their pods, e.g.:

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: efs-claim
spec:
  accessModes:
    - ReadWriteMany
  storageClassName: efs-sc
  resources:
    requests:
      storage: 5Gi

---

apiVersion: v1
kind: Pod
metadata:
  name: efs-app
spec:
  containers:
    - name: app
      image: rockylinux:8
      command: ["/bin/sh"]
      args: ["-c", "while true; do echo $(date -u) >> /data/out; sleep 5; done"]
      volumeMounts:
        - name: persistent-storage
          mountPath: /data
  volumes:
    - name: persistent-storage
      persistentVolumeClaim:
        claimName: efs-claim
```

The storage class points to the EFS created via the CF stack by reading the Stacks output variable. A mount point needs to be created for each region that the EFS is accessed from. It's protected by the worker nodes' security group. This was basically [done manually](https://cloud-platform.docs.zalando.net/howtos/enable-efs-storage-class/) before.

Each User volume is implemented by an "Access Point" which is managed by the EFS addon and it's using a unique subfolder within the EFS. Different PVCs will be completely isolated on the AWS side.

image

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -637,7 +637,6 @@ Resources:
       Path: /
       ManagedPolicyArns:
       - arn:aws:iam::aws:policy/service-role/AmazonEFSCSIDriverPolicy
-{{- if eq .Cluster.ConfigItems.eks_efs_filesystem_deployed "true" }}
   EFSFileSystem:
     Type: AWS::EFS::FileSystem
     Properties:
@@ -656,7 +655,6 @@ Resources:
       - !Ref EKSWorkerSecurityGroup
   {{ end }}
   {{ end }}
-{{- end }}
 {{- end }}
   EKSZalandoIAMProxyIAMRole:
     Type: AWS::IAM::Role
@@ -4007,7 +4005,7 @@ Outputs:
   #   Export:
   #     Name: "{{.Cluster.ID}}:eks-certificate-authority-data"
   #   Value: !GetAtt EKSCluster.CertificateAuthorityData
-  {{- if eq .Cluster.ConfigItems.eks_efs_filesystem_deployed "true" }}
+  {{- if eq .Cluster.ConfigItems.eks_efs_csi_support_enabled "true" }}
   EFSFileSystemID:
     Export:
       Name: '{{ .Cluster.ID}}:efs-filesystem-id'
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -637,6 +637,26 @@ Resources:
       Path: /
       ManagedPolicyArns:
       - arn:aws:iam::aws:policy/service-role/AmazonEFSCSIDriverPolicy
+{{- if eq .Cluster.ConfigItems.eks_efs_filesystem_deployed "true" }}
+  EFSFileSystem:
+    Type: AWS::EFS::FileSystem
+    Properties:
+      Encrypted: true
+      FileSystemTags:
+      - Key: Name
+        Value: "Default EFS for {{.Cluster.Alias}}"
+  {{ with $values := .Values }}
+  {{ range $index, $az := $values.availability_zones }}
+  EFSMountTargetZone{{$index}}:
+    Type: AWS::EFS::MountTarget
+    Properties:
+      FileSystemId: !Ref EFSFileSystem
+      SubnetId: "{{ index $values.subnets $az }}"
+      SecurityGroups:
+      - !Ref EKSWorkerSecurityGroup
+  {{ end }}
+  {{ end }}
+{{- end }}
 {{- end }}
   EKSZalandoIAMProxyIAMRole:
     Type: AWS::IAM::Role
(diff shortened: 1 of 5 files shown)


cluster/config-defaults.yaml
```diff
@@ -1344,7 +1340,8 @@ eks_zalando_iam_aws_proxy_hpa_cpu_target: "80"
 eks_zalando_iam_aws_proxy_hpa_memory_target: "80"
 eks_okta_identity_provider: "true"
 eks_fis_support_enabled: "false"
-eks_efs_csi_support_enabled: "false"
+eks_efs_csi_support_enabled: "true"
+eks_efs_filesystem_deployed: "true"
```
linki: just for e2e

cluster/cluster.yaml
```diff
@@ -637,6 +637,23 @@ Resources:
...
+    Type: AWS::EFS::FileSystem
+    Properties:
+      Encrypted: true
+  {{ with $values := .Values }}
+  {{ range $index, $az := $values.availability_zones }}
+  MountTargetResource{{$index}}:
+    Type: AWS::EFS::MountTarget
+    Properties:
+      FileSystemId: !Ref EFSFileSystem
+      SubnetId: "{{ index $values.lb_subnets $az }}"
```
mikkeloscar: I guess `lb_subnets` is not the right one to use

Today lb_subnets and node_subnets is the same in most clusters, but in the future it could be different (migration will be fun also considering EFS). At least we should use the node subnets here I would say.

linki: Thanks, I was just looking for anything that would yield the right values.

I changed it to `{{ index $values.subnets $az }}`.

linki:
👍

cluster/config-defaults.yaml
```diff
@@ -1344,7 +1344,15 @@ eks_zalando_iam_aws_proxy_hpa_cpu_target: "80"
...
 eks_okta_identity_provider: "true"
 eks_fis_support_enabled: "false"
+# {{if eq .Cluster.Environment "e2e"}}
+# Enables the Amazon EFS CSI driver addon
+eks_efs_csi_support_enabled: "true"
+# Sets up the default EFS for the cluster
+eks_efs_filesystem_deployed: "true"
+# {{else}}
 eks_efs_csi_support_enabled: "false"
+eks_efs_filesystem_deployed: "false"
```
katyanna:
```suggestion
```

I imagine you wanted these to be commented out/removed for testing?!
Also, what's the reasoning behind having both feature flags? Do we need the option of creating EFS storage without using the EFS CSI driver?

linki: > I imagine you wanted these to be commented out/removed for testing?!

My plan was to enable it in e2e tests (so the general setup is tested) but still keep it turned off by default for real clusters. It has the downside that we e2e test a configuration that differs from most clusters though.

linki: > Do we need the option of creating EFS storage without using the EFS CSI driver?

No, I thought there might be a use case for the opposite: Having only the driver but not the default EFS. That would mean users would need to "bring their own EFS and StorageClass". But I can also combine them if it makes more sense.

linki:
👍

cluster/config-defaults.yaml
linki: > Also, what's the reasoning behind having both feature flags?

I combined them into one config item.

cluster/config-defaults.yaml
```diff
@@ -1344,7 +1344,12 @@ eks_zalando_iam_aws_proxy_hpa_cpu_target: "80"
 eks_zalando_iam_aws_proxy_hpa_memory_target: "80"
 eks_okta_identity_provider: "true"
 eks_fis_support_enabled: "false"
+# {{if eq .Cluster.Environment "e2e"}}
+# Enables the Amazon EFS CSI driver addon and sets up the default EFS
+eks_efs_csi_support_enabled: "true"
+# {{else}}
 eks_efs_csi_support_enabled: "false"
+# {{end}}
```
katyanna:
```suggestion
{{ if eq .Cluster.Environment "e2e" }}
# Enables the Amazon EFS CSI driver addon and sets up the default EFS
eks_efs_csi_support_enabled: "true"
{{ else }}
eks_efs_csi_support_enabled: "false"
{{ end }}
```

katyanna: We should either uncomment the condition and have the flag enabled for e2e only or remove it and keep everything defaulted to `false`.

linki:
👍

katyanna:
👍

PR #10443, opened 2026-01-21
use cluster LocalID to be within maximum IAM role name length

tcondeixa:
Use cluster LocalID to be within the maximum IAM role name length allowed by AWS.
A new cluster was created within the maximum allowed name length, and this is failing since we are over the maximum size for the IAM role.
This will also affect the cluster with longer names that will be migrated to EKS.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -3371,6 +3371,7 @@ Resources:
       RoleName: "{{.Cluster.LocalID}}-audittrail-adapter"
     Type: 'AWS::IAM::Role'
 # {{- if and (eq .Cluster.Provider "zalando-eks") (eq .Cluster.ConfigItems.eks_ip_family "ipv6") }}
+# {{- if ne .Cluster.Name "business-partners-solutions-euc1-test" }}
   AWSLoadBalancerControllerIAMRole:
     Properties:
       AssumeRolePolicyDocument: !Sub
@@ -3580,6 +3581,7 @@ Resources:
           PolicyName: root
       RoleName: "aws-load-balancer-controller-{{.Cluster.Name}}"
     Type: 'AWS::IAM::Role'
+# {{- end }}
   AWSLBControllerIAMRole:
     Properties:
       AssumeRolePolicyDocument: !Sub
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -3578,7 +3578,7 @@ Resources:
               - elasticloadbalancing:SetRulePriorities
               Resource: "*"
           PolicyName: root
-      RoleName: "aws-load-balancer-controller-{{.Cluster.Alias}}"
+      RoleName: "aws-load-balancer-controller-{{.Cluster.Name}}"
     Type: 'AWS::IAM::Role'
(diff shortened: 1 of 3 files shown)


tcondeixa:
I think this approach work since the cloudformation is applied before the kubenetes manifests? I do't think there is an issue if we have a some seconds were the controller don't have access to AWS.
My main concern is how can we be sure the service account is updated before the deployment is updated? do we have any way to force explicit dependency? AFAIK the new pods need to start already with the temporary credentials from the new IAM role.

cluster/cluster.yaml
```diff
@@ -3578,7 +3578,7 @@ Resources:
               - elasticloadbalancing:SetRulePriorities
               Resource: "*"
           PolicyName: root
-      RoleName: "aws-load-balancer-controller-{{.Cluster.Name}}"
+      RoleName: "aws-load-balancer-controller-{{.Cluster.LocalID}}"
```
tcondeixa: I can also change `aws-load-balancer-controller` to `aws-lb-controller`, but it is still a change of the IAM role anyway.

mikkeloscar:
> I do't think there is an issue if we have a some seconds were the controller don't have access to AWS

I agree, this is fine.

> My main concern is how can we be sure the service account is updated before the deployment is updated? do we have any way to force explicit dependency? AFAIK the new pods need to start already with the temporary credentials from the new IAM role.

I don't think we can ensure this 100% no.

To 100% ensure it works, we would need two step "migration":

1. add the new role as an additional one, keeping the old. (and do the other changes of this PR)
2. Remove the old role and change/remove the force-deploy annotation to trigger another rotation

Maybe this is overkill and you can look for issues (e.g. in logs during roll out) and fix the few cases there likely will be by rotating pods. Up to you how complex you want to make it IMO.

tcondeixa:
> > I do't think there is an issue if we have a some seconds were the controller don't have access to AWS
> 
> I agree, this is fine.
> 
> > My main concern is how can we be sure the service account is updated before the deployment is updated? do we have any way to force explicit dependency? AFAIK the new pods need to start already with the temporary credentials from the new IAM role.
> 
> I don't think we can ensure this 100% no.
> 
> To 100% ensure it works, we would need two step "migration":
> 
> 1. add the new role as an additional one, keeping the old. (and do the other changes of this PR)
> 2. Remove the old role and change/remove the force-deploy annotation to trigger another rotation
> 
> Maybe this is overkill and you can look for issues (e.g. in logs during roll out) and fix the few cases there likely will be by rotating pods. Up to you how complex you want to make it IMO.

I can keep both roles for now, it's not a big issue. I will just exclude the new test cluster from the old role, so I can proceed with the cluster creation.

tcondeixa:
@mikkeloscar please check again. I'm keeping both roles, but the new name is slightly different because there are a few cluster where the LocalID is the same of the Cluster Name, and that would cause conflicts when creating the new IAM role.
I also excluded the new test cluster from the old IAM role to be able to continue with the cluster creation.

tcondeixa:
👍

mikkeloscar:
:+1:

PR #10446, opened 2026-01-21
[test PR] Skipper ingress Karpenter scheduling annotations

ayruslore:
Reverts zalando-incubator/kubernetes-on-aws#10374

Add node affinity for skipper-ingress scheduling
Add node affinity configuration for skipper-ingress pods when the
skipper_karpenter_node_pools_enabled flag is set.

When enabled:
- Adds requiredDuringSchedulingIgnoredDuringExecution node affinity
  with label zalando.org/node-pool: skipper-ingress to ensure pods
  only schedule on Karpenter-managed nodes
- Adds preferredDuringSchedulingIgnoredDuringExecution to prefer
  c7g instances (weight 100) over c6g/m6g instances (weight 50)

When disabled: the older ASG-managed worker nodes would be used.

This ensures skipper-ingress pods are scheduled on the correct node
pools based on the flag

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -92,6 +92,56 @@ spec:
               component: ingress
               application: skipper-ingress
 {{- end }}
+{{ if eq .Cluster.ConfigItems.skipper_karpenter_node_pools_enabled "false" }}
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: zalando.org/ingress-karpenter-managed
+                operator: DoesNotExist
+{{ else }}
+      affinity:
+       podAntiAffinity:
+         requiredDuringSchedulingIgnoredDuringExecution:
+         - labelSelector:
+             matchExpressions:
+             - key: application
+               operator: In
+               values:
+               - skipper-ingress
+             - key: component
+               operator: In
+               values:
+               - ingress
+           topologyKey: kubernetes.io/hostname
+       nodeAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            preference:
+              matchExpressions:
+              - key: node.kubernetes.io/instance-type
+                operator: In
+                values:
+                - c7g.large
+                - c7g.xlarge
+                - c7g.2xlarge
+                - c7g.4xlarge
+                - c7g.8xlarge
+                - m7g.large
+          - weight: 50
+            preference:
+              matchExpressions:
+              - key: node.kubernetes.io/instance-type
+                operator: In
+                values:
+                - c6g.large
+                - c6g.xlarge
+                - c6g.2xlarge
+                - c6g.4xlarge
+                - c6g.8xlarge
+                - m6g.large
+{{ end }}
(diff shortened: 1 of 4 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -478,12 +520,21 @@ spec:
...
 {{ end }}
 {{ if eq .Cluster.ConfigItems.enable_dedicate_nodepool_skipper "true"}}
+{{- if eq .Cluster.ConfigItems.skipper_karpenter_node_pools_enabled "true" }}
+      nodeSelector:
+        zalando.org/dedicated: skipper-ingress
+      tolerations:
+      - key: "zalando.org/dedicated"
+        operator: Exists
+        effect: NoSchedule
+{{ else }}
```
mikkeloscar: This part we should not have anymore as that is scheduling annotation related.

ayruslore: Yes I will remove it. I just created revert and now i am making changes locally.

cluster/manifests/skipper/deployment.yaml
```diff
@@ -91,6 +91,65 @@ spec:
...
+               operator: In
+               values:
+               - ingress
+           topologyKey: kubernetes.io/hostname
+       nodeAffinity:
+{{- if eq .Cluster.ConfigItems.skipper_karpenter_node_pools_enabled "true" }}
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: zalando.org/ingress-Karpenter-managed
```
mikkeloscar: I think this must be lowercase so: `zalando.org/ingress-karpenter-managed`

This could also just be an additional `nodeSelector`, simpler than a nodeAffinity.

cluster/manifests/skipper/deployment.yaml
```diff
@@ -91,6 +91,65 @@ spec:
...
+                values:
+                - "true"
+{{ else }}
+         requiredDuringSchedulingIgnoredDuringExecution:
+           nodeSelectorTerms:
+           - matchExpressions:
+             - key: zalando.org/ingress-ASG-managed
+               operator: In
+               values:
+               - "true"
```
mikkeloscar: I would leave the ingress-ASG-managed case out as this is the current behavior. As soon as we add the `zalando.org/ingress-Karpenter-managed` constraint it will automatically mean not ASG managed. We don't need both IMO, makes it more complex with more changes.

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -79,9 +79,11 @@ spec:
 #{{ end }}
 #{{ if and (eq .Cluster.ConfigItems.skipper_attach_only_to_skipper_node_pool "true") (eq (index .NodePool.ConfigItems "taints") "dedicated=skipper-ingress:NoSchedule") }}
     zalando.org/ingress-enabled: "true"
+    zalando.org/ingress-Karpenter-managed: "true"
```
mikkeloscar: I don't think it's strictly needed to have a tag with this value, but it doesn't hurt. If you keep it just make it lowercase to be consistent with the other suggested change.

ayruslore:
Test completed successfully. Closing the PR

PR #10451, opened 2026-01-21
Expose account clusters as deployment variable

linki:
Simple list of cluster aliases of the account and region. Returns the following in `stups-test` and `stups-euc1-test`.

```console
$ kubectl -n kube-system get cm deployment-config -o yaml | grep clusters
account-clusters: stups-test,stups-euc1-test
```

This is just a test to support [this use case](https://chat.google.com/room/AAAA1dSbk8M/X_i1a0RMEMk/X_i1a0RMEMk?cls=10) in a quick way.

### DIFF
--- a/cluster/manifests/deployment-service/01-config.yaml
+++ b/cluster/manifests/deployment-service/01-config.yaml
@@ -16,6 +16,7 @@ data:
   oidc-provider-arn: "{{.Cluster.OIDCProviderARN}}"
   oidc-subject-key: "{{.Cluster.OIDCSubjectKey}}"
   iam-role-trust-relationship-template: '{{.Cluster.IAMRoleTrustRelationshipTemplate}}'
+  account-clusters: '{{$sep := ""}}{{- range .Cluster.AccountClusters}}{{ if eq .LifecycleStatus "ready" }}{{$sep}}{{.Alias}}{{$sep = ","}}{{- end}}{{- end}}'
   s3-bucket-name: "{{ .Cluster.ConfigItems.deployment_service_bucket_name }}"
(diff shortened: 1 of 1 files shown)


cluster/manifests/deployment-service/01-config.yaml
```diff
@@ -16,6 +16,7 @@ data:
   oidc-provider-arn: "{{.Cluster.OIDCProviderARN}}"
   oidc-subject-key: "{{.Cluster.OIDCSubjectKey}}"
   iam-role-trust-relationship-template: '{{.Cluster.IAMRoleTrustRelationshipTemplate}}'
+  account-clusters: '{{$sep := ""}}{{- range .Cluster.AccountClusters}}{{ if eq .LifecycleStatus "ready" }}{{$sep}}{{.Alias}}{{$sep = ","}}{{- end}}{{- end}}'
```
tcondeixa, line 19: is this not adding an extra `,` at the end of the string?

tcondeixa:
can you just mention the use-case to introduce this? thanks

cluster/manifests/deployment-service/01-config.yaml
tcondeixa: you add the comma at the beginning and not the end, looks good

linki: No, it puts the comma before the value, except in the first iteration.

linki:
> can you just mention the use-case to introduce this? thanks

Done

tcondeixa:
👍

linki:
I need to re-confirm if that's really helping ACID.

linki:
This isn't what they need. Closing.

PR #10452, opened 2026-01-21
Enable scheduled-scaling-vpa by default

mikkeloscar:
Enables scheuled-scaling-vpa by default after it has been enabled via CCM (by channel) in all clusters.


mikkeloscar:
:+1:

PR #10453, opened 2026-01-22
Add node affinity for skipper-ingress scheduling 

ayruslore:
## Summary
Add Karpenter node pool support for skipper ingress scheduling with node affinity preferences.

## Changes
- Add `skipper_karpenter_node_pools_enabled` configuration flag (default: true)
- Add pod anti-affinity to spread skipper pods across different nodes
- Add node affinity rules to prefer c7g and m7g instances when Karpenter pools are enabled
- Maintain fallback scheduling for non-Karpenter environments (prevent scheduling on karpenter-managed nodes)
- Add Karpenter node pool with skipper-ingress taint for dedicated scheduling
- Add Karpenter provisioner requirements to label skipper-ingress nodes with `zalando.org/ingress-karpenter-managed`

## Testing
- Updated e2e cluster configuration with new Karpenter skipper ingress node pool

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -128,6 +128,7 @@ worker_sg_legacy_cluster: ""
 # skipper node-pool
 enable_dedicate_nodepool_skipper: "true"
 skipper_attach_only_to_skipper_node_pool: "true"
+skipper_karpenter_node_pools_enabled: "true"
 
 skipper_suppress_route_update_logs: "true"
 skipper_validate_query: "true"
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -92,6 +92,56 @@ spec:
               component: ingress
               application: skipper-ingress
 {{- end }}
+{{ if eq .Cluster.ConfigItems.skipper_karpenter_node_pools_enabled "true" }}
+      affinity:
+       podAntiAffinity:
+         requiredDuringSchedulingIgnoredDuringExecution:
+         - labelSelector:
+             matchExpressions:
+             - key: application
+               operator: In
+               values:
+               - skipper-ingress
+             - key: component
+               operator: In
+               values:
+               - ingress
+           topologyKey: kubernetes.io/hostname
+       nodeAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            preference:
+              matchExpressions:
+              - key: node.kubernetes.io/instance-type
+                operator: In
+                values:
+                - c7g.large
+                - c7g.xlarge
+                - c7g.2xlarge
+                - c7g.4xlarge
+                - c7g.8xlarge
+                - m7g.large
+          - weight: 50
+            preference:
+              matchExpressions:
+              - key: node.kubernetes.io/instance-type
+                operator: In
+                values:
+                - c6g.large
+                - c6g.xlarge
+                - c6g.2xlarge
+                - c6g.4xlarge
+                - c6g.8xlarge
+                - m6g.large
+{{ else }}
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: zalando.org/ingress-karpenter-managed
+                operator: DoesNotExist
+{{ end }}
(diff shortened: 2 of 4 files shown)


ayruslore:
A test PR verification was done by enabling the flag here in https://github.com/zalando-incubator/kubernetes-on-aws/pull/10446

cluster/config-defaults.yaml
```diff
@@ -128,6 +128,7 @@ worker_sg_legacy_cluster: ""
 # skipper node-pool
 enable_dedicate_nodepool_skipper: "true"
 skipper_attach_only_to_skipper_node_pool: "true"
+skipper_karpenter_node_pools_enabled: "false"
```
szuecs: why default to false?

szuecs:
:+1:

test/e2e/cluster_config.sh
```diff
@@ -130,6 +130,17 @@ EOFF
...
+    instance_types:
+    - "not-specified"
+    min_size: 0
+    max_size: 0
+    profile: worker-karpenter
+    name: skipper-ingress-karpenter-catch-all
+    config_items:
+      labels: dedicated=skipper-ingress
+      taints: dedicated=skipper-ingress:NoSchedule
+      requirements: "- key: zalando.org/ingress-karpenter-managed\n  operator: Exists\n"
```
mikkeloscar: Since we now went with the anti-affinity, we no longer need the requirement defined.

Same in cluster/node-pools/worker-karpenter/provisioners.yaml where the requirement is also defined, that is not needed.

ayruslore: removed the requirements and the label from the provisioners

cluster/config-defaults.yaml
```diff
@@ -128,6 +128,7 @@ worker_sg_legacy_cluster: ""
 # skipper node-pool
 enable_dedicate_nodepool_skipper: "true"
 skipper_attach_only_to_skipper_node_pool: "true"
+skipper_karpenter_node_pools_enabled: "true"
```
ponimas, line 131: `Add skipper_karpenter_node_pools_enabled config flag (default: false)`
?

cluster/manifests/skipper/deployment.yaml
```diff
@@ -480,6 +530,9 @@ spec:
 {{ if eq .Cluster.ConfigItems.enable_dedicate_nodepool_skipper "true"}}
       nodeSelector:
         dedicated: skipper-ingress
+{{ if eq .Cluster.ConfigItems.skipper_karpenter_node_pools_enabled "true" }}
+        zalando.org/ingress-karpenter-managed: "true"
```
madumalt, line 534: just curious, what is the significance of this?

cluster/config-defaults.yaml
ayruslore: updated the description. We need for it to be true to run e2e tests using karpenter nodepool and scheduling annotations

cluster/manifests/skipper/deployment.yaml
ayruslore: We dont want both node pools (ASG, Karpenter) to race or the scheduler to decide randomly which node pool to use. This new taint is the difference between ASG managed nodes and Karpenter nodes.

When the flag is false the anti affinity will ensure skipper schedules only on ASG nodes.
When the flag is true, the taint will require to be set ensuring only Karpenter managed nodes are used.

ASG nodes do not use this new taint. And Karpenter node pools add this here in startup https://github.com/zalando-incubator/kubernetes-on-aws/pull/10453/changes#diff-74af8dd6cb1e9f4e450b66a4de808737ccef4387b16b6a3a81824e973b5f2e2dR195

This way we control with flag which node pools are selected for skipper pods.

cluster/config-defaults.yaml
madumalt: in that case I assume we have already disabled it all the clusters where we do not wanna rollout karpenter node pools.

madumalt (approved):

cluster/config-defaults.yaml
ayruslore: Yes. Its disabled in all.

ponimas:
👍

cluster/config-defaults.yaml
ayruslore: right. its set to true so we can run e2e tests.

PR #10462, opened 2026-01-26
Enable OPA Status Plugin

x418:
## What & Why?

Part of [Expose OPA Status Metrics](https://github.com/zalando-infosec/corporate-iam-issues/issues/1331)

This change to the OPA configmap ensures that the PrometheusOverride hook added via https://github.com/zalando/skipper/pull/3631 enables the Status metrics that we plan to consume.


szuecs:
It looks like this was already merged

PR #10470, opened 2026-01-27
feat: add additional capability to WIZ sensor

vinaythupili:
To use WIZ run time response policies we have to enable the KILL capabilities to the WIZ sensor. 

- WIZ Sensor Capabilities : https://docs.wiz.io/docs/sensor-required-permissions?lng=en#linux-capabilities 
- WIZ Runtime Response Policies : https://docs.wiz.io/docs/runtime-response-policies 
- added wiz_enable_untime_sensor_response_capabilities to enable per cluster 

This feature is necessary to enable the Traffic restrictions based on Geo based threat detection rules.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -1333,6 +1333,8 @@ wiz_priority: "true"
 # to deploy the sensor and connector on specific nodes, by manually setting the node selector label on the nodes.
 # This is useful when you want to deploy the sensor and connector on specific nodes.
 wiz_node_feature_rollout : "false"
+# Enable runtime sensor response capabilities such as KILL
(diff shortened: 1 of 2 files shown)


mikkeloscar:
I would propose we add this behind a feature flag. Basically add an if-else (see other examples in the repo) to remove the drop or not. Then it can be enabled per cluster.

vinaythupili:
> I would propose we add this behind a feature flag. Basically add an if-else (see other examples in the repo) to remove the drop or not. Then it can be enabled per cluster.

Sure, Will make it as Optional Initially

mikkeloscar:
:+1:

vinaythupili:
👍

PR #10473, opened 2026-01-27
skipper: update canary to v0.24

szuecs:
skipper: update canary to v0.24

2 minor version jump because these were 2 security related default config breaking changes.
1. We know that we have disabled lua, so we change the default to "false" as all checks test for "true" it does not change anything, but we can pass the result to -enable-lua arg.
2. we set externalname service to enabled but use config item to remove it in clusters that do not use this feature. In our current case it does not matter much, but changing towards multi tenant config is always a good thing

replaces https://github.com/zalando-incubator/kubernetes-on-aws/pull/10416

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -1,10 +1,10 @@
 {{/* image-updater-bot detects *image variables so use name with suffix to disable it for the main image */}}
 
 {{ $main_image_updated_manually := "container-registry.zalando.net/teapot/skipper-internal:v0.22.223-1325" }}
-{{ $canary_image := "container-registry.zalando.net/teapot/skipper-internal:v0.22.223-1325" }}
+{{ $canary_image := "container-registry.zalando.net/teapot/skipper-internal:v0.24.5-1333" }}
 
 {{/* Optional canary arguments separated by "[cf724afc]" to allow whitespaces, e.g. "-foo=has a whitespace[cf724afc]-baz=qux" */}}
-{{ $canary_args := "" }}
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

cluster/manifests/skipper/deployment.yaml
```diff
@@ -1,10 +1,10 @@
 {{/* image-updater-bot detects *image variables so use name with suffix to disable it for the main image */}}
 
 {{ $main_image_updated_manually := "container-registry.zalando.net/teapot/skipper-internal:v0.22.223-1325" }}
-{{ $canary_image := "container-registry.zalando.net/teapot/skipper-internal:v0.22.223-1325" }}
+{{ $canary_image := "container-registry.zalando.net/teapot/skipper-internal:v0.24.5-1333" }}
 
 {{/* Optional canary arguments separated by "[cf724afc]" to allow whitespaces, e.g. "-foo=has a whitespace[cf724afc]-baz=qux" */}}
-{{ $canary_args := "" }}
+{{ $canary_args := "-enable-kubernetes-external-names={{ .Cluster.ConfigItems.skipper_externalname_service_enabled }}[cf724afc]-enable-lua={{ .Cluster.ConfigItems.skipper_lua_scripts_enabled }}" }}
```
madumalt, line 7: just curious, what is `[cf724afc]`? the way indicate a whitespace?

MustafaSaber: yes it's written in the comment above it

MustafaSaber (approved):

cluster/manifests/skipper/deployment.yaml
madumalt: ohh I didn't see it, thanks!

PR #10475, opened 2026-01-28
Use print function for canary_args string construction

ponimas:
Replace string interpolation with explicit print function call to construct the canary_args variable in the skipper deployment manifest.

fix for https://github.com/zalando-incubator/kubernetes-on-aws/pull/10473


ponimas:
👍

mikkeloscar:
:+1:

PR #10477, opened 2026-01-28
Update admission controller config for  eks-node-monitoring-agent

zaklawrencea:
Add admission controller exclusion for the eks-node-monitoring-agent. In EKS Clusters the agent runs as an EKS Addon and excluding it from the Pod admitter works around checks such as a non-compliant image (the image is not being republished for EKS Clusters).


zaklawrencea:
:+1:

katyanna:
👍

PR #10384, opened 2026-01-14
Move from /opt/podruntime/kubelet to /var/lib/kubelet

mikkeloscar:
Rename paths from `/opt/podruntime/kubelet` to `/var/lib/kubelet`

Depends on #10383 

* [x] dev
* [x] alpha
* [x] beta
* [x] stable
* [x] All clusters run latest AMI

PR #10479, opened 2026-01-28
s3-csi-driver: Update kubelet dir

mikkeloscar:
This updates the kubelet dir path to `/var/lib/kubelet` after it was changed as part of Kubernetes v1.34 roll out.

This is just for the s3-csi driver which is used only by 1 clusters so far.

The rest is done in https://github.com/zalando-incubator/kubernetes-on-aws/pull/10384

### DIFF
--- a/cluster/manifests/s3-csi-driver/daemonset.yaml
+++ b/cluster/manifests/s3-csi-driver/daemonset.yaml
@@ -47,7 +47,7 @@ spec:
             fieldRef:
               fieldPath: spec.nodeName
         - name: HOST_PLUGIN_DIR
-          value: /opt/podruntime/kubelet/plugins/s3.csi.aws.com/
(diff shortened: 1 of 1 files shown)


tcondeixa:
If the only 2 usecases are `playground` and this new EKS cluster why don't we use the EKS addon?

mikkeloscar:
:+1:

mikkeloscar:
> If the only 2 usecases are `playground` and this new EKS cluster why don't we use the EKS addon?

I'm not opposed to this, I was testing the v2 of the S3 CSI driver and not able to make it work in our setup due to two issues, one related to toleration handling, I created an upstream issue, and something related to mounting which I didn't investigate/understand yet.

I would propose to move this short-term fix forward, and look into using the addon as a second step.

tcondeixa:
👍

PR #10485, opened 2026-02-02
Add efs-provisoner to deletions.yaml so it can be turned off

linki:
Provide a way to turn off the legacy EFS provisioner when the old config item is flipped off.

This will remove all legacy EFS provisioner resources in the cluster (not the EFS itself) and it will also keep the PVC and PV resources. So, turning it off accidentally won't delete any data and it can be brought back by setting the previous value.


linki:
👍

tcondeixa (approved):

PR #10499, opened 2026-02-04
Allow to configure static prescale for Prometheus

mikkeloscar:
This adds a new config-item: `prometheus_static_scheduled_scaling_events` which allows defining a list of static scheduled scaling configurations for Prometheus. This is an alternative to `prometheus_scheduled_scaling_events` which assumes a dynamic scaling function.

The purpose of this is to handle cases where scheduled scaling is used with a static value and a scaling function is hard to create. Concrete examples in Zalando are ZPS and ZL-UJLT loadtests where a static value is used for the scheduled scaling event. In this case it makes sense to configure Prometheus to scale to the last known needed resources based on previous loadtests.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -558,9 +558,18 @@ prometheus_remote_max_backoff: "10s"
 # Comma-separated list of user ids allowed to access Prometheus UI
 prometheus_ui_users: ""
 
-# Scheduled scaling events for Prometheus VPA (in the format
-# <name>:<pre-start-min>:<cpu>:<memory>, multiple events separated by commas)
+# Dynamic Scheduled scaling events for Prometheus VPA (in the format
+# <event-name>:<pre-start-min>:<cpu>:<memory>, multiple events separated by commas)
+# <cpu> and <memory> must be a scaling function which takes scalingValue as a
+# parameter. Example for linear regression function:
+# ("<coef> * scalingValue + <constant>")
 prometheus_scheduled_scaling_events: ""
+
+# Static scheduled scaling events for Prometheus VPA (in the format 
+# <event-name>:<pre-start-min>:<cpu>:<memory>, multiple events separated by commas)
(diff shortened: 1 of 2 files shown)


tcondeixa:
Is it not possible just to pass a static value to the `prometheus_scheduled_scaling_events` or are we creating a new one to make it clear these are static?

mikkeloscar:
> Is it not possible just to pass a static value to the `prometheus_scheduled_scaling_events` or are we creating a new one to make it clear these are static?

The format is different on the VPA, so we would either need to "detect" function vs. static value in the templating or extend the format to have some way to indicate what it is.

tcondeixa:
> > Is it not possible just to pass a static value to the `prometheus_scheduled_scaling_events` or are we creating a new one to make it clear these are static?
> 
> The format is different on the VPA, so we would either need to "detect" function vs. static value in the templating or extend the format to have some way to indicate what it is.

do we already have the `minAllowed` supported or do we only have `unitResource`? IIRC the AST that we use should be able to parse a static value (Quantity) if we don't pass anything else, but probably there is something else in the logic.
If it's easy to change the VPA to accept both the dynamic and static IMO it would be better, so we provide a simpler interface and hide the complexity inside.

mikkeloscar:
> > > Is it not possible just to pass a static value to the `prometheus_scheduled_scaling_events` or are we creating a new one to make it clear these are static?
> > 
> > 
> > The format is different on the VPA, so we would either need to "detect" function vs. static value in the templating or extend the format to have some way to indicate what it is.
> 
> do we already have the `minAllowed` supported or do we only have `unitResource`? IIRC the AST that we use should be able to parse a static value (Quantity) if we don't pass anything else, but probably there is something else in the logic. If it's easy to change the VPA to accept both the dynamic and static IMO it would be better, so we provide a simpler interface and hide the complexity inside.

You're right, that is a good point and should likely already work (you just can't define the value as e.g. `10Gi` but we have a PR for that).

PR #10503, opened 2026-02-04
Enable scheduled scaling VPA by default

mikkeloscar:
This was successfully enabled in all current clusters so we can enable it by default.


tcondeixa (approved):

mikkeloscar:
:+1:

PR #10507, opened 2026-02-05
Allow disabling kube-cluster-autoscaler

mikkeloscar:
With skipper-ingress moving to karpenter based pools, we soon won't need to run CA anymore.

Add config-item to allow disabling kube-cluster-autoscaler per cluster (enabled by default).


mikkeloscar:
:+1:

katyanna:
:+1:

PR #10508, opened 2026-02-05
Use on-demand Karpenter nodes for skipper ingress in production

ayruslore:
## Summary
- Add conditional nodeSelector to use on-demand Karpenter capacity-type for skipper ingress in production environments
- Non-production environments continue to use spot instances

## Context
When `skipper_karpenter_node_pools_enabled` is true and the cluster environment is production, the deployment now specifies `karpenter.sh/capacity-type: on-demand` to ensure production traffic uses reliable on-demand instances rather than spot instances.

## Test plan
- [ ] Verify the change is applied only in production environments
- [ ] Confirm skipper ingress pods are scheduled on on-demand nodes in production
- [ ] Ensure non-production clusters continue to use spot instances


mikkeloscar:
:+1:

ayruslore:
👍

PR #10509, opened 2026-02-05
Sandbox e2e tests

ponimas:
implementation of the e2e tests for sandbox controller.


madumalt (approved):

cluster/config-defaults.yaml
```diff
@@ -1376,4 +1376,4 @@ aws_load_balancer_controller_mem_max: "4Gi"
 aws_load_balancer_controller_default_target_type: "instance"
 
 # configure if sandbox-controller should be deployed
-sandbox_controller_enabled: "false"
```
MustafaSaber, line 1379: Do we have the controller enabled by default everywhere?

MustafaSaber: Answered in another place, it's enabled by default so e2e can work but disabled by configItem

ayruslore (approved):

PR #10524, opened 2026-02-06
Add spot capacity type for skipper in non-production environments

ayruslore:
## Summary

- Add spot capacity type for skipper karpenter nodes in non-production environments
- Only applies to non-production clusters (maintains on-demand for production)
- Reduces infrastructure costs for non-production workloads


ayruslore:
👍

greeshma1196:
👍🏽

PR #10532, opened 2026-02-10
skipper: update to v0.24

szuecs:
2 minor version jump because these were 2 security related default config breaking changes.
    1. We know that we have disabled lua, so we change the default to "false" as all checks test for "true" it does not change anything, but we can pass the result to -enable-lua arg.
    2. we set externalname service to enabled but use config item to remove it in clusters that do not use this feature. In our current case it does not matter much, but changing towards multi tenant config is always a good thing


ponimas:
👍

ayruslore:
👍

PR #10540, opened 2026-02-11
enable c7g instance family for default karpenter node-pool

demonCoder95:
It's been request by some users that they wish to test the `c7g` instance family for their workloads. Enabling it behind a config-item to be able to turn it off easily. 

Once we determine it's fine to enable it by default in all clusters, we can drop the feature flag.

### DIFF
--- a/cluster/node-pools/worker-karpenter/provisioners.yaml
+++ b/cluster/node-pools/worker-karpenter/provisioners.yaml
@@ -218,6 +218,9 @@ spec:
             - "c6in"
             - "m6in"
             - "r6in"
+#{{ if eq .Cluster.ConfigItems.karpenter_instance_family_c7g_enabled "true" }}
+            - "c7g"
+#{{ end }}
             - "c7i"
             - "m7i"
             - "r7i"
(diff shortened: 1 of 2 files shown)


cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -218,6 +218,9 @@ spec:
             - "c6in"
             - "m6in"
             - "r6in"
+#{{ if eq .Cluster.ConfigItems.karpenter_instance_family_c7g_enabled "true" }}
+            - "c7g"
```
tcondeixa, line 222: this instance is graviton, right? this might affect other workload running in the same cluster.
IMO this should not be part of the pool, but should be enabled with annotations per deployment.

demonCoder95: Ah, you're right. This part is actually not needed. I forgot we also have the `catch-all` node-pool which allows ARM instances, so all we really need for the customer is an Affinity Predicate to prefer c7g. As we don't explicitly filter the `c7g` instance family from our setup, it should be provisioned by Karpenter, depending on the availability in the region, of course.

I'll close this PR. It's not necessary.

PR #10545, opened 2026-02-11
allow EKS worker to connect to legacy EFS

linki:
This expands the legacy EFS security group to allow access from the entire VPC.

This allows to mount the legacy EFS from the EKS cluster using NFS. It's already possible without any provisioner needed. The only thing missing is network connectivity.

FYI: Merging the two SGs would make this PR obsolete.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -1099,6 +1099,16 @@ Resources:
       SourceSecurityGroupId: !Ref WorkerSecurityGroup
       ToPort: 2049
     Type: 'AWS::EC2::SecurityGroupIngress'
+{{- if eq .Cluster.ConfigItems.efs_allow_vpc_connection "true" }}
+  EFSSecurityGroupIngressFromVPCCIDR:
+    Properties:
+      FromPort: 2049
+      GroupId: !Ref EFSWorkerSecurityGroup
+      IpProtocol: tcp
+      CidrIp: "{{.Values.vpc_ipv4_cidr}}"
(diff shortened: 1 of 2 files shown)


linki:
👍

cluster/cluster.yaml
```diff
@@ -1099,6 +1099,16 @@ Resources:
       SourceSecurityGroupId: !Ref WorkerSecurityGroup
       ToPort: 2049
     Type: 'AWS::EC2::SecurityGroupIngress'
+{{- if eq .Cluster.ConfigItems.efs_allow_vpc_connection "true" }}
+  EFSSecurityGroupIngressFromVPCCIDR:
+    Properties:
+      FromPort: 2049
+      GroupId: !Ref EFSWorkerSecurityGroup
+      IpProtocol: tcp
+      CidrIp: "{{.Values.vpc_ipv4_cidr}}"
```
tcondeixa, line 1108: do we connect to FES only with IPV4 or do we also need IPv6 for EKS?

linki: The `*.efs.eu-central-1.amazonaws.com` endpoint only resolves to IPv4 addresses at the moment.

I'll follow up with an IPv6 version if it's feasible.

tcondeixa: if we don't have IPv6 enabled (or it's not even provided by AWS) let's just use IPv4.
We can address that separated after the migration to EKS is finished.

tcondeixa (approved):

PR #10546, opened 2026-02-11
Replace sleep with poll when verifying mocked response

ponimas:
Use wait.PollUntilContextTimeout instead of a fixed 2-second sleep to wait for Skipper to pick up new routes, with a 30-second timeout and 2-second polling interval.

### DIFF
--- a/test/e2e/sandbox.go
+++ b/test/e2e/sandbox.go
@@ -429,7 +429,7 @@ var _ = describe("Sandbox Controller", func() {
 			})
 
 			By("Executing HTTP request to verify mocked response is returned")
-			err = wait.PollUntilContextTimeout(context.TODO(), 2*time.Second, 30*time.Second, false, func(ctx context.Context) (bool, error) {
+			err = wait.PollUntilContextTimeout(context.TODO(), 2*time.Second, 60*time.Second, false, func(ctx context.Context) (bool, error) {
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

test/e2e/sandbox.go
```diff
@@ -40,7 +41,7 @@ import (
 
 func waitForSandboxedRoutegrouop(ctx context.Context, c clientset.Interface, ns, originalRGName string) (*rgv1.RouteGroup, error) {
 	var sandboxedRG *rgv1.RouteGroup
-	err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 1*time.Minute, true, func(ctx context.Context) (done bool, err error) {
+	err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 5*time.Minute, true, func(ctx context.Context) (done bool, err error) {
```
MustafaSaber, line 44: 5min is a bit too much I think the pipeline is already so big that shouldn't matter, but it also feels that routegroup shouldn't take 5mins to be created, no?

szuecs: Cloud load balancer can take time.

szuecs (approved):

test/e2e/sandbox.go
ponimas: IMO that should be seconds too, but somehow lower limits didn't help.

ponimas:
👍

PR #10559, opened 2026-02-12
fix: enable console decision logging for validation webhook

x418:
## What & Why?

Override the discovery bundle's decision logging configuration to prevent OPA from attempting to upload logs to a remote service.

See [decision-logs](https://www.openpolicyagent.org/docs/configuration#decision-logs)'s `decision_logs.console` property


vlktna:
because of the way current discovery bundle is configured to hardcode the decision_logs.service property, there are spurious error logs (in the validation webhook's deployment) reporting that decision logs couldn't be sent.  This in itself is not a problem since the decision logs are only sent when there is an actual evaluation by OPA. The draft PR won't fix the problem

PR #10569, opened 2026-02-13
Update deletions order for cluster_autoscaler_enabled config item

zaklawrencea:
The current deletions order attempts to delete the `ServiceAccount` first, which fails as the `Daemonset` Pods are still running (and it's deletion is never reached as CLM discontinues on failure). This PR changes the order so the `Daemonset` and `Deployment`s are removed first.


ayruslore (approved):

mikkeloscar:
:+1:

PR #10581, opened 2026-02-16
e2e: Change limit-az test pool to karpenter

mikkeloscar:
Change the limit-az test node pool to be karpenter based as we're phasing out CA pools.


mikkeloscar:
:+1:

linki:
👍

PR #10582, opened 2026-02-16
log failed lookups for .cluster.local services

demonCoder95:
This enhances the coredns configuration to log failed lookups for cluster local service names with `.cluster.local` suffix.


demonCoder95:
Rebased the PR on `kube-1.35` branch so I can safely override `teapot` which is using `kube-1.35` as an override channel right now. Also, changed the base branch to `kube-1.35` for now.

PR #10585, opened 2026-02-16
skipper: migrate redis to valkey

szuecs:
skipper: migrate redis to valkey

Changing the name from redis to valkey is done on purpose such that valkey can start running and skipper-ingress rolling before skipper redis objects will be deleted. 

Right now only for load testing and change channel in our test cluster.
Test was successful

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -4,6 +4,14 @@ pre_apply:
 - name: deployment-service-controller
   kind: StatefulSet
   namespace: kube-system
+{{- if eq .Cluster.ConfigItems.skipper_redis_cleanup_enabled "true" }}
+- kind: StatefulSet
+  name: skipper-ingress-redis
+- kind: HorizontalPodAutoscaler
+  name: skipper-ingress-redis
+- kind: Service
+  name: skipper-ingress-redis
+{{- end }}
 
 # everything defined under here will be deleted after applying the manifests
 post_apply:
--- a/cluster/manifests/skipper/hpa-redis.yaml
+++ b/cluster/manifests/skipper/hpa-redis.yaml
@@ -1,32 +1,33 @@
 apiVersion: autoscaling/v2
 kind: HorizontalPodAutoscaler
 metadata:
-  name: skipper-ingress-redis
+  name: skipper-ingress-valkey
   namespace: kube-system
   labels:
-    application: skipper-ingress-redis
+    application: skipper-ingress
+    component: valkey
 spec:
   scaleTargetRef:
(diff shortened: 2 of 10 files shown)


cluster/manifests/deletions.yaml
```diff
@@ -4,6 +4,14 @@ pre_apply:
 - name: deployment-service-controller
   kind: StatefulSet
   namespace: kube-system
+{{- if eq .Cluster.ConfigItems.skipper_redis_cleanup_enabled "true" }}
```
MustafaSaber, line 7: I think we should also add valkey cleanup for now, then after migration remove it.

cluster/manifests/skipper/hpa-redis.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Cluster.ConfigItems.skipper_ingress_swarm_type "redis" }}
```
greeshma1196, line 1: Is this a syntax error `- if `, should it be `if`?

cluster/manifests/deletions.yaml
greeshma1196: Why do we need a valkey clean up if we are migrating to valkey?

MustafaSaber: In the migration process, if something went wrong and we need to revert it makes sense to also delete resources

MustafaSaber: for me
enable = deploy resources
disable = delete resources

greeshma1196: i agree on the naming, i think we should drop the word 'enabled'

szuecs:
the `-` is removing the whitespace in front.

cluster/manifests/deletions.yaml
szuecs: Do we want to run 1h e2e test for a spell issue?

```
grep enabled cluster/config-defaults.yaml W
104
```

I want to do a forward migration, steps:
1. deploy valkey
2. switch skipper
3. delete redis

MustafaSaber: I don't mind merging, my only concern is if something failed and want to fix something we will have both running for some time. The cost also won't be that much it's just personal preference

MustafaSaber (approved):

greeshma1196:
👍🏽

cluster/manifests/deletions.yaml
szuecs: Both are running on purpose for the same time!
The cleanup will be done if we are sure the system runs as expected.

PR #10596, opened 2026-02-18
Add additional SG to skipper-ingress nodes [1/2]

mikkeloscar:
Since skipper-ingress always runs on dedicated nodes we only need to open port 9998 and 9999 on those nodes instead of opening on _all_ Kubernetes nodes.

This adds a secondary Security Group `SkipperIngressNodeSecurityGroup` which will only be attached to skipper-ingress nodes.

After this is rolled out, we can drop the skipper-ingress ports from the main WorkerSecurityGroup (We need to do it in two steps to ensure availability during the migration).

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -132,6 +132,33 @@ Resources:
         - Key: 'karpenter.sh/discovery'
           Value: '{{ .Cluster.ID }}/WorkerNodeSecurityGroup'
       VpcId: "{{.Cluster.ConfigItems.vpc_id}}"
+  EKSIngressWorkerSecurityGroup:
+    Type: "AWS::EC2::SecurityGroup"
+    Properties:
+      GroupDescription: "{{ .Cluster.ID }}-ingress-worker-sg"
+      SecurityGroupIngress:
+        - CidrIp: "0.0.0.0/0"
+          FromPort: 9998
+          IpProtocol: tcp
+          ToPort: 9999
+        - CidrIpv6: "::/0"
+          FromPort: 9998
+          IpProtocol: tcp
+          ToPort: 9999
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

mikkeloscar:
This is failing because the aws-cloud-controller-manager (the one managing e.g. Service Type Load Balancer) doesn't handle if one or more instances of the cluster [has multiple security groups attached](https://github.com/kubernetes/cloud-provider-aws/blob/688a1dd816ac6ff01bdc1bf350c81f60d92b9099/pkg/providers/v1/aws_loadbalancer.go#L970-L973) which are also tagged with the cluster ID.

In theory we could work around this by not tagging one of the security groups with the cluster ID, but since the cluster ID tag comes from the CloudFormation stack it's quite tricky to exclude one security Group :shrug: Alternatively we could try to fix the controller, but this is run by EKS control plane so depends on AWS/EKS adopting a potential change....

szuecs:
@mikkeloscar maybe start by reporting the issue upstream.

tcondeixa:
not sure if it's relevant but if it's a tag we define ourselves in the CF stack, it's possible to override the tag value redefining the tag at the resource level to a different value.

mikkeloscar:
> not sure if it's relevant but if it's a tag we define ourselves in the CF stack, it's possible to override the tag value redefining the tag at the resource level to a different value.

Problem is that it's the tag `key` that must be different. The key has the format: `kubernetes.io/cluster/<cluster-id>` :shrug: 

With the current adoption of proxy protocol v2 for skipper-ingress (in EKS, but could also be non-EKS) we may be able to simply close this port altogether, and then this issue is not relevant any more.

PR #10609, opened 2026-02-20
Wrap WIZ Namespace behind config items

zaklawrencea:
Only create the WIZ Namespace when WIZ components are enabled. Currently if WIZ is disabled, the Namespace will be created and then subsequently cleaned up due to the deletions.


zaklawrencea:
:+1:

tcondeixa (approved):

zaklawrencea:
:+1:

PR #10614, opened 2026-02-23
Disable Cluster Autoscaler by default

mikkeloscar:
Disable the Cluster Autoscaler by default as there are no clusters using CA based pools anymore.

PR #10619, opened 2026-02-23
Dev to alpha custom

mikkeloscar:
Cherry-pick #10614 to alpha.

It doesn't work with e2e because of a race when deleting the deployments:

```
time="2026-02-23T14:04:14Z" level=fatal msg="Fail to provision: unable to delete: admission webhook \"serviceaccount-admitter.teapot.zalan.do\" denied the request: serviceaccount is still used by the following pods: kube-cluster-autoscaler-d499d586-csc9v"
```

CLM running on real clusters handle this by just rerunning the loop for the cluster, but in e2e it fails once and doesn't retry. Accept this as we only need to merge this through channels once and it's fixed.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -1,5 +1,5 @@
 # Autoscaling settings
-cluster_autoscaler_enabled: "true"
+cluster_autoscaler_enabled: "false"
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

zaklawrencea (approved):

PR #10633, opened 2026-02-24
Update to Kubernetes v1.35

Shinzu:
ref: https://github.com/zalando-build/cloud-platform/issues/7765


cluster/manifests/kube-proxy/configmap.yaml
```diff
@@ -23,7 +23,6 @@ data:
       tcpEstablishedTimeout: 24h0m0s
     enableProfiling: false
     featureGates:
-      SizeMemoryBackedVolumes: {{ .Cluster.ConfigItems.enable_size_memory_backed_volumes }}
 {{- if eq .Cluster.ConfigItems.enable_image_volumes "true" }}
       ImageVolume: true
```
linki: Let's also get rid of this feature gate. It's enabled by default in 1.35 and there's no reason why we shouldn't make it available to users.

https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

Shinzu: wdym?

Shinzu: is removed now

mikkeloscar:
:+1:

tcondeixa (approved):

PR #10635, opened 2026-02-24
Remove feature gate SizeMemoryBackedVolumes

Shinzu:
this is GA since 1.32 and default to true

https://github.com/kubernetes/kubernetes/pull/133720

general pr for v1.35 will remove any other occurrence

removing this here to fix 1.35 pipeline since upgrading the aws master nodes will fail with this still active since the kube-proxy cannot start on the master nodes when upgrading to 1.35

related logs from a current 1.34 cluster

```
W0224 16:11:21.469327       1 feature_gate.go:352] Setting GA feature gate SizeMemoryBackedVolumes=true. It will be removed in a future release.
I0224 16:11:21.469338       1 feature_gate.go:385] feature gates: {map[SizeMemoryBackedVolumes:true]}
```

### DIFF
--- a/cluster/manifests/kube-proxy/configmap.yaml
+++ b/cluster/manifests/kube-proxy/configmap.yaml
@@ -23,7 +23,6 @@ data:
       tcpEstablishedTimeout: 24h0m0s
     enableProfiling: false
     featureGates:
-      SizeMemoryBackedVolumes: {{ .Cluster.ConfigItems.enable_size_memory_backed_volumes }}
 {{- if eq .Cluster.ConfigItems.enable_image_volumes "true" }}
       ImageVolume: true
 {{- end }}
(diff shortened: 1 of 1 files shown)


cluster/manifests/kube-proxy/configmap.yaml
```diff
@@ -23,7 +23,6 @@ data:
       tcpEstablishedTimeout: 24h0m0s
     enableProfiling: false
     featureGates:
-      SizeMemoryBackedVolumes: {{ .Cluster.ConfigItems.enable_size_memory_backed_volumes }}
 {{- if eq .Cluster.ConfigItems.enable_image_volumes "true" }}
```
tcondeixa, line 25: I think it's better we move this if statement to  cover also the `featureGates`, otherwise we might have a feature gate without any nested key, and that might cause an error. This is at least my understanding from the current indentation just visually checking here in github.

linki:
👍

ni9hty:
:+1:

PR #10640, opened 2026-02-27
don't force cluster-lifecycle-controller to run on seed

tcondeixa:
don't force cluster-lifecycle-controller to run on seed, so it can also run in any worker node

This is a follow up on https://github.com/zalando-incubator/kubernetes-on-aws/pull/10117, where this specific component was not adjusted

### DIFF
--- a/cluster/manifests/cluster-lifecycle-controller/deployment.yaml
+++ b/cluster/manifests/cluster-lifecycle-controller/deployment.yaml
@@ -51,6 +51,9 @@ spec:
           value: "{{ .Cluster.Region }}"
 {{- if eq .Cluster.Provider "zalando-eks"}}
       tolerations:
+      - key: dedicated
+        value: cluster-seed
+        effect: NoSchedule
       - key: node.kubernetes.io/not-ready
         operator: Exists
 {{- else}}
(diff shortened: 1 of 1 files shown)


cluster/manifests/cluster-lifecycle-controller/deployment.yaml
```diff
@@ -50,12 +50,7 @@ spec:
         - name: AWS_REGION
           value: "{{ .Cluster.Region }}"
 {{- if eq .Cluster.Provider "zalando-eks"}}
-      nodeSelector:
-        dedicated: cluster-seed
       tolerations:
-      - key: dedicated
```
mikkeloscar, line 56: The toleration is still needed to _allow_ it to run on the seed node. The nodeSelector is what forces it, only that should be dropped.

tcondeixa: thanks, I adjusted

tcondeixa:
👍

tcondeixa:
JFYI @demonCoder95

demonCoder95:
Some unrelated e2es seem to fail, I retriggered the pipeline

tcondeixa:
👍

demonCoder95:
👍

PR #10647, opened 2026-03-02
kube-ingress-aws-controller: Add configurable CNI mode support for EKS clusters

ayruslore:
## Summary

Add configuration flags to support CNI mode targeting in kube-ingress-aws-controller for zalando-eks clusters. This enables flexible IP address type configuration (dualstack, IPv6) and CNI mode targeting through config items, allowing gradual rollout and easy rollback via cluster configuration management.

## Changes

- Add `kube_aws_ingress_controller_lb_ip_addr_type` config for load balancer IP address type
- Add `kube_aws_ingress_controller_target_group_ip_addr_type` config for target group IP address type  
- Add `kube_aws_ingress_controller_enable_cni_mode` config to enable pod-level targeting with AWS CNI
- Add necessary IAM permissions (VPC peering, pod discovery) for EKS clusters
- Add RBAC permissions for pod watching when CNI mode is enabled

## Test plan

- [ ] Verify config items are properly templated
- [ ] Review RBAC and IAM permission additions
- [ ] Test with dualstack LB configuration on zalando-eks cluster
- [ ] Verify graceful rollback via config changes

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -2082,6 +2082,11 @@ Resources:
               - Action: 'ec2:DescribeVpcs'
                 Effect: Allow
                 Resource: '*'
+{{- if eq .Cluster.Provider "zalando-eks" }}
+              - Action: 'ec2:DescribeVpcPeeringConnections'
+                Effect: Allow
+                Resource: '*'
+{{- end }}
               - Action: 'iam:CreateServiceLinkedRole'
                 Effect: Allow
                 Resource: '*'
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -72,7 +72,13 @@ kube_aws_ingress_controller_cert_polling_interval: "2m"
 kube_aws_ingress_default_lb_type: "application"
 # cert filter
 kube_aws_ingress_controller_cert_filter_tag: "kubernetes=enabled"
-
+# ip address type for LB, can be "ipv4", "dualstack"
+kube_aws_ingress_controller_lb_ip_addr_type: "ipv4"
+# ip address type for target group, can be "ipv4" or "ipv6"
+# Note: if LB ip address type is "ipv4", target group ip address type must be "ipv4".
+kube_aws_ingress_controller_target_group_ip_addr_type: "ipv4"
+# when enabled, the ingress controller will use AWS CNI to attach ENIs to the ingress controller pods and use pod IPs as targets in the target group.
+kube_aws_ingress_controller_enable_cni_mode: "false"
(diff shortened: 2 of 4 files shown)


ayruslore:
👍

linki:
How does this affect load balancers in the existing EKS clusters?

ayruslore:
> How does this affect load balancers in the existing EKS clusters?

this will recreate the load balancers. right now EKS clusters use IPv4 lbs.  With this change a new LB of type Dual Stack will be created and will stop using the IPv4 LB.

tcondeixa:
> > How does this affect load balancers in the existing EKS clusters?
> 
> this will recreate the load balancers. right now EKS clusters use IPv4 lbs. With this change a new LB of type Dual Stack will be created and will stop using the IPv4 LB.

Will this not create downtime? or what is the strategy here?
This might have more implications on the overall network setup only affecting some use-cases, so IMO this should be rolled out with a config-item. This allows not only to control the rollout but also quickly rollback (still need to recreate the LB)

ayruslore:
> > > How does this affect load balancers in the existing EKS clusters?
> > 
> > 
> > this will recreate the load balancers. right now EKS clusters use IPv4 lbs. With this change a new LB of type Dual Stack will be created and will stop using the IPv4 LB.
> 
> Will this not create downtime? or what is the strategy here? This might have more implications on the overall network setup only affecting some use-cases, so IMO this should be rolled out with a config-item. This allows not only to control the rollout but also quickly rollback (still need to recreate the LB)

So the downtime problem has been addressed and fixed here https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/753. So the LB recreation has been simplified. We have a delay before switching from old LB to the new LB, this was added as fix in the controller. And the unused or the old LBs set to delete 1hr later. So for the rollback if done before the lb. is deleted it will just use that. otherwise will create a new IPv4 LB again and wait to switch.

The LB recreation process has been tested many times and works correctly now. Also this change has been load tested with 60K rps and made sure there was no request / data drop or loss.

But I do agree with the config item rollout. Will update the PR. It would be easy to revert with it in which ever cluster an issue occurs fails. As there are multiple steps involved.

ayruslore:
@linki @tcondeixa I have updated the PR now to reflect config items which can be used to do a step wise controlled gradual rollout for enabling CNI mode. Please review

tcondeixa:
I don't know much about the kube-ingress-aws-controller itself and the configuration, but the overall plan looks good.

tcondeixa:
👍

szuecs:
:+1:

ayruslore:
👍

PR #10650, opened 2026-03-02
fix OPA e2e test: add retry strategy

szuecs:
fix e2e test: retry strategy because of unrelated component seems to be not ready, yet

### DIFF
--- a/test/e2e/aaas.go
+++ b/test/e2e/aaas.go
@@ -7,14 +7,15 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	"net/http"
+	"time"
+
 	netv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
 	"k8s.io/kubernetes/test/e2e/framework/ingress"
 	admissionapi "k8s.io/pod-security-admission/api"
-	"net/http"
-	"time"
 )
 
 var _ = Describe("Ingress tests for OPA filters", func() {
@@ -59,6 +60,7 @@ var _ = Describe("Ingress tests for OPA filters", func() {
 
 		url := "https://" + hostName + authorizationEnforcedPath
 		req, err := http.NewRequest("GET", url, nil)
+		framework.ExpectNoError(err)
 		resp, err := getAndWaitResponseWithInterval(rt, req, 60*time.Second, 5*time.Second, http.StatusForbidden)
 		framework.ExpectNoError(err)
 		Expect(resp.StatusCode).To(Equal(http.StatusForbidden))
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

test/e2e/aaas.go
```diff
@@ -145,7 +148,20 @@ func updateIngressAndWait(serviceName, hostName, path, ingressRoute string, port
 		},
 		port,
 	)
-	ingressUpdate, err := cs.NetworkingV1().Ingresses(ingressCreate.ObjectMeta.Namespace).Update(context.TODO(), updatedIng, metav1.UpdateOptions{})
+
+	var (
+		err           error
+		ingressUpdate *netv1.Ingress
+	)
+	// skipper validation webhook fails very often with status code 500, maybe it's not ready for some reason
```
mikkeloscar, line 156: It looks like this will just hide the underlying problem. The webhook is configured with a 5s timeout which is what times out and results in 500, at least in the cases I have observed. This would suggest the webhook is often not responding within 5s. The e2e already implicitly retry by running twice so it would mean it fails already twice in those cases.

The webhook is relatively recently introduced in e2e, I wonder if there is some special condition for this filter that makes it slow? E.g. does the opa filter lazy load or so 🤔

szuecs: Yes opa filter is likely concurrently loading at route creation and is not blocking. I think a retry strategy here will help. I am not sure if the e2e retries are too slow and pod will be deleted and moved somewhere else.

szuecs: In general we see that out of 1500 tests basically all have 1 flake during the last weeks.

mikkeloscar: > Yes opa filter is likely concurrently loading at route creation and is not blocking. I think a retry strategy here will help.

Doesn't this show a problem that could also happen in real clusters as well then?

szuecs: It's designed as a fix for slow routing updates caused by opa infrastructure blocking royte updates on filter creation because it needs to load bundles from s3....
So what we have right now is the best we can do right now.
I think the best would be to move the validation webhook into skipper-ingress data plane, which could resolve some issues and maybe create some new ones (?) but right now it's split out as separate deployment.

szuecs: Let's tru to merge it and see if it helps for the current issues. I am all in to fix the whole e2e test issues and invest some time to fix biggest issues we have

szuecs: Here also when it's green:
```
 [FLAKEY TEST - TOOK 2 ATTEMPTS TO PASS] [165.796 seconds]

Ingress tests for OPA filters opaServeResponse filter should return body [Ingress] [Opa]

/workspace/test/e2e/aaas.go:76
```

All the opa tests took 2 attempts, so I think it's a good idea to merge this pr

madumalt (approved):

szuecs:
We fixed the problem and it seems we do not need a retry here anymore

PR #10664, opened 2026-03-04
Enable S3 CSI Driver EKS Addon

linki:
Allow to enable the S3 CSI Driver EKS addon.

Some resources overlap with the self-managed driver setup which can also be enabled (not at the same time though). So deletion will only happen when both config items are turned off.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -330,6 +330,17 @@ Resources:
         - RoleArn: !GetAtt EKSEFSCSIDriverIAMRole.Arn
           ServiceAccount: efs-csi-controller-sa
 {{- end }}
+{{- if eq .Cluster.ConfigItems.eks_s3_csi_support_enabled "true" }}
+  EKSAddonS3CSIDriver:
+    Type: AWS::EKS::Addon
+    Properties:
+      AddonName: aws-mountpoint-s3-csi-driver
+      AddonVersion: "v2.3.0-eksbuild.1"
+      ClusterName: !Ref EKSCluster
+      PodIdentityAssociations:
+        - RoleArn: !GetAtt EKSS3CSIDriverIAMRole.Arn
+          ServiceAccount: s3-csi-driver-sa
+{{- end }}
   EKSAddonAmazonVPCCNI:
     Type: AWS::EKS::Addon
     Properties:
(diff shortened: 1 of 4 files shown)


cluster/cluster.yaml
```diff
@@ -664,6 +675,30 @@ Resources:
...
+                "Action": [
+                    "sts:AssumeRole",
+                    "sts:TagSession"
+                ]
+            }
+        ]
+      }
+      Path: /
+      ManagedPolicyArns:
+      - arn:aws:iam::aws:policy/AmazonS3FullAccess
```
linki, line 701: This comes from AWS' docs. But maybe we need to discuss it.

tcondeixa: probably you can do all sort of operations in the bucket, right? no sure if we can avoid this.

cluster/manifests/deletions.yaml
```diff
@@ -240,7 +240,7 @@ post_apply:
 - name: skipper-canary-conrtroller
   kind: CronJob
   namespace: kube-system
-{{- if ne .Cluster.ConfigItems.s3_csi_driver "true" }}
+{{- if and (ne .Cluster.ConfigItems.s3_csi_driver "true") (ne .Cluster.ConfigItems.eks_s3_csi_support_enabled "true") }}
```
tcondeixa, line 243: I also did a suggestion in the chat, but we can just combine the existing config-item with the provider value. This should allow us to keep using only one config-item without the need to think each one to use.

cluster/cluster.yaml
```diff
@@ -329,6 +329,17 @@ Resources:
...
+{{- if eq .Cluster.ConfigItems.eks_s3_csi_support_enabled "true" }}
+  EKSAddonS3CSIDriver:
+    Type: AWS::EKS::Addon
+    Properties:
+      AddonName: aws-mountpoint-s3-csi-driver
+      AddonVersion: "v2.3.0-eksbuild.1"
+      ClusterName: !Ref EKSCluster
+      PodIdentityAssociations:
+        - RoleArn: !GetAtt EKSS3CSIDriverIAMRole.Arn
+          ServiceAccount: s3-csi-driver-sa
```
tcondeixa, line 342: is this service account created automatically by AWS? I don't see it defined anywhere.

linki: Yes, it is! And also cleaned up when disabled.

tcondeixa (approved):

cluster/cluster.yaml
linki: In all buckets I guess.

@mikkeloscar had some thoughts about this I think.

cluster/manifests/deletions.yaml
linki: For reference: We currently have two S3 CSI implementations in EKS (v1 and v2 via addon), hence two config items. Once we migrate the v1 cases to v2 we can combine the config item into one.

linki:
👍

tcondeixa (approved):

PR #10669, opened 2026-03-05
Change the Wiz sensor component to be enabled by default

zaklawrencea:
Now that the WIZ Sensor has been enabled in all clusters, update the default value to be `true`.


hagenmathis:
:+1:

zaklawrencea:
:+1:

PR #10674, opened 2026-03-05
Expose cluster's channel in deployment-config

linki:
Expose cluster's channel as field in deployment-config so it can be queried via kubectl et al.


linki:
👍

zaklawrencea:
:+1:

PR #10686, opened 2026-03-06
don't force skipper on spot for test clusters

tcondeixa:
This PR remove the nodeSelector that forces skipper to use `spot`.
The issue is we are forcing Karpenter to use `spot` instances even when it's not the best option.
This sometimes results in higher costs when choosing bigger instances that stay for longer times.

More details in https://github.com/zalando-build/cloud-platform/issues/7799


szuecs:
:+1:

ayruslore:
👍

tcondeixa:
👍

PR #10689, opened 2026-03-06
Update memory limit from 0.3Gi to 300Mi for sandbox-controller

linki:
Using this format leads to warnings in the logs. Let's use the format that we use in other files as well.

image


szuecs:
:+1:

mikkeloscar:
:+1:

PR #10705, opened 2026-03-09
deployment-service-controller: migrate to deployment with VPA

tcondeixa:
Migrate the `deployment-service-controller` to deployment with a single replica with a VPA.

- This will help to reduce costs since most clusters don't really need a static 1Gi for memory.
- The change to `deployment` instead of `statefulset` is more aligned with the purpose of the resources in kubernetes, and there were a few issue in using statefulsets with VPA in the past we want to avoid.
- Remove the config items that does not have any effect anymore with VPA.

The statefulset will be deleted before applying (pre-apply) the new resource to ensure we don't have concurrency and be on the safe side, this might create a small downtime but the deployment should pick the create task after created

### DIFF
--- a/cluster/manifests/deployment-service/controller-vpa.yaml
+++ b/cluster/manifests/deployment-service/controller-vpa.yaml
@@ -16,7 +16,5 @@ spec:
   resourcePolicy:
     containerPolicies:
     - containerName: "deployment-service-controller"
-      maxAllowed:
-        memory: 2Gi
       minAllowed:
         memory: "{{.Cluster.ConfigItems.deployment_service_controller_memory_min}}"
--- a/cluster/manifests/deployment-service/controller-vpa.yaml
+++ b/cluster/manifests/deployment-service/controller-vpa.yaml
@@ -1,0 +1,20 @@
+apiVersion: autoscaling.k8s.io/v1
+kind: VerticalPodAutoscaler
+metadata:
+  name: "deployment-service-controller"
+  namespace: "kube-system"
+  labels:
+    application: "deployment-service"
+    component: "controller"
+spec:
+  targetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: "deployment-service-controller"
+  updatePolicy:
+    updateMode: Recreate
+  resourcePolicy:
+    containerPolicies:
+    - containerName: "deployment-service-controller"
+      maxAllowed:
+        memory: 2Gi
--- a/cluster/manifests/deployment-service/controller-vpa.yaml
+++ b/cluster/manifests/deployment-service/controller-vpa.yaml
@@ -18,3 +18,5 @@ spec:
     - containerName: "deployment-service-controller"
       maxAllowed:
         memory: 2Gi
+      minAllowed:
+        memory: "{{.Cluster.ConfigItems.deployment_service_controller_memory_min}}"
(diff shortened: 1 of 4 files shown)


cluster/manifests/deployment-service/controller-vpa.yaml
```diff
@@ -0,0 +1,20 @@
...
+    apiVersion: apps/v1
+    kind: Deployment
+    name: "deployment-service-controller"
+  updatePolicy:
+    updateMode: Recreate
+  resourcePolicy:
+    containerPolicies:
+    - containerName: "deployment-service-controller"
+      maxAllowed:
+        memory: 2Gi
```
mikkeloscar: It would be good to use `.Cluster.ConfigItems.deployment_service_controller_memory` here so we have the option to manually scale up in case the VPA doesn't react fast enough, just as a fail-safe. See how it's done for Prometheus as an example.

tcondeixa: do you mean on the maxAllowed or to introduce a minAllowed that we can increase in case it's needed?

tcondeixa: is it good practice to keep the maxAllowed? or should I just remove it and let the VPA to adjust to what is needed?

tcondeixa: @mikkeloscar added the minAllowed, but not 100% sure that was your suggestion.

linki:
I suggest to retrigger the latest master pipelines for deployment-service once this is rolled out to `stups`/`stups-test` accounts respectively. It tests the deployment process quite well.

https://sunrise.zalando.net/cdp/gh/zalando-build/deployment-service

cluster/manifests/deployment-service/controller-vpa.yaml
linki: Mikkel just wants to get rid of a hard-coded fixed cap in case it's not enough. Prometheus _only_ uses `minAllowed` so we can also go with that.

linki: https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/external-dns/vpa.yaml#L19-L20

ExternalDNS uses _only_ `maxAllowed` but it's configurable without code changes. This is also an option.

tcondeixa: I'm removing the maxAllowed to avoid we end up reaching it (although it was hardcoded for a long time) since it would case issues for users to deploy.
I'm leaving the minAllowed if we need to force it at some point to quickly increase memory.

tcondeixa:
> I suggest to retrigger the latest master pipelines for deployment-service once this is rolled out to `stups`/`stups-test` accounts respectively. It tests the deployment process quite well.
> 
> https://sunrise.zalando.net/cdp/gh/zalando-build/deployment-service

thanks, I will do it. I was already considering to create a deployment myself, but it's good we already have it.

tcondeixa:
@linki @mikkeloscar can you please check again. thanks

cluster/manifests/deployment-service/controller-vpa.yaml
mikkeloscar: I thought I commented on minAllowed, and that is what I was thinking about to have a minAllowed we can manually bump should it be needed.

I see you changed it like this now so lgtm.

mikkeloscar:
:+1:

tcondeixa:
👍

PR #10707, opened 2026-03-09
Add RBAC config for Scaling Console

Jan-M:
Add RBAC config for Scaling Console

Grant Read-Only access to the new console.

### DIFF
--- a/cluster/manifests/roles/build-scaling-console.yaml
+++ b/cluster/manifests/roles/build-scaling-console.yaml
@@ -1,0 +1,75 @@
+apiVersion: "v1"
+kind: "ServiceAccount"
+metadata:
+  name: "build-scaling-console"
+  namespace: "default"
+  labels:
+    application: "build-scaling-console"
+    team: builder-infrastructure
+---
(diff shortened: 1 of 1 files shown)


cluster/manifests/roles/build-scaling-console.yaml
```diff
@@ -0,0 +1,75 @@
+apiVersion: "v1"
+kind: "ServiceAccount"
+metadata:
+  name: "build-scaling-console"
+  namespace: "default"
+  labels:
+    application: "build-scaling-console"
+    team: builder-infrastructure
+---
```
zaklawrencea: With the app not actually being deployed to each cluster, you only need the ClusterRole/Binding deployed to all destination clusters. The SA should only exist in the source cluster where your app will run.

Jan-M: Fixed, agree.

zaklawrencea (approved):

mikkeloscar:
:+1:

PR #10710, opened 2026-03-09
Disable amd instance types in production by default

mikkeloscar:
### Description

We want to remove the use of `default-karpenter` pool to fix scheduling annotation: `zalando.org/compute-instance-architecture` with the values `prioritize-arm64` and `any` as they currently don't work when the `default-karpenter` pool with only x86 instances have higher priority (see internal ticket for more context).

Doing this has the side-effect of using the `catch-all` pools which allow any instance types including amd based instances (e.g. c5a). We generally don't use AMD instance types as they historically have a worse cost/performance ratio compared to intel, based on user feedback, therefore we didn't include them in the `default-karpenter` pool, but they are currently optional in the catch-all pools.

While we could/should re-evaluate cost/performance of amd instance types for newer generation, for now simply disable them to avoid a potential performance degradation across the fleet while we work on the initial goal to remove `default-karpenter` pool and fix `prioritize-arm64` and `any` scheduling annotations.

Currently only a subset of clusters use amd instances because they use scheduling annotations for a subset of pods.

This adds a requirement to node pools to exlcude amd instance types.

The change _only_ applies to _production_ clusters as for test we don't care as much about performance but just want the cheapest setup as there is generally very little load in test for it to matter.

Additionally do the same for `flex` instance types with the same reason as for AMD. Flex types provide baseline CPU performance with burst capability so it's unpredictable how they perform. We want to avoid those by default in production.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -69,6 +69,18 @@ karpenter_amd_cpu_types_enabled: "false"
 karpenter_amd_cpu_types_enabled: "true"
 {{end}}
 
+# configure whether we allow flex instance types for Karpenter nodes
+# flex instance types have a baseline CPU performance with burst capability.
+# This provides an unpredictable performance, hence we want to avoid running
+# those by default in production.
+{{if eq .Cluster.Environment "production"}}
+karpenter_flex_types_enabled: "false"
+{{else}}
+karpenter_flex_types_enabled: "true"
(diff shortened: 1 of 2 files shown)


tcondeixa:
👍

mikkeloscar:
Closing in favor of https://github.com/zalando-incubator/kubernetes-on-aws/pull/10730

PR #10712, opened 2026-03-09
Add missing deletions for legacy S3 CSI Driver setup

linki:
These resources are deployed but not added to deletions.yaml.

If they aren't cleaned up, the addon-based v2 implementation cannot be turned on.


linki:
👍

tcondeixa (approved):

tcondeixa (approved):

PR #10722, opened 2026-03-10
Update to AMI with increased nofile limit

mikkeloscar:
This updates the AMI to a version with an increased nofile limit:

```
cat /etc/systemd/system.conf.d/02-nofile-limit.conf‎
[Manager]
DefaultLimitNOFILE=1048576
```

This limit increase is done to ensure both amd64 and arm64 instances have the same limit set. Before, the arm64 instances had a much lower limit leading to issues for applications that don't automatically increase the soft limit on startup and need many file descriptors because of high load e.g. envoy.

This is targeting the Kubernetes v1.35 branch to piggy-back on the AMI roll out for that.


mikkeloscar:
:+1:

tcondeixa:
👍

PR #10723, opened 2026-03-10
Adjust skipper resources to reduce costs

tcondeixa:
Reduce skipper costs with the changes on test clusters that have small load.
- add smaller instances since load is small in test clusters
- reduce resources (memory and cpu) to half in test clusters since most of the time they are doing nothing.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -227,11 +227,7 @@ skipper_redis_max_conns: 100
 
 skipper_ingress_redis_swarm_enabled: "true"
 skipper_ingress_redis_swim_enabled: "false"
-{{if eq .Cluster.Environment "production"}}
 skipper_ingress_redis_target_average_utilization_cpu: "30"
-{{else}}
-skipper_ingress_redis_target_average_utilization_cpu: "60"
-{{end}}
 skipper_ingress_redis_target_average_utilization_memory: "60"
 skipper_ingress_redis_min_replicas: "1"
 skipper_ingress_redis_max_replicas: "100"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -104,11 +104,10 @@ skipper_ingress_max_replicas: "50"
 {{end}}
 {{if eq .Cluster.Environment "production"}}
 skipper_ingress_cpu: "1000m"
-skipper_ingress_memory: "1500Mi"
 {{else}}
 skipper_ingress_cpu: "500m"
-skipper_ingress_memory: "750Mi"
 {{end}}
+skipper_ingress_memory: "1500Mi"
 
 skipper_topology_spread_enabled: "true"
 skipper_topology_spread_timeout: "7m"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -222,7 +222,11 @@ skipper_redis_max_conns: 100
 
 skipper_ingress_redis_swarm_enabled: "true"
 skipper_ingress_redis_swim_enabled: "false"
+{{if eq .Cluster.Environment "production"}}
 skipper_ingress_redis_target_average_utilization_cpu: "30"
+{{else}}
+skipper_ingress_redis_target_average_utilization_cpu: "60"
+{{end}}
 skipper_ingress_redis_target_average_utilization_memory: "60"
 skipper_ingress_redis_min_replicas: "1"
 skipper_ingress_redis_max_replicas: "100"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -102,8 +102,13 @@ skipper_ingress_max_replicas: "300"
 skipper_ingress_min_replicas: "2"
 skipper_ingress_max_replicas: "50"
 {{end}}
+{{if eq .Cluster.Environment "production"}}
 skipper_ingress_cpu: "1000m"
 skipper_ingress_memory: "1500Mi"
+{{else}}
+skipper_ingress_cpu: "500m"
+skipper_ingress_memory: "750Mi"
+{{end}}
 
 skipper_topology_spread_enabled: "true"
 skipper_topology_spread_timeout: "7m"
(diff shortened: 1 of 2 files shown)


ayruslore (approved):

cluster/manifests/skipper/deployment.yaml
```diff
@@ -114,24 +114,36 @@ spec:
               - key: node.kubernetes.io/instance-type
                 operator: In
                 values:
+                {{- if eq .Cluster.Environment "test" }}
+                - c7g.medium
+                {{- end}
```
ayruslore:
```suggestion
                {{- end}}
```

cluster/manifests/skipper/deployment.yaml
```diff
@@ -114,24 +114,36 @@ spec:
...
                 - c7g.large
                 - c7g.xlarge
                 - c7g.2xlarge
                 - c7g.4xlarge
                 - c7g.8xlarge
                 - m7g.large
+                {{- if eq .Cluster.Environment "test" }}
+                - m7g.medium
+                {{- end}
```
ayruslore:
```suggestion
                {{- end}}
```

cluster/manifests/skipper/deployment.yaml
```diff
@@ -114,24 +114,36 @@ spec:
...
           - weight: 50
             preference:
               matchExpressions:
               - key: node.kubernetes.io/instance-type
                 operator: In
                 values:
+                {{- if eq .Cluster.Environment "test" }}
+                - c6g.medium
+                {{- end}
```
ayruslore:
```suggestion
                {{- end}}
```

cluster/manifests/skipper/deployment.yaml
```diff
@@ -114,24 +114,36 @@ spec:
...
                 - c6g.large
                 - c6g.xlarge
                 - c6g.2xlarge
                 - c6g.4xlarge
                 - c6g.8xlarge
                 - m6g.large
+                {{- if eq .Cluster.Environment "test" }}
+                - m6g.medium
+                {{- end}
```
ayruslore:
```suggestion
                {{- end}}
```

cluster/manifests/skipper/deployment.yaml
```diff
@@ -114,24 +114,36 @@ spec:
               - key: node.kubernetes.io/instance-type
                 operator: In
                 values:
+                {{- if eq .Cluster.Environment "test" }}
+                - c7g.medium
```
mikkeloscar, line 117: Can it actually fit with daemonset overhead and default CPU requests for skipper-ingress? I think this was the blocker in the past from using those

mikkeloscar: The change is fine, it could just be that it doesn't materialize if it doesn't fit.

tcondeixa: It's a good point, I will check. I noticed the nodes have a small load but we might need to adjust the requested memory and cpu from skipper to fit it in medium nodes.

tcondeixa: I'm reducing the cpu and memory to half in test cluster to keep the ration. It should horizontal scale if it's temporarily needed, but most of the time it's almost doing nothing.

szuecs:
:-1:

cluster/config-defaults.yaml
```diff
@@ -102,8 +102,13 @@ skipper_ingress_max_replicas: "300"
 skipper_ingress_min_replicas: "2"
 skipper_ingress_max_replicas: "50"
 {{end}}
+{{if eq .Cluster.Environment "production"}}
 skipper_ingress_cpu: "1000m"
 skipper_ingress_memory: "1500Mi"
+{{else}}
+skipper_ingress_cpu: "500m"
```
tcondeixa, line 108: @ayruslore do we see any issue in reducing the resources to half since they are not doing anything most of the time? We still have HPA to horizontal scale in case of temporary load.

szuecs: We can not set it to the half here, but we can check which clusters could have a change to half it.

tcondeixa: is the issue with reducing the cpu, the memory or both?

cluster/manifests/skipper/deployment.yaml
szuecs: We do no really to guard the test/prod, or do we need this?

cluster/manifests/skipper/deployment.yaml
```diff
@@ -114,24 +114,36 @@ spec:
...
+                {{- end }}
                 - c7g.large
                 - c7g.xlarge
                 - c7g.2xlarge
                 - c7g.4xlarge
                 - c7g.8xlarge
                 - m7g.large
+                {{- if eq .Cluster.Environment "test" }}
```
szuecs: I would drop the guard of test only, same for the changes below on 6th gen

cluster/config-defaults.yaml
szuecs: both

szuecs: At least in some clusters it will lead to crashing instances

cluster/manifests/skipper/deployment.yaml
tcondeixa: true, if it does not fit the node Karpenter will not choose it anyway

cluster/config-defaults.yaml
tcondeixa: I reverted the memory and only adjust cpu, there is enough memory available in medium instances

tcondeixa:
there are concerns to reduce the cpu for skipper in test clusters, so I will close the PR. There is not point on trying to use medium instance if the requested cpu does not fit. The other option to reduce resources would be to not have dedicated node for skipper, we we would avoid the extra system overhead in the 2 extra nodes. Most of the time the test cluster are not using the resource but this need to be tackle later since this seems not to be a low hanging fruit.

PR #10725, opened 2026-03-10
Disable flex instance types in production

mikkeloscar:
Flex instances provide burstable CPU performance leading to unpredictable performance.

To make the roll out of catch-all node pools more safe we disable flex instance types from production by default to avoid getting those instance types as they are cheaper than non-flex counterpart.


tcondeixa (approved):

mikkeloscar:
:+1:

PR #10728, opened 2026-03-10
remove AWS permissions from S3 CSI driver in favour of pod authentication source

linki:
This should work when using `authenticationSource: pod` on the `PersistentVolume`.


tcondeixa (approved):

linki:
👍

PR #10730, opened 2026-03-10
Exclude 5th gen AMD instances in prod

mikkeloscar:
Similar to #10725 exclude 5th gen AMD instances in production as we historically had reports of low performance which we want to introduce when disabling the more restricted `default-karpenter` pool.

#10710 aimed to disable _all_ amd instances but this doesn't work for several reasons:

* Some users explicitly request AMD instance types of newer generations
* GPU types like g6 which are regularly used are amd instances, those we definitely want to allow.


tcondeixa:
👍

mikkeloscar:
:+1:

PR #10731, opened 2026-03-10
add vpa for cluster lifecycle controller

tcondeixa:
add vpa for cluster lifecycle controller

Most clusters don't need more than 10/20 Mi of memory, so this should save at least 50Gi across all clusters.


tcondeixa:
👍

mikkeloscar:
:+1:

PR #10685, opened 2026-03-06
Exclude S3 CSI Driver Addon components from being processed by admission-controller

linki:
It's a precondition for https://github.com/zalando-incubator/kubernetes-on-aws/pull/10664 to pass its e2e tests.

PR #10734, opened 2026-03-10
[STABLE: cherry-pick] exclude S3 CSI Driver Addon components from being processed by admission-controller

mikkeloscar:
This cherry-picks #10685 to stable such that it's not part of #10670

This is to make sure e2e can pass for #10670 which otherwise fails because #10685 and #10664 are combined there which was not intended.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #10741, opened 2026-03-11
add vpa for static-egress-controller

tcondeixa:
add vpa for static-egress-controller

### DIFF
--- a/cluster/manifests/kube-static-egress-controller/vpa.yaml
+++ b/cluster/manifests/kube-static-egress-controller/vpa.yaml
@@ -1,0 +1,20 @@
+apiVersion: autoscaling.k8s.io/v1
+kind: VerticalPodAutoscaler
+metadata:
+  name: "kube-static-egress-controller"
+  namespace: "kube-system"
+  labels:
+    application: kubernetes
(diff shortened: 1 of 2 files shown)


tcondeixa:
👍

tcondeixa:
@linki whenever you have time please check this one, thanks
Do you think we should only deploy the VPA if the feature is enabled?

cluster/manifests/kube-static-egress-controller/vpa.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: autoscaling.k8s.io/v1
```
tcondeixa, line 1: I have doubts if we should only deploy the VPA in case the feature is enabled.

linki: Let's keep it like this. The way we disable egress-controller is unusual and doing something else for VPA just gets confusing.

The Deployment exists in both cases so if it doesn't make any problems then I'm fine having the VPA as well.

linki:
👍

tcondeixa:
👍

PR #10747, opened 2026-03-11
add vpa for fabric-gateways-controller

tcondeixa:
add vpa for fabric-gateways-controller.
This allow both to reduce cost adjusting resources, but also scale up automatically in case it's needed.
- Removed config items with memory and cpu only used now on the first deployment
- Added config-item for minAllowed memory in case we need to manually increase at some point

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -482,8 +482,7 @@ skipper_open_policy_agent_jwt_cache_max_num_entries: "1000"
 #   - true: runs the controller
 #
 fabric_gateway_controller_enabled: "true"
-fabric_gateway_controller_cpu: "50m"
-fabric_gateway_controller_memory: "150Mi"
+fabric_gateway_controller_memory_min: "10Mi"
 fabric_gateway_controller_allow_all_filters: "false"
 fabric_gateway_controller_ssl_policy: ""
 fabric_gateway_controller_log_level: "INFO"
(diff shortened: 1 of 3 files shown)


cluster/manifests/fabric-gateway/vpa.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: autoscaling.k8s.io/v1
```
tcondeixa, line 1: not sure if we should only deploy the VPA in case the feature is enabled, but since the deployment is still there with 0 replicas it should be fine.

linki:
👍

tcondeixa:
👍

tcondeixa:
@szuecs or @ayruslore please check in case you have any concerns.

cluster/config-defaults.yaml
```diff
@@ -482,8 +482,7 @@ skipper_open_policy_agent_jwt_cache_max_num_entries: "1000"
 #   - true: runs the controller
 #
 fabric_gateway_controller_enabled: "true"
-fabric_gateway_controller_cpu: "50m"
-fabric_gateway_controller_memory: "150Mi"
+fabric_gateway_controller_memory_min: "10Mi"
```
szuecs, line 485: where does this number come from?

szuecs: It's lower than the min of all clusters max usage, so we will see a lot of them crashing.

tcondeixa: in fact we can increase it to 50Mi since it's the minimum memory defined in the VPA.
The main idea is just to have a config-item that we can easily increase in case of issues, the VPA should choose the value per cluster.

tcondeixa: we can have a config-item for 1 week while the VPA collect metrics to reduce risk, but if we have a min of 50Mi it should not create any issue.

tcondeixa:
the owning team preferred to manually adjust the resources for this use-case

PR #10748, opened 2026-03-11
add vpa for efs provisioner

tcondeixa:
add vpa for efs provisioner
This not only helps to reduce a bit the costs but it also allow to automatically scale up if needed.
Since we don't define the requests (only limits) some internal tools have some issues, so I'm defining the request equal to the limits.


linki:
👍

tcondeixa:
👍

PR #10751, opened 2026-03-11
Fix: Add proper target-access-mode configuration for CNI and non-CNI modes

ayruslore:
## Summary
This fix simplifies the target-access-mode configuration for EKS clusters by always using the CNI configuration logic:

- When CNI mode enabled: Use AWSCNI with pod targeting
- When CNI mode disabled: Use HostPort mode with instance targeting

This addresses the issue where non-CNI mode on EKS clusters was not properly configuring the target-access-mode, causing ingress controller to fail to connect to targets.

## Changes
- Removed provider-specific logic (zalando-aws vs zalando-eks check)
- Simplified to always apply CNI configuration logic
- Always set --target-access-mode based on CNI enablement

## Test plan
- [x] Verify changes work for EKS clusters with CNI enabled
- [x] Verify changes work for EKS clusters with CNI disabled  
- [x] Confirm ingress controller can reach targets correctly
- [x] Validate simplified approach is cleaner and maintainable

## Related
- Fixes IPv6 target groups issue with missing client IP preservation
- Relates to PR #10647

### DIFF
--- a/cluster/manifests/ingress-controller/deployment.yaml
+++ b/cluster/manifests/ingress-controller/deployment.yaml
@@ -37,15 +37,14 @@ spec:
         - name: controller
           image: "{{ $image }}"
           args:
-            - --target-access-mode=HostPort
-{{- if eq .Cluster.Provider "zalando-eks" }}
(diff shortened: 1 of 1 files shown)


ayruslore:
👍

MustafaSaber:
👍

PR #10756, opened 2026-03-12
Role policy to give bucket access

wisinghe:
Opa decision logs from the skipper ingress can be directed to a given S3 bucket by default.

This PR adds the role policies so that skipper-ingress and skipper-validation-webhook will have necessary write access to the bucket


MustafaSaber:
Was this tested on test cluster?

wisinghe:
> Was this tested on test cluster?

Was tested in corporate-iam-test cluster

MustafaSaber (approved):

wisinghe:
👍

PR #10759, opened 2026-03-12
zalando-iam-aws-proxy: adjust ration memory to cpu

tcondeixa:
adjust ration memory to cpu from current metrics
After checking the previous 7 days from several clusters (including the ones with more load) it's clear the HPA is always scaling based on CPU and we are waisting most of the memory per pod.
The initial idea was to introduce VPA with 2 replicas but the gains should be similar with less risk and disruption.
Some clusters frequently scale up and down during the days, so probably HPA is better for this use-case

Also added the resource requests equal to limits to be more explicits and some internal tools might assume there is not requests (kubernetes should use the limits for requests when not defined)


tcondeixa:
👍

linki:
👍

PR #10763, opened 2026-03-12
kube-node-ready: 1 replica with vpa and rollingUpdate

tcondeixa:
kube-node-ready: 1 replica with vpa and rollingUpdate

NOTE: we could replace the daemonset by a deployment in the legacy cluster to simplify, but it would not bring any cost savings. We can always do it in a separated PR later if we prefer.


tcondeixa:
👍

mikkeloscar:
:+1:

PR #10767, opened 2026-03-13
adjust min memory for event-logs the same as minimum VPA memory

tcondeixa:
adjust min memory for event-logs the same as minimum VPA memory.
Even 50Mi is more than enough for all cluster, and at least we keep the minimum possible value that is enforced by VPA.


tcondeixa:
👍

zaklawrencea (approved):

PR #10769, opened 2026-03-13
Exclude 6a family intance types

mikkeloscar:
Similar to #10730 also include standard 6a instance family types

We do this to reduce the risk of introducing "worse performing" instances

There are a small subset of workloads requesting these types explicitly, will filter those clusters out of this change for now and work with the teams to update.


tcondeixa:
👍

mikkeloscar:
:+1:

PR #10784, opened 2026-03-16
feature: allow parca-agent to run on different node-pools

szuecs:
feature: allow parca-agent to run on different node-pools


szuecs:
@mikkeloscar will propose some other way

PR #10793, opened 2026-03-17
Upgrade EKS Node Monitoring agent version to v1.5.2

zaklawrencea:
Bump the EKS Node Monitoring Agent version to `v1.5.2-eksbuild.1`.

According to the docs, Addons can only be updated by a single minor version at a time:

> You can only update one minor version at a time. For example, if your current version is 1.28.x-eksbuild.y and you want to update to 1.30.x-eksbuild.y, then you must update your current version to 1.29.x-eksbuild.y and then update it again to 1.30.x-eksbuild.y.

The update to `v1.6.0-eksbuild.1`, which allows setting Pod Labels will follow afterwards.

PR #10794, opened 2026-03-17
Upgrade EKS Node Monitoring agent version to v1.6.1

zaklawrencea:
Relies on: https://github.com/zalando-incubator/kubernetes-on-aws/pull/10793.

Bump the EKS Node Monitoring Agent version to `v1.6.1-eksbuild.1`.

This version allows setting Pod labels on the EKS Node Monitoring Agent plugin.


linki:
👍

zaklawrencea:
:+1:

mikkeloscar:
`do-not-merge` :)

#10793 is only rolled to alpha right now so this one shouldn't catch up or the split roll out is lost.

zaklawrencea:
@mikkeloscar I checked and merged the previous one to alpha before this and thought that's good enough for it not to catch up if we stay on top of the rollouts. Could always be more safe though. Also forgot to remove the label :facepalm:

PR #10795, opened 2026-03-17
Re-order skipper validation webhook manifests

zaklawrencea:
e2es currently fail on EKS cluster creation as the VPA CRD (folder `02-vertical-pod-autoscaler`) does not yet exist when the VPA skipper-validation-webhook manifest is applied. This change moves the webhook folder from `02-` to `03-` to ensure they are applied in the correct order.


tcondeixa:
👍

zaklawrencea:
:+1:

PR #10820, opened 2026-03-18
Optionally Give poweruser role access to modify endpointslices

mikkeloscar:
Add the possibility to manage endpointslices similar to how it's possible with endpoints.

This is requested by one team with a special use case where they manage endpoints directly and they want to migrate to endpointslices.

Keep it behind a flag, disabled by default, to gauge the need for this in general as it can be quite powerful.


linki:
👍

mikkeloscar:
:+1:

PR #10822, opened 2026-03-18
Allow polarsignals to run on more dedicated workloads

mikkeloscar:
This is a more complex version of #10784

It introduces the config-item: `polarsignals_workloads` which is a comma-separated list of "workload" names mapping to dedicated node names.

For each workload defined in the list it will create a daemonset with the right nodeSelector and tolerations for that to only run next to the workloads with that same dedicated node name.

* [ ] This does NOT handle cleanup of daemonsets if the workload value is removed from the list, right now a manual delete of the daemonset in the cluster would be needed.


szuecs:
:+1:

ponimas:
👍

PR #10824, opened 2026-03-18
skipper-canary: enable zone aware routing

greeshma1196:
Enable zone aware routing only for canary if flag skipper_ingress_zone_aware_routing_enabled is set to true. Currently it is disabled in Tier 1 & Tier 2 clusters.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -162,6 +162,7 @@ skipper_edit_route_placeholders: ""
 skipper_ingress_inline_routes: ""
 skipper_ingress_refuse_payload: ""
 skipper_endpointslices_enabled: "true"
+skipper_ingress_zone_aware_routing_enabled: "true"
 skipper_externalname_service_enabled: "true"
 
 skipper_kubernetes_annotation_predicates: ''
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -9,7 +9,10 @@
 {{ end }}
 
 {{/* Optional canary arguments separated by "[cf724afc]" to allow whitespaces, e.g. "-foo=has a whitespace[cf724afc]-baz=qux" */}}
-{{ $canary_args := "-routes-urls=http://skipper-ingress-routesrv.kube-system.svc.cluster.local/routes/$(KUBE_NODE_ZONE)" }}
+{{ $canary_args := "" }}
+{{ if eq .Cluster.ConfigItems.skipper_ingress_zone_aware_routing_enabled "true" }}
+{{ $canary_args = "-routes-urls=http://skipper-ingress-routesrv.kube-system.svc.cluster.local/routes/$(KUBE_NODE_ZONE)" }}
+{{ end }}
 
 {{ $cluster_internal_cidrs := list "10.2.0.0/15" .Values.vpc_ipv4_cidr }}
 {{ if eq .Cluster.Provider "zalando-eks" }}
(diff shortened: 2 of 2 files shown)


cluster/manifests/skipper/deployment.yaml
```diff
@@ -615,7 +621,7 @@ spec:
 {{- end }}
       containers:
       - name: routesrv
-        image: container-registry.zalando.net/teapot/skipper:{{ $main_version }}
+        image: container-registry-test.zalando.net/teapot/skipper-test:pr-3117-24
```
szuecs: this is wrong

cluster/config-defaults.yaml
```diff
@@ -291,6 +292,8 @@ skipper_routesrv_target_average_utilization_memory: "80"
 skipper_ingress_routesrv_scaling_schedules: ""
 skipper_routesrv_log_level: "INFO"
 
+
+
```
szuecs: cleanup

cluster/config-defaults.yaml
```diff
@@ -162,6 +162,7 @@ skipper_edit_route_placeholders: ""
 skipper_ingress_inline_routes: ""
 skipper_ingress_refuse_payload: ""
 skipper_endpointslices_enabled: "true"
+skipper_ingress_zone_aware_routing_enabled: "true"
```
szuecs, line 165: we have disabled in tier-1 and tier-2 clusters via ccm

szuecs:
:+1:

cluster/config-defaults.yaml
greeshma1196: https://github.com/zalando-build/teapot-cluster-config-manager/pull/1459

szuecs:
:+1:

greeshma1196:
👍🏽

cluster/manifests/skipper/deployment.yaml
```diff
@@ -212,7 +219,9 @@ spec:
           - "-validate-query={{ .Cluster.ConfigItems.skipper_validate_query }}"
           - "-validate-query-log={{ .Cluster.ConfigItems.skipper_validate_query_log }}"
 {{ if eq .Cluster.ConfigItems.skipper_routesrv_enabled "exec" }}
+{{ if ne "{{ .name }}" "skipper-ingress-canary" }}
```
MustafaSaber, line 222: I think there is no need to have this condition, if you have

```
-routes-urls=URL1
-routes-urls=URL2
```

You can test with

```bash
# In terminal 1
~ ./bin/skipper -inline-routes='r1: Path("/routes1") -> inlineContent("* -> status(201) -> <shunt>") -> <shunt>; r2: Path("routes2") -> inlineContent("* -> status(202) -> <shunt>") -> <shunt>;'

# In terminal 2
~ ./bin/skipper -address=:9091 -routes-urls=http://127.0.0.1:9090/routes1 -routes-urls=http://127.0.0.1:9090/routes2
```

greeshma1196: I will need to test this, but in this case how does it know which routes-url to use? If this condition is not there:
```
-routes-url=ALL_ENDPOINTS_URL
-routes-url=ZONE_AWARE_ENDPOINTS_URL
```

MustafaSaber: last flag value is used, and we always add canary flags at the end, in this specific case flag values overrides (since this is not a multivalue flag)

greeshma1196: Ohh nice, I did not know this. This PR is merged now, should I create a follow up to fix it or we can leave this as is, at any point we need to fix this deployment I will make sure to correct this.

MustafaSaber: It's okay, we can follow up in another PR

szuecs: We would have duplicate routes and one set of routes with zone aware and one set without zone aware, so yes we need this condition :)

szuecs: Oh no it's not a good idea to rely on such behavior which we could also say is a bug

PR #10825, opened 2026-03-18
patched karpenter 1.9.0 with option to select min for spot to spot consolidation

tcondeixa:
patched karpenter 1.9.0 with option to select min for spot to spot consolidation

To be on the safe side I decided to run the new patched image only for non production environment for 1/2 weeks, just in case we have any issue in this critical component.
I'm already updating the CRDs since it's an optional field that we set the value under a config item, so we only create it for node pools in the test environment

### DIFF
--- a/cluster/node-pools/worker-karpenter/provisioners.yaml
+++ b/cluster/node-pools/worker-karpenter/provisioners.yaml
@@ -322,7 +322,7 @@ spec:
 #{{  end }}
     # Minimum number of instance types for spot-to-spot consolidation
 #{{ if index .NodePool.ConfigItems "minInstanceTypesForSpotToSpotConsolidation"}}
-    minInstanceTypesForSpotToSpotConsolidation: {{.NodePool.ConfigItems.minInstanceTypesForSpotToSpotConsolidation}}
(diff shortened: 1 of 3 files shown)


tcondeixa:
@mikkeloscar wdyt about the different options to rollout the new version?

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -319,6 +319,10 @@ spec:
 #{{  else }}
     consolidationPolicy: WhenEmptyOrUnderutilized
     consolidateAfter: {{ or (index .NodePool.ConfigItems "consolidate_after") "0s" }}
+#{{  end }}
+    # Minimum number of instance types for spot-to-spot consolidation
+#{{ if index .NodePool.ConfigItems "minInstanceTypesForSpotToSpotConsolidation"}}
```
mikkeloscar: config-items should be `snake_case`

mikkeloscar:
@tcondeixa FYI I rebased on dev after we merged #10829

tcondeixa:
👍

tcondeixa:
@mikkeloscar please review my approach for the rollout in the description in case you have any concerns. thanks

mikkeloscar:
:+1:

PR #10836, opened 2026-03-18
hotfix: skipper spp issues

szuecs:
optimize lb algorithm consistentHash to perform well with more than 400 backend pods


szuecs:
Looks like in production it's worse than in the benchmark

PR #10846, opened 2026-03-19
Provide cluster IPv4 CIDR in deploy config

mikkeloscar:
Utilize deployment config to make the IPv4 CIDR easily discoverable.

The current use case is that postgres-migrator needs this in order to set up the right LB SG rules. Having it available in the cluster makes it simpler to discover.


mikkeloscar:
:+1:

linki:
👍

PR #10852, opened 2026-03-20
ShadowTraffic: Introduces e2e tests for RouteGroup as source route object

madumalt:
Once a user creates a ST resource with a RouteGroup(RG) as the source routing object, the shadow-traffic-controller should kick in and create the shadow RG. When the user deletes the ST resource, then the controller should delete the created shadow RG.


cluster/manifests/shadow-traffic-controller/20-rbac.yaml
```diff
@@ -20,6 +20,7 @@ rules:
       - zalando.org
     resources:
       - shadowtraffics
+      - "shadowtraffics/status"
```
madumalt, line 23: Created a [separate PR](https://github.com/zalando-incubator/kubernetes-on-aws/pull/10864) for this change. Having this change here temporary to see whether the e2e tests pass with out any other issue.

szuecs:
lgtm from test code

madumalt:
👍

test/e2e/ingress.go
```diff
@@ -800,129 +800,6 @@ var _____ = describe("Ingress tests paths", func() {
 	})
 })
 
-var ______ = describe("Ingress tests custom routes - 2", func() {
```
szuecs, line 803: why do you delete this?

test/e2e/ingress.go
```diff
@@ -488,7 +488,7 @@ var ___ = describe("Ingress tests paths", func() {
 	})
 })
 
-var ____ = describe("Ingress tests custom routes", func() {
+var ____ = describe("Ingress tests custom routes - 1", func() {
```
szuecs, line 491: Why this change?

szuecs:
:+1:

test/e2e/ingress.go
madumalt: with the updated dependencies, two test can't have the same name, they must be unique. In Ingress.go there is two like that and I appended '- 1' to one.

test/e2e/ingress.go
madumalt: (it is an old commit, and is already reverted back)

PR #10866, opened 2026-03-20
Enable AAAA record creation in external-dns for IPv6 DNS support

ayruslore:
## Summary
- Enable external-dns to create AAAA (IPv6) DNS records for dualstack ingress
- Allows IPv6 clients to discover and connect to services via native IPv6
- Fixes client IP preservation for IPv6 client connections

## Test plan
- Verify DNS resolves to AAAA records: `dig test-predicates.pg9-test.zalan.do AAAA`
- Test IPv6 curl: `curl -6 https://test-predicates.pg9-test.zalan.do/test-client-ip-ipv6`
- Confirm ClientIP predicates match actual IPv6 client address
- Validate both A and AAAA records exist for ingress domains


szuecs:
:+1:

mikkeloscar:
We had an incident when upgrading external-dns to a version that creates AAAA records (#11984).

I will close this one for now, we have an internal ticket about properly enabling IPv6/AAAA records in the platform.

PR #10895, opened 2026-03-23
Delete kube-cluster-autoscaler resources

mikkeloscar:
Remove all configuration related to `kube-cluster-autoscaler`

It has been disabled in all clusters since a month and we're not intending to switch back from Karpenter.


linki:
👍

mikkeloscar:
:+1:

PR #10947, opened 2026-04-01
allow to exclude instance size by node pool to experiment

tcondeixa:
Allow to exclude instance size by node pool to experiment and see the cost impact.
The main goal is to validate the cost impact if we exclude the `large` instance from test clusters.
The goal would be to reduce the overhead cost from daemonsets that use a significant part of the `large` instance nodes.


tcondeixa:
👍

linki:
👍

PR #10976, opened 2026-04-07
vpa-updater: expose metrics to prometheus

tcondeixa:
This exposes the vpa-updater metrics to prometheus, so we can have visibility on the pod eviction.
This also allow us to start measuring how frequently we will be able to use the `InPlace` mode compared to the `Recreate`

NOTE: I don't see any significant advantage to configure the liveness and readiness probes, but we can also do it.

TEST: I tested with a pet cluster and I was able to have both the metrics from the `/metrics` endpoint and from prometheus.


cluster/manifests/02-vertical-pod-autoscaler/updater-service.yaml
```diff
@@ -0,0 +1,17 @@
...
+    component: vpa-updater
+spec:
+  ports:
+    - port: 8943
+      targetPort: 8943
+      protocol: TCP
+  selector:
+    application: kubernetes
+    component: vpa-updater
+  type: ClusterIP
```
mikkeloscar: It' unclear what's the purpose of this service is?

tcondeixa: this is for the prometheus to scrape the metrics, but probably the annotations do it already for all pods? We only have one pod for vpa-updater, and this setup should continue like this, so we don't really need the service.

mikkeloscar: yeah, prometheus discovers the pods so the service doesn't do anything in this regard. Let's remove it if not needed explicitly.

tcondeixa: done

tcondeixa:
👍

cluster/manifests/02-vertical-pod-autoscaler/updater-deployment.yaml
```diff
@@ -43,5 +47,5 @@ spec:
             cpu: {{.Cluster.ConfigItems.vpa_cpu}}
             memory: {{.Cluster.ConfigItems.vpa_mem}}
         ports:
-        - containerPort: 8080
+        - containerPort: 8943
```
demonCoder95, line 50: Why the change of port number?

tcondeixa: this is the default port number for vpa-updater configured on the image, but since the port was never used before probably no one noticed it was wrong.
https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/deploy/updater-deployment.yaml

demonCoder95: Ah, good catch!

demonCoder95:
👍

PR #10987, opened 2026-04-08
Remove ExecProbe feature gate logic

mikkeloscar:
Remove ExecProbe feature gate logic, it can't be configured since Kubernetes v1.35 (default enabled only) so we don't need the feature flag logic anymore.


mikkeloscar:
:+1:

tcondeixa (approved):

PR #10995, opened 2026-04-09
Add toggle to turn off skipper-ingress eastwest topology-mode

mikkeloscar:
This adds a config-item that allows turning off the `service.kubernetes.io/topology-mode: auto` setting on the skipper-ingress internal (eastwest) service.

This is done as we see a correlation with issues and this feature in at least a subset of clusters. This allows to easily turn it off to check the effect.


mikkeloscar:
👍

szuecs:
:+1:

PR #11002, opened 2026-04-12
Add support for skipper route metrics flag

trkohler:
I'd like to enable `skipper_serve_route_duration_seconds` skipper metric in some clusters because it would ease us monitoring during EKS migration


szuecs:
I don't see the need for this. Discussion internal

szuecs:
:-1:

PR #11010, opened 2026-04-13
fix: switch zone aware *.ingress.cluster.local ClusterIP AZ aware config

szuecs:
fix: switch zone aware *.ingress.cluster.local ClusterIP to be zone aware without automatic guardrails

My assumption is that it either flaps all the time or is enabled all the time. In both cases it would be a fix not a feature.
If it's disable and now enabled it should not matter either because skipper-ingress will handle the load anyways.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -356,8 +356,10 @@ skipper_eastwest_dns_log_enabled: "false"
 # if enabled adds port 8080 as svc port to eastwest svc
 skipper_ingress_eastwest_additional_port: "false"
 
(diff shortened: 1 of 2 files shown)


szuecs:
@mikkeloscar what do you think of the change?

The eks cluster had some hpa e2e error

mikkeloscar:
:+1:

mikkeloscar:
We agreed to disable Tier 1+2 prod initially via CCM so it only gets rolled out to Tier 3+4 via this PR

szuecs:
:+1:

PR #11025, opened 2026-04-15
[PoC] Expose all config items via ConfigMap

linki:
Quick PoC that "logs" all config items used for the current cluster configuration.


linki:
It worked.

tcondeixa:
> It worked.

just to learn, are the secret listed in the config map still encrypted or decrypted

PR #11027, opened 2026-04-15
fix: Add http redirect proxy protocol env for skipper

ayruslore:
Add environment variable to control or enable proxy protocol for the http redirect skipper process


ayruslore:
👍

MustafaSaber (approved):

PR #11028, opened 2026-04-16
Templating KRO and ACK under a config-item to enable in another clusters

tcondeixa:
templating KRO and ACK under a config-item to enable in other clusters

- ACK with the same permissions as the deployment services for AWS resources
- KRO with the same `poweruser` and `poweruser-read-secrets` as the deployment service in EKS clusters
- Both KRO and ACK are deployed in the master nodes (managed by AWS), but they still use the `kube-system` for leader election, configmaps and other things. So we need to unblock the KRO user to have access to the `kube-system` in the admission controller, but later we should restrict the actions.

These were tested in a pet cluster.

https://docs.aws.amazon.com/eks/latest/userguide/capabilities-security.html#kro-resource-permissions


cluster/manifests/02-admission-control/teapot.yaml
```diff
@@ -648,9 +648,11 @@ webhooks:
       - name: 'exclude-eks-components'
         expression: '!request.userInfo.username.startsWith("eks:")'
       # Restricted to cloud-euc1-test until we decide to make it production ready
-      {{- if eq .Cluster.Alias "cloud-euc1-test"}}
+      {{- if eq .Cluster.ConfigItems.kro_enabled "true"}}
```
linki: How about naming this `capability_kro_enabled` (as well as the other)?

tcondeixa: yes, sounds like a good approach

cluster/manifests/02-admission-control/teapot.yaml
```diff
@@ -648,9 +648,11 @@ webhooks:
       - name: 'exclude-eks-components'
         expression: '!request.userInfo.username.startsWith("eks:")'
       # Restricted to cloud-euc1-test until we decide to make it production ready
-      {{- if eq .Cluster.Alias "cloud-euc1-test"}}
+      {{- if eq .Cluster.ConfigItems.capability_kro_enabled "true"}}
```
tcondeixa, line 650: @linki I'm not sure if this is the best approach. Would this not allow KRO create resources in the kube-system?

cluster/cluster.yaml
```diff
@@ -319,6 +319,69 @@ Resources:
...
       KubernetesGroups:
         - zalando:readonly
+{{- if eq .Cluster.ConfigItems.capability_kro_enabled "true"}}
+  EKSAccessEntryKRO:
+    Type: "AWS::EKS::AccessEntry"
+    Properties:
+      ClusterName: !Ref EKSCluster
+      PrincipalArn: !GetAtt EKSCapabilityKRORole.Arn
+      Type: "STANDARD"
+      KubernetesGroups:
```
tcondeixa: I'm not entirely sure the best approach here, but assuming we want KRO for abstract things for users it's probably better to only give it the same permissions as the deployment-service? wdyt @linki ? any suggestion?

diegoroccia: It's just for us for experimentation for now, zalando:engineer should be fine

cluster/manifests/02-admission-control/teapot.yaml
diegoroccia: It's just for us for experimentation for now, should be fine

cluster/cluster.yaml
tcondeixa: I don't think engineer will work for the production accounts. This is still mainly copy-and-past, I'm reading the docs

tcondeixa: IIUC the KRO only have access to the RGDs, but we need to provide more permissions

tcondeixa:
kro requires permissions to create and manage the resources defined in your ResourceGraphDefinitions. By default, kro can only watch and manage RGDs themselves.

```
kro arbitrary resource permissions

kro requires permissions to create and manage the resources defined in your ResourceGraphDefinitions. By default, kro can only watch and manage RGDs themselves.

To grant kro permissions to create resources:

Option 1: Access entry policies

Associate pre-defined access entry policies like AmazonEKSAdminPolicy or AmazonEKSEditPolicy to the capability’s access entry.

Option 2: Kubernetes RBAC

Create a ClusterRoleBinding that grants the capability’s Kubernetes user the necessary permissions.
```

https://docs.aws.amazon.com/eks/latest/userguide/capabilities-security.html#kro-resource-permissions

cluster/manifests/02-admission-control/teapot.yaml
linki: Yes, but we treat it as a "trusted" system component since it's provided by AWS, so we can allow it.

tcondeixa:
I would still like to restrict the permission of the KRO user in the `kube-system` namespace, but we need to confirm with AWS what exactly the KRO capabilities requires (or check docs and do try and error).
For now I would leave it like this to unblock the PR and be able to proceed, but we should do it later when we enable it for more clusters.

tcondeixa:
👍

cluster/manifests/kro/kro-rbac.yaml
```diff
@@ -0,0 +1,33 @@
...
+    component: kro
+roleRef:
+  kind: ClusterRole
+  name: poweruser
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: User
+  name: "arn:aws:sts::{{.Cluster.InfrastructureAccountID}}:assumed-role/{{.Cluster.Alias}}-KRO-CapabilityRole/KRO"
+  apiGroup: rbac.authorization.k8s.io
+
```
linki: Shouldn't there be a `---` here to separate the two resources?

cluster/manifests/02-admission-control/teapot.yaml
```diff
@@ -647,10 +647,11 @@ webhooks:
         expression: '!(request.userInfo.username in ["system:kube-controller-manager", "system:kube-scheduler", "zalando-iam:zalando:service:k8sapi_credentials-provider"])'
       - name: 'exclude-eks-components'
         expression: '!request.userInfo.username.startsWith("eks:")'
-      # Restricted to cloud-euc1-test until we decide to make it production ready
-      {{- if eq .Cluster.Alias "cloud-euc1-test"}}
+      {{- if eq .Cluster.ConfigItems.capability_kro_enabled "true"}}
       - name: 'exclude-eks-kro-capability'
         expression: '!request.userInfo.username.startsWith("arn:aws:sts::{{.Cluster.InfrastructureAccountID}}:assumed-role/{{.Cluster.Alias}}-KRO-CapabilityRole/")'
```
linki: Below, in the RBAC definitions, the full user name is `arn:aws:sts:: [...] KRO-CapabilityRole/KRO`(ending in `/KRO`). Why do we use `startsWith` here instead of matching on the full user name?

cluster/cluster.yaml
```diff
@@ -319,6 +319,61 @@ Resources:
...
+            Action:
+              - 'sts:AssumeRole'
+              - 'sts:TagSession'
+      ManagedPolicyArns:
+      - !Sub "arn:aws:iam::${AWS::AccountId}:policy/ZalandoCloud-DenyEc2UsApSa"
+      - !Sub "arn:aws:iam::${AWS::AccountId}:policy/ZalandoCloud-DenyModifyShibboleth"
+      - !Sub "arn:aws:iam::${AWS::AccountId}:policy/ZalandoCloud-DenyDefault"
+      - !Sub "arn:aws:iam::${AWS::AccountId}:policy/ZalandoCloud-AllowPowerUser"
+      - !Sub "arn:aws:iam::${AWS::AccountId}:policy/ZalandoCloud-AllowPowerUserCustom"
+      {{- if eq .Cluster.Name "playground" }}
```
demonCoder95, line 366: Should this be `.Cluster.Alias`?

cluster/cluster.yaml
```diff
@@ -319,6 +319,61 @@ Resources:
...
+              - 'sts:AssumeRole'
+              - 'sts:TagSession'
+  EKSCapabilityKRO:
+    Type: 'AWS::EKS::Capability'
+    Properties:
+      ClusterName: !Ref EKSCluster
+      CapabilityName: "KRO"
+      Type: 'KRO'
+      RoleArn: !GetAtt EKSCapabilityKRORole.Arn
+      DeletePropagationPolicy: 'RETAIN'
```
linki, line 344: Would this _not_ delete the resources in the cluster? Why do we want to keep them?

cluster/manifests/kro/kro-rbac.yaml
tcondeixa: yes, I did not notice it because I apply first time only with poweruser, and the second time with both. It seems that CLM used the latest one defined in the file when there is no separator.

demonCoder95:
> For now I would leave it like this to unblock the PR and be able to proceed, but we should do it later when we enable it for more clusters.

We should have an issue to track this work.

cluster/cluster.yaml
tcondeixa: It's the only option available at the moment for managed capabilities, I hope they change this in the future.
https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-eks-capability.html#cfn-eks-capability-deletepropagationpolicy

tcondeixa:
> > For now I would leave it like this to unblock the PR and be able to proceed, but we should do it later when we enable it for more clusters.
> 
> We should have an issue to track this work.

I can create while the PR progress, this is the downside of managed capabilitiies.

cluster/cluster.yaml
tcondeixa: I copied this from the deployment-service and I see we have both available. I could change to Alias but in the cluster.yaml file Name is way more frequent, so I will keep it for now.

cluster/manifests/02-admission-control/teapot.yaml
tcondeixa: I can change this to comparison directly, I was not sure if KRO was always used for the Session Name. Can I use `!=` in the expression?

tcondeixa: done, please check again. Applying in my pet cluster seems to be working fine.

cluster/cluster.yaml
tcondeixa: I added a comment since it might help in the future

tcondeixa:
JFYI I found the current ClusterRole definition in the manifests from the github repository, but my main concern is KRO is automatically updated by AWS and trying to have fine grain control might end up causing an issues in case the permissions are updated and we are blocking with admission controller
```
# Source: kro/templates/cluster-role.yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  annotations:
    kubernetes.io/description: |
      This ClusterRole grants access for the kro controller to resources it always needs access to.
  labels:

    app.kubernetes.io/name: kro
    app.kubernetes.io/instance: kro
    app.kubernetes.io/version: "v0.9.1"
    app.kubernetes.io/component: controller
    app.kubernetes.io/part-of: kro
    rbac.kro.run/aggregate-to-controller: "true"
  name: kro:controller:static
rules:
- apiGroups:
  - kro.run
  resources:
  - resourcegraphdefinitions
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - kro.run
  resources:
  - resourcegraphdefinitions/finalizers
  verbs:
  - update
- apiGroups:
  - kro.run
  resources:
  - resourcegraphdefinitions/status
  verbs:
  - get
  - patch
  - update
- apiGroups:
  - internal.kro.run
  resources:
  - graphrevisions
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - internal.kro.run
  resources:
  - graphrevisions/finalizers
  verbs:
  - update
- apiGroups:
  - internal.kro.run
  resources:
  - graphrevisions/status
  verbs:
  - get
  - patch
  - update
- apiGroups:
  - apiextensions.k8s.io
  resources:
  - customresourcedefinitions
  verbs:
  - create
  - get
  - list
  - watch
  - patch
  - update
  - delete
- apiGroups:
  - ""
  resources:
  - configmaps
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - patch
  - delete
- apiGroups:
  - coordination.k8s.io
  resources:
  - leases
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - patch
  - delete
- apiGroups:
  - ""
  resources:
  - events
  verbs:
  - create
  - patch
  ```

cluster/manifests/02-admission-control/teapot.yaml
linki: nit: Can we also do it for the `ACK-CapabilityRole`? (maybe later)

linki:
👍

demonCoder95:
👍

tcondeixa:
👍

PR #11054, opened 2026-04-17
Drop kubenurse configuration

zaklawrencea:
Clean up the Kubenurse configuration as it has now been disabled across all clusters.


zaklawrencea:
:+1:

cluster/config-defaults.yaml
```diff
@@ -1257,18 +1257,14 @@ open_sg_ingress_ranges: ""
...
-network_monitoring_check_api_server_dns: "false"
-network_monitoring_check_kubenurse_service: "false"
-network_monitoring_check_kubenurse_ingress: "false"
-network_monitoring_check_neighborhood: "true"
-network_monitoring_check_unschedulable_nodes: "true"
-network_monitoring_check_interval: "1m"
-network_monitoring_separate_prometheus: "false"
+# Enable/Disable the s3-csi-driver daemonset:
+# https://github.com/awslabs/mountpoint-s3-csi-driver
+s3_csi_driver: "false"
```
tcondeixa: why is this here? are you missing to rebase on dev?

tcondeixa:
when we disabled the kubenurse was everything cleaned in all clusters? From the deletions it seems everything probably was already deleted but just want to confirm

cluster/config-defaults.yaml
tcondeixa: just want to confirm, it's fine we merge with these changes to avoid running all the tests again

zaklawrencea: Was an oversight in the rebase previously. I remove it, better to have it as intended now rather than needing to remove it again later.

tcondeixa:
👍

katyanna:
👍

PR #11056, opened 2026-04-17
Make the skipper-forward-backend-url configurable

mikkeloscar:
Make the `skipper_ingress_migration_forward_backend_url` configurable via a config-item. The default is what it was before.

Only relevant if `skipper_eks_migration_enabled: true`

The reason for this is in special cross account migration setups we can't use the predefined URL pattern as it assumes same account migration.


mikkeloscar:
:+1:

szuecs (approved):

mikkeloscar:
:+1:

mikkeloscar:
👍

szuecs:
:+1:

PR #11113, opened 2026-04-24
fix: canary opt-out non zone aware routes

szuecs:
fix: canary opt-out non zone aware routes

We saw that we get both args in canary and somehow it uses the non zone aware routes.

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -223,7 +223,7 @@ spec:
           - "-validate-query={{ .Cluster.ConfigItems.skipper_validate_query }}"
           - "-validate-query-log={{ .Cluster.ConfigItems.skipper_validate_query_log }}"
 {{ if eq .Cluster.ConfigItems.skipper_routesrv_enabled "exec" }}
-{{ if ne {{ .name }} "skipper-ingress-canary" }}
+{{ if ne .name "skipper-ingress-canary" }}
           - "-routes-urls=http://skipper-ingress-routesrv.kube-system.svc.cluster.local/routes"
 {{ end }}
           - "-normalize-host"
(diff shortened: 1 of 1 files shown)


greeshma1196:
👍🏽

cluster/manifests/skipper/deployment.yaml
```diff
@@ -223,7 +223,7 @@ spec:
           - "-validate-query={{ .Cluster.ConfigItems.skipper_validate_query }}"
           - "-validate-query-log={{ .Cluster.ConfigItems.skipper_validate_query_log }}"
 {{ if eq .Cluster.ConfigItems.skipper_routesrv_enabled "exec" }}
-{{ if ne "{{ .name }}" "skipper-ingress-canary" }}
+{{ if ne {{ .name }} "skipper-ingress-canary" }}
```
a4180p: i don't think it will work
before that it was string so it was always evaluating to `true` and now it's template inside template

a4180p: IIUC that will work
```suggestion
{{ if ne .name "skipper-ingress-canary" }}
```

a4180p: ```time="2026-04-24T10:29:23Z" level=fatal msg="Fail to provision: error rendering template main/cluster/manifests/skipper/deployment.yaml: template: main/cluster/manifests/skipper/deployment.yaml:226: unexpected \"{\" in operand"```

greeshma1196:
👍🏽

cluster/manifests/skipper/deployment.yaml
szuecs: thanks for the fix

a4180p:
👍

PR #11121, opened 2026-04-27
Decouple OPA Prometheus metrics from decision log plugin

x418:
Introduce a separate config flag `skipper_open_policy_agent_prometheus_metrics_enabled` to control the OPA `status.prometheus` block independently of `skipper_open_policy_agent_decision_log_export_enabled`.

### Why
Previously, the `status: prometheus: true` configuration was nested inside the same conditional branch as the `eopa_dl` decision log plugin. Given we still have some open question sregarding dl plugin, this enables us to at least already start receiving prometheus metrics from OPA


demonCoder95:
👍

PR #11126, opened 2026-04-28
expose the vpa update mode for kube-state-metrics and aws-load-balancer-controller

tcondeixa:
expose the vpa update mode for kube-state-metrics and aws-load-balancer-controller


tcondeixa:
👍

zaklawrencea:
:+1:

PR #11138, opened 2026-04-28
bugfix: canary deployment no route-url

greeshma1196:
Canary deployment doesn't get any `routes-url` if `skipper_ingress_zone_aware_routing_enabled` is set to false.


cluster/manifests/skipper/deployment.yaml
```diff
@@ -12,6 +12,8 @@
 {{ $canary_args := "" }}
```
szuecs: should not we have canary_args set here

tcondeixa:
@MustafaSaber @szuecs should we proceed with this change to dev, so you can validate and start unblocking channels?

ayruslore (approved):

MustafaSaber:
👍

PR #11167, opened 2026-05-04
Update etcd AMI

demonCoder95:
This update drops the snapd package manager from the AMI that has vulnerabilities. This is required for vulnerability resolution.

Also updates to etcd `v3.5.30`


demonCoder95:
Have to wait before merging this, we have an etcd version update in the AMI also in progress.

demonCoder95:
👍

mikkeloscar:
:+1:

PR #11180, opened 2026-05-05
Configure lifecycle-metrics with statefile

mikkeloscar:
This updates the kubernetes-lifecycle-metrics component to a version that supports storing it's state in a statefile on a PVC.

The purpose of this feature is to persist state across rescheduling of the `kubernetes-lifecycle-metrics` pods and thereby get rid of the usage of `karpenter.sh/do-not-disrupt: "true"` which was introduced to avoid the pod from being re-scheduled too often.

Before the component would replay some events if it was rescheduled leading to multi-counting of the same pod scheduling events impacting the SLO metrics for Pod Scheduling SLO. With the state file the replay of events is avoided.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -630,6 +630,7 @@ kube_state_metrics_vpa_update_mode: "InPlaceOrRecreate"
 
 kubernetes_lifecycle_metrics_mem_max: "4Gi"
 kubernetes_lifecycle_metrics_mem_min: "120Mi"
+kubernetes_lifecycle_metrics_state_file: "true"
 
 kube_node_ready_controller_cpu: "50m"
 kube_node_ready_controller_memory: "200Mi"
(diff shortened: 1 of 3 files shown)


linki:
Does it support running with `2` replicas to make it highly available?

mikkeloscar:
> Does it support running with `2` replicas to make it highly available?

It does not, but with the statefile we solve the problem of replaying events multiple times. In terms of loosing events, events are stored some time in the api so there is some times where it could be unavailable and still catch up. If a pod is created/scheduled/deleted in the time the pod is not running it would be lost, but the risk and impact of loosing those kind of events are minimal IMO.

We can do it in a follow up, but I would start with what is enough to resolve the most pressing problem of this component.

tcondeixa:
maybe an idea for the future would be to send the metrics directly to the metrics provider, so we don't need to keep state locally.

tcondeixa:
👍

tcondeixa (approved):

mikkeloscar:
:+1:

PR #11205, opened 2026-05-07
Disable poweruser networkpolicy access by default

mikkeloscar:
By default PowerUsers should not have access to crea/temodify networkpolicies. Put it behind a config-item to allow us to enable it per cluster e.g. in some test clusters where users want to explore the feature.

This is only relevant for EKS and only when: `aws_vpc_cni_enable_network_policy: true`


tcondeixa (approved):

linki:
👍

PR #11209, opened 2026-05-07
Inject default do-not-disrupt grace-period in pods

mikkeloscar:
### Description

This implement automatic injection of an annotation `karpenter.sh/do-not-disrupt: "30m"` when `pod.prevent-scale-down.grace-period-default: 30m` is set and none of the following conditions are true:

* Pod is a CronJob pod (gets `karpenter.sh/do-not-disrupt: "true"`)
* Pod is orphaned (gets `karpenter.sh/do-not-disrupt: "true"`)
* Pod has `zalando.org/prevent-scale-down: "true"` set (gets `karpenter.sh/do-not-disrupt: "true"`)
* Target node pools are not Karpenter (currently only implemented for this)

The idea is that all pods get this by default which essentially ensures a minimum runtime of that grace-period before they can be disrupted.
This uses a feature introduced in [Karpenter v1.12.0](https://github.com/kubernetes-sigs/karpenter/releases/tag/v1.12.0)

If users attempt to set a custom value of e.g `karpenter.sh/do-not-disrupt: "1h"` it will be rejected as we don't allow users to set a value themselves.

The value is set to 30m in production and 15m in test for now.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -802,6 +802,11 @@ teapot_admission_controller_runtime_policy_spot_hard_assignment: "false"
 
 # Enable and configure prevent scale down annotation
 teapot_admission_controller_prevent_scale_down_enabled: "true"
+{{if eq .Cluster.Environment "production"}}
+teapot_admission_controller_prevent_scale_down_grace_period_default: "30m"
+{{else}}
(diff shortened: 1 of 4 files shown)


tcondeixa (approved):

mikkeloscar:
:+1:

PR #11216, opened 2026-05-08
Run node DNS with static interface [1/4]

mikkeloscar:
## Problem

Today the per node DNScache+CoreDNS listens on 0.0.0.0/::0 and kubelet is dynamically configured to map the private IP of the node to the clusterDNS. This means the `/etc/resolve.conf` of the pods will point to the IP of the local node.

This dynamic setup puts a requirement of dynamically configuring kubelet on startup which is currently encoded in the custom AMI. This makes it more complicated to switch to another AMI without carrying over this complexity.

## Solution

Instead of using the _dynamic_ node IP we can simple define a static IP in the 'link-local' range '169.254.0.0/16' for IPv4 or from the 'Unique Local Address' range in IPv6 'fd00::/8'. This is inspired by how the [NodeLocal DNScache](https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/) is configured.
That way we can configure a static clusterDNS address for every AMI which is supported out of the box e.g. in EKS optimized AMIs/Karpenter.

To solve this, this PR introduces a sidecar container `dnscache-interface` to the CoreDNS daemonset pods. The purpose of this container is to setup a dummy interface and map a static IP to it which is then used for DNS.

Essentially the container does the equivalent of:

```bash
IFACE="dnscache"
IP="169.254.20.10"

ip link add ""$IFACE" type dummy
# Bring it up
ip link set "$IFACE" up
# Assign IP
ip addr add "$IP" dev "$IFACE"
```

Additionally the resources are changed from being per container in the daemonset pods, to be [at the pod level](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#pod-level-resource-specification) (Feature since Kubernetes v1.34). This simplifies the resource settings as the containers can just share the resources and we don't need to tweak per container. Likely we can reduce the default resources this way. For now the resources are just the sum of the previous per container resources, with the assumption that the new container, `dnscache-interface`, can simply use the spare resources already allocated, it does so little that it should not affect anything.

### Roll out

The CoreDNS daemonset pods are tied to the lifecycle of a node, because we can't just restart the pods without downtime of DNS. The implication of this design is that we need to roll the nodes in multiple stages in order to ensure a zero downtime roll out.

The roll out stages are as follows:

1. The new dnscache-interface container is enabled in the daemonset, this will setup the static interface on all new pods/nodes
2. The unbound (dnscache) config is changed to listen on both 0.0.0.0/::0 and the static ipaddress/interface. This ensures that kubelet still can connect to 0.0.0.0 which is required until the AMI/kubelet config is changed.
3. The AMI/Kubelet is updated to use the static ipaddress meaning DNS traffic will stop going over 0.0.0.0
4. 0.0.0.0/::0 is removed from the unbound (dnscache) config as it's no longer needed

#### Why not just use upstream NodeLocal DNScache?

The current DNS cache setup is based on extensive performance testing also comparing NodeLocal DNS cache where unbound + CoreDNS was the best performing setup in terms of latency and resource usage

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -1062,12 +1062,23 @@ external_dns_mem: "4Gi"
 expirimental_dns_unbound_liveness_probe: "true"
 
 # DNS container resources
-dns_unbound_cpu: "100m"
-dns_unbound_mem: "50Mi"
-dns_unbound_exporter_cpu: "10m"
-dns_unbound_exporter_mem: "45Mi"
-dns_coredns_cpu: "50m"
-dns_coredns_mem: "100Mi"
+dns_coredns_cpu: "160m"
+dns_coredns_mem: "195Mi"
+
+# Roll out of static dns interface
+#
+# The rollout needs to be in multiple stages to avoid disruption of node local
+# DNS. Each stage must fully complete i.e. all CoreDNS pods must rotate (thus
+# all nodes must rotate) before going to the next stage.
+#
+# stages:
+#
+#  interface - static interface container runs
+#  listening - dns cache listens to static interface
+#  kubelet - kubelet configures static dns IP address - worker node rotation!
+#  only_interface - dns cache listens _only_ on the static interface
+#
+dns_static_interface: "interface"
 
 # special roles for test/pet clusters
 {{if eq .Cluster.Environment "e2e"}}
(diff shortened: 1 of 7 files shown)


szuecs:
nice

mikkeloscar:
👍

zaklawrencea (approved):

PR #11217, opened 2026-05-08
Run node DNS with static interface [2/4]

mikkeloscar:
This is the second phase of the roll out described in #11216 

All clusters have the stage defined at cluster level so this will only affect e2e (and new clusters).


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #11259, opened 2026-05-12
Shorten EKS Control Plane SubscriptionFilter / DeliveryStream names

zaklawrencea:
To make sure new accounts/clusters do not hit naming max length limitations, shorten the resource name of the EKS Control Plane logging subscription filter/delivery stream.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -172,13 +172,13 @@ Resources:
     Properties:
       LogGroupName: !Ref ControlPlaneLogGroup
       RoleArn: !GetAtt EKSControlPlaneCWtoFirehoseRole.Arn
-      FilterName: "EKSCtrlPlaneLogs-{{.Cluster.Name}}"
+      FilterName: "EKS-{{.Cluster.Name}}"
       FilterPattern: ""
       DestinationArn: !GetAtt EKSControlPlaneLogsDataFirehose.Arn
   EKSControlPlaneLogsDataFirehose:
     Type: AWS::KinesisFirehose::DeliveryStream
     Properties:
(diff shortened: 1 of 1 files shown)


cluster/cluster.yaml
```diff
@@ -245,7 +245,7 @@ Resources:
                   - "firehose:PutRecord"
                   - "firehose:PutRecordBatch"
                 Resource:
-                  - "arn:aws:firehose:*:{{.Cluster.InfrastructureAccountID}}:deliverystream/EKSCtrlPlaneDeliveryStream-{{.Cluster.Name}}"
+                  - "arn:aws:firehose:*:{{.Cluster.InfrastructureAccountID}}:deliverystream/EKS-{{.Cluster.Name}}"
       RoleName: "EKSCtrlPlaneCWtoFirehose-{{.Cluster.Name}}"
```
tcondeixa: is this name not too big for some cases?

zaklawrencea: I don't think the ARN length is a problem. And adding 4 characters to the cluster name `EKS-<cluster-name>` for the delivery stream should be reasonable.

tcondeixa: I was not taking about the arn length but the RoleName itself that is also prepended with `EKSCtrlPlaneCWtoFirehose-`

zaklawrencea: Do you have a suggestion for a shorter name? `EKSCWtoFirehose-` / `EKSFirehoseS3-`. (trying to preserve some context)

tcondeixa: I think `EKSCWtoFirehose-` should be enough

tcondeixa (approved):

zaklawrencea:
:+1:

PR #11262, opened 2026-05-12
add toggle to drop lightstep envar injection in admission-controller

demonCoder95:
As we are decommissioning lightstep, we need to drop this environment variable injection in the admission-controller. 

This PR introduces a toggle with the default behavior of injection allowed. In a follow-up, we can adjust the e2e tests to drop checking for injection of these variables.


szuecs:
:+1:

tcondeixa:
@demonCoder95 did you confirm with observability they are not using the same env variables for Dash0? my guess is this env variables are still in-use, otherwise they would need to force everyone to update the otel sdks.

PR #11267, opened 2026-05-15
Sandbox Controller update

a4180p:
Updating:
- SandboxEgress CRD to new version
- Sandbox Controller to new version
- Egress e2e tests to use Stack reference instead of ConfigMap referens


szuecs (approved):

a4180p:
👍

PR #11290, opened 2026-05-19
fix: skipper-ingress PM action-item ensure expected dataclients

szuecs:
This change makes sure skipper-ingress data plane has the expected dataclients configured and if not we stop the deployment by fatal log.

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -228,6 +228,7 @@ spec:
         args:
           - "run.sh"
           - "skipper"
+          - "-ensure-dataclients=inline,eskipfile-remote"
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

cluster/manifests/skipper/deployment.yaml
```diff
@@ -228,6 +228,7 @@ spec:
         args:
           - "run.sh"
           - "skipper"
+          - "-ensure-dataclients=inline,eskipfile-remote"
```
MustafaSaber, line 231: I think we can also have a configItem with default value here

MustafaSaber (approved):

cluster/manifests/skipper/deployment.yaml
szuecs: Less config item the better. If we will change the dataclients we will see that it fails in e2e test on create/update cluster.

PR #11294, opened 2026-05-20
new vpa version with config-item

tcondeixa:
new vpa version with patch to restart pod on OOM with `InPlaceOrReplace`.
This is a temporary patch while kubernetes fixes the issue in one of the next releases.
I'm using a config-item to be easily/fast to rollout and rollback if needed.
I reused the existing `vertical_pod_autoscaler_version` since I did not find it used anywhere, probably this was never cleaned.


mikkeloscar:
:+1:

tcondeixa:
👍

PR #11314, opened 2026-05-21
feature: zone aware routing

greeshma1196:
Enable zone aware routing in canary


greeshma1196:
Closing this PR, will use CCM to do the rollouts.

PR #11315, opened 2026-05-21
Remove misleading annotations from EKS migration service

linki:
Based on our docs https://cloud.docs.zalando.net/reference/kubernetes-infrastructure/eks-migration/howtos/loadbalancer-service/#noteworthy-annotations-on-load-balancer-services this removes annotations that don't work anymore or that are already on the default.

Main driver was the removal of `service.beta.kubernetes.io/aws-load-balancer-extra-security-groups` which doesn't have any effect and could lead to people having a false sense of security. This spec was also used for a Service on a legacy cluster where it fails with: `error: BYO extra security group annotation "service.beta.kubernetes.io/aws-load-balancer-extra-security-groups" is not supported by NLB` in the latest Kubernetes version.

The other annotations are changed from their deprecated version or are dropped because they are the default value (on EKS). None of these changes here should have any effect on the final load balancer setup. 

EDIT: it will tighten the SG to only be accessible from the defined SG, not from everywhere. Not so relevant for internal load balancers but more accurate overall this way.

---

Note, I assume this service is only relevant on EKS clusters. If not, please let me know.

### DIFF
--- a/cluster/manifests/skipper/service-eks-internal.yaml
+++ b/cluster/manifests/skipper/service-eks-internal.yaml
@@ -8,13 +8,10 @@ metadata:
     service.beta.kubernetes.io/aws-load-balancer-attributes: dns_record.client_routing_policy=availability_zone_affinity,load_balancing.cross_zone.enabled=false
     service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
     # SG of the old cluster worker nodes
-    service.beta.kubernetes.io/aws-load-balancer-extra-security-groups: {{ .Cluster.ConfigItems.worker_sg_legacy_cluster }}
     service.beta.kubernetes.io/aws-load-balancer-healthcheck-path: /kube-system/healthz
     service.beta.kubernetes.io/aws-load-balancer-healthcheck-protocol: http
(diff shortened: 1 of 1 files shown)


linki:
[UPDATE] I replaced the "extra" annotation with the `service.beta.kubernetes.io/aws-load-balancer-security-groups` annotation. The difference is that "extra" added to the default list whereas the replacement annotation is used to define the full list of allowed SGs.

@szuecs This is supported and might be even better suited in the first place, if the legacy cluster is really the only client to that service.

mikkeloscar:
Just to be sure, this does not result in replacing the existing LBs or anything like that right? That would be the biggest risk as some apps may depend on this connection (if they are actively migrating).

linki:
Correct. It's an in-place update and it merely updates the SGs attached to the load balancer.

It will exclusively allow traffic coming from the `worker_sg_legacy_cluster` which I think is the intended behaviour anyways.

szuecs:
I think the other is also fine.
No need to change.
I think your change reduces the number of SGs attached to nlb.

linki:
👍

mikkeloscar:
:+1:

PR #11363, opened 2026-06-02
feat: Enable Runtime Response Capabilties for Wiz Sensor

vinaythupili:
- Enable Runtime Response Capability for Wiz Sensor on all clusters.
- We have tested the Runtime Response Capability with wiz sensor on test cluster and as runtime response policies need to be rolled out across all cluster for threat prevention this need to enabled by default on all projects.


zaklawrencea:
:+1:

vinaythupili:
👍

PR #11366, opened 2026-06-02
Update etcd image

Shinzu:
updates AMI for etcd to latest version which includes amazon-ssm-agent again, which was removed with te removal of snap


linki:
👍

Shinzu:
:+1:

PR #11367, opened 2026-06-02
EBS CSI Plugin: Set memory through variable

katyanna:
Replace hardcoded 40Mi memory requests/limits with configurable template variable via `ebs_csi_plugin_memory` config item, allowing cluster-specific memory tuning without manifest modification.


linki:
👍

katyanna:
👍

PR #11404, opened 2026-06-09
allow to configure per cluster the min sync for iptables and ipvs

tcondeixa:
This allows to configure per cluster the minSync period for iptables and ipvs.

There is one cluster with a Client Side Load Balancer accessing directly the pods and when we have cases of syncronized node boots is causes issues with packet delays. If we configure a value of `20s` for this cluster it should at least to help spread across a wider window.

https://github.com/zalando-build/zooport/issues/7288


szuecs:
@tcondeixa we should switch the implementation from iptables to nftables. nftables is the modern replacement for iptables and ipvs will be not supported by kubernetes soon.

tcondeixa:
this seems not to be need and it might not solve the problem, so closing

PR #11442, opened 2026-06-16
Remove wrongly configured SG for internal service

linki:
Following up on https://github.com/zalando-incubator/kubernetes-on-aws/pull/11315.

The change in the PR prevents connection from the legacy app to the EKS cluster's proxy service.

Turns out the annotation is wrongly used. Configuring the SG doesn't mean the LB is accessible from nodes having that SG. It means the LB is configured with the LB. In our case it would give the LB the worker SG which is not what we want. We want to access the LB _from_ the worker which is not the same.

Let's just specify no SG which allows traffic from everywhere. Since it's an internal LB it's fine. However, as additional precaution, we're setting the IPv4 CIDR (legacy cluster nodes) which adds it as the only inbound rule.

### DIFF
--- a/cluster/manifests/skipper/service-eks-internal.yaml
+++ b/cluster/manifests/skipper/service-eks-internal.yaml
@@ -7,8 +7,6 @@ metadata:
     service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags: application=skipper-ingress,component=ingress
     service.beta.kubernetes.io/aws-load-balancer-attributes: dns_record.client_routing_policy=availability_zone_affinity,load_balancing.cross_zone.enabled=false
     service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
-    # SG of the old cluster worker nodes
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

linki:
👍

mikkeloscar:
:+1:

linki:
👍

cluster/manifests/skipper/service-eks-internal.yaml
```diff
@@ -1,4 +1,4 @@
-# {{ if eq .Cluster.ConfigItems.skipper_eks_migration_svc_enabled "true" }}
+# {{ if eq .Cluster.Provider "zalando-eks" }}
```
mikkeloscar: Ideally we want zalando-eks AND siblingCluster != nil so we don't create it for single EKS clusters without a legacy sibling.

linki: Done.

linki:
👍

tcondeixa:
👍

PR #11476, opened 2026-06-22
feat: allow mutatingwebhookconfigurations for poweruser role

alkagin:
This is needed for the Dash0 operator. It's hidden behind a feature flag as Dash0 is for a PoC.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -1340,6 +1340,8 @@ role_sync_controller_enabled: "true"
 role_sync_controller_enabled: "false"
 {{ end }}
 
+role_grant_mutatingwebhookconfigurations_permission: "false"
+
 # Access control configs
 # Enable users with the poweruser role to modify EndpointSlices.
 access_control_poweruser_modify_endpointslices: "false"
--- a/cluster/manifests/roles/poweruser-role.yaml
+++ b/cluster/manifests/roles/poweruser-role.yaml
@@ -94,6 +94,9 @@ rules:
   - admissionregistration.k8s.io
   resources:
   - validatingwebhookconfigurations
+{{- if eq .Cluster.ConfigItems.role_grant_mutatingwebhookconfigurations_permission "true" }}
+  - mutatingwebhookconfigurations
+{{- end }}
   verbs:
   - create
   - delete


cluster/manifests/deployment-service/controller-rbac.yaml
```diff
@@ -216,6 +216,51 @@ subjects:
...
+metadata:
+  name: "deployment-service-executor-tracing-rbac"
+  labels:
+    application: "deployment-service"
+    component: "controller"
+rules:
+  - apiGroups:
+      - "admissionregistration.k8s.io"
+    resources:
+      - mutatingwebhookconfigurations
```
mikkeloscar: The right model would be to add these permisisons (likely behind a config-item) to the poweruser role. That will mean that CDP/deployment-service and Manual users automatically gain those permissions: https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/roles/poweruser-role.yaml

cluster/manifests/deployment-service/controller-rbac.yaml
```diff
@@ -216,6 +216,51 @@ subjects:
...
+    verbs:
+      - patch
+      - update
+  - apiGroups:
+      - "monitoring.coreos.com"
+    resources:
+      - prometheusrules
```
mikkeloscar: If this CRD is deployed, the permissions will automatically be made available in the respective cluster, so these should not be defined here.

cluster/manifests/deployment-service/status-service-rbac.yaml
```diff
@@ -60,6 +60,11 @@ rules:
     resources:
       - flinkdeployments
     verbs: ["get", "list", "watch"]
+  - apiGroups:
+      - "monitoring.coreos.com"
+    resources:
+      - prometheusrules
+    verbs: ["get", "list", "watch"]
```
mikkeloscar: Same here, no need define it explicitly, it will be made available if the respective CRD is deployed.

cluster/manifests/deployment-service/controller-rbac.yaml
alkagin: done at [3fba83a](https://github.com/zalando-incubator/kubernetes-on-aws/pull/11476/commits/3fba83a7d4337a76888fe3bdee18331ce0744346)

cluster/manifests/deployment-service/controller-rbac.yaml
alkagin: done at [3fba83a](https://github.com/zalando-incubator/kubernetes-on-aws/pull/11476/commits/3fba83a7d4337a76888fe3bdee18331ce0744346)

cluster/manifests/deployment-service/status-service-rbac.yaml
alkagin: done at [3fba83a](https://github.com/zalando-incubator/kubernetes-on-aws/pull/11476/commits/3fba83a7d4337a76888fe3bdee18331ce0744346)

alkagin:
@mikkeloscar could you add a minor label?

mikkeloscar:
:+1:

diegoroccia:
:+1:

tcondeixa:
@mikkeloscar is this still needed if the dash0 operator (and other resources) are deployed with CLM?

alkagin:
> @mikkeloscar is this still needed if the dash0 operator (and other resources) are deployed with CLM?

Yes, it still needed. Check conversation in chat.

tcondeixa:
> > @mikkeloscar is this still needed if the dash0 operator (and other resources) are deployed with CLM?
> 
> Yes, it still needed. Check conversation in chat.

This should not be needed unless you deploy with CDP, that's why it's added to the poweruser (used by cdp and the users with higher privileges). CLM has enough permission for this action without any change (it's a different role)

PR #11489, opened 2026-06-24
stackset: Allow per stack public hostname in EKS

mikkeloscar:
Introduce a config-item that allows the automatic public hostname per Stack for StackSets.

It was previously disable for any EKS cluster to facility EKS migration as it can't be enabled in two clusters of the same account, however, there is no way for users to get the feature in EKS now.

Moving it to a config-item makes it still disabled by default in EKS, but can optionally be enabled for valid use cases, when there is no non-EKS cluster in the account.

Ideally we make the feature opt-in per StackSet and disable it by default, but this is a temporary solution.


tcondeixa:
👍

mikkeloscar:
:+1:

PR #11497, opened 2026-06-25
Make kube-aws-iam-controller deployment conditional on enablement config

katyanna:
In preparation for the decommissioning of kube-aws-iam-controller, this makes the controller explicitly opt-in, allowing clusters to disable it before complete removal.

- Add `kube_aws_iam_controller_enabled` config flag (defaults to ~false~true)
- Wrap controller deployment, RBAC, VPA, and admission control configs with enablement check
- Update e2e tests to enable the controller for test execution by setting env var in init()

Ref.: https://github.com/zalando-build/cloud-platform/issues/7886

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -628,7 +628,7 @@ metrics_server_metric_resolution: "15s"
 
 kube_aws_iam_controller_cpu: "5m"
 kube_aws_iam_controller_mem: "50Mi"
-kube_aws_iam_controller_enabled: "false"
+kube_aws_iam_controller_enabled: "true"
 
 kube_state_metrics_cpu: "100m"
 kube_state_metrics_mem_max: "4Gi"
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -628,6 +628,7 @@ metrics_server_metric_resolution: "15s"
 
 kube_aws_iam_controller_cpu: "5m"
 kube_aws_iam_controller_mem: "50Mi"
+kube_aws_iam_controller_enabled: "false"
 
 kube_state_metrics_cpu: "100m"
 kube_state_metrics_mem_max: "4Gi"
--- a/cluster/manifests/02-admission-control/config.yaml
+++ b/cluster/manifests/02-admission-control/config.yaml
@@ -105,7 +105,9 @@ data:
   pod.pod-security-policy.privileged-service-accounts.kube-system_node-monitor: ""
   pod.pod-security-policy.privileged-service-accounts.kube-system_nvidia: ""
   pod.pod-security-policy.privileged-service-accounts.kube-system_audittrail-adapter: ""
+{{- if eq .Cluster.ConfigItems.kube_aws_iam_controller_enabled "true"}}
   pod.pod-security-policy.privileged-service-accounts.kube-system_kube-aws-iam-controller: ""
+{{- end }}
   pod.pod-security-policy.privileged-service-accounts.kube-system_kube2iam: ""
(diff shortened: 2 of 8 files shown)


test/e2e/aws_iam.go
```diff
@@ -28,6 +29,10 @@ import (
 	admissionapi "k8s.io/pod-security-admission/api"
 )
 
+func init() {
+	os.Setenv("KUBE_AWS_IAM_CONTROLLER_ENABLED", "true")
```
linki: @katyanna Do the e2e tests fail when the env var isn't set to true?

cluster/manifests/02-admission-control/config.yaml
```diff
@@ -105,7 +105,9 @@ data:
   pod.pod-security-policy.privileged-service-accounts.kube-system_node-monitor: ""
   pod.pod-security-policy.privileged-service-accounts.kube-system_nvidia: ""
   pod.pod-security-policy.privileged-service-accounts.kube-system_audittrail-adapter: ""
+{{- if eq .Cluster.ConfigItems.kube_aws_iam_controller_enabled "true"}}
   pod.pod-security-policy.privileged-service-accounts.kube-system_kube-aws-iam-controller: ""
+{{- end }}
```
linki: For simplicity, let's omit this `if` statement if it's not needed. The `efs-provisioner`, for example, is also turned off sometimes but still mentioned here unconditionally.

cluster/manifests/deletions.yaml
```diff
@@ -399,3 +399,21 @@ post_apply:
...
+  kind: ClusterRole
+- name: kube-aws-iam-controller
+  kind: ClusterRoleBinding
+- name: kube-aws-iam-controller-privileged-psp
+  namespace: kube-system
+  kind: RoleBinding
+- name: kube-aws-iam-controller
+  namespace: kube-system
+  kind: VerticalPodAutoscaler
+{{- end }}
```
linki, line 423: This list misses the CRD but I think we can leave that out for now. This way we preserve the role definitions that the users have submitted in case we need to roll something back. But at some point we should clean up the CRD as well.

linki: I added it but commented it out. This way we can easily recover if this feature gets turned off to early. Once we don't have any AWSIAMRoles anywhere, we can uncomment it.

cluster/config-defaults.yaml
```diff
@@ -628,6 +628,7 @@ metrics_server_metric_resolution: "15s"
 
 kube_aws_iam_controller_cpu: "5m"
 kube_aws_iam_controller_mem: "50Mi"
+kube_aws_iam_controller_enabled: "true"
```
linki, line 631: Enabled by default to match the current setup. This way we can merge the PR and control it via config item later.

linki:
This is now safe to merge when it's green. It doesn't disable the controller by default.

linki:
👍

cluster/manifests/deletions.yaml
```diff
@@ -399,3 +399,25 @@ post_apply:
...
+  namespace: kube-system
+  kind: ServiceAccount
+- name: kube-aws-iam-controller
+  kind: ClusterRole
+- name: kube-aws-iam-controller
+  kind: ClusterRoleBinding
+- name: kube-aws-iam-controller-privileged-psp
+  namespace: kube-system
+  kind: RoleBinding
+- name: kube-aws-iam-controller
```
tcondeixa, line 416: it's probably better to move the vpa to the top

linki:
👍

tcondeixa:
👍

PR #11498, opened 2026-06-26
Enable node-exporter ENA metrics

mikkeloscar:
This enables the node exporter experimental metrics which includes ENA metrics (e.g. packet drops on EC2 Network interfaces).

This was previously set in 173 clusters per cluster, but newer clusters didn't have it enabled. Since we ran it in so many clusters already it's safe to enable by default.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #11534, opened 2026-07-01
Allow IAM role for cdp-steps to be assumed via OIDC provider

linki:
Allow IAM role for cdp-steps to be assumed using OIDC for the migration

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -1856,23 +1856,54 @@ Resources:
           Arn: !GetAtt KarpenterInterruptionQueue.Arn
   DeploymentIAMRole:
     Properties:
-      AssumeRolePolicyDocument:
-        Statement:
-          - Action:
-              - 'sts:AssumeRole'
-            Effect: Allow
-            Principal:
-              Service:
-                - ec2.amazonaws.com
-          - Action:
-              - 'sts:AssumeRole'
-            Effect: Allow
-            Principal:
-              AWS: !Join
-                - ''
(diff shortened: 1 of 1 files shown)


cluster/cluster.yaml
```diff
@@ -1856,23 +1856,55 @@ Resources:
...
+                }
+              }
+            ]
+          }
+{{- if eq .Cluster.Provider "zalando-eks" }}
+        - OIDC: !Select [1, !Split ["//", !GetAtt EKSCluster.OpenIdConnectIssuerUrl]]
+{{- else }}
+        - OIDC: "{{.Cluster.LocalID}}.{{.Values.hosted_zone}}"
+{{- end }}
+          WorkerIAMRole: !Sub 'arn:aws:iam::${AWS::AccountId}:role/${WorkerIAMRole}'
```
tcondeixa: Why don't we define the arn of the WorkerIAMRole directly in the template? just like passing directly `'arn:aws:iam::${AWS::AccountId}:role/${WorkerIAMRole}'` .

IMO it's also a bit confusing because we are using the same name for both variables

linki: Fixed

tcondeixa (approved):

tcondeixa:
let's move with this one since it needs to reach stable before we adjust cdp-steps role

tcondeixa:
do we use the same role for both the `cdp-steps` and the `deployment-service`?

ni9hty:
:+1:

linki:
> do we use the same role for both the `cdp-steps` and the `deployment-service`?

Yes, I think so.

PR #11584, opened 2026-07-07
Allow access to ValidatingAdmissionPolicies

mikkeloscar:
Allow conditional access to ValidatingAdmissionPolicies for the poweruser/cdp role per cluster

This is done to enable it for special cases i.e. ZAP

ref: https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/


mikkeloscar:
:+1:

jmcs:
:+1:

PR #11588, opened 2026-07-08
increase kubelet_summary_metrics_hpa_max_replicas to 20

demonCoder95:
We suffered OOMkills on some `kubelet-summary-metrics` Pods last evening because the HPA had maxed out to `10` replicas and the deployment couldn't scale further. Increasing this for all clusters to avoid this in the future.


tcondeixa (approved):

demonCoder95:
👍

PR #11605, opened 2026-07-09
Allow PowerUser create tokenreviews permissions at cluster level

mikkeloscar:
This adds a config-item to enable tokenreviews create permissions to PowerUser/CDP by cluster. It's disabled by default.

This is done to enable ZAP where the AgentGateway depends on it.


jmcs:
:+1:

mikkeloscar:
:+1:

PR #11615, opened 2026-07-10
Add servicequotas:GetServiceQuota to zmon permissions

Shinzu:
Extend zmon permissions for ServiceQuotas by `GetServiceQuota`, with this we can query single quotas instead 

ref: https://linear.app/zalando/issue/OBS-1692/limit-aws-servicequota-detection-to-the-ones-that-are-actually-used-by


mikkeloscar:
:+1:

Shinzu:
:+1:

PR #11643, opened 2026-07-13
Support running EKS AMIs

mikkeloscar:
Enable support for running EKS AMIs as alternative to the Custom Ubuntu AMI.

This puts it behind a config-item such that it can be enabled per cluster or even node pool. The default is still Ubuntu AMI (`custom`).

This is e2e tested with bottlerocket in #11838


mikkeloscar:
:+1:

zaklawrencea (approved):

PR #11644, opened 2026-07-13
Set backend-zone-metrics in skipper-ingress deployment

greeshma1196:
Add config item backend-zone-metrics to skipper-ingress deployment


greeshma1196:
👍🏽

MustafaSaber (approved):

PR #11660, opened 2026-07-14
e2e: Only set efs_id in non-EKS

mikkeloscar:
Bottlerocket is not compatible with the legacy EFS provisioner as it requires nfs tools on the instance (#11643)

Since we don't use this feature in any EKS cluster, simply disable it from e2e test in EKS clusters as well.

### DIFF
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -371,9 +371,6 @@ post_apply:
 - name: efs-provisioner
   kind: ConfigMap
   namespace: kube-system
-- name: efs-provisioner
-  kind: ServiceAccount
-  namespace: kube-system
 - name: efs-provisioner-runner
   kind: ClusterRole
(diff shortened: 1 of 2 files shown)


mikkeloscar:
The build fails because of:

```
Fail to provision: unable to delete: admission webhook \"serviceaccount-admitter.teapot.zalan.do\" denied the request: serviceaccount is still used by the following pods: efs-provisioner-5f9b776ccd-gsqkj"
```

This is because the pod of the deleted deployment is not yet cleaned up before it gets to the serviceAccount deletion. In normal operation CLM would just rerun and eventually succeed. For e2e this doesn't happen.

I would suggest we simply merge this through knowing the e2e will fail for these cases. Once it's merged through all the channels it's solved.

mikkeloscar:
:+1:

katyanna:
:+1:

PR #11665, opened 2026-07-15
e2e: Only allow IAM test role access to test bucket

mikkeloscar:
Fixes a bug in the e2e test where getting the worker role credentials would also pass the test. We explicitly want to catch such issues and fail the tests.


mikkeloscar:
:+1:

katyanna:
:+1:

PR #11675, opened 2026-07-15
e2e: Add a test for ephemeral storage

zaklawrencea:
Add a new e2e test to verify pod scheduling with ephemeral storage. This should utilise RAID0 and be configured by Bottlerocket natively via `instanceStorePolicy: RAID0`. In our Ubuntu AMI we handle the RAID0 setup manually.

### DIFF
--- a/test/e2e/instance_storage.go
+++ b/test/e2e/instance_storage.go
@@ -1,0 +1,89 @@
+package e2e
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+	admissionapi "k8s.io/pod-security-admission/api"
+)
+
+var _ = describe("Instance storage", func() {
+	f := framework.NewDefaultFramework("instance-storage")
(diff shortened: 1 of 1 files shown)


test/e2e/instance_storage.go
```diff
@@ -0,0 +1,87 @@
...
+					Resources: v1.ResourceRequirements{
+						Requests: v1.ResourceList{
+							v1.ResourceCPU:              resource.MustParse("1m"),
+							v1.ResourceMemory:           resource.MustParse("50Mi"),
+							v1.ResourceEphemeralStorage: resource.MustParse("1Gi"),
+						},
+						Limits: v1.ResourceList{
+							v1.ResourceCPU:              resource.MustParse("1m"),
+							v1.ResourceMemory:           resource.MustParse("50Mi"),
+							v1.ResourceEphemeralStorage: resource.MustParse("1Gi"),
```
mikkeloscar: We need to request north of 100Gi to ensure we don't just land on a node with a root EBS. DO 500Gi or so to ensure we actually get an instance type with local ssd where RAID0 is set up.

We ideally also find a way from the pod to validate if it actually can see 500Gi space and not just that it can write a few bytes.

test/e2e/instance_storage.go
```diff
@@ -0,0 +1,89 @@
...
+	var cs kubernetes.Interface
+
+	BeforeEach(func() {
+		cs = f.ClientSet
+	})
+
+	It("Should schedule a pod with ephemeral storage [Zalando]", func(ctx context.Context) {
+		ns := f.Namespace.Name
+
+		By("Creating a pod requesting 500Gi ephemeral storage on the worker-instance-storage pool")
```
mikkeloscar:
```suggestion
		By("Creating a pod requesting 500Gi ephemeral storage")
```

mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #11676, opened 2026-07-15
[HOTFIX]: revert platform-iam-tokeninfo image in production

demonCoder95:
This reverts image update in #11670 

This new version is causing 1/2 Pods to constantaly crash loop, and there are no Pod logs or events available to debug easily either. Let's revert, and investigate in beta channel clusters where we have the same issue.


demonCoder95:
👍

linki:
👍

PR #11659, opened 2026-07-14
ACK needs to be able to tag EKS resources

diegoroccia:
Signed-off-by: Diego Roccia <diego.roccia@gmail.com>

PR #11689, opened 2026-07-16
Revert "ACK needs to be able to tag EKS resources"

mikkeloscar:
Reverts zalando-incubator/kubernetes-on-aws#11659

The action was added under assumerole policy which is invalid. Needs to be an actual inline policy permission


mikkeloscar:
:+1:

demonCoder95:
👍

PR #11698, opened 2026-07-16
Annotate DNS pods to track progress

mikkeloscar:
For the rollout of #11216 we need a way to track if a cluster is fully rolled out for a stage before proceeding to the next stage. To simplify this, simply annotate the CoreDNS pods such that we know if the given stage was applied or not and can count how many pods/nodes are missing per cluster.


mikkeloscar:
:+1:

zaklawrencea (approved):

PR #11703, opened 2026-07-17
Karpenter: Add missing permissions

mikkeloscar:
Observed in e2e that Karpenter tries to do `DescribeInstanceStatus` and failing. Add missing permissions from upstream.


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #11705, opened 2026-07-17
Fix e2e aws iam test bucket policy

mikkeloscar:
The bucket policy introduced in https://github.com/zalando-incubator/kubernetes-on-aws/pull/11665 was so strict we can't cleanup the bucket after the e2e is run.

Change the policy to allow any action except for `s3:PutObject` for all roles in the account like before. We only allow `s3:PutObject` for the e2eawsiamrole to validate that only this can put objects which are part of the e2e test.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -3429,18 +3429,7 @@ Resources:
       PolicyDocument:
         Statement:
           - Action:
-              - s3:ListBucket
               - s3:PutObject
-              - s3:GetObject
-              - s3:DeleteObject
-            Effect: Allow
-            Resource:
-              - !Sub "arn:aws:s3:::${E2EAWSIAMTestBucket}"
-              - !Sub "arn:aws:s3:::${E2EAWSIAMTestBucket}/*"
-            Principal:
-              AWS:
-                - !GetAtt E2EAWSIAMTestRole.Arn
-          - Action: "s3:*"
             Effect: Deny
             Resource:
               - !Sub "arn:aws:s3:::${E2EAWSIAMTestBucket}"


mikkeloscar:
:+1:

cluster/cluster.yaml
```diff
@@ -3429,18 +3429,7 @@ Resources:
...
               - s3:PutObject
-              - s3:GetObject
-              - s3:DeleteObject
-            Effect: Allow
-            Resource:
-              - !Sub "arn:aws:s3:::${E2EAWSIAMTestBucket}"
-              - !Sub "arn:aws:s3:::${E2EAWSIAMTestBucket}/*"
-            Principal:
-              AWS:
-                - !GetAtt E2EAWSIAMTestRole.Arn
```
tcondeixa, line 3442: don't we still need to give permission for the e2e role to do all the operations? or are we doing it in the role itself?

tcondeixa: You probably want to keep the allow and only change the deny action from `s3:*` to `s3:PutObject`

mikkeloscar: My understanding is that any role can do all s3 actions except `s3:PutObject`, and e2e role can do all

I have not tested, only asked LLM

tcondeixa: if the role already has enough permissions for these actions in the bucket and it's in the same account it's fine. Here we are just denying but not really give any permission.

tcondeixa: Just checked the role and it seems to have enough permissions
https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/cluster.yaml#L3371-L3411

tcondeixa (approved):

cluster/cluster.yaml
mikkeloscar: The role has the permissions: https://github.com/zalando-incubator/kubernetes-on-aws/blob/49bf62e5d66134765060c347a1d519d300dd5a69/cluster/cluster.yaml#L3397-L3405

I think you anyway need both directions always.

1. IAM role allows it so it doesn't do actions it should be able to
2. Bucket Policy allows it or restricts even if the iam role allow things not allowed on _that_ bucket.

PR #11711, opened 2026-07-17
Use RollingUpdate for CoreDNS DaemonSet

a4180p:
Use RollingUpdate for CoreDNS DaemonSet

### DIFF
--- a/cluster/manifests/01-coredns-local/daemonset-coredns.yaml
+++ b/cluster/manifests/01-coredns-local/daemonset-coredns.yaml
@@ -8,7 +8,7 @@ metadata:
     component: coredns
 spec:
   updateStrategy:
-    type: OnDelete
+    type: RollingUpdate
   selector:
     matchLabels:
       daemonset: coredns


mikkeloscar:
This is intentionally not done because it would mean a downtime every time an update is rolled out.

a4180p:
@mikkeloscar I was trying to measure an effect of that on the create-cluster step running time. And effect is huge.

mikkeloscar:
> @mikkeloscar I was trying to measure an effect of that on the create-cluster step running time. And effect is huge.

Yes, we explicitly roll all nodes if the CoreDNS daemonset needs an update to "simulate" the update during e2e.

I'm working on a few changes related to static DNS interface (https://github.com/zalando-incubator/kubernetes-on-aws/pull/11216) which requires rolling the daemonset/nodes, but outside of that e.g. unrelated changes should not need to rotate the nodes for this and thus have not impact on the time.

a4180p:
> Yes, we explicitly roll all nodes if the CoreDNS daemonset needs an update to "simulate" the update during e2e.

Can we create and use a separate cluster for that?
It would really help with waiting times (~30 minutes of saved time)

PR #11724, opened 2026-07-20
Set default routes-url for canary to the zone aware endpoint

greeshma1196:
The config-item `skipper_ingress_canary_routes_urls` is set to the zone aware endpoint in all clusters, this PR cleans up the config-item and makes it the default endpoint.


vlktna:
👍

MustafaSaber (approved):

PR #11731, opened 2026-07-20
karpenter: Allow enabling NodeRepair

mikkeloscar:
Allow enabling NodeRepair feature gate via a config-item.

This would allow Karpenter to perform automatic node repair actions as outlined in https://karpenter.sh/docs/concepts/disruption/#node-auto-repair

This feature will only work in EKS as it needs to be paired with e.g. eks-node-monitoring-agent which can indicate "broken" nodes.


mikkeloscar:
👍

katyanna:
👍

PR #11819, opened 2026-07-30
Update OPA bundle path to refer v2

Pushpalanka:
Direct OPA agents to retrieve bundles from v2

Rollback plan would be to switch back to V1.
Currently Skipper is running on V1, but the content is V2 and this is making it transparent.


demonCoder95:
Closing this, as pipelines are not running on a PR based on a fork

PR #11852, opened 2026-08-03
set karpenter maxPods for EKS clusters

demonCoder95:
Currently, there is no limit set for `maxPods` per node in the kubelet configuraiton for Karpenter nodes. This results in some larger instance sizes (e.g `96xlarge`) to schedule 100s of Pods. This results in unpredictable behavior and is not supported.

This change limits the `maxPods` for EKS clusters to `110`, which is the maximum supported by Kubernetes: https://kubernetes.io/docs/setup/best-practices/cluster-large/

### DIFF
--- a/cluster/node-pools/worker-karpenter/provisioners.yaml
+++ b/cluster/node-pools/worker-karpenter/provisioners.yaml
@@ -53,11 +53,13 @@ spec:
     # {{ end }}
     cpuCFSQuota: false
 #{{ if ne .Cluster.Provider "zalando-eks" }}
-# {{ if ne .Cluster.ConfigItems.karpenter_max_pods_per_node "" }}
+#   {{ if ne .Cluster.ConfigItems.karpenter_max_pods_per_node "" }}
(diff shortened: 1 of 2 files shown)


demonCoder95:
The default of `110` is always injected for every EKS cluster, so the concern by LLM is not relevant ^

cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -53,11 +53,13 @@ spec:
     # {{ end }}
     cpuCFSQuota: false
 #{{ if ne .Cluster.Provider "zalando-eks" }}
```
mikkeloscar: This part is no longer needed as it's done on config-defaults side. So drop this to make the condition simpler

demonCoder95: Updated in the next commit

mikkeloscar:
:+1:

demonCoder95:
👍

PR #11864, opened 2026-08-04
Enable Scheduling Controls by default

mikkeloscar:
It's set as default in all environments via CCM, simply set it as default here so it also applies to pet clusters.


mikkeloscar:
👍

katyanna:
:+1:

PR #11850, opened 2026-08-03
Update AMI to Kubernetes v1.35.7

mikkeloscar:
This updates the AMI to Kubernetes v1.35.7 and resolves a critical CVE in the current AMI.

PR #11865, opened 2026-08-04
Update to AMI with open NVIDIA drivers

mikkeloscar:
Updates the AMI to a version using Open NVIDIA drivers needed for g7/g7e instance type support.


mikkeloscar:
This was used to validate the new NVIDIA driver pass e2e.

The actual AMI update is in #11850

PR #11880, opened 2026-08-04
 Grant ACK capability role for S3 buckets

alkagin:
Add a set of permissions to test the (ACK) S3 controller. I may miss some actions which will be caught during the PoC. After the PoC, we can revisit it again.

### DIFF
--- a/cluster/cluster.yaml
+++ b/cluster/cluster.yaml
@@ -470,6 +470,10 @@ Resources:
           - "eks:AssociateAccessPolicy"
           - "eks:DisassociateAccessPolicy"
           - "eks:ListAssociatedAccessPolicies"
+          - "s3:CreateBucket"
+          - "s3:PutBucket*"
+          - "s3:ListBuckets"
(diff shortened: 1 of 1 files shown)


cluster/cluster.yaml
```diff
@@ -470,11 +471,22 @@ Resources:
           - "eks:AssociateAccessPolicy"
           - "eks:DisassociateAccessPolicy"
           - "eks:ListAssociatedAccessPolicies"
-          - "bedrock:CreateInferenceProfile"
-          - "bedrock:GetInferenceProfile"
-          - "bedrock:DeleteInferenceProfile"
-          - "bedrock:ListInferenceProfiles"
           Resource: "*"
+  # Temporary PoC permissions for ACK S3 controller.
+  # Revisit scope after PoC is complete; remove or tighten before GA.
```
demonCoder95, line 476: For temporary use cases like these, it would be better to put this behind a config-item so we can selectively enable it in PoC clusters instead of rolling it out everywhere.

PR #11883, opened 2026-08-04
Allow AWS fluent-bit image in production

mikkeloscar:
Follow up to #11770 

Allow the AWS fluent-bit image in production by ignoring the image check in admission-controller. This was missed and only done in e2e in #11770

The motivation for doing this is to avoid re-hosting the AWS provided image on ECR. This is similar to EKS addons where we also rely on and trust the AWS provided images.


demonCoder95:
👍

katyanna:
👍

PR #11890, opened 2026-08-05
Revert fluent bit

mikkeloscar:
Revert the fluent-bit journald-reader setup introduced in #11770 and #11883 

This setup only works on bottlerocket making it not attractive to run on the short term. The assumed reason is the version of systemd in the fluent-bit image is only compatible with the systemd on bottlerocket and not on Ubuntu AMI. The result is no logs (and no errors) on Ubuntu.


mikkeloscar:
:+1:

ni9hty:
:+1:

PR #11891, opened 2026-08-05
Make journald-reader bottlerocket compatible

mikkeloscar:
Make the journald-reader container compatible with Bottlerocket. The main changes are:

* Implementation is Go based so it's not depending on python on the host
* `LD_LIBRARY_PATH` is set and extended to cover paths used on bottlerocket such that journalctl and related libraries can be found.

* [x] Update to production built image


mikkeloscar:
:+1:

demonCoder95:
👍

PR #11918, opened 2026-08-10
AMI: neigh-gc fix

mikkeloscar:
This updates the AMI to a version that increases the IPv6 Neightbor Discovery limits similar to how it's done for IPv4 (the ARP equivalent)

Without this EKS IPv6 clusters can reach table limits and drop connections:

    [Fri Aug  7 12:57:04 2026] neighbour: ndisc_cache: neighbor table overflow!


mikkeloscar:
:+1:

ni9hty:
:+1:

PR #11919, opened 2026-08-10
AMI: neigh-gc fix

mikkeloscar:
This updates the AMI to a version that increases the IPv6 Neightbor Discovery limits similar to how it's done for IPv4 (the ARP equivalent)

Without this EKS IPv6 clusters can reach table limits and drop connections:

    [Fri Aug  7 12:57:04 2026] neighbour: ndisc_cache: neighbor table overflow!
    
The AMI update is scoped to EKS clusters only as this is the only place where the setting matters. We want to avoid rolling the nodes in non-EKS when it's not needed (this change will be anyway in all Kubernetes v1.36 AMI #11918)


mikkeloscar:
:+1:

ni9hty (approved):

PR #11938, opened 2026-08-10
Scope karpenter_ami_family_alias to EKS

mikkeloscar:
Scope `karpenter_ami_family_alias` to EKS to avoid configuring bottlerocket in non-EKS which is not supported.

This allows things like setting bottlerocket to all skipper-ingress nodes in test clusters but it will only apply to EKS.


mikkeloscar:
:+1:

demonCoder95 (approved):

demonCoder95:
👍

PR #11956, opened 2026-08-11
Allow ZMON readaccess from outside

mikkeloscar:
Extend the zmon external access to readonly permissions. This allows making ZMON checks that run from a zmon-worker in a different cluster to monitor core functionality that could prevent zmon _inside_ the cluster from working in case there are issues.

This allows us to work around the current issue where the local zmon is used to monitor the local cluster and if some part of the local cluster is broken, we won't get notified.

Concrete examples to monitor:

* Monitor DNS from the outside as the local zmon depends on it to work
* Monitor Karpenter from the outside and local zmon depends on it to scale and work.

### DIFF
--- a/cluster/manifests/roles/zmon-external-check-rbac.yaml
+++ b/cluster/manifests/roles/zmon-external-check-rbac.yaml
@@ -1,28 +1,19 @@
-# This role is used to check coredns endpoints of a cluster from the outside.
+# This gives zmon read access to clusters from the outside. This allows checking
+# various core functionality like coredns availability and Karpenter
+# availability from a zmon-worker in another cluster.
+#
 # The intention is to monitor a cluster from the outside in case DNS is broken
 # and ZMON inside the cluster would not work.
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: zmon-external-check
-rules:
-- apiGroups:
-  - ""
-  resources:
-  - endpoints
-  resourceNames:
-  - coredns
-  verbs:
-  - get
----
+# Or Karpenter is not working and can't scale up nodes, so ZMON inside the
+# cluster would not work.
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

cluster/manifests/roles/zmon-external-check-rbac.yaml
```diff
@@ -1,28 +1,19 @@
...
-  verbs:
-  - get
----
+# Or Karpenter is not working and can't scale up nodes, so ZMON inside the
+# cluster would not work.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: zmon-external-check
+  name: zmon-external-reader
```
demonCoder95, line 12: Where is this `ClusterRole` defined? 🤔

mikkeloscar: It binds to the `readonly` role which already exists: https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/roles/readonly-role.yaml

This is this "magic" role that is extended with read permissions if new CRDs are added to the cluster. The one attached also to human ReadOnly role.

demonCoder95: I mean the role we renamed below to `readonly` but the binding here refers to the role as `zmon-external-reader`. That's a mismatch, no?

mikkeloscar: They don't have to match, so the question is if it's confusing to us.

The intention of the binding is to give zmon read access from outside, hence the name `zmon-external-reader`. This binds zmon identity to the generic `reader` role.

demonCoder95: Ah, okay. That makes sense then. Thanks!

demonCoder95:
👍

PR #11961, opened 2026-08-11
ebs-csi controller must not run on cluster-seed node

mikkeloscar:
Change the ebs-csi controller to not require running on the cluster-seed node. It still can run there, but doesn't have to. This allows it to run _somewhere_ even if the cluster-seed is rotating.


mikkeloscar:
:+1:

Shinzu (approved):

PR #11965, opened 2026-08-12
Change Karpenter VPA updateMode to 'Recreate'

zaklawrencea:
Currently we observe Karpenter ending up in an unhealthy state (OOMing), especially during node rotations. The `InPlace` update functionality doesn't appear to be stable in our setup at the moment so try reverting it back to only `Recreate`.


mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #11967, opened 2026-08-13
e2e: Disable non-compliant image tests

mikkeloscar:
The test docker-meta instance used in these tests was changed so the tests no longer work. Disable for now until we have a better solution to avoid broken e2es.


zaklawrencea:
:+1:

mikkeloscar:
:+1:

PR #11983, opened 2026-08-14
kube-aws-iam-controller must not run on cluster-seed node

mikkeloscar:
Similar to #11961

Change the kube-aws-iam-controller to not require running on the cluster-seed node. It still can run there, but doesn't have to. This allows it to run somewhere even if the cluster-seed is rotating.


tcondeixa:
👍

mikkeloscar:
:+1:

PR #11985, opened 2026-08-14
VPA teapot admission controller

tcondeixa:
Add VPA for teapot admission controller to save on resources and to automatically scale up when resources are not enough. The updateMode is set to `Off` to collect metrics for a week before enable it with `InPlaceOrRecreate`

The VPA CRD was moved before the admission controller so they are there when creating the VPA, otherwise we will have an error for new clusters.


cluster/manifests/02-admission-control/vpa.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if eq .Cluster.Provider "zalando-eks"}}
```
demonCoder95, line 1: I think it would be better to put this behind another feature flag for testing, we don't want it enabled for all EKS clusters right away. Safer to run it for a while on some and test.

tcondeixa: what is the main concern here? this is configured with updateMode `Off` for now, so it's only collecting metrics.

tcondeixa:
👍

mikkeloscar:
:+1:

PR #12003, opened 2026-08-17
e2e: Give EKS seed node high priority for rotation

mikkeloscar:
Observed in e2e that the cluster-seed node has same priority as other nodes meaning other nodes can be rotated first and let the seed node unscheduled for a long time. This is not desired.

Additionally change the taint value set by the daemonset updater as it was a non-nonsensical taint value in the context.


tcondeixa:
👍

tcondeixa:
is this something simple we could also apply not just for e2e but for any node rotation in the clusters? We would need to still work on this topic, but could be a quick and easy win.

mikkeloscar:
:+1:

PR #12024, opened 2026-08-18
karpenter: Fix don't bind var logs pods to ssd

mikkeloscar:
Follow up to #11976 fixing the Karpenter patch to not bind mount `/var/log/pods` on local SSD instance types.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #12027, opened 2026-08-18
feature: inject "zalando.org/topology-spread: true" into the pod template annotation in stacksets

szuecs:
feature: inject "zalando.org/topology-spread: true" into the pod template annotation in stacksets

off by default, rolling out via ccm

ref: https://github.com/zalando-incubator/kubernetes-on-aws/pull/11996 should be deployed, before we enable this


mikkeloscar:
:+1:

szuecs:
:+1:

PR #12048, opened 2026-08-19
Configure HPAScaleToZero feature gate on kube-controller-manager

mikkeloscar:
This configures the HPAScaleToZero feature gate on the kube-controller-manager. This is needed since Kubernetes v1.36 where the HPA controller is checking for this feature gate to decide on the ScaledToZero feature introduced in https://github.com/kubernetes/kubernetes/pull/135118

Before it was enough to set HPAScaleToZero feature gate on the kube-apiserver.


mikkeloscar:
:+1:

demonCoder95:
👍

PR #12053, opened 2026-08-19
external-dns: Disable creation of AAAA records

mikkeloscar:
Upgrading external-dns in #11984 implicitly enabled creation of AAAA records. This uncovered that our platform is not 100% IPv6 capable on the related paths and caused an incident.

To allow updating external-dns without this side-effect, disable AAAA records by default.

This also lets us enable them per cluster once we are ready to test and support it properly.


mikkeloscar:
👍

linki:
👍

PR #12077, opened 2026-08-24
Default kube_aws_iam_controller_enabled to false

linki:
## Summary
- kube-aws-iam-controller / AWSIAMRole is deprecated in favor of OIDC-based IAM roles.
- Flip the default of `kube_aws_iam_controller_enabled` from `"true"` to `"false"` in `cluster/config-defaults.yaml`.
- Clusters still relying on AWSIAMRole CRDs have been switched to explicit opt-in (`kube_aws_iam_controller_enabled: true`) in zalando-build/teapot-cluster-config-manager#1874, so no existing cluster loses the controller as a result of this change.

## Test plan
- [x] Confirm zalando-build/teapot-cluster-config-manager#1874 merges (or has merged) before/alongside this, so opted-in clusters keep the controller running
- [x] Verify no other cluster still depends on kube-aws-iam-controller without an explicit override

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -676,7 +676,7 @@ metrics_server_metric_resolution: "15s"
 
 kube_aws_iam_controller_cpu: "5m"
 kube_aws_iam_controller_mem: "50Mi"
-kube_aws_iam_controller_enabled: "true"
+kube_aws_iam_controller_enabled: "false"
 
 kube_state_metrics_cpu: "100m"
 kube_state_metrics_mem_max: "4Gi"
--- a/cluster/manifests/deletions.yaml
+++ b/cluster/manifests/deletions.yaml
@@ -409,10 +409,8 @@ post_apply:
 - name: kube-aws-iam-controller
   namespace: kube-system
   kind: ServiceAccount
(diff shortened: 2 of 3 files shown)


linki:
👍

linki:
👍

linki:
@zalando-robot clusters have been updated accordingly in https://github.com/zalando-build/teapot-cluster-config-manager/pull/1874

linki:
👍

tcondeixa (approved):

cluster/manifests/03-kube-aws-iam-controller/deployment.yaml
```diff
@@ -8,7 +8,11 @@ metadata:
     application: kubernetes
     component: aws-iam-controller
 spec:
+{{ if eq .Cluster.ConfigItems.kube_aws_iam_controller_enabled "true" }}
   replicas: 1
+{{ else }}
+  replicas: 0
+{{ end }}
```
linki: The idea is that it scales down the deployment during the apply phase so that when the deletions kick in there's no running pod that blocks serviceaccount deletion.

linki:
👍

tcondeixa (approved):

PR #12101, opened 2026-08-25
add config-item to inject AWS_REGION and AWS_DEFAULT_REGION to pods

tcondeixa:
add option to inject AWS_REGION to pods with config-item, so we can gradually rollout to clusters.
The default is set to false and we using the config-item rollout to a set of cluster at the time.

This ensure that region is set for the IRSA, and avoid to use the metadata api through the kube2iam proxy to get that information.


szuecs (approved):

mikkeloscar:
👍

PR #12120, opened 2026-08-26
skipper-ingress: update main-fleet to v0.27.71

szuecs:
- update: skipper-ingress: update main-fleet to v0.27.71
- feature: h2c enabled in test clusters, disabled in production
- refactor: metric flags that are set by default an not changed in cluster-registry do not need a toggle
- fix: skipper_combined_response_metrics is set to true by default, because it is set to true via cluster-registry to a lot of clusters so we can use this by default


madumalt (approved):

cluster/config-defaults.yaml
```diff
@@ -142,7 +142,11 @@ skipper_ingress_memory: "1500Mi"
 skipper_topology_spread_enabled: "true"
 skipper_topology_spread_timeout: "7m"
 
-skipper_copy_stream_pool_enabled: "true"
+{{if eq .Cluster.Environment "production"}}
```
madumalt, line 145: I was talking about this, we usually set the default feature flag to true and then disable/enable as appropiate in the CLM. 

guideline doc: https://pg9.docs.zalando.net/howtos/safe-channel-promotion-and-rollout-strategies/#feature-flagscluster-config-manager-ccm

szuecs: Yeah this will be just dropped to have only "true" by default after the migration.
It's just a speedup to deliver fast, as zamp is aiming 1st September, I do small quick hacks that do not change reliability.

szuecs:
:+1:

cluster/config-defaults.yaml
MustafaSaber: I don't think it's worth it to enable here TBH even as a hack, just merge and have CCM PR ready to disable this for all prod

PR #12138, opened 2026-08-27
skipper-ingress: set otel sampling to always_on

madumalt:
From v0.27.69 [1] onwards skipper has the capability to configure the otel sampler mode. The default mode is parentbased_always_on. We want to change it to always_on as we want to propagate sampled=1 always from skipper-ingress.

This make sure that even after a upstream service decides to change the sampled=0, after skipper-ingress it will be overridden to sampled=1. As http traffic almost always goes through skipper-ingress before they reach the application in Zalando, it avoids cascading of trace sampling decisions that individual application make to other applications.

References:
  [1] https://github.com/zalando/skipper/releases/tag/v0.27.69

### DIFF
--- a/cluster/manifests/skipper/deployment.yaml
+++ b/cluster/manifests/skipper/deployment.yaml
@@ -303,6 +303,7 @@ spec:
               cloud.availability_zone: $(KUBE_NODE_ZONE)
               otel-exporter: true
             propagators: [tracecontext, ottrace, b3multi, baggage]
+            sampler: always_on
             batchSpanProcessor:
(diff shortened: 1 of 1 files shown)


madumalt:
👍

szuecs:
https://github.com/zalando-incubator/kubernetes-on-aws/pull/12120 needs to be in stable before we can merge this, because the pre condition of the mentioned version is not met, yet.

madumalt:
> #12120 needs to be in stable before we can merge this, because the pre condition of the mentioned version is not met, yet.

it is in stable. Now we can merge this.

madumalt:
@szuecs let's merge this

szuecs:
:+1:

PR #12160, opened 2026-08-31
Make pod image-check namespaces and ignored-images configurable

linki:
## Summary
- Extract the previously hardcoded, e2e-only values for `pod.image-check.namespaces` and `pod.image-check.ignored-images` into config items: `teapot_admission_controller_validate_pod_images_namespaces` and `teapot_admission_controller_validate_pod_images_ignored_images`.
- These config items are only set for e2e clusters, matching the previous hardcoded behavior. For other environments they are left unset.
- `cluster/manifests/02-admission-control/config.yaml` guards both settings the same way as the existing `teapot_admission_controller_validate_pod_images_soft_fail_namespaces` config item, so they are simply omitted from the rendered ConfigMap (falling back to the admission controller's own defaults: match everything / match nothing) when not configured.
- Added commented-out examples in the production branch of `config-defaults.yaml`, next to the existing soft-fail-namespaces hint, documenting these as a temporary escape hatch during incidents.

## Test plan
- [x] Verified `config-defaults.yaml` only defines the two config items in the e2e branch (no duplicate definitions).
- [x] Verified `config.yaml` renders `pod.image-check.namespaces`/`pod.image-check.ignored-images` only when the corresponding config item is set.

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -795,6 +795,12 @@ teapot_admission_controller_validate_pod_images: "true"
 # if docker-meta is down, do not reject container images running in `kube-system`
 # teapot_admission_controller_validate_pod_images_soft_fail_namespaces: "^kube-system$"
 
+# As a temporary escape hatch during incidents, you can limit the namespaces where container image
+# checks are required (defaults to all namespaces) and/or always allow certain container images
+# regardless of namespace (defaults to none), e.g.:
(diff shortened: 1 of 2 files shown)


tcondeixa:
👍

tcondeixa:
@linki can you add these 2 options to the playbook after merged and validated?

demonCoder95:
👍

PR #12185, opened 2026-09-01
Add pod.image-check.ignore-namespaces config item

linki:
## Summary
- Follows up on #12160: admission-controller also supports `pod.image-check.ignore-namespaces` to exclude namespaces from container image compliance checks (complementing the existing `pod.image-check.namespaces` allow-list).
- Wires this up as `teapot_admission_controller_validate_pod_images_ignored_namespaces` in `cluster/manifests/02-admission-control/config.yaml`, matching the "ignored" naming used by the sibling `..._ignored_images` config item.
- Note: admission-controller's own flag is `pod.image-check.ignore-namespaces` (no "d"), so there's a small naming mismatch between our config item and the underlying flag — called out with a comment since we can't change admission-controller's flag name here.
- Documents the new escape-hatch config item in `cluster/config-defaults.yaml`.

## Test plan
- [ ] Verify rendered admission-controller configmap includes `pod.image-check.ignore-namespaces` when the config item is set

### DIFF
--- a/cluster/config-defaults.yaml
+++ b/cluster/config-defaults.yaml
@@ -795,10 +795,10 @@ teapot_admission_controller_validate_pod_images: "true"
 # if docker-meta is down, do not reject container images running in `kube-system`
 # teapot_admission_controller_validate_pod_images_soft_fail_namespaces: "^kube-system$"
 
-# As a temporary escape hatch during incidents, you can limit the namespaces where container image
-# checks are required (defaults to all namespaces) and/or always allow certain container images
-# regardless of namespace (defaults to none), e.g.:
-# teapot_admission_controller_validate_pod_images_namespaces: "^visibility$"
+# As a temporary escape hatch during incidents, you can exclude certain namespaces from the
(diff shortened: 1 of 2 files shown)


linki:
👍

tcondeixa:
I still see the manifest templates with `teapot_admission_controller_validate_pod_images_namespaces` but there is not default value. Do we still need to expose this config-item since we have now a way to exclude namespaces?

linki:
We can go back to this approach from before the first PR: https://github.com/zalando-incubator/kubernetes-on-aws/blob/e4735a744d274615e9016e9629ece016321015be/cluster/manifests/02-admission-control/config.yaml#L99

It sets the relevant inclusive field for admission-controller without exposing it as a config item. In this way it becomes less likely that anyone of us misuses this setting. But on the other hand it makes it harder for someone that knows what they do to set it (at least for multiple clusters).

linki:
👍

tcondeixa:
👍

PR #12193, opened 2026-09-02
Karpenter: discover tagged EC2 Capacity Reservations and offer `reserved` capacity

elgalu:
## Why

With the `ReservedCapacity` feature gate on (fleet-wide since the v1.8.0 bump) Karpenter launches every node with
`CapacityReservationPreference=none`, so an open EC2 Capacity Reservation is consumed only if the EC2NodeClass selects it, and nothing
did. 2026-09-02, search-test: three `g7e.2xlarge` nodes launched on-demand while the open reservation sat at 1/1 available until
cancelled. The gate-off override (`karpenter_feature_reserved_capacity: false`, used by deepthought and fashion-proposition-tech)
consumes open reservations blindly and leaves Karpenter unable to prefer, consolidate into or report on them; this PR makes the
gate-on mode work instead.

Companion: zalando-build/admission-controller#419, which lets `require-on-demand` pods land on `reserved` nodes. Without it this PR
serves only pods that carry no such policy.

## What

| File | Change |
| --- | --- |
| `cluster/node-pools/worker-karpenter/provisioners.yaml` | EC2NodeClass selects reservations tagged `karpenter.sh/discovery: <cluster-id>/CapacityReservation` (the `securityGroupSelectorTerms` convention); NodePool `capacity-type` gains `reserved` |
| `cluster/config-defaults.yaml` | `karpenter_feature_reserved_capacity` comment rewritten as the two-mode matrix (tagged-and-modelled vs open-and-blind) |
| `cluster/manifests/z-karpenter/deployment.yaml` | `FEATURE_GATES` comment points at that matrix |

Both additions render unconditionally: a reservation without the tag is never selected, and on a gate-off cluster Karpenter skips the
reservation reconciler and the reserved-offering resolver, so they are inert there (verified in karpenter-provider-aws v1.14.1,
`pkg/controllers/nodeclass/controller.go` and `pkg/providers/instancetype/offering/reserved_capacity_resolver.go`).
`capacityReservationSelectorTerms` is `hash:"ignore"`, so no existing node drifts.

What a gate-on cluster gets once a reservation carries the tag:

- Karpenter launches into it by id, ahead of on-demand: a reserved offering is priced at on-demand / 10^7.
- An existing matching on-demand node is consolidated into it, which replaces the manual `zkubectl delete nodeclaim` step.
- The node is labelled `karpenter.sh/capacity-type: reserved`; Capacity Blocks and interruptible reservations carry their own labels.
- Karpenter falls back to on-demand once the reservation is full.

IAM needs nothing: the controller policy already grants `ec2:DescribeCapacityReservations` and `capacity-reservation/*` on
`RunInstances` / `CreateFleet`. No config-manager change: the selector is tag-scoped, so consuming a reservation is self-service at
purchase time. The tag value is per cluster, so a reservation shared by two clusters in one account is discoverable by one of them.

## Verification

Render emulation of both templates parses to an `EC2NodeClass` carrying the selector plus the provider-conditional `userData` chain
from #12185's Bottlerocket branch, and a `NodePool` with `capacity-type In [spot, on-demand, reserved]`.


cluster/node-pools/worker-karpenter/provisioners.yaml
```diff
@@ -112,6 +112,13 @@ spec:
 #    {{ end}}
 #  {{ end}}
 #{{ end}}
+  # An EC2 Capacity Reservation tagged `karpenter.sh/discovery: <cluster-id>/CapacityReservation` (the securityGroupSelectorTerms
+  # convention) becomes a `reserved` offering of every pool using this class: Karpenter launches into it by id ahead of on-demand and
+  # consolidates a matching on-demand node into it. Inert until a reservation carries the tag, and inert while
+  # `karpenter_feature_reserved_capacity` is false (the reconciler is skipped); the two modes are documented on that config item.
+  capacityReservationSelectorTerms:
+    - tags:
+        karpenter.sh/discovery: "{{ .Cluster.ID }}/CapacityReservation"
```
mikkeloscar: Maybe this needs to be configurable, for special cases, but my intuitions says we want some simple tag _without_ the cluster.ID in it.

It should be straight forward to create and consume and ODCR without having to think about what the cluster.ID is, also if we share an ODCR for multiple clusters it won't work to require a cluster.ID (unless it's part of the tag key).

I would also not make the tag name karpenter specific, from the UX perspective it's enabling the Kubernetes platform to consume it. `zalando.org/kubernetes: "true"` or something generic like this.

elgalu: Agreed, switched to `zalando.org/kubernetes: "true"` with no cluster id, so buying one takes no cluster lookup and clusters in the account can share it. Left it hardcoded rather than configurable

PR #12209, opened 2026-09-04
skipper-ingress: removes b3multi from otel propagators list

madumalt:
As explained in the test cases in PR [[1]](https://github.com/zalando/skipper/pull/4247), having b3 propagator results in disconnected traces.

With opentracing this was not supported. With opentracing we only use the lightstep propagator and it is as same as the ottrace propagator.

Therefore removing the b3 propagator restores the trace behaviour closer to the previous behaviour.


szuecs (approved):

a4180p:
👍

PR #12252, opened 2026-09-09
Update to stackset v1.4.141

mikkeloscar:
Update to stackset v1.4.141

**Changes from v1.4.128 to v1.4.141**
- Fixed traffic pre-scaling for stacks forwarding traffic to another cluster. Forwarded stacks now skip deployment readiness and replica-based pre-scaling checks.
- Upgraded Kubernetes dependencies from v0.35.4 to v0.36.4, along with Go, Prometheus, x/*, and related dependencies.
- Regenerated Kubernetes clients/informers with context-aware startup, cache synchronization, informer metrics identity, and watch-list support.
- Refreshed Stack and StackSet CRD documentation for newer Kubernetes API schemas.
- Migrated container publishing toward GitHub Container Registry and added a multi-architecture amd64/arm64 GitHub Actions workflow.
- Replaced the old Zalando static base image with gcr.io/distroless/static-debian12.
- Updated CI actions, Go 1.26 tooling, and golangci-lint v2 configuration.
- Removed the explicit IPv6/dual-stack network restriction.
- Improved error handling in the traffic CLI and Kubernetes event formatting.
- Added Dependabot grouping for GitHub Actions and refreshed build/release configuration.


mikkeloscar:
:+1:

demonCoder95:
I see a lot of fields are losing their description section entirely in the CRD definition. Is this an intended change? If users of stacksets rely on `kubectl explain` or something, this could be an issue for them, also for us to lose some embedded CRD documentation.

mikkeloscar:
> I see a lot of fields are losing their description section entirely in the CRD definition. Is this an intended change? If users of stacksets rely on `kubectl explain` or something, this could be an issue for them, also for us to lose some embedded CRD documentation.

This is automatic via this script during build: https://github.com/zalando-incubator/stackset-controller/blob/master/hack/crd/trim.go we trim the CRD at the deepest level first to reduce the chance of impacting `kubectl explain` as much as possible.

demonCoder95:
Ah, I was wondering if this was somehow related to reducing the CRD size 🙂 Alright, makes sense.

demonCoder95:
👍

PR #12260, opened 2026-09-11
eks: Fix reverse DNS lookup config

mikkeloscar:
Fix reverse DNS SERVFAIL responses in EKS clusters by scoping CoreDNS reverse zones to the cluster IP family.

- EKS IPv4 claims only in-addr.arpa.
- EKS IPv6 claims only ip6.arpa.
- Non-cluster reverse queries are forwarded through /etc/resolv.conf.
- Non-EKS DNS behavior remains unchanged.

This prevents IPv4 reverse queries from being handled as Kubernetes-owned records in IPv6 clusters, and vice versa.

Preventing log spamming (and potential real issues) like this:

```
[1789134927] unbound[1:0] error: SERVFAIL <21.136.154.10.in-addr.arpa. PTR IN>: all the configured stub or forward servers failed, at zone in-addr.arpa. from 127.0.0.1 got SERVFAIL
```

### DIFF
--- a/cluster/manifests/01-coredns-local/configmap-local.yaml
+++ b/cluster/manifests/01-coredns-local/configmap-local.yaml
@@ -56,6 +56,7 @@ data:
       # disable the default protection in unbound to allow reverse lookup
       # queries to pass through to CoreDNS
       # https://github.com/NLnetLabs/unbound/blob/5c84bb573f9728c10bcb3592dbd12be403d362de/doc/example.conf.in#L804-L850
+{{- if eq .Cluster.ConfigItems.eks_ip_family "ipv6" }}
       local-zone: "d.f.ip6.arpa." nodefault
       local-zone: "8.e.f.ip6.arpa." nodefault
       local-zone: "9.e.f.ip6.arpa." nodefault
@@ -63,8 +64,10 @@ data:
       local-zone: "b.e.f.ip6.arpa." nodefault
       local-zone: "8.b.d.0.1.0.0.2.ip6.arpa." nodefault
       local-zone: "ip6.arpa." transparent
+{{- else }}
       local-zone: "10.in-addr.arpa." nodefault
       local-zone: "in-addr.arpa." transparent
+{{- end }}
 {{- else }}
       local-zone: "2.10.in-addr.arpa." transparent
       local-zone: "3.10.in-addr.arpa." transparent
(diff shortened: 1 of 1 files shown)


cluster/manifests/01-coredns-local/configmap-local.yaml
```diff
@@ -144,7 +150,7 @@ data:
 
     # Defines that this server is authority for reverse
     # lookups for these ranges.
-    cluster.local:9254 {{if eq .Cluster.Provider "zalando-eks"}}in-addr.arpa:9254 ip6.arpa:9254{{else}}10.2.0.0/15:9254 10.5.0.0/16:9254{{end}} {{ if eq .Cluster.ConfigItems.tracing_coredns_route_traces_to_local_zone "true"}}{{ range $src := split .Cluster.ConfigItems.tracing_coredns_global_traces_endpoint  "," }}{{ $src }}:9254 {{ end }} {{ end }} {
+    cluster.local:9254 {{if eq .Cluster.Provider "zalando-eks"}}{{if eq .Cluster.ConfigItems.eks_ip_family "ipv6"}}ip6.arpa:9254{{else}}in-addr.arpa:9254{{end}}{{else}}10.2.0.0/15:9254 10.5.0.0/16:9254{{end}} {{ if eq .Cluster.ConfigItems.tracing_coredns_route_traces_to_local_zone "true"}}{{ range $src := split .Cluster.ConfigItems.tracing_coredns_global_traces_endpoint  "," }}{{ $src }}:9254 {{ end }} {{ end }} {
```
Shinzu, line 153: what is the reasoning about the separation of ipv4 and ipv6?

an ipv6 still has ipv4 node addresses or?

mikkeloscar: This block is aiming to scope the Kubernetes ranges, in IPv6 EKS, there are no IPv4 cluster ranges, so there is no need to scope it here and reversely for EKS IPv4 (we don't run it, but keep the theoretical support).

In non-EKS we have a known IPv4 range, hence we scope to: `10.2.0.0/15:9254 10.5.0.0/16:9254`, in EKS the range is dynamic, well, its static in the context of the cluster, but it's not a fixed value that is the same across all clusters, so we use the full range and fallthrough.

Shinzu: > This block is aiming to scope the Kubernetes ranges

ahh you mean the pod ip

szuecs:
we should just drop all PTR requests and make sure we do not run broken software :)

mikkeloscar:
:+1:

demonCoder95:
Did we look into the forwarding loop concern pointed out by LLM?

> [High] forward . /etc/resolv.conf in the reverse zone block may cause forwarding loops


Otherwise looks okay

mikkeloscar:
> Did we look into the forwarding loop concern pointed out by LLM?
> 
> > [High] forward . /etc/resolv.conf in the reverse zone block may cause forwarding loops
> 
> Otherwise looks okay

Yes, it's wrong about it, because the user pod `/etc/resolv.conf` that points to unbound->coreDNS, is not the host level `/etc/resolv.conf` which is highlighted here. I even asked my coding-agent LLM about it and it fixed it by introducing the exact loop it talks about 😂

katyanna:
:+1:

PR #12286, opened 2026-09-16
Add an e2e test case for HPA Scaling from 0 Replicas

demonCoder95:
This comes as an action item from the incident where the HPAScaleToZero behavior was not working but e2e didn't catch this.

Extending the current test suite that only covers scale down behavior to also check for scaling up from 0 replicas.


demonCoder95:
The PR currently reverts the fix on purpose to break the feature and then edit the current test case to trigger it to fail and correctly catch the broken feature.

This will later be extended to introduce a new set of tests separate from the existing ones and once the testing is complete, the fix will be put back in place.

demonCoder95:
The test failed as expected:

```
...
  I0916 13:11:03.371750     135 util.go:724] waiting for 1 replicas (current: 0)
...
  [FAIL] [zalando] [HPA] Horizontal pod autoscaling (scale resource: Custom Metrics from kube-metrics-adapter) [It] should scale with external metric based on hostname RPS [CustomMetricsAutoscaling] [Zalando]
```

Will now restore this test case and add new ones that fail in a similar way.

demonCoder95:
I've now added the feature-gate back to the kube-controller-manager and the test will re-run. This should pass, as it verifies that scaling from 0 is working as expected.

The last step will be to restore this testcase, and add new ones that cover the scaling from 0 behavior.

demonCoder95:
The test case logic actually triggers a "scale down" as the target rps metric value (20) is less than metric value of 10. Will adjust this to test the logic correctly to reflect the desired test case.

PR #12289, opened 2026-09-17
Support denying container registries

mikkeloscar:
This PR adds a feature to deny container registries by prefix e.g. deny pods (or higher level resources like deployment/statefulsets etc.) from defining images if the registry is on the deny list. The list is controlled via `teapot_admission_controller_deny_listed_registries: "pierone.stups.zalan.do/,registry.opensource.zalan.do/"`. The intention of this feature is to block pierone usage in test clusters as it's being phased out.


linki:
👍

mikkeloscar:
:+1:

cluster/config-defaults.yaml
```diff
@@ -768,6 +768,12 @@ teapot_admission_controller_deployment_default_max_unavailable: "1"
 teapot_admission_controller_inject_aws_waiter: "true"
 teapot_admission_controller_parent_resource_hash: "true"
 teapot_admission_controller_stackset_inject_topology_spread_enable: "false"
+{{ if eq .Cluster.Environment "test" }}
+# Comma-separated list of image registries denied in test clusters.
+teapot_admission_controller_deny_listed_registries: "pierone.stups.zalan.do/,registry.opensource.zalan.do/"
```
jmcs: OS registry should still be allowed.

mikkeloscar: alright, removed

mikkeloscar:
:+1:

cluster/config-defaults.yaml
```diff
@@ -768,6 +768,12 @@ teapot_admission_controller_deployment_default_max_unavailable: "1"
 teapot_admission_controller_inject_aws_waiter: "true"
 teapot_admission_controller_parent_resource_hash: "true"
 teapot_admission_controller_stackset_inject_topology_spread_enable: "false"
+{{ if eq .Cluster.Environment "test" }}
+# Comma-separated list of image registries denied in test clusters.
+teapot_admission_controller_deny_listed_registries: "pierone.stups.zalan.do/"
```
tcondeixa: it's a bit strange we need to terminate the registry with a `/`, maybe we can make it part of the template itself? if we expose this to cluster-config it has some room for issues.

mikkeloscar: It's not needed, it's a prefix match so we can just drop the `/`

mikkeloscar:
:+1:

tcondeixa (approved):

PR #12290, opened 2026-09-17
Update to Bottlerocket admin container v0.21.5

mikkeloscar:
https://github.com/bottlerocket-os/bottlerocket-admin-container/releases/tag/v0.21.5


demonCoder95:
Does this have the potential to rotate nodes? 🤔 

In which case we might need to label it appropriately and plan the rollout.

demonCoder95:
We also have the containerd v2.4.0 update to the AMI, we can roll it out together with this update.

PR #12292, opened 2026-09-17
Enable running EKS seed nodes on Bottlerocket

zaklawrencea:
Introduce a new `seed-default` profile, utilising AutoScalingGroups to enable running EKS Seed Nodes on Bottlerocket. The new profile also includes userdata to add relevant labels/taints and to enable the use of debugging via our admin container image.


cluster/node-pools/worker-combined/stack.yaml
```diff
@@ -7,11 +7,18 @@ Metadata:
     component: "shared-resource"
     "kubernetes.io/cluster/{{.Cluster.Name}}": "owned"
 
+{{- if eq .NodePool.ConfigItems.node_ami_family "bottlerocket" }}
+Parameters:
+  BottlerocketAMI:
+    Type: 'AWS::SSM::Parameter::Value<AWS::EC2::Image::Id>'
+    Default: '/aws/service/bottlerocket/aws-k8s-1.36/{{if eq .Values.InstanceInfo.Architecture "amd64"}}x86_64{{else}}arm64{{end}}/latest/image_id'
```
mikkeloscar: Instead of `latest` here we should use a version, possible set via config-item if that is simpler. We can use v1.65.0 which is the latest available

zaklawrencea: I added a config item: `bottlerocket_ami_version` with a default value of the current latest version `1.65.0` as you mentioned. It makes the string a little less readable with the added templating but improves the flexibility. In the future this could potentially be merged with the `karpenter_ami_family_alias` which currently sets `bottlerocket@<version>` in the config manager. If you have a different opinion on naming etc then just let me know.

zaklawrencea:
:+1:

cluster/node-pools/worker-combined/stack.yaml
```diff
@@ -7,11 +7,18 @@ Metadata:
     component: "shared-resource"
     "kubernetes.io/cluster/{{.Cluster.Name}}": "owned"
 
+{{- if eq .NodePool.ConfigItems.node_ami_family "bottlerocket" }}
+Parameters:
+  BottlerocketAMI:
+    Type: 'AWS::SSM::Parameter::Value<AWS::EC2::Image::Id>'
+    Default: '/aws/service/bottlerocket/aws-k8s-1.36/{{if eq .Values.InstanceInfo.Architecture "amd64"}}x86_64{{else}}arm64{{end}}/{{.Cluster.ConfigItems.bottlerocket_ami_version}}/image_id'
```
tcondeixa: I would not use parameters here, because the default only works for the creation, if you change the default later it will not update all the existing ones. We would need to pass the parameters on every call to cloudformation but we don't have that capability in CLM.

tcondeixa: does the `aws-k8s-1.36` in the path of the AMI needs to match the EKS cluster version?

PR #12320, opened 2026-09-21
set instance types for skipper-ingress deployments

greeshma1196:
add config items to enable and set instance types for skipper-ingress deployments


greeshma1196:
Closing this PR, will control the instance via CPU and memory

PR #12330, opened 2026-09-22
IPv6 explicit opt out for IPv4 clusters

mikkeloscar:
This introduces a config-item: `disable_ipv6_address_assignment` (default `false`) which when set to `true` will explicitly disable IPv6 address assignment to EC2 nodes via `Ipv6AddressCount: 0` on the respective launch templates.

The motivation for this feature is to be able to enable `AssignIpv6AddressOnCreation` on subnets, which is required by EKS node groups and EKS automode, while still allowing to run legacy nodes in the same subnets.

We want to explicitly opt-out of IPv6 for workloads that should not need it to reduce the risk when enabling `AssignIpv6AddressOnCreation`.

### TODO

* [ ] Use prod karpenter image


szuecs:
@mikkeloscar does this impact ingress/routegroup backend type `forward` ?
IIRC we have to add the IPv6 on skipper-ingress nodes to allow the roundtrip to the eks cluster.

